PR #441, opened 2014-01-31
[admin] Commands to append/remove values to config options

saltire:
If a config option contains a list, we can use the .addto and
.removefrom commands to modify the list without having to retype the
whole thing.

Also, after making a change the bot will confirm the new setting with a reply.

### DIFF
--- a/willie/modules/admin.py
+++ b/willie/modules/admin.py
@@ -136,10 +136,10 @@ def invite_join(bot, trigger):
 @willie.module.priority('low')
 def hold_ground(bot, trigger):
     """
-    This function monitors all kicks across all channels willie is in. If it
+    This function monitors all kicks across all channels Willie is in. If it
     detects that it is the one kicked it'll automatically join that channel.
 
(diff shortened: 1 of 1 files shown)


willie/modules/admin.py
```diff
@@ -189,29 +191,39 @@ def set_config(bot, trigger):
...
+            setattr(getattr(bot.config, section), option, value)
+
+        elif command in ('addto', 'removefrom'):
+            # Convert this option to a list, and try to add or remove the value.
+            vlist = getattr(bot.config, section).get_list(option)
+            try:
+                vlist.append(value) if command == 'addto' else vlist.remove(value)
+                setattr(getattr(bot.config, section), option, vlist)
+            except ValueError: # tried to remove value not in list
+                pass
```
embolalia: This should have some sort of error message.

embolalia:
This probably won't work with whatever we come up with for #708, so I'm going to close it and make sure the functionality gets included when that one gets done.

PR #448, opened 2014-02-07
[modules] Added tinyurl module

Ethcelon:
Hello
I added a tinyurl url shorten command. Please let me know if any changes are required.


embolalia:
On second thought, this would belong in -extras. As far as new standards, the things that pop out are that you need to use decorators rather than assigning attributes, and the bot object passed to the callable should be called `bot`, not `willie`. Generally, it should follow the same conventions as everything else.

PR #598, opened 2014-08-16
improve wikipedia module to print head of the wikipedia pages

roidelapluie:


embolalia:
I've implemented the feature, with the formatting changes and some tweaks to reduce repetition. Thanks for the PR.

PR #627, opened 2014-09-28
Add utc option to chanlogs module

moises-silva:
Although UTC is most of the time the sane option, at times you just need logs that will be displayed in the local time of the server

### DIFF
--- a/willie/modules/chanlogs.py
+++ b/willie/modules/chanlogs.py
@@ -36,6 +36,7 @@ def configure(config):
         config.add_option("chanlogs", "by_day", "Split log files by day", default=True)
         config.add_option("chanlogs", "privmsg", "Record private messages", default=False)
         config.add_option("chanlogs", "microseconds", "Microsecond precision", default=False)
+        config.add_option("chanlogs", "utc", "Use UTC time", default=True)
         # could ask if user wants to customize message templates,
         # but that seems unnecessary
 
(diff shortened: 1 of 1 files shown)


embolalia:
@moises-silva, can you address the above concerns, fix the merge conflicts, and make sure the build passes so we can merge this?

moises-silva:
Sorry guys, somehow I missed the initial update by @elad661 back in November (did not see the email notification).

@elad661, I thought that by adding config.add_option() and setting a default there that would set to be the default *even if not specified in the configuration explicitly by the user*. From your comment I must infer now that my assumption was wrong? 

How is the new configuration option I added any different from the existing 'by_day' option?

by_day sets the default to True, using the same API config.add_option(), then used in get_fpath() by checking if bot.config.chanlogs.by_day

Which is basically the same I am doing (though I do it in one line for convenience).

Is that configuration option also inconsistent then? what's the proper way to setup the 'real' default?

tyrope:
Python 2.7.4:
```
>>> dt = 1 if True else 2
>>> print dt
1
>>> dt = 1 if False else 2
>>> print dt
2
```
Seems like your code syntax is correct, but perhaps read incorrectly by the others? I'm uncertain.

Either way, the merge conflict is definitely stopping from this being pulled.

moises-silva:
I just had a look at master, and the new option 'localtime' is a superset of this option, so this merge request no longer applies.

Thanks!

PR #662, opened 2014-11-23
Nice help

qsantos:
A nicer and more intuitive help function

### DIFF
--- a/willie/bot.py
+++ b/willie/bot.py
@@ -306,6 +306,9 @@ class Willie(irc.Bot):
         self.callables = set()
         self.shutdown_methods = set()
 
+        # module name: list(command names)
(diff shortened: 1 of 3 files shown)


embolalia:
@qsantos are you going to address the above?

qsantos:
Sorry for the long wait. I have removed the commits that were off-topic and use a dictionnary of available commands per modules instead of a dictionnary of modules.

Also, *register() takes the list symbols as parameters while I need the module name. I have thus added *register_help() which just takes the module. An alternative would be to change *register() so that they take a module as a parameter.

embolalia:
@qsantos: Sorry for leaving this open for so long. The change has been implemented and will be in 6.0. Thanks for the contribution.

PR #667, opened 2014-11-27
Have the bot use the correct prefix in examples, even if it's hard coded

LeftyBC:
Basically this just looks at the ```@example``` decorator's text to ensure that if the prefix set in config is different or missing, it gets correctly prepended to the help text.

My users were very confused when I changed the prefix to @ but the help commands still said to use "." :)

Also, I *love* this bot.  Switching from Hubot to Willie was a huge time-saver for me!

### DIFF
--- a/willie/bot.py
+++ b/willie/bot.py
@@ -485,7 +485,17 @@ class Willie(irc.Bot):
                     else:
                         # The new format is a list of dicts.
                         example = func.example[0]["example"]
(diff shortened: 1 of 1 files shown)


embolalia:
Added config.core.help_prefix in ddb8a5421654dae31d2fbc18865c26dd21b9caf2 to let it be changed.

PR #672, opened 2014-11-30
allow find_url to parse the content directly

anarcat:
this makes find_title() a little more useful for outside tools.

### DIFF
--- a/willie/modules/url.py
+++ b/willie/modules/url.py
@@ -210,12 +210,15 @@ def check_callbacks(bot, trigger, url, run=True):
     return matched
 
 
-def find_title(url):
+def find_title(url=None, content=None):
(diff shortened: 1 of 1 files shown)


anarcat:
bump.

embolalia:
What do you mean by "outside tools"? Can you provide the use case for this?

anarcat:
the use case is for contrib module to use this. i have tried to use it in the bookie plugin for example.

maxpowa:
The import system is currently too unreliable to make this reasonable. If the module that imports url is loaded before url itself, it will fail to import. This already happens with occasionally with xkcd.py because it imports search. There's been no activity on this for over a year so I'm going to close it.

Issue #658, opened 2014-11-11
Use Python logging, and replace the debug function with it


embolalia:
It should be possible to write a handler for the built-in logging which sends to an IRC channel, which would mean we could get rid of `debug()` and use something standard instead.

PR #675, opened 2014-12-04
properly log output of willie in systemd

anarcat:
with the --quiet flag, the verbose core config setting has zero effect
and there would be no way for admins to diagnose bot startup problems.

removing the flag has no harmful effects: willie's output is dutifully
logged to systemd's journal files, according to the proper log levels
defined in the config file. since willie doesn't support syslog
logging (see #658 for that), this is pretty much the *only* way to
have logs of willie's activities and problems outside of IRC.


embolalia:
Is this still relevant since #658 is done?

anarcat:
that's pretty neat! i'll need to upgrade and test this first...

thanks!

anarcat:
@embolalia so to answer your question: yes and no!

no: it seems that output is silenced *regardless* of whether `--quiet` is used now, which is hardly what i was hoping for. :package: 

yes: since it doesn't matter what if you put `--quiet` or not, why not abstract this away from systemd?

i'm not sure how to fix logging here... it would be nice if this would still go to stdout even if started from systemd...

embolalia:
It doesn't restart if it's misconfigured, which is correct. There's no point in continuing to restart it over and over if it'll never run successfully.

anarcat:
my experience with this is that some situations would make willie disappear. i couldn't find out why (because logs don't show up in `systemctl status`), but the previously mentionned patch fixed that issue for me. it was not a configuration problem, it was a properly configured bot that was running, that would crash and not be restarted by systemd.

Issue #300, opened 2013-07-05
Quit after timeout from server


devtea:
I've run into an issue where the bot times out from the server (freenode) and on reconnecting, the bot will quit because it sees that the nick is still in use. A couple possible solutions would be to ghost the existing nick, or retry a finite number of times before disconnecting to give the server a chance to disconnect the nick.

dgw:
I suspect this will be easier to implement with the IRC backend refactor.

Exirel:
Thinking about some solutions:

* behind a config option to enable this behavior, use nickserv's ghost command when this case is detected; this would require other options to be configured (such as the `auth_*` options)
* a job that periodically check the availability of the bot's wanted nick (but maybe only if the ghost option is not set?)

The problems are:

* making sure `bot.nick` is always the relevant one
* nothing is using the nick from the config file instead of the `bot.nick`

Exirel:
So far I didn't use a job at all, and instead relied on listening to the `QUIT` event. This can be seen in #1930.

PR #678, opened 2014-12-07
try to find a new nick if previous failed

anarcat:
we really need to have a config option for the max here, but i want to
test the waters before going any further at fixing #300, since it was
proposed to only use GHOST.

those approaches should be complementary: if we fail at using GHOST
with nickserv, we should fallback to this dumb routine.

choosing a nick should probably be encapsulated in a function instead
of just being thrown in the terminator routine. i'm also unsure about
the variables names and would welcome feedback on that.

### DIFF
--- a/willie/irc.py
+++ b/willie/irc.py
@@ -444,7 +444,7 @@ class Bot(asynchat.async_chat):
             else:
                 self.nick_try += 1
             if self.nick_try >= self.nick_try_max:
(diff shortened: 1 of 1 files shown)


anarcat:
bump, anything new here? is it worth rebasing?

embolalia:
No. I don't think it's worth the complexity at this point.

PR #683, opened 2014-12-12
Weather tweaks: include time in response, reverse wind direction arrows

Winter-Stormer:
This alters the response of the .weather command so that it includes the time of the weather report.  A typical response intro will change from "Yahoo! Weather - Austin, TX:" to "Austin, TX at 11:50 pm CST:", which seems nicer and more useful to me.

This also reverses all the arrows displayed for wind direction.  Currently, for example, 337.5 - 22.5 degrees is represented by an up arrow, suggesting motion towards the north, but those numbers mean that the wind is coming *from* the north.

### DIFF
--- a/willie/modules/weather.py
+++ b/willie/modules/weather.py
@@ -144,7 +144,7 @@ def weather(bot, trigger):
     query = web.urlencode({'w': woeid, 'u': 'c'})
     url = 'http://weather.yahooapis.com/forecastrss?' + query
     parsed = feedparser.parse(url)
-    location = parsed['entries'][0]['title'].replace("Conditions for ", "")
+    location = parsed['feed']['title']
 
     cover = get_cover(parsed)
     temp = get_temp(parsed)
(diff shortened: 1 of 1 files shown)


embolalia:
I've mentioned a number of times my view toward dependencies in this project. TL;DR, until the world is on 3.4 and we can rely on every user having pip, they are to be avoided if at all possible.

Regarding formatting, I'd like to move to a more standard format across modules. Pretty much everything does some form of `[Thing] stuff | stuff | stuff`. I've created #709 for deciding on and implementing a standard. That would dictate something like `[Yahoo! Weather] For Brooklyn, NY, Sun, 11 January 2015, 10:49:00 EST | Cloudy, -2 ℃ | Humidity: 28% | Light breeze, 3.1m/s (←)`. Perhaps you could amend this to fix the direction, and make another PR once the details of #709 are settled?

Winter-Stormer:
Done; sorry about the delay.

embolalia:
Thanks!

PR #699, opened 2015-01-02
adding simple proxy support

FodT:
Added simple proxy awareness.

add proxy and proxyport entries to [core] config
e.g
proxy=www-cache.proxy.com
proxyport=80


elad661:
I don't think there's a usecase for this, and it adds quite a bit of complexity

FodT:
Well, I suppose you can run it through tsocks but this filled a usecase I had.

embolalia:
I don't see the limited use cases for this outweighing the maintenance burden. Thank you, but I don't think this is something that should be merged.

PR #705, opened 2015-01-09
add support for &bitlbee IDENTIFY.

alanpost:
model it after the NickServ support.

### DIFF
--- a/willie/coretasks.py
+++ b/willie/coretasks.py
@@ -57,6 +57,12 @@ def startup(bot, trigger):
             'IDENTIFY %s' % bot.config.core.nickserv_password
         )
 
+    if bot.config.core.bitlbee_password is not None:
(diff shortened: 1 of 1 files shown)


alanpost:
I couldn't find the documentation in the repository, so that has not been updated to reflect a new configuration parameter.  I'm happy to fix the documentation as well, if you can point me in the right direction.

embolalia:
Since it works identically to NickServ, and there are other things that NickServ can be called, I'm closing this in favor of just being able to configure what NickServ is called.

alanpost:
I think that's a fantastic idea, would you like a new patch?

tyrope:
That would be very helpful, @alanpost. Looking forward to it. ;)

embolalia:
It's already done; that's why I closed the PR.

PR #703, opened 2015-01-07
[youtube] Fixed video duration format when running on python3

flashcactus:
Compare: 
python 2: 
[YT Search] Title: Test Brytyjskich Ciast | Uploader: paei100 | **Duration: 9mins 6secs** | Uploaded: 05/01/2015, 17:49 | Views: 109,733 | Link: http://youtu.be/p9a0LgQv_Gw
python 3:
[YT Search] Title: Test Brytyjskich Ciast | Uploader: paei100 | **Duration: 0.15166666666666667hours 6secs** | Uploaded: 05/01/2015, 17:49 | Views: 109,733 | Link: http://youtu.be/p9a0LgQv_Gw

This is caused by python 3 assuming float division even when both operands are integer. Changing the division to be explicitly integer fixes the issue.

PR #706, opened 2015-01-10
[youtube] Fixed video duration format when running on python3 in 5.0

flashcactus:
This is identical to PR #703, but the fix is in the master branch as opposed to 4.6.


embolalia:
Thanks.

PR #725, opened 2015-01-28
[seen] fix #653 make .seen persist over restarts

### LINKED ISSUE #653: Make .seen persist over bot restarts
Currently, when a Willie-based bot is restarted, it loses all the info about who it saw. It is quite inconvenient, as restarts are required quite often, especially on networks where there are lots of netsplits going on and the bot loses its nick to them all the time. The proposed solution would be to keep a persistent DB containing the relevant info from which old records may or may not be auto-deleted at regular intervals.

pmrowla:
- use willie.db instead of a Ddict to store seen information for a nick

### DIFF
--- a/willie/modules/seen.py
+++ b/willie/modules/seen.py
@@ -11,10 +11,8 @@ from __future__ import unicode_literals
 
 import time
 import datetime
-from willie.tools import Ddict, Identifier, get_timezone, format_time
+from willie.tools import Identifier, get_timezone, format_time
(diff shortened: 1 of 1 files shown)


flashcactus:
Awesome. But is the .format() going to work with py2?

embolalia:
Nice, thanks!

pmrowla:
@flashcactus string.format() works in python 2. I changed the string formatting here to make it consistent with the current willie core.

PR #727, opened 2015-02-01
Adding two new modules.

robguderian:
autoadmin takes a whitelist of users and ops/voices them
passwdadmin takes messages from users, and ops or voices them
depending on the password provided.

### DIFF
--- a/willie/modules/autoadmin.py
+++ b/willie/modules/autoadmin.py
@@ -5,9 +5,9 @@ Security hole: if the users in the whitelist aren't
 protected by a nickserv, someone malicious could take your chat room
 """
 
+from __future__ import unicode_literals
 import willie
 import time
-from __future__ import unicode_literals
 
 # the whitelist is a hash of arrays
 # the key for the hash is the room, the array is the whitelist of users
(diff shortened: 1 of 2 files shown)


tyrope:
Willie was never intended to be used as a channel management bot, rather as a utility one. I'll go through the code and maybe comment here or there but I don't think this should be pulled here .Maybe in [-extras](https://github.com/embolalia/willie-extras). I'll leave that up to @embolalia.

One comment (as you can see by the failing build) is that the modules failed style check. Although functionally unimportant, please do change it to hold the Willie codebase up to standards.

robguderian:
I added rate-limiting to the the passwdadmin module.
I disagree about the comment about 'utility'. Autoadmin and old-school password admin are 2 utility functions that most IRC bots have, and one I was surprised to not find here. I agree they could be moved to -extra, but it seems like a feature that any/all IRC bots should have.
The 1 password thing is secure out-of-the-box. The admin chooses to add the master key.

embolalia:
NAK. Willie is not network services. This is fine in -extras, but does not belong in the main project.

PR #737, opened 2015-02-10
Add the option to use the bot's preferred timezone for logs.

izzy84075:
This adds an option to use the clock module's timezone for storing logs.

I'm not sure if this is done properly as I'm not very well versed in Python, but I figured I'd give it a shot since everybody in my channel is in America/Los_Angeles and was complaining about logs being split in the middle of when we tend to be active. It appears to work properly, when pytz is installed, and falls back to the old behaviour when it isn't.

### DIFF
--- a/willie/modules/chanlogs.py
+++ b/willie/modules/chanlogs.py
@@ -46,18 +46,20 @@ def configure(config):
         # could ask if user wants to customize message templates,
         # but that seems unnecessary
 
+
 def get_datetime(bot):
     """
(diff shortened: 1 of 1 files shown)


izzy84075:
I'm not sure where the remaining style errors are coming from. Running checkstyle.sh locally turns up nothing, and I didn't touch any of the files Travis is complaining about.

tyrope:
Don't worry about those style errors. They're from the master code.

PR #749, opened 2015-02-20
[trigger] Fix CTCP -> IRCv3 Intent conversion

maxpowa:
CTCP messages that are just single word messages weren't properly being converted into intents, because the current system was expecting a following section. Some CTCP messages (TIME, USERINFO) wouldn't usually be sent with trailing information, causing the old regex to miss converting them into intents.


maxpowa:
Probably will be a good idea to add a decorator for intent too, I will work on that.

PR #763, opened 2015-03-21
Write unicode to ascii log file

rogosher:
Fix for a possible bug involving logging of unicode characters to ascii text files.


rogosher:
Eh probably could come up with something better than this, but it at least stops the error from being output

embolalia:
That's exactly what it should be doing. Thanks.

PR #771, opened 2015-04-05
Fix unquiet.

harrowed:
Unquiet was using op instead of channel, so to unquiet you were requried to .unquiet channel name, instead of .unquiet name

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -7,3 +7,4 @@ praw
 feedparser
 pyenchant
 pep8==1.4.6
+urllib3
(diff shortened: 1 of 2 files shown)


tyrope:
The original behaviour sounds more like the intended behaviour, as you can unquiet people outside of the channel they've been quieted in. I'd like opinions on this.

harrowed:
If that were the case, then quiet would also have this functionality, would it not?

tyrope:
You raise a valid point.

Regarding 6bf52a0, why'd you add urllib3 to the dev-requirements to this PR?

harrowed:
Travis was failing on 2.7 due to a dummy package. Was hoping it would resolve to allow for a pass.
Evil Reddit package :(

embolalia:
This is now merged on 6.0.

Issue #119, opened 2012-09-14
Additional Reddit features


embolalia:
I made note of these when writing the module, but forgot about them:

* Allow per-channel setting to kick on NSFW tag
* Add post creation time
* Detect user's cake day

elad661:
Almost 3 years later, closing this issue as it was fixed by pull request #775

9cf2dfdf7140790eb965ef7e7cf30057d9755df1

PR #775, opened 2015-04-09
[WIP] Additional reddit features, addressing Issue #119

whonut:
I've implemented creation time inclusion and cake day detection, as mentioned in Issue #119. I've also made the module more PEP8 compliant and added an example to .redditor.

Cake days present a bit of a problem. Their exact implementation on reddit is explained here: http://www.reddit.com/r/AdviceAnimals/comments/1v6xam/seriously/cepjnfg (I linked the submission, not the comment in commit 2dbaf77). The API doesn't make accessible a user's last login time, so we cannot delay cake days as reddit.com does. Unfortunately this means that users may have the cake day flair on reddit.com but not be labelled as having their cake day by Willie, or vice versa.

I have yet to add the 'kick on NSFW tag' feature from #119 as I was unsure of exactly what was meant. If @embolalia could clarify their intentions, I'd be happy to work on it.

Feedback on the quality of my code, commits etc. would be greatly appreciated.

### DIFF
--- a/willie/modules/reddit.py
+++ b/willie/modules/reddit.py
@@ -103,15 +103,32 @@ def redditor_info(bot, trigger, match=None):
         # Fail silently if it wasn't an explicit command.
 
     message = '[REDDITOR] ' + u.name
+    now = dt.datetime.utcnow()
+    cakeday_start = dt.datetime.utcfromtimestamp(u.created_utc)
+    cakeday_start = cakeday_start.replace(year=now.year)
+    day = dt.timedelta(days=1)
+    year_div_by_400 = now.year % 400 == 0
+    year_div_by_100 = now.year % 100 == 0
+    year_div_by_4 = now.year % 4 == 0
+    is_leap = year_div_by_400 or ((not year_div_by_100) and year_div_by_4)
+    if (not is_leap) and ((cakeday_start.month, cakeday_start.day) == (2, 29)):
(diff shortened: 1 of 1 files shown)


embolalia:
At a quick glance, this all looks good. I'm not too worried about the precision of cake days; what you have is close enough. (Especially since it's documented!)

As for the NSFW kick, I was thinking basically a `.setsfw` or some such thing (we really need guidelines for command names…) would create an entry in the db for the channel. When an NSFW link was posted, the module would check if that setting was enabled for the channel and, if it was, kick the user who posted the reddit link. (I also liked the suggestion in #119 of not showing the link from the submission.)

This is good work (on an issue that has been open for an embarrassingly long time). I'll try it out a bit, maybe this weekend, and then merge it. Thanks for the contribution. ^_^

whonut:
Thanks very much. Do you have a preference for the default value of the SFW 'flag'? The way I have it currently it would default to false because the db returns None for a non-existent entry.

If you want it to default to true, I think it'd be best to write a helper that returns True if there's no value set for SFW. Then you'd just always use that to check the value. I assume you'd want something like that in willie.tools?

whonut:
Apologies. Commit 25d4a51 also contains the getsfw command. I messed up my staging.

embolalia:
Great work. Thanks!

PR #776, opened 2015-04-10
Do not check if running as root

Stealthii:
This commit removes the checks for uid != 0, which prevent running
willie in environments without users (such as Docker, or embedded
systems)

### DIFF
--- a/willie.py
+++ b/willie.py
@@ -91,18 +91,6 @@ def main(argv=None):
                             dest="version", help="Show version number and exit")
         opts = parser.parse_args()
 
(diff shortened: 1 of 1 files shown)


embolalia:
Many Willie users are unaware of the security implications of running things as root. Willie is far from secure, which makes doing so *extremely* dangerous. Checking that the uid is non-zero is a very common security check, and I would be shocked if Docker doesn't have some accommodation for it (and I'm pretty sure I've heard of people running Willie on Docker, so I imagine it does). Regardless, explicit support for Docker is not a priority for this project. WONTFIX

PR #783, opened 2015-04-16
Various fixes and improvements

matthijskooijman:
Here's a bunch of commits I created while setting up my bot. It's a mix of bugfixes and small improvements, some of which I actually needed and some of which I just noticed going over the code.

This PR is a bit of everything, but if it would help to separate some of these changes into their own PRs (e.g. to separate the easy commits from ones that need a bit of discussion), let me know.

PR #786, opened 2015-04-16
Drop locking methods from WillieMemory

matthijskooijman:
Drop some useless locking methods on WillieMemory and WillieMemoryWithDefault. Instead of these methods, the memory.lock object can be used directly.

### DIFF
--- a/willie/tools/__init__.py
+++ b/willie/tools/__init__.py
@@ -324,11 +324,6 @@ class WillieMemory(dict):
         """Backwards compatability with 3.x, use `in` operator instead."""
         return self.__contains__(key)
 
-    def unlock(self):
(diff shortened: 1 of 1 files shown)


matthijskooijman:
In #783, where these commits originally lived, @embolalia said:

> the memory changes have the potential to be breaking and might need to be held off until 6.0.

This is probably true, though the lock() method that was removed didn't actually work (calling it would try to call the lock object, not the method). And without a working lock() method, it seems unlikely that people are using the unlock() method right now. Still, might be good form to hold this off. I don't actually rely on this change, it's just something I came across.

embolalia:
This is now merged on 6.0.

PR #791, opened 2015-04-19
Tweaks to nice help PR (#662) so we can finally merge it

firerogue:
In addition to @qsantos' original commits, I did the following things to hopefully make it mergeable:
 * Code edits to improve readability
 * Tweaked functions output
 * Remove the 'others' category in output of all commands
 * Make commands an alias for help because its functionality was being replicated but done better in help.

### DIFF
--- a/willie/modules/help.py
+++ b/willie/modules/help.py
@@ -12,6 +12,8 @@ from __future__ import unicode_literals
 from willie.module import commands, rule, example, priority
 from willie.tools import iterkeys
 
+import textwrap
+
 
 def setup(bot=None):
     if not bot:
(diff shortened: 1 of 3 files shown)


willie/modules/help.py
```diff
@@ -44,34 +65,21 @@ def help(bot, trigger):
...
 @rule('$nick' r'(?i)help(?:[?!]+)?$')
 @priority('low')
 def help2(bot, trigger):
     response = (
-        'Hi, I\'m a bot. Say ".commands" to me in private for a list ' +
+        'Hi, I\'m a bot. Say "%scommands" to me in private for a list ' +
         'of my commands, or see http://willie.dftba.net for more ' +
         'general details. My owner is %s.'
-    ) % bot.config.owner
+    ) % (bot.config.core.help_prefix, bot.config.owner)
```
tyrope: Shouldn't this be bot.config.core.owner?

firerogue: Yes it should. Thanks for catching it

PR #793, opened 2015-04-22
[coretasks] Make the anti spam module aware of passing time (ish)

firerogue:
The recent issue opened by the one who wanted to have willie say the same thing every 15 seconds reminded me of a previous concern I had regarding willie's anti-spam detection. Namely, that if something was said 5 times 10 hours ago and nothing has been said since, the next time it is said willie will print '...'. At that point its not spam protection, its just annoying. This clips the last message willie remembers for each channel every 10 minutes (this should be changed to a more useful number) to prevent that without affecting spam detection too much

### DIFF
--- a/willie/coretasks.py
+++ b/willie/coretasks.py
@@ -502,3 +502,16 @@ def blocks(bot, trigger):
             return
     else:
         bot.reply(STRINGS['huh'])
+
+@willie.module.interval(600)
+def flood_decay(bot):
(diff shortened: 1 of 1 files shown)


firerogue:
Related questions because the answers could be folded into this PR:
A. Why does bot.stack use lists and not deques?
B. Why do you clip to 10 messages if only the last 8 are used for spam detection?

firerogue:
Missed the line where this does happen, but the questions still stand

PR #806, opened 2015-05-19
Allow willie to be run as root

Cadair:
I was trying to run willie from inside a docker container and the uid check caught me out. This gives an option to turn it off.


elad661:
Don't run willie as root.

root on the container means root on the host, so using docker is no excuse either.

http://www.slideshare.net/jpetazzo/docker-linux-containers-lxc-and-security
http://mjg59.dreamwidth.org/33170.html

Closing.

embolalia:
See also #776.

Cadair:
Thanks for the info.

Issue #824, opened 2015-06-25
Long-running commands not rate-limited until after completion


dgw:
In the very exact opposite of the problem described in #293, I have a callable ([see code](https://github.com/dgw/BombBot/blob/b5e3d72ec79bb3e65e39a1b9b66ff6aae51f42c6/bomb.py#L24)) that isn't being rate-limited even though it's decorated with `@rate(600)` and is being called by users not present in `config.core.admins`.

While some code branches do return `NOLIMIT`, the main branch of the callable has no return statement and shouldn't be bypassing rate-limiting.

This was observed with willie 5.3.0, freshly pulled from pip just in case (`sudo pip install --upgrade willie`).

dgw:
After further investigation and debugging, it seems the rate-limiting does kick in eventually.

My current theory is that using `sched` blocks the completion of the callable. Per [this code in bot.py](https://github.com/dgw/BombBot/blob/b5e3d72ec79bb3e65e39a1b9b66ff6aae51f42c6/bomb.py#L24) the last-used time isn't updated until the callable returns. Therefore, until the callable returns, it can be run any number of times by the same user.

embolalia:
I don't know that there's a good solution to this. We need to wait on the return value in order to stop limiting. In the vast majority of cases, it's not a problem. You're right, `sched` is blocking here, and causing your problems. There's probably a better way to do that, whereas there's probably not a better way to do this.

dgw:
Yeah, I fixed my particular case now that I know what's causing it (switched to threading.Timer()).

There really isn't a good way I can think of to fix it in willie core. And as you said, it's irrelevant to the vast majority of callables, so if you want to just close this and ignore it that's fine. I'll try to find a good place in the wiki to point it out, though.

PR #825, opened 2015-06-25
Add note to docs about using sched with @rate

dgw:


dgw:
I'm revisiting this from the future to note that it was spurred by behavior described in #824.

This note seems not to apply any more after the rule-system rewrite for 7.x; the callable's start time is now used for rate-limiting if it hasn't completed yet:

https://github.com/sopel-irc/sopel/blob/1c6aeb93a269a3997a9d55ec102a9ed77e393402/sopel/plugins/rules.py#L485-L492

<details>
<summary>Quick and dirty plugin to test <tt>sched</tt>'s behavior with rate-limiting in 8.0</summary>

```py
import sched
import time

from sopel import plugin


s = sched.scheduler(time.time, time.sleep)


@plugin.rate(60)
@plugin.command('testrate')
def rate_limited(bot, trigger):
    s.enter(30, 2, bot.say, argument=('hi',))
    s.run()
```

</details>

PR #826, opened 2015-06-28
Merge nickname IDs when merging nick groups

dgw:
Also fixed broken test for merge_nick_groups() which allowed this bug to persist.

PR'd per @embolalia's response on IRC.

### DIFF
--- a/test/test_db.py
+++ b/test/test_db.py
@@ -212,8 +212,9 @@ def test_merge_nick_groups(db):
 
     db.merge_nick_groups(aliases[0], aliases[1])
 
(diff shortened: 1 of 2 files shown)


dgw:
Don't know why the Travis CI build failed, but I didn't touch anything related to JSON. Given that #825 (which changed comments only) also failed, I'm guessing it wasn't me.

embolalia:
A PR with a test is the best! :heart: (And yeah, the build failure is Google being a :hankey: and the build not accounting for it; not a problem)

PR #842, opened 2015-07-17
[config] Try to parse ListAttribute as list before excepting (fix #841)

### LINKED ISSUE #841: [admin] 5.4.1 can't set admins list from IRC
From PM:
````
<dgw> .set admins dgw,list,of,other,nicks
<willie> ValueError: ListAttribute value must be a list. (file
  "/usr/local/lib/python2.7/dist-packages/willie/config/types.py",
  line 224, in serialize)
````

dgw:
ListAttribute.serialize() should try to parse the value it is given into a list before giving up. This fixes #841 and allows the admin module to properly set list values like `core.admins`.

### DIFF
--- a/willie/config/types.py
+++ b/willie/config/types.py
@@ -221,7 +221,9 @@ class ListAttribute(BaseValidated):
 
     def serialize(self, value):
         if not isinstance(value, list):
-            raise ValueError('ListAttribute value must be a list.')
+            value = self.parse(value)
+            if not isinstance(value, list):
+                raise ValueError('ListAttribute value must be a list.')
         return ','.join(value)
 
     def configure(self, prompt, default):


embolalia:
The point of `serialize` is to take an object and make it into a string for the config file. If the object is not appropriate, it should not be accepted; the exception is intentional. The actual issue with #841 is that it does not parse what's given in the command first. That is to say it should parse what's given in the command (with exception handling to give a good error message if it's invalid), and then set the config value to the object that the parsing returns (whereupon serialize will turn it back into a string).

dgw:
I'll look into this for real tonight. Sorry, I misunderstood what the code was doing. Expect a real PR within the next day or two.

PR #845, opened 2015-07-21
Treat error in connect as a disconnect

fatalis:
This is so that Willie attempts to reconnect when your network goes down, instead of exiting.


fatalis:
updated to latest code

PR #850, opened 2015-08-01
Fix None saved to config being loaded as 'None'

balu92:


embolalia:
This would not be good, because it would mean you couldn't store the string None in the config intentionally. I've pushed a better solution.

balu92:
Yep, that was the alternate solution i thought about, but I didn't see a case, when I'd wanted to store None as a string. Another alternate solution would have been to 'encrypt' None to something like %NoneType%, that has pretty much zero chance that someone would want to store as a string, and when parsing: if setting is "%NoneType%": return None elif: return setting.

PR #866, opened 2015-08-21
[adminchannel] Fix .kick & .kickban

dgw:
[adminchannel] Fix .kick & .kickban

1. .kick reason was being truncated
2. .kickban didn't work because local var channel was used before assignment


embolalia:
Nice catch! Thanks!

Issue #872, opened 2015-08-30
ListAttribute should be striped


Deedasmi:
```extra=/test, /test2```

Will fail and cause sopel to not launch, while:

```extra=/test,/test2 ```

Will not.

tyrope:
Although this will fix the problem in `[core]extra=`, it could introduce a new problem to (the unintended feature) `[core]channel=#channel,#foo password,#bar` unless executed properly.

Deedasmi:
This is my first time contributing to a project, and I haven't totally figured out how Sopel works on the inside yet, so.... Do you have any suggestions on a better/safer way to fix it? It just makes sense to me that the config file should not be dependent on (the lack of) whitespace.

embolalia:
This wouldn't harm the channel password thing (since it only removes from the ends of each item), but it would make it impossible to have list items that have spaces. It'll have to be optional, that is `ListAttribute` takes an argument `strip`, and stripping is only done if it's `True` (which I think can be the default).

Deedasmi:
> "would make it impossible to have list items that have spaces."

I don't understand this part. As you mention in the first line, this is not a str.replace(" ", ""). It will simply remove every space immediately before and after a comma, which in a comma separated string should be pretty harmless.

If you'd still prefer the default argument, I can go ahead and update the pull request in a bit.

embolalia:
Sorry, I meant list items that start or end with a space. It's an edge case, but handling edge cases well is important.

Deedasmi:
#873 now has a default parameter.

PR #873, opened 2015-08-30
Added a .strip() to ListAttributes

Deedasmi:
Potential fix for #872. Not thoroughly tested, but didn't notice any issues or logged errors.


embolalia:
The argument needs to be to the type's constructor; `parse` won't get an argument like that. I've merged the idea, though. Thanks.

PR #876, opened 2015-09-05
[ping] Stop Sopel from replying rudely to the bot's owner.

DMzda:
Stop Sopel from replying rudely to the bot's owner in ping.py:

    <@dmzda> hi dmbot 
    <dmbot> Fuck off, dmzda

### DIFF
--- a/sopel/modules/ping.py
+++ b/sopel/modules/ping.py
@@ -12,10 +12,7 @@ from sopel.module import rule, priority, thread
 
 @rule(r'(?i)(hi|hello|hey),? $nickname[ \t]*$')
 def hello(bot, trigger):
-    if trigger.owner:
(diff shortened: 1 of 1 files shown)


dgw:
Could be a setting, but the current behavior is amusing enough that I'd rather not see it gone. :-1:

embolalia:
The number of people who would bother configuring it, rather than just replacing the module, is rather small, I think. And while it might be amusing to some people, it's annoying to others. It doesn't belong in the default module set.

Thanks for the PR, @DMzda.

DMzda:
I had written the code to make this a config option called `be_nice` before I saw that this was merged, is that something you're interested in @embolalia ?

embolalia:
I don't think it would be worthwhile to include in the default module set. If you want to make your own module for that, feel free, but I don't think we'll pull it in.

PR #894, opened 2015-09-17
SQLAlchemy Compat for db.get_uri()

cerealcable:
SQLAlchemy requires that sqlite absolute paths begin with three forward slashes not just two. This fixes an expectation that opening a file is relative to root instead of absolute to root (which self.filename is already checked to be absolute).

### DIFF
--- a/sopel/db.py
+++ b/sopel/db.py
@@ -99,7 +99,7 @@ class SopelDB(object):
     def get_uri(self):
         """Returns a URL for the database, usable to connect with SQLAlchemy.
         """
-        return 'sqlite://{}'.format(self.filename)
+        return 'sqlite:///{}'.format(self.filename)
 
     # NICK FUNCTIONS
 


embolalia:
Hm, I think the real solution here is to make the `db_filename` config attribute a `FilenameAttribute`. I'm not sure why [it isn't already](https://github.com/sopel-irc/sopel/blob/master/sopel/config/core_section.py#L76)…

cerealcable:
Why not both? db.get_uri() is returning data that isn't what is expected so that's a problem.

Though in fairness I'm having concurrency issues with sqlite so I'm going to have to do something else anyways. Not sure what my plan is, but it definitely involved dropping sqlite in favor of something that can properly handle threading without entire database locking to do most operations.

Would you be in favor of an option to move to SQLAlchemy which could in theory support multiple database types, and default to sqlite?

dgw:
I am going to merge this. Nothing in Sopel's core uses this function, admittedly, so it's tricky to test. But my research on how DB URIs are expected to look says this is the correct thing to do, so if it ends up causing problems in one of the (likely few) third-party modules using this then we'll adapt quickly.

PR #898, opened 2015-09-25
Fix the config section lookup and help URL for meetbot

jamesmcdonald:
Couple of small fixes. Running .startmeeting threw:
 AttributeError: 'Config' object has no attribute 'has_section'

I have changed the lookup to use hasattr instead, which works correctly with the Config object.

The help URL was still pointing to https://github.com/embolalia/sopel/wiki/Using-the-meetbot-module, so I updated it to the new location.

### DIFF
--- a/sopel/modules/meetbot.py
+++ b/sopel/modules/meetbot.py
@@ -148,7 +148,7 @@ def ischair(nick, channel):
 def startmeeting(bot, trigger):
     """
     Start a meeting.
-    https://github.com/embolalia/sopel/wiki/Using-the-meetbot-module
+    https://github.com/sopel-irc/sopel/wiki/Using-the-meetbot-module
     """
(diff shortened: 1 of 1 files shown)


jamesmcdonald:
That doesn't actually do quite what I intended; it uses the defaults if there's no section in the config file. I'll leave this open for now, as there's still a bug here and I'm not sure what behaviour you intend.

embolalia:
It using the defaults is the intended behavior. I think that hasattr will always return true, because the section gets defined regardless of whether there are values set. That bit about it not being configured can just be dropped.

zoell:
Is this fixed? If I change has_section to hasattr in my version the error message disappeared but nothing is happening.

jamesmcdonald:
Closing this because these issues have been resolved already.

PR #904, opened 2015-10-01
Fix safety module configuration

Ohcanep:
Configuring the module would previously bomb out with a TypeError because `default` was actually passed as `serialize`.


embolalia:
Thanks!

PR #908, opened 2015-10-05
make message rate configurable, including burst mode

larsks:
This adds several knobs that control how fast Sopel will send multiple
messages to the same recipient:
- `bucket_burst_tokens` -- this controls how many messages may be sent
  in "burst mode", i.e., without any inter-message delay.
- `bucket_refill_rate` -- once exhausted, burst tokens are refilled at a
  rate of `bucket_refill_rate` tokens/second.
- `bucket_empty_wait` -- how long to wait between sending messages when
  not in burst mode.

This permits a bot to return a few lines of information quickly, while
still engaging rate limiting to prevent flood protection from kicking
in.

How it works:

When sending to a new recipient, we initialize a token counter to
`bucket_burst_tokens`.  Every time we send a message, we decrement this
by 1.  If the token counter reaches 0, we engage the rate limiting
behavior (which is identical to the existing code, except that the
minimum wait of 0.7 seconds is now configurable).

When sending a new message to an existing recipient, we check if the
token counter is exhausted.  If it is, we refill it based on the elapsed
time since the last message and the value of `bucket_refill_rate`, and
then proceed as described above.

### DIFF
--- a/sopel/config/core_section.py
+++ b/sopel/config/core_section.py
@@ -183,3 +183,16 @@ class CoreSection(StaticSection):
 
     verify_ssl = ValidatedAttribute('verify_ssl', bool, default=True)
     """Whether to require a trusted SSL certificate for SSL connections."""
+
+    bucket_burst_tokens = ValidatedAttribute('bucket_burst_tokens', int,
+                                             default=4)
+    """How many messages can be sent in burst mode."""
+
+    bucket_refill_rate = ValidatedAttribute('bucket_refill_rate', int,
+                                            default=1)
+    """How many tokens/second to add to the token bucket."""
+
+    bucket_empty_wait = ValidatedAttribute('bucket_empty_wait', float,
+                                           default=0.7)
+    """How long to wait before sending a messaging when not in burst
+    mode."""
--- a/sopel/irc.py
+++ b/sopel/irc.py
@@ -418,18 +418,28 @@ class Bot(asynchat.async_chat):
 
             recipient_id = Identifier(recipient)
 
-            if recipient_id not in self.stack:
-                self.stack[recipient_id] = []
-            elif self.stack[recipient_id]:
-                elapsed = time.time() - self.stack[recipient_id][-1][0]
-                if elapsed < 3:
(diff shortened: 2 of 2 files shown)


maxpowa:
Using this change on my test bot -- works very well, have yet to encounter any issues.

thomastanck:
If you're still active, could you resolve the new merge conflicts that have popped up? +1

larsks:
I'll take a look.  It looks as if in this case "merge conflict" means "all the code my patch touched has been relocated".

embolalia:
I'm really hesitant to touch the rate limiting code without doing some serious testing on it first. This is getting discussed in #952. It looks like this is a replacement of the existing PRIVMSG limit system following that model, which is good. When I get the chance, I'd like to run some stress tests with it on a bunch of networks to see how it holds up in practice.

I'm definitely okay with merging this system for just PRIVMSG now, and then figuring out how to pull it in for everything else later. However, I don't really like having configs for these internals. There's no way for a user to know what to put for them, whether they've set it right, and whether their configuration will lead to unexpected and confusing errors later on. Moreover, I think it's highly unlikely that (assuming the stress tests work out with the defaults), anyone would ever really change these (let alone to something that wouldn't be quietly broken).

larsks:
I'm closing this old PR.  Cheers!

dgw:
@larsks Would you like to continue working on this, or should someone else take over? This commit (f58fa9f) could be @deathbybandaid's starting point for #1342, if you're done with it. (Anything built on this will include the existing commit & authorship info, I'll make sure of it. :smile_cat:)

If you're not done with it, feel free to reopen. I took over maintaining Sopel a few months ago, so if you closed this because it stalled, it can be un-stalled now.

dgw:
I'm going to reopen this just so we don't lose track of it, since closed PRs basically disappear into a black hole unless you know exactly what you're looking for.

If someone wants to replace/extend it, we'll label this appropriately and close it again at that time.

deathbybandaid:
I think this one could be closed since #1518 takes care of this.

dgw:
Definitely. I just forgot to close it and add the "Replaced" label.

PR #910, opened 2015-10-09
[help.py] Sort output alphabetically by module

maxpowa:
Minor tweak, but it makes a ton of difference when you have a lot of modules loaded.


embolalia:
Nice, thanks!

PR #912, opened 2015-10-14
Fix blocking functionality

maxpowa:
Every function is currently forced unblockable, as the unblockable decorator doesn't allow a boolean value, and the default, when there is no decorator, is True. This PR fixes that behavior by making the default, when there is no decorator, False.


embolalia:
Yikes; good catch!

Issue #885, opened 2015-09-12
ValueError: No JSON object could be decoded


stopspazzing:
Command: 

```
.g #
```

ValueError: No JSON object could be decoded (file "/opt/rh/python33/root/usr/lib64/python3.3/json/decoder.py", line 370, in raw_decode)

PR #913, opened 2015-10-15
A quick fix to issue #885

matiasv:
Issue #885 (https://github.com/sopel-irc/sopel/issues/885) was caused by the fact that the duckgogo query wasn't urlencoded previously which caused the API to return its HTML documentation page instead of a JSON document when the query contained characters that should be urlencoded.

I did a quick fix to the issue by urlencoding the query string with urllib.parse module's quote_plus() in /modules/search.py duck_api function. Now it doesn't crash on queries such as "#" or "100% sugar-free".

### DIFF
--- a/sopel/modules/search.py
+++ b/sopel/modules/search.py
@@ -54,7 +54,7 @@ def duck_api(query):
     if '!bang' in query.lower():
         return 'https://duckduckgo.com/bang.html'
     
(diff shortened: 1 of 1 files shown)


maxpowa:
For py 2.7 compat, just switch your import to 
```
import sys
    from urllib import quote_plus
else:
    from urllib.parse import quote_plus
```

maxpowa:
+1

embolalia:
Awesome, thanks!

PR #914, opened 2015-10-16
Ensure help output matches the example correctly

maxpowa:
When populating the functions' doc value, we were incorrectly checking if it started with just `prefix` instead of `help_prefix`. `prefix` is usually regex, so it's better to use `help_prefix`


BrianGallew:
Thanks, this looks like it will improve the situation.

BrianGallew:
+1

ddstreet:
This commit doesn't fully fix the problem, I just added a pull request that includes a complete fix.

PR #916, opened 2015-10-19
fix handling of help_prefix

ddstreet:
Currently any .help requests for command details will show examples with the first
character replaced by the configured prefix; by default, \.

e.g.:
<ddstreet> .help rand
<sopel> ddstreet: Replies with a random number between first and second argument.
<sopel> ddstreet: e.g. \.and 10 99

That is due to 2 things:

1. the local help_prefix var is set to config.core.prefix instead of
   config.core.help_prefix.

2. the first character in the example string is replaced by the help_prefix;
   instead the help_prefix should be prefixed to the example string.

This sets help_prefix correctly, adds the help_prefix to example correctly,
and uses example.startswith(help_prefix) instead of only comparing the first
character of example to help_prefix.

### DIFF
--- a/sopel/loader.py
+++ b/sopel/loader.py
@@ -142,7 +142,7 @@ def clean_callable(func, config):
     puts them in func._docs, and sets defaults"""
     nick = config.core.nick
     prefix = config.core.prefix
-    help_prefix = config.core.prefix
+    help_prefix = config.core.help_prefix
     func._docs = {}
     doc = trim_docstring(func.__doc__)
     example = None
@@ -173,8 +173,8 @@ def clean_callable(func, config):
         if hasattr(func, 'example'):
             example = func.example[0]["example"]
             example = example.replace('$nickname', nick)
-            if example[0] != help_prefix:
-                example = help_prefix + example[len(help_prefix):]
+            if not example.startswith(help_prefix):
+                example = help_prefix + example
         if doc or example:
             for command in func.commands:
                 func._docs[command] = (doc, example)


maxpowa:
This still doesn't fix the issue correctly. `@example('!cmd arg')` would still have incorrect output, assuming your help_prefix is .

That is the reason why it was replacing the first char regardless.

ddstreet:
What's !cmd for?  I don't see anything in the code that would do anything with that.

It's also generally broken regarding $nickname examples; any examples using ('$nickname cmd arg') will get prefixed too, and switching the $nickname replacement to before the prefixing won't help as help_prefix is single-mindedly expecting only non-nickname examples; even if help_prefix is set to $nickname and the ordering is fixed, that will prefix "$nickname" to all regular commands.  I was going to fix that by removing the help_prefix checking entirely when the example had a leading $nickname, but decided to keep this change simple.

maxpowa:
It was an example of how the example decorator can be used. The example
decorator can be used with any prefix, neither solution (mine or yours)
fully corrects the issue.

ddstreet:
Wait, you're saying that if prefix is set to \! and help_prefix is set to . it won't work right?  Uh yeah of course it won't.  help_prefix needs to be set to the non-regex equivalent of prefix.  Maybe sopel config check should verify that re.match(prefix, help_prefix).group() == help_prefix.  Getting them to agree has nothing to do with the code in loader though.

And all the examples using hardcoded . prefixes is wrong too.  Those should not include any prefix; the actual prefix will be prefixed to the example.

maxpowa:
Not really, you misunderstand. `@example` _should_ include prefix, otherwise you're breaking the test functionality that comes along with the `@example` decorator.

embolalia:
I don't know that `!` in the `@example` will work properly; I haven't tested it. But having the `.` prefix in the `@example` is definitely intended; that is not incorrect. The docs aren't clear about that, but it was the intent. I've merged @maxpowa's PR, so I'm closing this one.

PR #923, opened 2015-10-29
Replace lxml with a pure python solution

maxpowa:
lxml is one of the harder parts of Sopel to get running correctly, and it's really a lot more than what's needed for the situations it has been used in.


maxpowa:
If someone could test the bugzilla changes, that'd be great. I don't personally use the bugzilla module so I have no clue if the changes will actually work.

embolalia:
This is amazing. I'd wanted to do something like this for a while (I'm not sure why there's not an issue for it). This is going to help people sooo much with installing the bot. Really, thank you so much for doing this.

PR #924, opened 2015-10-30
Fixed page not found error message to main chat. Strips invalid chars…

crtrmxwll:
Users often surround URLs in brackets. Most commonly, I see <code>< ></code> used. Sopel includes the <code>></code> character at the end of the URL string and in some cases, it causes a Page Not Found error in main chat instead when it should announce the page title. This change strips common invalid characters from the end of the URL before it asks for the title. 

An example of a URL enclosed in <code>< ></code> that this would fix:
http://www.joelonsoftware.com/articles/ThePerilsofJavaSchools.html

### DIFF
--- a/sopel/modules/seen.py
+++ b/sopel/modules/seen.py
@@ -35,12 +35,12 @@ def seen(bot, trigger):
         timestamp = format_time(bot.db, bot.config, tz, trigger.nick,
                                 trigger.sender, saw)
 
-        msg = "I last saw {} at {}".format(nick, timestamp)
+        msg = "I last saw {} at {} UTC".format(nick, timestamp) #UTC is hardcoded. Careful!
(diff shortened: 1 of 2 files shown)


maxpowa:
Could cause issues with URLs that are supposed to end with those characters though. Some of those characters you gave *should* be URL encoded, but not all of them have to be. It's perfectly legal to have `<>|{}^` in your URL, unencoded. Maybe attempt both? If the first attempt (all chars) 404s, try with the stripped chars? Of course, that only works if the site doesn't redirect you to a 'Page not found' page instead of surfacing the 404.

yoosi:
Ah, yeah that makes sense. At this time I don't have the necessary skills to tackle your suggestion. Is someone else up for the task?

PR #925, opened 2015-10-30
update movie.py to use omdbapi as imdbapi is non-functional

reverieeee:
it seems the imdb api has been taken down entirely- there is no mention of it on their site and any connections to the site time out. [omdb](http://www.omdbapi.com/) provides the exact same functionality, and i have tested these changes and confirmed them working on my own installation.

### DIFF
--- a/sopel/modules/movie.py
+++ b/sopel/modules/movie.py
@@ -35,7 +35,7 @@ def movie(bot, trigger):
             LOGGER.warning(
                 'Got an error from the OMDb api, search phrase was %s; data was %s',
                 word, str(data))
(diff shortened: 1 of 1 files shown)


dasu:
Just noticed this over the weekend as well, can confirm these changes also work on my end.

blancNoir:
Noticed this as well. I don't think the original url was official based on comments in the following stackoverflow page. Here's what appears to be a statement about the name change by the author as well as description of the official api in an answer above.

http://stackoverflow.com/a/4269579

embolalia:
I hadn't noticed this. Thanks!

PR #927, opened 2015-11-02
Resolve #926

### LINKED ISSUE #926: Remove feedparser dependency
The weather module needlessly uses `feedparser` for some things, which adds a needless (python3-incompatible) dependency. It should be done with straight XML processing, instead.

maxpowa:
Also cleaned up dict value retrieval a little, the .get() calls were a bit unnecessary.

### DIFF
--- a/contrib/rpm/willie.spec.in
+++ b/contrib/rpm/willie.spec.in
@@ -17,7 +17,6 @@ BuildRequires:  dos2unix
 BuildRequires: systemd
 
 Requires:       pytz
-Requires:       python-feedparser
 Requires:       python-enchant
 Requires:       pyOpenSSL
 Requires:       python-praw
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,3 @@
-feedparser
 xmltodict
 pytz
 praw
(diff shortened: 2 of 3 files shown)


dgw:
Should probably be squashed into a single commit that says "Removed feedparser dependency (closes #926)" or something like that. But :+1:

maxpowa:
But muh commiting from github web

dgw:
That sounds painful lol

embolalia:
The `get` calls are necessary, because the result will not always have those keys in it. `get` will return `None`, rather than raising a `KeyError`.

maxpowa:
Ah good call, didn't even think about that. Should I edit this PR or create a new one to add them back in?

embolalia:
You can just edit this one. Thanks!

embolalia:
I got this exception while testing this with `.weather new york, ny`:
```
  File "/usr/local/lib/python3.4/dist-packages/xmltodict.py", line 246, in parse
    parser.ParseFile(xml_input)
TypeError: argument must have 'read' attribute

During handling of the above exception, another exception occurred:

  File "/home/embo/src/willie/sopel/bot.py", line 262, in call
  File "/home/embo/src/willie/sopel/modules/weather.py", line 147, in weather
    parsed = xmltodict.parse(url).get('rss')
  File "/usr/local/lib/python3.4/dist-packages/xmltodict.py", line 248, in parse
    parser.Parse(xml_input, True)
xml.parsers.expat.ExpatError: not well-formed (invalid token): line 1, column 5
```

maxpowa:
Yeaaaah, don't ask how that one slipped through.

embolalia:
Thanks!

PR #931, opened 2015-11-08
Resolve #929

### LINKED ISSUE #929: [config] FilenameAttribute throws exceptions when configuring
https://github.com/sopel-irc/sopel/blob/master/sopel/config/types.py#L314-L340

Serialize and parse require extra arguments that they are not passed during `sopel --configure-modules`

maxpowa:
Ensures that FilenameAttribute's parse and serialize always have the parameters they need.

### DIFF
--- a/sopel/config/types.py
+++ b/sopel/config/types.py
@@ -111,7 +111,7 @@ class BaseValidated(object):
         self.name = name
         self.default = default
 
-    def configure(self, prompt, default, *args):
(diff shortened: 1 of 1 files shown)


embolalia:
Great, thanks!

maxpowa:
This caused an issue which I've fixed with https://github.com/sopel-irc/sopel/pull/938

PR #932, opened 2015-11-08
Resolve #899

### LINKED ISSUE #899: Unable to reload folder modules
Using Sopel 6.0.0 installed by Pip, reloading a directory module within a directory in `extras`:
```
  File "/home/embo/prog/env/local/lib/python2.7/site-packages/sopel/bot.py", line 257, in call
  File "/home/embo/prog/env/local/lib/python2.7/site-packages/sopel/modules/reload.py", line 49, in f_reload
    bot.unregister(obj)
  File "/home/embo/prog/env/local/lib/python2.7/site-packages/sopel/bot.py", line 169, in unregister
    for rule in obj.rule:
TypeError: 'function' object is not iterable
```

maxpowa:
This is a bit of a hacky fix, the problem is that the obj at this point could be any part of the module file, including an import or docstring or whatever. This just checks if obj is callable and then unregisters in that case. If it's not callable, we just ignore it. (Should be safe enough to ignore)

According to loader.py we only register callables, doing the same callable(obj) check.

### DIFF
--- a/sopel/modules/reload.py
+++ b/sopel/modules/reload.py
@@ -46,7 +46,13 @@ def f_reload(bot, trigger):
 
     old_callables = {}
     for obj_name, obj in iteritems(vars(old_module)):
-        bot.unregister(obj)
+        # This is a bit of a hacky fix, the problem is that the obj at this
+        # point could be any part of the module file, including an import
+        # or docstring or whatever. This just checks if obj is callable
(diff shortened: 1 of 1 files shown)


embolalia:
There were a few cases that this fix missed. I've committed something that accounts for them. Thanks, though!

PR #933, opened 2015-11-09
[isup] fix bad indexes preventing bot from recognizing http protocols

firerogue:
`'http://'` is 7 characters long, so `site[:6]` can't ever match it. ditto for `'https://'` and `site[:7]`


embolalia:
Thanks!

PR #936, opened 2015-11-13
Fix IncompleteRead

anewmanRH:
Address "IncompleteRead" raised by Bugzilla integration.


embolalia:
Is there a way I can reproduce the error you're fixing here?

anewmanRH:
The issue popped up because of a read timeout issue from a server that I have no control or rights to.
The only other thing I can think of is to retry the read vs returning an incomplete read.

dgw:
There is no way to get in touch with the author of this patch any more. GitHub will not merge it, and I cannot fetch the changes (cleanly) into a local branch to merge manually. With so littler information on this alleged error, I would rather not pull this in.

PR #941, opened 2015-11-18
[coretasks] Sopel account tracking

maxpowa:
Requires IRCv3.1 support on the network. Allows module devs to check the user account before performing any potentially harmful actions.

Probably needs some more thorough testing, I've just tested enough to know it won't crash most of  the time ;)

~~TODO: Suppress WHO on join if `account-notify` and/or `extended-join` aren't available.~~

### DIFF
--- a/sopel/bot.py
+++ b/sopel/bot.py
@@ -99,6 +99,17 @@ class Sopel(irc.Bot):
         bitwise integer value, determined by combining the appropriate constants
         from `module`."""
 
+        self.accounts = dict()
+        """A dictionary of channels to their users' accounts and hostmasks.
+
+        The value associated with each channel is a dictionary of hostmask
+        to account strings, determined via IRCv3.1 account-notify and
+        extended-join.
+
+        Requires the server to support IRCv3.1 `account-notify` and 
+        `extended-join`. If the server does not support these, you may see 
+        inaccurate or out-of-date information."""
+
         self.db = SopelDB(config)
         """The bot's database."""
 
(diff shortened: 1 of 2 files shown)


embolalia:
Definitely some good stuff here (and I'd been planning on implementing this at some point, because auth by nick, and even hostmask, is not really good enough), but I think there's probably more to do here to make the API for it nice.

The big thing is, I think optimally you shouldn't need the channel to get the user's account. That way it could used in PMs as well (though with the obvious limitation that, until `account-tag` is common, the user and bot will have to share a channel). It'll be a bit hard to do this nicely. I had been considering maybe some sort of subclass of `Identifier` with the list of channels and account name as attributes. Or rather, as read-only property, to make it harder to get into trouble with that.

Also, I'm not sure what that random part is for? And what's the global variable there for?

Lastly, what networks can I test this against?

maxpowa:
The global and random are for tracking the WHO requests we make, specifying `t` in our WHO output allows us to send a token that will be sent with the response so we can ensure that the WHO responses are saved to the correct channel. This really is just a shim, because ideally, like you said, we would be able to use account-tag and not have to track accounts per-channel. I considered not tracking per-channel, but there are problems relating to PART specifically (If a user parts a channel the bot is in, but remains on the network, we no longer receive ACCOUNT updates) which would have created a lot more complexity.

As for testing, I've done all mine on Espernet, but it *should* work on every IRCd that supports `account-notify`, `extended-join` and extended WHO.

maxpowa:
#961 is heading in a better direction than this, so I'll close this for now.

PR #944, opened 2015-11-20
FIX: Bugzilla module issues

anewmanRH:


maxpowa:
Might be a bit cleaner to just do 
```
if error:
    print('Bugzilla error: ' + error)
    return
```
instead of cramming all the code inside the if block. Also, since the function is listening for urls, it should be fine to not say anything back to the channel (see url.py when it fails to find a title in the page).

embolalia:
Yeah, I think @maxpowa is right here. I think probably the best thing to do would be to do `if error: return` right away, rather than outputting a message.

anewmanRH:
Updated as suggested.

maxpowa:
+1

elad661:
Can you squash the commits for easier reviewing and cleaner git history?

anewmanRH:
Squashed as requested.

anewmanRH:
Fixed the logger issue, please merge.

PR #954, opened 2015-12-02
fix minutes calculation

yoosi:
This change properly calculates the minutes portion of the countdown.


yoosi:
Never mind. I done goofed.

PR #958, opened 2015-12-04
[unicode_info] Fallback if input is None (fix #957)

### LINKED ISSUE #957: AttributeError: 'NoneType' object has no attribute 'strip' (file "/usr/local/lib/python2.7/dist-packages/sopel/modules/unicode_info.py", line 23, in codepoint)
[01:04am] <Ant> .u
01:04AM <Sopel> AttributeError: 'NoneType' object has no attribute 'strip' (file "/usr/local/lib/python2.7/dist-packages/sopel/modules/unicode_info.py", line 23, in codepoint)
01:04AM <Sopel> Ant: Sopel v. 6.1.1

This is in my Debian oldstable with Python v2.7.3. :(

dgw:


embolalia:
Will this work? It will evaluate `trigger.group(2).strip()` before getting to the added or, which will still break. It'd need to be `(trigger.group(2) or '').strip()` (and idk that group 2 can have whitespace anyway, so it might be better still to just do `arg = trigger.group(2)` and then `if arg:…`)

dgw:
Sorry, finally got around to this. Force-pushed an updated version that does away with `.strip()`, and checks for `if not arg:`.

maxpowa:
Probably still want to strip, just after we check if trigger.group(2) exists. I've posted two fail cases below, the first one would start with your code and the second happens right now, but would be fixed by your PR.
```
<maxpowa> .u ☃  
<Sopel> That's not a valid code point.
```
```
<maxpowa> .u  
<Sopel> maxpowa: What code point do you want me to look up?
```
` ` both of the the examples is a whitespace character (U+1680 ogham space mark) that gets removed by strip.

Ideally it would check first if group 2 exists, then check the length after the strip and decide what the best action is. If the length is 1, go ahead and process that character. If the length is 0, check the length without the strip & process.

dgw:
Wow, I really forgot about this. But here, force-pushed another commit that should do what you explained, @maxpowa (unless the fatigue is getting to me and I'm wrong…it *is* past 1am).

PR #961, opened 2015-12-06
Add tracking of users and their accounts

embolalia:
This uses some of the code that @maxpowa wrote for #941, but gives a somewhat more intuitive API. It also paves the way for potentially adding direct support for away-notify and metadata-notify.

Since this being broken could have big security repurcussions (especially since I'd like to be able to make admins/owner pull from account tracking in the future), I'd like a few people to pick this over and see what I've broken before merging it.

### DIFF
--- a/sopel/tools/target.py
+++ b/sopel/tools/target.py
@@ -46,6 +46,7 @@ class Channel(object):
         assert isinstance(user, User)
         self.users[user.nick] = user
         self.privileges[user.nick] = 0
+        user.channels[self.name] = self
 
     def rename_user(self, old, new):
         if old in self.users:
--- a/sopel/tools/target.py
+++ b/sopel/tools/target.py
@@ -1,0 +1,64 @@
+# coding=utf-8
+from __future__ import unicode_literals, absolute_import, print_function, division
+
+import functools
+from sopel.tools import Identifier
+
+
+@functools.total_ordering
+class User(object):
+    def __init__(self, nick, user, host):
+        assert isinstance(nick, Identifier)
+        self.nick = nick
+        self.user = user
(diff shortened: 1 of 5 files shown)


embolalia:
`bot.channels_` is a sucky name, but idk if anyone is relying on the type of `bot.channels`. Something will have to be done about that before merging.

maxpowa:
Privileges could probably be merged into the user object too, so you would do user.privileges[channel] or something along those lines. Would simplify the privileges usage a bit and streamline both paths.

sopel/tools/target.py
```diff
@@ -0,0 +1,64 @@
...
+        self.host = host
+        self.channels = {}
+
+    hostmask = property(lambda self: '{}!{}@{}'.format(self.nick, self.user,
+                                                       self.host))
+
+    def __eq__(self, other):
+        if not isinstance(other, User):
+            return NotImplemented
+        return self.name == other.name
```
maxpowa: name is not defined for user, same thing on line 28

embolalia:
@elad661, @maxpowa: I've addressed the comments, and added some extra features. Do either of you have an objection to replacing the old `bot.channels` with the new dict version? It'll *probably* still work for *most* use cases of the old one, but I'm not sure if it counts as a breaking change. I suppose we could just call it 7.0…

embolalia:
I've updated this to get rid of the channels list and replace it with the channels dictionary.

embolalia:
@elad661: I kind of figured that's the only kind of use that's common, and it still works (because it's iterating over the keys, which are the same). You're probably right about changing the name of `SopelMemory`.

Issue #967, opened 2015-12-09
Say a preconfigured greeting upon joining a channel


Ganon11:
It would be fun if a Sopel admin could (optionally) define a greeting in the Config file that the bot would say upon joining a channel.

For instance, I run a Sopel instance that I've given the nick "WorfBot". A previous incarnation of WorfBot would say `I am WorfBot, son of MoghBot.` With this feature, I could define a setting such as `core.greeting` to be `I am WorfBot, son of MoghBot.`. If the greeting was undefined, no greeting would be used upon channel joining.

Similarly, a departure message might be fun, though used less often.

maxpowa:
You can do that pretty easily with a plugin, just listen for event 'JOIN'
and trigger.nick == bot nick.

dgw:
I'll publish a plugin that can do the join greeting tonight. Will look into how to do the part greeting then, if possible.

Ganon11:
I also filled this issue with the intention of doing the change myself. I can either add a `greeting` module, or modify `irc.py` to look for and say a greeting in `join`

maxpowa:
In my opinion it would be better as a module as it's not required for functionality of Sopel.

Ganon11:
I have a module doing the join message properly. The parting message does not work, since by the time the 'PART' event happens, the bot has already parted the channel (thus the bot can't say a message to that channel).

I also realized that you can already give a part_message to the .part command, so this may not be needed. I'll prepare a pull request with the greeting module.

embolalia:
While a few people might find this useful, I think a lot of others would find it unnecessary (or even spammy). The required configuration is also a problem for including in the mainline. The [general guideline](https://github.com/sopel-irc/sopel/wiki/Module-inclusion-guidelines) is not to allow this, because it makes setting the bot up harder.

PR #968, opened 2015-12-10
[greeting] Add greeting.py module

Ganon11:
This module allows admins to define a join_message in a new [greeting] config section. If provided, this message will be said by the bot to any channel it joins.

Addresses feature request #967

### DIFF
--- a/sopel/modules/greeting.py
+++ b/sopel/modules/greeting.py
@@ -1,4 +1,4 @@
-# coding=utf8
+# coding=utf-8
 """
 greeting.py - Greeting Module
 Copyright 2015, Michael Stark, <stark3@gmail.com>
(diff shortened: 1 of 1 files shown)


maxpowa:
Couple nitpicky/code style things:
- The coding declaration at the top does not support Windows, it has to be `utf-8` for Windows support. 
- `__future__` imports should conform to the standard for the rest of Sopel's components
- The `define_section` in configure should have `validate=False` as an argument, or configuring will fail when no value is in the config.
- Checking for `bot.config.greeting` on line 33 is not required, because it's put in by setup.

Otherwise, looks good.

embolalia:
The code looks good, but per my comment in #967 I don't think it belongs as a default module.

Issue #874, opened 2015-09-02
dice not accepting upper-case


ventris:
The dice function wont accept upper-case D in number of dices.

eg. 1D100

answer from bot: "SyntaxError, eval(1D100), unexpected EOF while parsing (<unknown>, line 1)"

Has only been tried on python 3.

Deedasmi:
http://i.imgur.com/dubuZwB.png?1

It also doesn't handle invalid input very gracefully. I'll look into it more this weekend.

maxpowa:
Appears to have been resolved by #969

PR #969, opened 2015-12-10
[dice] Case insensitivity and better error handling for negative numbers.

Ganon11:
## Case insensitivity
Instead of matching for the literal `d` or `v`, match `[dD]` and `[vV]` appropriately.

## Negative numbers
Match an optional `-` character before each possible number, and handle negative inputs in the `_roll_dice` method. There was already a check for negative `dice_type`, but because the `-` wasn't being captured, this code wasn't being used. I added a check for `dice_num < 0` (which returns `None`) and `drop < 0` (which ignores the `drop_lowest` value entirely).

Partially addresses #874


embolalia:
Thanks!

PR #987, opened 2016-01-05
[reload] Prevent reloading nonexistent module

### LINKED ISSUE #977: Key error when reloading a nonexistent module
`KeyError: 'time' (file "/opt/rh/python33/root/usr/lib/python3.3/site-packages/sopel/modules/reload.py", line 62, in f_reload)` should be a sane error message

maxpowa:
Fixes #977

### DIFF
--- a/sopel/modules/reload.py
+++ b/sopel/modules/reload.py
@@ -40,7 +40,7 @@ def f_reload(bot, trigger):
         return bot.reply('done')
 
     if name not in sys.modules:
-        return bot.reply('%s: not loaded, try the `load` command' % name)
+        return bot.reply('"%s" not loaded, try the `load` command' % name)
 
     old_module = sys.modules[name]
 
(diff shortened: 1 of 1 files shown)


embolalia:
Would it make more sense to put the new logic up where the old logic is? Or maybe even replace the old logic with the new.

maxpowa:
I dunno. The logic in the upper part is for unregistering the modules, and the logic in the lower part is for loading the file. With something like `.reload time`, sys.modules has `time` because of imports in other files - so it passes the first logic - but then when it reaches the `enumerate_modules`, there is no `time.py` in the modules folders - causing it to fail out there. The first logic probably needs some extra checks to see if the module in sys.modules is actually a sopel module, dunno how reliably we could check that though. 

Having both checks makes sense though. What if the file was loaded, deleted and then you attempted to reload? It would make it though the first logic, then fail out on the second. (Makes me realize the second error should be along the lines of `"xyz" not found, are you sure it's in the proper location?`)

maxpowa:
I guess in `clean_module` we could add a `__sopel_module__` attribute to the modules. That would help indicate for the first logic if we are a sopel module or not, and would prevent us from unloading system modules.

tpltnt:
The code looks good to me.

PR #990, opened 2016-01-08
[loader] Properly walk module directories

maxpowa:
Tested fairly thoroughly, works with reloading, fresh starts. Should ignore compiled python (`__pycache__`) directories, as well as both hidden files and directories. Works with pypi installed third party modules, ~~should work with third party packaged modules put directly in the modules directory~~. Standalone python files in the modules directory work as before.

Edit: It will currently ignore symlinks, dunno if we should do that or not.

Edit2: It doesn't work if you drop a python package in the modules directory, not really sure what subtlety is messing that up.

### DIFF
--- a/sopel/loader.py
+++ b/sopel/loader.py
@@ -32,11 +32,33 @@ def get_module_description(path):
 
 def _update_modules_from_dir(modules, directory):
     # Note that this modifies modules in place
-    for path in os.listdir(directory):
-        path = os.path.join(directory, path)
-        result = get_module_description(path)
-        if result:
-            modules[result[0]] = result[1:]
+    for root, dirs, files in os.walk(directory):
+        # Ignore hidden folders/files and __ prefixed (__pycache__)
+        files = [f for f in files if not f[0] == '.']
+        dirs[:] = [d for d in dirs if not d[0] == '.']
+        dirs[:] = [d for d in dirs if not d.startswith('__')]
+
+        # Test if folders are package folders. If they are, we
+        # don't need to walk into them so remove them from the
+        # `dirs` list
+        non_pkg_dirs = []
+        for folder in dirs:
+            path = os.path.join(root, folder)
+            result = get_module_description(path)
+            if result:
+                print(path)
+                modules[result[0]] = result[1:]
+            else:
+                non_pkg_dirs.append(folder)
+        dirs[:] = non_pkg_dirs
+
+        # Assuming all files are modules, try to validate
+        for module in files:
+            path = os.path.join(root, module)
+            result = get_module_description(path)
+            if result:
+                print(path)
+                modules[result[0]] = result[1:]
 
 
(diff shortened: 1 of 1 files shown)


thomastanck:
+1 not sure why this could hurt, and it only allows for better organisation of custom modules

I will test this on Saturday.

dgw:
@thomastanck Is it Saturday yet? :smile_cat: (Translation: Are you still able to test this?)

thomastanck:
Sorry about that... it's Sunday already :(
(Translation: sorry I completely forgot about this, and now it's so late I no longer have sopel set up for testing)

dgw:
Punting to next major update: There is currently no one with prior experience to review this, and the known issues (ignoring symlinks, failing on packaged modules) really should be solved before merging.

sopel/loader.py
```diff
@@ -32,11 +32,31 @@ def get_module_description(path):
...
     # Note that this modifies modules in place
-    for path in os.listdir(directory):
-        path = os.path.join(directory, path)
-        result = get_module_description(path)
-        if result:
-            modules[result[0]] = result[1:]
+    for root, dirs, files in os.walk(directory):
+        # Ignore hidden folders/files and __ prefixed (__pycache__)
+        files = [f for f in files if not f[0] == '.']
+        dirs[:] = [d for d in dirs if not d[0] == '.']
```
Exirel: I know this syntax works, but here it is equivalent to this less confusing one:

```
dirs = [d for d in dirs if not d[0] == '.']
```

sopel/loader.py
```diff
@@ -32,11 +32,31 @@ def get_module_description(path):
...
+        dirs[:] = [d for d in dirs if not d.startswith('__')]
```
Exirel: It is possible to merge both filter in a single iteration:

```
dirs = [
    d for d in dirs
    if not d.startswith('.') and not d.startswith('__')
]
```

sopel/loader.py
```diff
@@ -32,11 +32,31 @@ def get_module_description(path):
...
+        # `dirs` list
+        non_pkg_dirs = []
+        for folder in dirs:
+            path = os.path.join(root, folder)
+            result = get_module_description(path)
+            if result:
+                modules[result[0]] = result[1:]
+            else:
+                non_pkg_dirs.append(folder)
+        dirs[:] = non_pkg_dirs
```
Exirel: I'm not sure we need to store `non_pkg_dirs` and why it replaces `dirs` either.

dgw:
@Exirel Do you want to take this over in a new PR? You could just `git rebase -i master` and "fixup" 601936d into the initial commit (which I would do, but I can't push a revised history to Max's branch), then add your own stuff onto there. That way we can un-stall this PR _and_ keep the original authorship info!

Exirel:
@dgw let's do this then!

dgw:
Closing in favor of #1479

PR #995, opened 2016-01-09
[tests] Add a bunch of tests

maxpowa:
Adds a bunch of tests for trigger/modules/formatting and makes some minor cleanup to db tests. 

Updated the .travis.yml to test style, but ignore the result. Made the tests run in verbose mode so you can see the exact test that failed. Runs coveralls after a successful build. 

It will currently fail the tests, as there's an issue in web.py.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,3 +25,6 @@ willie.egg-info/*
 .settings
 .project
 .pydevproject
+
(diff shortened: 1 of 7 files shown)


embolalia:
Tests! You're officially awesome.

PR #1000, opened 2016-01-15
Make _ssl_recv always return bytes

### LINKED ISSUE #937: Python 3 incompatibility bug? TypeError: can't concat bytes to str
Using the admin module quit command running sopel on archlinux with python 3.5:

> ERROR:sopel.modules.sopel.irc:ERROR recieved from server: Closing Link: mechasqueak-test[p200300630A2FCF79AE220BFFFE4D6336.dip0.t-ipconnect.de] (Quit: Quitting on command from DukeOfEarl)
> Calling shutdown for 0 modules.
> Closed!
> Traceback (most recent call last):
>   File "/usr/lib64/python3.5/asyncore.py", line 83, in read
>     obj.handle_read_event()
>   File "/usr/lib64/python3.5/asyncore.py", line 423, in handle_read_event
>     self.handle_read()
>   File "/usr/lib64/python3.5/asynchat.py", line 126, in handle_read
>     self.ac_in_buffer = self.ac_in_buffer + data
> TypeError: can't concat bytes to str
> 
> ERROR:sopel.modules.sopel.irc:Fatal error in core, please review exception log

It seems there's a `decode()` missing.

duk3luk3:
_ssl_recv returned empty strings instead of the empty bytes object if
the socket was closed or upon ENOENT.

This lead to exceptions when running sopel with python3 because
asynchat.handle_read expects byte objects.

This commit fixes #937.

Note - b'' is available from python 2.7 onwards. I tested this change with python 3.5.0 and 2.7.10.

### DIFF
--- a/sopel/irc.py
+++ b/sopel/irc.py
@@ -310,7 +310,7 @@ class Bot(asynchat.async_chat):
             data = self.socket.read(buffer_size)
             if not data:
                 self.handle_close()
-                return ''
+                return b''
             return data
         except ssl.SSLError as why:
             if why[0] in (asyncore.ECONNRESET, asyncore.ENOTCONN,
@@ -319,7 +319,7 @@ class Bot(asynchat.async_chat):
                 return ''
             elif why[0] == errno.ENOENT:
                 # Required in order to keep it non-blocking
-                return ''
+                return b''
             else:
                 raise
 


embolalia:
Awesome, thanks. (Also, 1000 issues and PRs! Woah!)

PR #1007, opened 2016-01-22
woeid_search uses geo.places fixing issue 1006

caseytb:
Yahoo seems to have shut down geo.placefinder ( receiving 503 responses
from the api ). This is fixed by using geo.places instead.

### DIFF
--- a/sopel/modules/weather.py
+++ b/sopel/modules/weather.py
@@ -2,12 +2,14 @@
 # Copyright 2008, Sean B. Palmer, inamidst.com
 # Copyright 2012, Elsie Powell, embolalia.com
 # Licensed under the Eiffel Forum License 2.
+
 from __future__ import unicode_literals, absolute_import, print_function, division
 
 from sopel import web
 from sopel.module import commands, example, NOLIMIT
 
 import xmltodict
+import collections
 
 
 def woeid_search(query):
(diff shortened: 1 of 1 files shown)


sopel/modules/weather.py
```diff
@@ -1,14 +1,19 @@
...
-# Copyright 2012, Elsie Powell, embolalia.com
-# Licensed under the Eiffel Forum License 2.
+"""
+weather.py - Sopel Yahoo! Weather Module
+Copyright 2008, Sean B. Palmer, inamidst.com
+Copyright 2012, Edward Powell, embolalia.net
+Licensed under the Eiffel Forum License 2.
+
+http://sopel.chat
+"""
```
embolalia: I'm guessing you hadn't pulled recently, because you're wiping out the change from ee86f38ddf07213d1dafb4939b71dea7a47809f3 here, I assume unintentionally.

sopel/modules/weather.py
```diff
@@ -16,16 +21,29 @@ def woeid_search(query):
...
     results = parsed.get('results')
-    if results is None or results.get('Result') is None:
+    if type(results) is collections.OrderedDict:
+        place = results.get('place')
+        if type(place) is list:
+            return place[0]
+    if not (results):
+        return None
+    if type(results) is list:
+	first_result = result[0]
```
embolalia: There's a tab character here (and possibly also elsewhere?). Indentation must always be in multiples of four spaces. (It won't even load on Python 3 if you mix them.) Also, `result` isn't defined, and `first_result` is assigned to, but not used.

sopel/modules/weather.py
```diff
@@ -138,7 +156,7 @@ def weather(bot, trigger):
         return bot.reply("I don't know where that is.")
 
     query = web.urlencode({'w': woeid, 'u': 'c'})
-    raw = web.get('http://weather.yahooapis.com/forecastrss?' + query,
+    raw = web.get('http://weather.yahooapis.com/forecastrss?' + query,
```
embolalia: Trailing space here should be removed.

embolalia:
Thanks! This is good, but there are some things that'll need to be fixed.

eanxgeek:
Is this going to fix anything at this point, it looks like Yahoo has disabled public access to its geolocation service.  Perhaps I am mistaken, but weather hasn't worked in the past 24+ hours.

caseytb:
I have it on a running sopel bot and it has been working fine since
geo.placefinder stopped functioning yesterday - woeids can be queried again
by using geo.location.

As far as "fixing" anything in the long run I haven't had time to figure
out how long geo.location is supposed to be support, if it is supported at
all. For now this provides an at least temporary fix that doesn't require
an API key.

maxpowa:
Lines 29-32 and 38-45 seem very similar, is there a reason why you'd check twice?

caseytb:
Yep I had modified weather.py from an older version of Sopel (i think it was weather.py that was in sopel distributed by pypi before the January 17 update, but I could be wrong on that) and copy pasted it in my forked version (whoops). So the copyright line is consistent with the rest of Sopel. That's what I get for committing while tired.

The weather module should work ok now. I think its actually a little bit more liberal in finding matches to a query string than the older API, but I never really tried too hard to break the older weather API. Currently it seems only terrible mis-spellings will trigger the "I don't know where that is" response.

elad661:
Can you please squash all those commits to one single commit? For easier reviewing and cleaner git history

caseytb:
Done.

PR #1013, opened 2016-01-28
Add SNI support to VerifiedHTTPSConnection

calzoneman:
Previously, things like `.title https://redd.it/nmo0q` would fail because the HTTPS request wouldn't use SNI to specify the hostname, (the servers behind redd.it return the certificate for reddit.com by default if the client does not support SNI).  This patch detects when the `ssl` module supports SNI and includes the hostname in the `wrap_socket` parameters, so `.title` and friends work on servers sharing multiple certificates on different virtualhosts.

Should degrade gracefully if `ssl.HAS_SNI = False` (I wasn't able to test this since both Pythons on my system support SNI).  Tested in Python 3.5.1 and 2.7.11.

### DIFF
--- a/sopel/web.py
+++ b/sopel/web.py
@@ -175,11 +175,14 @@ class VerifiedHTTPSConnection(httplib.HTTPConnection):
             if not ca_certs or not os.path.exists(ca_certs):
                 sock.close()
                 raise Exception('CA Certificate bundle %s is not readable' % ca_certs)
-            self.sock = ssl.wrap_socket(sock,
-                                        ca_certs=ca_certs,
-                                        cert_reqs=ssl.CERT_REQUIRED)
-            ssl.match_hostname(self.sock.getpeercert(), self.host)
-
+            context = ssl.SSLContext(ssl.PROTOCOL_SSLv23)
+            context.verify_mode = ssl.CERT_REQUIRED
+            context.check_hostname = True
+            context.load_verify_locations(ca_certs)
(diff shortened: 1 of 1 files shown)


maxpowa:
+1 this would probably close #988 too.

elad661:
web.py is deprecated. We shouldn't be adding any functionality to it. Instead, modules which use web.py should be ported to requests, and web.py should either be removed or reworked as a compatibility layer to requests.

Therefore, I'm closing this pull request and not merging it. Sorry.

calzoneman:
I consider this a bug fix, not a feature. SNI should be a basic requirement of anything making an HTTPS request in 2016.

Regardless of deprecation, there are still core modules that rely on this that I expect will not all be ported to requests overnight. It seems more useful to merge a 5 line patch than tell everyone "yeah that's just going to remain broken until we port each of the modules that depend on it".

That said, I'm content to just run a fork of sopel; I was mostly providing this PR in case it was useful to someone else.

That's my 2 cents.

Calvin

maxpowa:
@elad661 I agree with calzoneman on this one, it should be fixed since web.py has not yet been removed. Yes, the modules are going to be switched to requests, but until then this should be fixed.

elad661:
SNI support is a feature. An important one, but still, a feature. You don't add features to deprecated modules.

Anyway this discussion is moot as of 35a543d4c8fbc5476a828db783de771d033b884b because this is the *correct* solution to this problem.

Issue #365, opened 2013-11-11
Privilege handling is just generally shit. (AKA NAMES is ignored, KeyError on someone leaving something, etc.)


ict:
_Repo owner edit: This is just one facet of this error. Just in general, privilege handling doesn't work properly anywhere. There have been probably half a dozen issues for this, so I'm crunching them all in here._

---

Willie seems to not parse the `NAMES` List that irc.fu-berlin.de is sending. 

Raw output from the server (bot's name is Amdahl):

```
<<1384209258.33 :Amdahl!~willie@koala24.de JOIN :#mytestchannel
<<1384209258.33 :fu-berlin.de 353 Amdahl = #mytestchannel :Amdahl @ict
<<1384209258.33 :fu-berlin.de 366 Amdahl #mytestchannel :End of NAMES list.
```

This leads to error messages when someone uses `.op` or similar:

```
Signature: KeyError: u'ict' (file "/home/ict/build/willie/willie/modules/adminchannel.py", line 28, in op)
from #mytestchannel at 2013-11-11 23:27:34.803790:
Message was: <ict> .op
Traceback (most recent call last):
  File "/home/ict/build/willie/willie/bot.py", line 697, in call
    exit_code = func(willie, trigger)
  File "/home/ict/build/willie/willie/modules/adminchannel.py", line 28, in op
    if (bot.privileges[trigger.sender][trigger.nick] >= OP) or trigger.admin or trigger.owner:
KeyError: u'ict'
```

If I part and rejoin the channel while the bot is in the channel, everything works.

Version information et al:

```
08:45:13    ict | .version
08:45:14 Amdahl | ict: Willie v. 4.1.0-git at commit:                                                                                                                                                                               
08:45:15 Amdahl |   commit 03fa4ef744cf00776f8efb099e27d866ec0cbfb2                                                                                                                                                                 
08:45:15 Amdahl |   Author: Edward Powell <git@embolalia.com>                                                                                                                                                                       
08:45:16 Amdahl |   Date:   Sun Nov 10 17:49:04 2013 -0500
```

I'm running Debian Wheezy  and willie is a direct checkout of this repository, the dependencies have been installed with pip. Despite having `verbose = verbose` in the default.cfg, I can see no debug output at all. This is the startup output:

```
Welcome to Willie. Loading modules...


Error loading reddit-info: No module named praw (modules/reddit-info.py:11)


Registered 35 modules,
1 modules failed to load


[...]

PR #1014, opened 2016-01-29
Prevent KeyError when untracked user leaves

### LINKED ISSUE #1005: When an IRC person leaves & returns, I get a traceback error...
Hello.

For an example:
...
��� SignOff Flexo: #bots,#linux (Quit: restarting, back in 5...)
10:51PM ��� Flexo [~rbot@[deleted its IP address] has joined #linux
...

$ Traceback (most recent call last):
  File "/home/ant/SopelPython3/lib/python3.4/site-packages/sopel/bot.py", line 441, in call
  File "/home/ant/SopelPython3/lib/python3.4/site-packages/sopel/coretasks.py", line 375, in track_quit
    channel.clear_user(trigger.nick)
  File "/home/ant/SopelPython3/lib/python3.4/site-packages/sopel/tools/target.py", line 63, in clear_user
    user = self.users[nick]
KeyError: Identifier('Flexo')

Thank you in advance. :)

thomastanck:
Note that untracked users and their privileges are currently horribly handled. See #365 and its derivatives.

Technically fixes #1005, but we really should handle untracked users a lot better. See handle_names in coretasks.py

Before: (5 seconds online time)
```
gracefu:sopel gracefu$ python sopel.py -c test.cfg
Could not open CA certificates file. SSL will not work properly.

Welcome to Sopel. Loading modules...


Error loading ipython: No module named IPython (modules/ipython.py:23)


Registered 42 modules,
1 modules failed to load


Connecting to irc.ppy.sh:6667...
Connected.
  File "/Users/gracefu/sopel/sopel/bot.py", line 441, in call
  File "/Users/gracefu/sopel/sopel/coretasks.py", line 375, in track_quit
    channel.clear_user(trigger.nick)
  File "/Users/gracefu/sopel/sopel/tools/target.py", line 63, in clear_user
    user = self.users[nick]
KeyError: Identifier('[GoGri]')

  File "/Users/gracefu/sopel/sopel/bot.py", line 441, in call
  File "/Users/gracefu/sopel/sopel/coretasks.py", line 375, in track_quit
    channel.clear_user(trigger.nick)
  File "/Users/gracefu/sopel/sopel/tools/target.py", line 63, in clear_user
    user = self.users[nick]
KeyError: Identifier('Tayzen')

  File "/Users/gracefu/sopel/sopel/bot.py", line 441, in call
  File "/Users/gracefu/sopel/sopel/coretasks.py", line 375, in track_quit
    channel.clear_user(trigger.nick)
  File "/Users/gracefu/sopel/sopel/tools/target.py", line 63, in clear_user
    user = self.users[nick]
KeyError: Identifier('darrylshin198')

  File "/Users/gracefu/sopel/sopel/bot.py", line 441, in call
  File "/Users/gracefu/sopel/sopel/coretasks.py", line 375, in track_quit
    channel.clear_user(trigger.nick)
  File "/Users/gracefu/sopel/sopel/tools/target.py", line 63, in clear_user
    user = self.users[nick]
KeyError: Identifier('Neke-2001')

^CKeyboardInterrupt
```

After: (5 minutes online time)
```
gracefu:sopel gracefu$ python sopel.py -c test.cfg
Could not open CA certificates file. SSL will not work properly.

Welcome to Sopel. Loading modules...


Error loading ipython: No module named IPython (modules/ipython.py:23)


Registered 42 modules,
1 modules failed to load


Connecting to irc.ppy.sh:6667...
Connected.
^CKeyboardInterrupt
```

### DIFF
--- a/sopel/coretasks.py
+++ b/sopel/coretasks.py
@@ -355,6 +355,7 @@ def track_join(bot, trigger):
     user = bot.users.get(trigger.nick)
     if user is None:
         user = User(trigger.nick, trigger.user, trigger.host)
+        bot.users[trigger.nick] = user
     bot.channels[trigger.sender].add_user(user)
 
     if len(trigger.args) > 1 and trigger.args[1] != '*' and (
--- a/sopel/tools/target.py
+++ b/sopel/tools/target.py
@@ -60,10 +60,10 @@ class Channel(object):
         be compared to appropriate constants from ``sopel.module``."""
 
     def clear_user(self, nick):
-        user = self.users[nick]
-        user.channels.pop(self.name, None)
-        del self.users[nick]
-        del self.privileges[nick]
+        user = self.users.pop(nick, None)
+        self.privileges.pop(nick, None)
+        if user != None:
+            user.channels.pop(self.name, None)
 
     def add_user(self, user):
         assert isinstance(user, User)


embolalia:
This is unrelated to #365. That refers to the old (now deprecated) permissions system, which this doesn't touch (and it was closed ages ago).

I think, other than it hiding #1005, there's  probably nothing wrong with merging this. I'm not 100% sure, now that I think of it, that sending QUIT to a user who shares no channels with the quitting user is in violation of the spec, and if that happened it would cause the error that this does fix.

maxpowa:
+1 We should be robust enough to handle some minor spec discrepancies like this one.

thomastanck:
A friend suggested adding ```__missing__``` methods to dicts so that things like bot.users[nick] don't fail.

e.g.
```python
Python 3.4.3 (default, Jul 13 2015, 12:18:23)
[GCC 4.2.1 Compatible Apple LLVM 6.1.0 (clang-602.0.53)] on darwin
>>> from sopel.tools.target import User, Identifier
>>> test = dict()
>>> test[Identifier('maximumpower')]
KeyError: Identifier('maximumpower')
>>> class Default(dict):
...     def __missing__(self, key):
...         self[key] = User(key, None, None)
...         return self[key]
...
>>> test = Default()
>>> test[Identifier('maximumpower')]
<sopel.tools.target.User object at 0x106099f98>
>>> test.items()
dict_items([(Identifier('maximumpower'), <sopel.tools.target.User object at 0x106099f98>)])
```

embolalia:
@thomastanck: Having a default here wouldn't solve the original issue, and would make the API semantics confusing. `bot.users[nick]` for a nick the bot isn't aware of *should* fail; the bot should not pretend to know things it doesn't know.

I'd like to see what, if anything, comes of #1005 before merging this. From what's been discussed there so far, it isn't a minor spec discrepancy but a complete failure to implement mandatory components. Handling blatant non-compliance from every esoteric broken implementation quickly becomes a maintenance burden, and is beyond the scope of this project.

thomastanck:
@embolalia Here's what I've seen so far in my investigations:

bot.users is properly handled 99.9% of the time. It is populated in:
1) track_join*
2) account_notify
3) recv_whox
4) track_notify (AWAY status)

bot.users is depopulated in:
1) _remove_from_channel (PART/KICK)
2) track_quit (basically _remove_from_channel except for every channel)

bot.users is mutated in:
1) track_nicks (the key is changed if the user is already in bot.users: ```if old in bot.users: bot.users[new] = bot.users.pop(old)``` else it doesn't add or remove anything. Also note that the User information is now outdated, ```bot.users[new].nick != new```)

*Only after merging this. track_join has a legitimate bug in the current version. It never triggered an error because we always .pop when removing users. This commit removes a del in Channel.clear_user(), but we actually do populate Channel.users in track_join and everything else in the list above. None of the modules actually use bot.users. The module I was working on did, which is a big part of the reason why I found the error in the first place.

So why does the error occur for me? It's because bot.users and bot.channels are NOT populated in handle_names, and the IRCd I'm on is returning NAMREPLY instead of WHOREPLY. What happens is that when an untracked user QUITs, ```for channel in bot.channels.values(): channel.clear_user(trigger.nick)``` is called. In contrast with _remove_from_channel, during a QUIT no existence checks are made. That then results in the del lines inside target.py raising exceptions.



TL;DR:
This commit has a legit bug fix (track_join) and suppresses exceptions when a IRCd does not reply with WHOREPLY, and instead uses NAMREPLY, resulting in bot.channels not being populated. Other than the track_join bug, the whole system is probably consistently populating both bot.users and bot.channels.

I wouldn't say this is a complete failure to implement mandatory components, since the only thing left out was track_join.

You're right that handling noncompliant servers can get out of hand quickly, so there's no need to go beyond what's already committed. Note that even after merging, we still will not populate bot.users/bot.channels after a NAMREPLY, since compliant IRCds should reply with WHOREPLY, which we already handle.

PR #1016, opened 2016-01-31
url: handle capitalized URLs

calzoneman:
Trigger rules are case-insensitive regexes, so the auto title responder
will be triggered even for capitalized URLs such as "Http://google.com"
(which can happen, for example, when a mobile device attempts to
auto-capitalize the beginning of a sentence).  Match URLs case
insensitively for title lookup purposes and add error handling in case
no URLs could be extracted from the match.


elad661:
Looks good, thanks.

PR #1017, opened 2016-01-31
reddit: fetch posts by submission_id

calzoneman:
Previously, the reddit module fetched posts by the full URL of the post.
This led to RedirectExceptions in some cases, for example when someone
links a naked reddit.com URL instead of www.reddit.com.  Instead, match
only the post ID and pass it to get_submission.


elad661:
Merged, thanks.

PR #1020, opened 2016-02-04
round currency exchange value result

yoosi:
Before:
`<user> .cur 22 SEK in USD`
`<sopel> 22.0 SEK (Swedish krona) = 2.6187908496732026 USD (U.S. dollar)`

After:
`<user> .cur 22 SEK in USD`
`<sopel> 22.0 SEK (Swedish krona) = 2.62 USD (U.S. dollar)`

### DIFF
--- a/sopel/modules/currency.py
+++ b/sopel/modules/currency.py
@@ -83,7 +83,7 @@ def display(bot, amount, of, to):
 
     result = amount / of_rate * to_rate
     bot.say("{} {} ({}) = {} {} ({})".format(amount, of.upper(), of_name,
-                                             result, to.upper(), to_name))
+                                             round(result, 2), to.upper(), to_name))
 
 
 @commands('btc', 'bitcoin')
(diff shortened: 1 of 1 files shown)


dgw:
I'm not sure I'm in favor of rounding this. Best to know the actual result, IMO.

maxpowa:
I agree with dgw here, currency exchange shouldn't be rounded.

embolalia:
The service doesn't seem to generally give more than 4 to 6 significant figures, so anything more than that in our output is really false precision. I'm in favor of rounding in principal, but this isn't a good enough implementation. It breaks badly for something like `.cur 1 JPY in USD`. This is currently ~$0.0084612, but which this shows as $0.01, which is a huge error. I'd be willing to merge something that rounds to say 5 sig figs, rather than just the nearest hundredth.

yoosi:
@embolalia:
>I'd be willing to merge something that rounds to say 5 sig figs, rather than just the nearest hundredth.

Done

PR #1023, opened 2016-02-09
Added missing import to xkcd.py

nfearnley:
xkcd.py was missing the import for the requests module when it was ported over to requests.

This commit adds that missing import.


elad661:
Oops. Thanks.

PR #1027, opened 2016-02-16
Use unicode() instead of str() to decode HTML page titles.

### LINKED ISSUE #1021: url: UnicodeEncodeError
UnicodeEncodeError: 'ascii' codec can't encode character u'\u2014' in position 352:    | ordinal not in range(128) (file "/usr/local/lib/python2.7/dist-packages/sopel/modules/url.py", line | 191, in find_title)	

https://github.com/sopel-irc/sopel/blob/master/sopel/modules/url.py#L191

br0ziliy:
This fixes #1021

Following the existing code, `errors='ignore'` causes decoding errors to be
silently ignored.  Also we catch LookupError exception, which is raised by unicode() if it does not know the encoding and/or is not able to decode an object properly.

### DIFF
--- a/sopel/modules/url.py
+++ b/sopel/modules/url.py
@@ -188,10 +188,10 @@ def find_title(url):
     try:
         content = ''
         for byte in response.iter_content(chunk_size=512, decode_unicode=True):
-            content += str(byte)
+            content += unicode(byte,errors='ignore')
             if '</title>' in content or len(content) > max_bytes:
                 break
-    except UnicodeDecodeError:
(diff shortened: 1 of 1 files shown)


elad661:
`unicode` does not exist in Python 3. Please change your code to make sure it supports python 3 as well. Python 3 support is more important than python 2 support because python3 is the default for Sopel

br0ziliy:
@elad661 Ooops, you're right. Closing this one, will open another pull request with a proper fix.

PR #1039, opened 2016-03-07
Typo correction

lunaticitizen:
d**ea**mon -> d**ae**mon


elad661:
Please squash those commits together and use a proper commit message instead of "update <filename>.py".

lunaticitizen:
Done. That's what happens when you're not really paying attention ;)

PR #1040, opened 2016-03-09
Use unicode strings to avoid UnicodeEncodeError

jbnance:
The URL module chokes when looking up the title of web pages that include unicode with the following error:

UnicodeEncodeError: 'ascii' codec can't encode character u'\xb7' in position 403: ordinal not in range(128) (file "/opt/willie/venv/lib/python2.7/site-packages/sopel/modules/url.py", line 191, in find_title)

(have it look up https://github.com/sopel-irc/sopel-extras/pull/78 for an example)

Avoid this issue by handling strings as unicode.

### DIFF
--- a/sopel/modules/url.py
+++ b/sopel/modules/url.py
@@ -186,9 +186,9 @@ def find_title(url):
     """Return the title for the given URL."""
     response = requests.get(url, stream=True)
     try:
-        content = ''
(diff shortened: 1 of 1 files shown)


elad661:
I've already merged a different pull request which fixed this issue ( #1028 ).

Issue #1041, opened 2016-03-09
Wolfram|Alpha now requires API key


dgw:
````
<dgw> .wa test
<Sopel> [WOLFRAM ERROR]This feature, WolframAlpha, has been deprecated due to
        WolframAlpha's strict new updates and policies.
````

I see two ways forward. One is to simply delete the Wolfram code from `calc.py`. The other is to move it to a new module in sopel-extras where it can be rewritten to comply with the new API requirements.

I'd be happy to take a stab at a rewrite, but before I make any PRs I'd like to see which direction people want to go with this, especially @embolalia.

elad661:
I think wolfram is a popular enough functionality so it should live in its own repository, and not in extras (which is mostly a pile of unmaintained stuff of little use).

sopel allows installing modules via PyPi, so there's no reason to put something like that in -extras.

The code in calc.py never called wolfram directly, but rather via a 3rd party service, so I doubt any of it will be useful for a new module.

I'm sure many people would appreciate a working Wolfram Alpha module, so if you have time to try writing one, it'd be great. (we could also probably import it into the sopel github organization later if you really want it to look "official")

maxpowa:
I've already written[ a module ](https://github.com/maxpowa/inumuta-modules/blob/sopel-compat/wolfram.py)to interface directly with Wolfram|Alpha, if someone wants to do the work of throwing it into a cookiecutter repo I'd be fine with that. It does require an API key and it uses the `wolframalpha` PyPi package. I'd do the cookiecutter myself but I'm currently pretty busy with other things.

dgw:
I'm working on the cookiecutter bit right now, @maxpowa. Created the repo and added you as a collaborator. Would you like me to try and `git-filter-branch` the history of this particular module out to preserve it in the new repo, or just link from the file header to the original code?

maxpowa:
Hmm, I think its simple enough code that we can just link to the original in the header. Thanks for putting together the cookiecutter.

dgw:
OK, it's basically done. I'm going to finish tweaking the README, but once that's done it should be good to go. I already tested `pip install` with a ZIP of the master branch straight from GitHub and it Just Works™ once I put the `app_id` in the config and restart.

If there are no objections I'll push a release to PyPI after I finish and commit the README.

dgw:
Having given ample time for objections, and seeing that there have been none, the package's initial release is now [available on PyPI](https://pypi.python.org/pypi/sopel-modules.wolfram).

[...]

PR #1045, opened 2016-03-16
[calc] Remove .wa, as API now requires a key

dgw:
Will be moved to an external module that supports the new API, which I intend to cookiecutter from @maxpowa's code later tonight after I get to a computer. See #1041


dgw:
External module now [available on PyPI](https://pypi.python.org/pypi/sopel-modules.wolfram).

PR #1051, opened 2016-03-27
fix .set command for non filename attributes

neonobjclash:
The parse() method only expects one argument, whereas the .set command gives it two. This is my fix for that.


maxpowa:
Good catch, thanks.

PR #1052, opened 2016-03-29
Add nick alias config option for $nick and $nickname in @rule regex

calebj:
I have a bot named Pangubot, and sometimes people refer to it as just pangu. With this code, it recognizes its name either way.

### DIFF
--- a/sopel/loader.py
+++ b/sopel/loader.py
@@ -115,11 +115,14 @@ def compile_rule(nick, pattern, alias_nicks):
     if isinstance(pattern, _regex_type):
         return pattern
 
-    nick = re.escape(nick)
     if alias_nicks:
-        nicks = alias_nicks
+        nicks = alias_nicks.copy()
         nicks.append(nick)
+        nicks = map(re.escape, nicks)
         nick = '(%s)' % '|'.join(nicks)
+    else:
+        nick = re.escape(nick)
+
     pattern = pattern.replace('$nickname', nick)
     pattern = pattern.replace('$nick', r'{}[,:]\s+'.format(nick))
     flags = re.IGNORECASE
--- a/sopel/loader.py
+++ b/sopel/loader.py
@@ -117,7 +117,9 @@ def compile_rule(nick, pattern, alias_nicks):
 
     nick = re.escape(nick)
     if alias_nicks:
-        nick = '(%s)' % '|'.join(alias_nicks)
+        nicks = alias_nicks
+        nicks.append(nick)
+        nick = '(%s)' % '|'.join(nicks)
     pattern = pattern.replace('$nickname', nick)
     pattern = pattern.replace('$nick', r'{}[,:]\s+'.format(nick))
     flags = re.IGNORECASE
--- a/sopel/loader.py
+++ b/sopel/loader.py
@@ -110,12 +110,14 @@ def enumerate_modules(config, show_all=False):
     return modules
 
 
-def compile_rule(nick, pattern):
+def compile_rule(nick, pattern, alias_nicks):
     # Not sure why this happens on reloads, but it shouldn't cause problems…
     if isinstance(pattern, _regex_type):
         return pattern
 
     nick = re.escape(nick)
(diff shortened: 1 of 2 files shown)


sopel/loader.py
```diff
@@ -116,6 +116,9 @@ def compile_rule(nick, pattern):
         return pattern
 
     nick = re.escape(nick)
+    botflt = re.match('(?i)(.*)([-_.]?bot$)', nick)
```
dgw: What does this variable name mean?

dgw:
I'm not sure this sort of thing will be wanted in core, but at the same time I can't think of a way to do it in a module. (That said, I've also been awake for far too long, so maybe fatigue is keeping me from an idea that would work.)

sopel/loader.py
neonobjclash: bot filter

dgw: Why not name it `bot_filter` then? It's much easier to read, and there's no penalty for long variable names in Python.

neonobjclash:
I don't think you can. I haven't played with decorators, but a custom @rule might work. This code is called from several other places, which is why I put it here instead of trying to write another @rule for the module.

sopel/loader.py
neonobjclash: Done.

maxpowa:
I agree with @dgw on this, not really sure it's something that fits into core. A lot of IRC clients support tab completion too, so I think that users being lazy isn't really a reason to make the existing behavior inconsistent. 

Something that might be better is adding an 'aliases' config option which expands the nick regex or something along those lines.

neonobjclash:
Changed to use alias config

maxpowa:
LGTM, can you please squash the commits?

neonobjclash:
Done.

dgw:
:+1:

sopel/loader.py
```diff
@@ -110,12 +110,16 @@ def enumerate_modules(config, show_all=False):
...
+def compile_rule(nick, pattern, alias_nicks):
     # Not sure why this happens on reloads, but it shouldn't cause problems…
     if isinstance(pattern, _regex_type):
         return pattern
 
     nick = re.escape(nick)
+    if alias_nicks:
+        nicks = alias_nicks
+        nicks.append(nick)
+        nick = '(%s)' % '|'.join(nicks)
```
maxpowa: Just realized a situation where this could fail, if one of the alias nicks contains a pipe or any other regex special character, it may parse the nick differently. For example, if someone were to add `sopel|gone` as an alias nick, it wouldn't match `sopel|gone` but `sopel` and `gone` individually.

calebj:
Addressed the issue @maxpowa highlighted without changing the line in question and rebased.

dgw:
Hey @calebj, could you amend the last commit (690ffec) to remove the whitespace at `loader.py` L125? I think the linter was made more strict since your last push to this branch. This is good to merge once that's done (and I can easily amend it myself if it's not convenient for you to do so, just shout).

I'm glad I tested the merge locally before hitting the big green button. Already "broke" the build once this week due to a style issue, and I don't want another angry email from Travis. :joy_cat:

dgw:
Bing is such a diva, I swear. Kicking Travis (restarting the failed job) fixed it though. I probably ought to just reduce the check to making sure `sopel.chat` is present…

Thanks for rebasing, this will go in soon unless I find an issue in manual testing at the last minute.

dgw:
Found no issues in manual testing, so decided to :shipit: ship it. Thanks!

dgw:
As the above references from other issues & commits can attest, I didn't quite test this thoroughly enough. So far I've discovered that this patch:

  1. Changed match groups in `trigger` passed to nickname command callables when `alias_nicks` actually contains at least one alias
  2. Doesn't work under Python 2 because `list.copy()` wasn't added until Python 3.3

I've fixed both issues locally (they're trivial code changes). Better test coverage will come along with the fixes when I'm ready to push them up.

Issue #831, opened 2015-07-05
Jobs weird on reload?


embolalia:
5.x reload uses `bind_commands`, which [clears the queue](https://github.com/embolalia/willie/blob/5.x/willie/bot.py#L269). This (I think) cancels all the jobs, meaning all scheduled stuff probably breaks when any module is reloaded. The 6.0 branch doesn't do anything with jobs, so old versions of jobs keep running. I haven't verified this behavior, but it's probably the case.

embolalia:
Tweaked so that this is no longer any worse than what previous versions had, so this no longer blocks 6.0 from being released. It's still not right, though, so I'm leaving it open.

firerogue:
Might be a complicated fix, but what if we saved some sort of callback to every job a module sets up so that when the module is unloaded/reloaded we can kill it? add some function to the JobScheduler that takes a Job object (or the ID of said object, or some other unique indicator) and removes all instances of said job from the queue?

Exirel:
Will be fixed in #1479.

Exirel:
It was already fixed in 6.5.x (or was it 6.6.x?, @dgw will know) but not yet fully available in master.

dgw:
> It was already fixed in 6.5.x (or was it 6.6.x?, @dgw will know) but not yet fully available in master.

Nope, 6.6.x still clears the whole queue:

https://github.com/sopel-irc/sopel/blob/684c0b48298618dac7234b759923726708cb28d1/sopel/bot.py#L219-L222

Nevertheless, it's fixed on master now with #1479 merged, so… yay.

PR #1053, opened 2016-03-30
Refactor module (un|re)loading

calebj:
_Commit summary stolen for use as PR description, since @calebj didn't enter one._ — @dgw

````
Core changes:
Now there's a single place to (un)register modules and keep track of
registered modules, called by both reload.py and bot.setup(). Please
let me know if I forgot to unload/reset something. It hasn't broken
on me yet.

The command categories are structured slightly differently. Instead
of storing only the first command alias, it stores a list of all of
them. I also changed the help.py module to work with this structure.

I added an unload command to the reload module. Since the bot uses a
seperate structure to store loaded modules, the user can't do something
stupid like 'unload sopel'.

Scheduler changes:
Not much here, just added another method that removes a job. Sadly,
PriorityQueue doesn't support this natively, so I just rebuild the
queue, skipping the job in question. This is used in bot.unregister()

Reload module changes:
Since the command is included in trigger.groups(2), making seperate
rules for PM and nick commands that call the same function won't work
without ignoring the first word when triggered by nickname_commands.
It's easier to make it use standalone command syntax, which works in
both PMs and in a channel.
Also, I replaced the admin tests with their decorator counterparts.
````

### DIFF
--- a/sopel/bot.py
+++ b/sopel/bot.py
@@ -280,7 +280,7 @@ class Sopel(irc.Bot):
             self.memory['url_callbacks'].pop(func.url_regex, None)
 
     def unregister_module(self, module):
-        relevant_parts = sopel.loader.clean_module(module, self.config)
+        relevant_parts = sopel.loader.clean_module(module, self.config, modify=False)
         self.unregister(*relevant_parts)
 
         if hasattr(module, "teardown"):
--- a/sopel/bot.py
+++ b/sopel/bot.py
@@ -17,7 +17,7 @@ import time
 from sopel import tools
 from sopel import irc
 from sopel.db import SopelDB
-from sopel.tools import stderr, Identifier
+from sopel.tools import stderr, Identifier, iteritems
 import sopel.tools.jobs
 from sopel.trigger import Trigger
 from sopel.module import NOLIMIT
@@ -89,6 +89,12 @@ class Sopel(irc.Bot):
         """A set containing the IRCv3 capabilities that the bot has enabled."""
         self._cap_reqs = dict()
         """A dictionary of capability names to a list of requests"""
+
+        self._modules = dict()
+        """A dictionary of modules currently registered by the bot. sys.modules
+        doesn't work because although del <module> removes it from the namespace,
+        it remains in sys.modules. Thus, we need another way to keep track of 
+        what has been (un)loaded."""
 
         self.privileges = dict()
         """A dictionary of channels to their users and privilege levels
@@ -177,6 +183,7 @@ class Sopel(irc.Bot):
 
             try:
                 module, _ = sopel.loader.load_module(name, path, type_)
+                self.register_module(module)
             except Exception as e:
                 error_count = error_count + 1
                 filename, lineno = tools.get_raising_file_and_line()
@@ -184,23 +191,7 @@ class Sopel(irc.Bot):
                 raising_stmt = "%s:%d" % (rel_path, lineno)
                 stderr("Error loading %s: %s (%s)" % (name, e, raising_stmt))
             else:
-                try:
-                    if hasattr(module, 'setup'):
-                        module.setup(self)
-                    relevant_parts = sopel.loader.clean_module(
-                        module, self.config)
-                except Exception as e:
-                    error_count = error_count + 1
-                    filename, lineno = tools.get_raising_file_and_line()
-                    rel_path = os.path.relpath(
-                        filename, os.path.dirname(__file__)
-                    )
-                    raising_stmt = "%s:%d" % (rel_path, lineno)
-                    stderr("Error in %s setup procedure: %s (%s)"
-                           % (name, e, raising_stmt))
-                else:
-                    self.register(*relevant_parts)
-                    success_count += 1
+                success_count += 1
 
         if len(modules) > 1:  # coretasks is counted
             stderr('\n\nRegistered %d modules,' % (success_count - 1))
@@ -208,20 +199,36 @@ class Sopel(irc.Bot):
         else:
             stderr("Warning: Couldn't load any modules")
 
+    def unregister_module(self, module):
+        for obj_name, obj in iteritems(vars(module)):
+            self.unregister(obj)
+        callables, _, _ = sopel.loader.clean_module(module, self.config)
+        if hasattr(module, "setup"):
+            delattr(module, "setup")
+        del self._modules[module.__name__]
+
     def unregister(self, obj):
         if not callable(obj):
             return
+        if hasattr(obj, 'commands'):
+            module_name = obj.__module__.rsplit('.', 1)[-1]
+            category = getattr(obj, 'category', module_name)
+            self._command_groups[category].remove(obj.commands)
+            if len(self._command_groups[category]) is 0:
+                del self._command_groups[category]
+            for command, docs in obj._docs.items():
+                del self.doc[command]
+
         if hasattr(obj, 'rule'):  # commands and intents have it added
             for rule in obj.rule:
                 callb_list = self._callables[obj.priority][rule]
                 if obj in callb_list:
                     callb_list.remove(obj)
         if hasattr(obj, 'interval'):
-            # TODO this should somehow find the right job to remove, rather than
-            # clearing the entire queue. Issue #831
-            self.scheduler.clear_jobs()
-        if (getattr(obj, '__name__', None) == 'shutdown' and
-                    obj in self.shutdown_methods):
+                job = sopel.tools.jobs.Job(obj.interval, obj)
+                self.scheduler.del_job(job)
+        if (getattr(obj, '__name__', None) == 'shutdown'
+                and obj in self.shutdown_methods):
             self.shutdown_methods.remove(obj)
 
     def register(self, callables, jobs, shutdowns, urls):
@@ -239,7 +246,7 @@ class Sopel(irc.Bot):
                 # TODO doc and make decorator for this. Not sure if this is how
                 # it should work yet, so not making it public for 6.0.
                 category = getattr(callbl, 'category', module_name)
-                self._command_groups[category].append(callbl.commands[0])
+                self._command_groups[category].append(callbl.commands)
             for command, docs in callbl._docs.items():
                 self.doc[command] = docs
         for func in jobs:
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Comment by calebj** [comment]
Dunno if this patch still works, but I rebased it anyway. Will probably test later today.

**Comment by dgw** [comment]
@calebj Are you still available to work on this/#1054? They'll need some rebasing work to catch up with changes to the help module made in the mean time, so I'm holding off on code review for now.

**Comment by calebj** [comment]
Yeah, I'm still available. Should be rebased now.

File: sopel/bot.py [hunk 1]
```diff
@@ -177,49 +183,52 @@ def setup(self):
...
         if len(modules) > 1:  # coretasks is counted
             stderr('\n\nRegistered %d modules,' % (success_count - 1))
             stderr('%d modules failed to load\n\n' % error_count)
         else:
             stderr("Warning: Couldn't load any modules")
 
+    def unregister_module(self, module):
+        for obj_name, obj in iteritems(vars(module)):
+            self.unregister(obj)
+        callables, _, _ = sopel.loader.clean_module(module, self.config)
```
dgw: This causes a `ValueError`; `clean_module()` only returns one value.

sopel/modules/help.py
```diff
@@ -63,6 +63,7 @@ def help(bot, trigger):
             for category, cmds in collections.OrderedDict(sorted(bot.command_groups.items())).items():
                 category = category.upper().ljust(name_length)
                 cmds = set(cmds)  # remove duplicates
+                cmds = ['|'.join(cg) for cg in cmds]
```
dgw: Maybe let's split command-group help into its own PR?

sopel/modules/reload.py
```diff
@@ -11,20 +11,19 @@
...
 import time
-from sopel.tools import iteritems
+from sopel.tools import iteritems, get_raising_file_and_line
 import sopel.loader
 import sopel.module
 import subprocess
+import os
 
 
-@sopel.module.nickname_commands("reload")
```
dgw: I'm not a fan of changing the command behavior just for the hell of it. `Sopel: reload modulename` should continue to work.

sopel/modules/reload.py
```diff
@@ -33,54 +32,40 @@ def f_reload(bot, trigger):
             'medium': collections.defaultdict(list),
             'low': collections.defaultdict(list)
         }
+        bot.shutdown_methods.clear()
+        bot.scheduler.clear_jobs()
```
dgw: With the refactor in `bot.py` to remove only the specified job, why is the `reload` module now clearing the whole queue? Why is `reload` clearing all shutdown methods for the entire bot instead of only the specified module? Keeping #831 in mind, this refactor is a good opportunity to stop clearing things that shouldn't be.

calebj:
I think reloading should work for you now. Both of my PRs are having issues with that URL check, but it tests fine locally. I would like to squash and add a 'fixes' keyword if these are acceptable to merge.

dgw:
Kicked Travis, test passed. Bing can be an utter diva. It'd be easier to fix if it failed consistently…

Squashing sounds great! But not just yet, sorry. Don't have as much time as I'd like right now to go back and do another in-depth test of this, and the moved functions make even reviewing the diff hard. I will, however, put this (and your other rebased PR) at the top of my to-do list, so it shouldn't be more than a few days before I have a chance to look at both of them. :smile_cat:

dgw:
Currently blocked on #1343, a regression caused by me fulfilling the first half of my promise and merging #1052 without (apparently) testing it as thoroughly as I should have. Sorry @calebj, I'll still get to this ASAP.

dgw:
OK, I'm happy enough now with the #1343 situation that this came back to the top of my list.

Merging #1053 (specifically, 72dd79e) directly onto current `master` (currently a0c567c<sup>[1]</sup>) gave me no change in the reloading behavior with a packaged module in my `.modules/` folder. It still adds two more repetitions for every time the packaged module is reloaded, and doesn't reload anything outside the module's `__init__.py` file.

This is the part where I find out how happy _you_ are that I came back to this, @calebj… What would you say to me merging #1314 to get reloading working as expected<sup>[2]</sup> and you rebase this on top? I can warn you that there would be a sizable merge conflict to resolve in `modules/reload.py`—that much I know from trying to test both PRs together.

I've unintentionally discouraged PR authors before by asking them to spend too much time making too many tweaks in too many rounds of feedback, and I don't want to do it again. At the same time, while this refactor looks really nice to have (and enables #1054)—I have an eagerly awaited bug fix (#1314) that's ready to go now but directly conflicts with some of the changes made here.

**But:** Now that I've shared my point of view, please don't let it discourage you! If you'd really rather not have to do a(nother) rebase, I'll have a look at what it would take to merge this first. We can also see about integrating the two PRs somehow, maybe through a conversation on IRC (in `#sopel` or PM).

Let me know? I promise not to merge #1314 (or anything else that I know will conflict with this) until I get your go-ahead.<sup>[3]</sup> :smile_cat:

----

  1. To reproduce my test branch, assuming your local repo is up-to-date with upstream: `git checkout -b test-1053 a0c567c` then `git merge refactorreload`. `git fetch upstream` (or whatever you named the remote for this repo) if needed first, to make sure you have the latest commits from `master`.
  2. See https://github.com/sopel-irc/sopel/issues/1106#issuecomment-378742293 and the later comments of #1056 for context. But to summarize: @ralienpp is using Sopel with some custom modules on a fleet of headless remote systems. Not being able to cleanly reload packaged modules is a real PITA there, so merging #1314 first would help immensely.
  3. Or if I don't hear from you for several weeks. Can't keep @ralienpp waiting forever, can I? :laughing:

dgw:
Reassigning to Sopel 7, along with its sister PR, at least until @calebj responds.

calebj:
Rebased, and appears to be working with core modules, wolfram and youtube on Python 2.7, 3.6, and 3.7 :tada:

I implemented some of the ideas from #1314, but used code I adapted from an existing module in another project. This should fix #1324 and, of course, #1056. #990 still applies cleanly.

@ralienpp, would you mind testing this branch with your own modules?

calebj:
Weird. That happens on my end too... but not if I run it from the repo, or pip install from the repo folder. Something isn't making it to the repo, even though it says my working tree is clean. Will investigate.

calebj:
Found it. you used `#` when you should have used `@`. Running this:
```sh
pip install git+https://github.com/sopel-irc/sopel.git@a970dee4336133bc5c5c9dece33c5b41301c8bed --upgrade
```
works as expected. You should get a `Reloaded: ...` message after the module spec and mtime.

ralienpp:
This is what I get when I run the command with the `#` symbol replaced by a `@`:

```
/pip install git+https://github.com/sopel-irc/sopel.git@a970dee4336133bc5c5c9dece33c5b41301c8bed --upgrade
Collecting git+https://github.com/sopel-irc/sopel.git@a970dee4336133bc5c5c9dece33c5b41301c8bed
  Cloning https://github.com/sopel-irc/sopel.git (to a970dee4336133bc5c5c9dece33c5b41301c8bed) to /tmp/pip-YpPR5y-build
  Could not find a tag or branch 'a970dee4336133bc5c5c9dece33c5b41301c8bed', assuming commit.
fatal: reference is not a tree: a970dee4336133bc5c5c9dece33c5b41301c8bed
Command "git checkout -q a970dee4336133bc5c5c9dece33c5b41301c8bed" failed with error code 128 in /tmp/pip-YpPR5y-build
```

This has happened before, please see the history here: https://github.com/sopel-irc/sopel/issues/1056#issuecomment-379499863

calebj:
Try passing it my fork URL instead:
```sh
pip install --user git+https://github.com/calebj/sopel.git@a970dee4336133bc5c5c9dece33c5b41301c8bed --upgrade
```

ralienpp:
I had to adjust your command and remove the `--user` part, to install it. The experiment was successful, I reloaded it a couple of times - still got the expected behaviour. Thanks!

I will let this picture express my feelings in a more dramatic way https://cdn.theatlantic.com/static/mt/assets/science/JPL.gif

dgw:
Hey @calebj, any ideas on this behavior?

````
15:27:57 <~dgw>  Sopel: reload wolfram
15:27:58 <Sopel>  dgw: <module 'wolfram' from '/home/dgw/.sopel/modules/wolfram/__init__.py'> (version: 2018-08-20 20:27:55)
15:27:59 <Sopel>  Reloaded: wolfram
15:28:01 <~dgw>  .wa2
15:28:01 <Sopel>  wa2 command added
15:28:11 <~dgw>  Sopel: reload wolfram
15:28:12 <Sopel>  dgw: <module 'wolfram' from '/home/dgw/.sopel/modules/wolfram/__init__.py'> (version: 2018-08-20 20:28:09)
15:28:12 <Sopel>  Reloaded: wolfram
15:28:18 <~dgw>  (removed wa2 command)
15:28:20 <~dgw>  .wa2
15:28:20 <Sopel>  wa2 command added
15:28:21 <Sopel>  wa2 command added
15:28:22 <Sopel>  wa2 command added
15:28:23 <Sopel>  wa2 command added
15:28:23 <Sopel>  ...
````

(The last `...` is the repetition protection blocking out a fifth copy of the same line.)

In this test, `.wa2` triggered the following:

````python
@commands('wa2')
def wa2(bot, trigger):
    bot.say('wa2 command added')
````

And to test removal, I simply removed the whole function and its decorators.

FWIW, my reloading overhaul in #1314 also bugs out on removed methods, but it only adds one repetition. Yours seems to add four, and I'm not sure why. (Not that I'm yet sure why mine adds even one…)

calebj:
The whole practice of modifying the command function object directly isn't the best thing to work with, but I fixed it in my latest push.

In a future update, I would advocate moving to a single `@command` decorator that takes keyword arguments for `rule=` and so on. The decorator would then create a new object with the necessary attributes and leave the original function object alone.

dgw:
I'm more than a little squeamish about doing anything like `del sys.modules[name]`… https://justus.science/blog/2015/04/19/sys.modules-is-dangerous.html comes to mind.

Reloading support is probably not worth completely rejiggering the decorators, honestly. Keeping multiple versions of callable objects around doesn't tickle my fancy either. :/

calebj:
The handling [here](https://github.com/Rapptz/discord.py/blob/async/discord/ext/commands/bot.py#L741) has been robust in my experience, and only Sopel modules are removed from `sys.modules`. Full reloading needs a full GC due to the modifications made by changing the function object directly, and the cached item in `sys.modules` prevents that.

dgw:
I need to dig way into this at a computer, hmm. There must be a good way to have our cake and eat it too re: modifying functions during load…

ralienpp:
Having used that patched version for a while, I have found that _the problem persists_, unfortunately. My first thought is that the initial test, where I use `.reload` to trigger the reloading action - is not enough. That test passes, but over time something else happens that leads to this behaviour.

1. What other testing technique could I try?
2. Is there something else I can do to help speed up the resolution of this issue?

dgw:
@ralienpp Does the problem persist with #1314 as well? Because if not, I'm happy to merge that for inclusion in the next release and let work on this refactor continue for the future.

dgw:
So, @HumorBaby is putting a lot of work into the reloading logic, and @Exirel is working on refactoring the plugin/module system itself. Between the two of them, they're likely to duplicate a lot of this PR.

Given that this PR has kind of stalled, I'm going to say it'll wind up getting closed unmerged unless @calebj reappears to continue work on it relatively soon.

dgw:
Closing in favor of #1479

PR #1054, opened 2016-03-30
Add help command that displays docs for all commands in a category

neonobjclash:


sopel/modules/help.py
```diff
@@ -60,6 +60,40 @@ def help(bot, trigger):
...
+            if len(bot._command_groups[name]) > threshold: # 3 commands doesn't always mean 3 lines, but meh
+                if trigger.nick != trigger.sender:  # don't say that if asked in private
+                    bot.reply('The documentation for this module is too long; I\'m sending it to you in a private message.')
+                msgfun = lambda l: bot.msg(trigger.nick, l)
+            else:
+                msgfun = bot.say
+
+            cmds = bot._command_groups[name]
+            for cmd in cmds:
+                cmd_names = name.upper() + ' : ' + '|'.join(cmd)
```
maxpowa: This causes command names to have | between each character, not sure if that's intended.

```
<@maxpowa> .module help
<Sopel> HELP : m|o|d|u|l|e : no documentation available for this command.	
<Sopel> HELP : h|e|l|p : no documentation available for this command.	
```

sopel/modules/help.py
```diff
@@ -60,6 +60,40 @@ def help(bot, trigger):
...
+                cmd = cmd[0] # Docs from first/only alias
```
maxpowa: Breaks documentation for commands because it'll only look for the first character in the command name. In the example below it's only ever looking for `m` and `h` so it never sees the documentation for module or help.
```
<@maxpowa> .module help
<Sopel> HELP : m|o|d|u|l|e : no documentation available for this command.   
<Sopel> HELP : h|e|l|p : no documentation available for this command.   
```

neonobjclash:
The patch assumes that cmds is a list of strings that are aliases for a command, not just a string. That change is in my other PR.

maxpowa:
Blocked by #1053

calebj:
Rebased both this and #1053

dgw:
Reassigning to Sopel 7, along with its sister PR, at least until @calebj responds.

dgw:
Current development direction looks like someone will have to take this and possibly rework it on top of whatever changes happen after @Exirel and @HumorBaby finish reworking the (re)loading of plugins. We're definitely into "let's wait and see" territory, at this point.

dgw:
It's a good call, since @calebj hasn't been active in a while. I worry that we'll lose track of this without the open PR, though… but if it's on your own list, @Exirel, then I will not worry! :grin:

PR #1059, opened 2016-04-09
Support multiple instances with systemd, defaults to 'sopel'

TheReverend403:
Usage: `systemctl start sopel@sopel`

### DIFF
--- a/contrib/sopel@.service
+++ b/contrib/sopel@.service
@@ -2,12 +2,13 @@
 Description=Sopel IRC bot
 Documentation=http://sopel.chat/
 After=network.target
+DefaultInstance=sopel
 
 [Service]
 Type=simple
 User=sopel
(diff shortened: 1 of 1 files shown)


maxpowa:
I don't have much experience with systemd, so it'd be great if someone with experience could review this.

Exirel:
Why not both?

We can keep `sopel.service` file, and provide a `sopel@.service` template unit file too for anyone who need/want one.

With a bit of documentation, I think this should be fine!

dgw:
Decided to use a small part of my Sunday morning on fixing this up. Merged by f2838a0 in modified form, leaving the original `sopel.service` file alone. Unfortunately, modifying the PR means GitHub can't tell it was merged, so this will go into the log as "Closed" instead.

PR #1064, opened 2016-04-12
Add Dockerfile

maxpowa:
This picture is out of date, but it's quite relevant.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,22 @@
+FROM python:alpine
+
+RUN mkdir -p /usr/src/app
+WORKDIR /usr/src/app
+
+RUN adduser -S sopel
+
+COPY requirements.txt /usr/src/app/
+RUN pip install --no-cache-dir -r requirements.txt
+
+COPY . /usr/src/app
+
+USER sopel
+
+# OS X users will crash with permission errors if they use a
+#     volume over /home/sopel/.sopel - nothing we can do about it.
+# Workaround: Don't use OS X
+# https://github.com/boot2docker/boot2docker/issues/581
+# https://github.com/docker/kitematic/issues/351
+VOLUME [ "/home/sopel/.sopel" ]
+
+CMD [ "python", "./sopel.py" ]


RX14:
:+1: i've been memed

Koshatul:
My PR  #1095 is helpful for this, so you don't need to put the configuration in the sopel.cfg and build it in, you can use environment variables.

Dockerfile
```diff
@@ -0,0 +1,22 @@
+FROM python:alpine
+
+RUN mkdir -p /usr/src/app
+WORKDIR /usr/src/app
+
+RUN adduser -S sopel
+
+COPY requirements.txt /usr/src/app/
+RUN pip install --no-cache-dir -r requirements.txt
```
HumorBaby: This `RUN pip ...` is no longer sufficient for sopel to run. `pyenchant` requires the `enchant` library, so a `apk add --no-cache enchant` will need to be added somewhere

Dockerfile
```diff
@@ -0,0 +1,22 @@
+FROM python:alpine
+
+RUN mkdir -p /usr/src/app
+WORKDIR /usr/src/app
+
+RUN adduser -S sopel
+
+COPY requirements.txt /usr/src/app/
```
HumorBaby: his can be changed to `COPY . /usr/src/app` since `requirements.txt` is already in `.`. Will save on image size (a super tiny amount, but one less layer)

Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+RUN pip install --no-cache-dir -r requirements.txt
+
+COPY . /usr/src/app
+
+USER sopel
+
+# OS X users will crash with permission errors if they use a
```
HumorBaby: Is this still valid since the transition from boot2docker to Docker Machine for Mac?

Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+#     volume over /home/sopel/.sopel - nothing we can do about it.
+# Workaround: Don't use OS X
+# https://github.com/boot2docker/boot2docker/issues/581
+# https://github.com/docker/kitematic/issues/351
+VOLUME [ "/home/sopel/.sopel" ]
```
HumorBaby: Because ~/.sopel does not exist when the `VOLUME` directive is run, docker created the directory owned by `root`. This leads to the following error on run: `ValueError: Invalid value for core.logdir: Value must be an existing or creatable directory.` May be best to just make it `VOLUME [ "/home/sopel" ]` since it should be empty (or super close to); or could explicitly `mkdir /home/sopel/.sopel` before the current `VOLUME` call (but after  `USER sopel`)

HumorBaby:
@maxpowa: would love to see a Dockerfile with Sopel 7.0.0, so in the spirit of making that happen, I am reaching out with some changes that will need to be made to it for it to work since it was initially submitted.  I have opened maxpowa/Inumuta#14 to address these.

@maxpowa do you think you will be able to follow up on this? 

Thanks! :smile:

HumorBaby:
Just as a note, travis-ci is failing because now `enchant` lib is needed to run sopel. This was a big reason in itself for updated the Dockerfile here.

dgw:
Set up a new [sopel-irc/sopel-docker](https://github.com/sopel-irc/sopel-docker) repo here and a [Docker Hub entry for Sopel](https://hub.docker.com/r/sopelirc/sopel), which @HumorBaby will develop and manage because I know nothing about Docker.

@maxpowa Thank you for opening this PR, even though it ultimately never got merged. Having it around seems to have motivated someone else to pick up the torch!

Issue #718, opened 2015-01-20
Allow willie.module.rate To Apply To All Nicks


prestonconnors:
Hello,

I think a nice feature in willie would have some flag that is applicable to willie.module.rate to allow rate limiting of certain commands to only be activated by any/all nicks.

Example situation is a large chat room where I would like the 'hello' command to bot.reply('hello world!') only once every 300 seconds no matter how many people typed '.hello' after the first person activated the '.hello' command. If there is a current way of doing this please let me know and disregard this feature request!

hroafelme:
This would be very good. I can't belive the bot does not have a global post rate. 

When you run the bot in a channel with over 1000 users it can be crazy, for example twich.tv

tyrope:
@hroafelme The bot DOES have a global rate limiter. But it limits the entirety of the bot's IRC output.

dgw:
How about a new `@globalrate` decorator?

PR #1065, opened 2016-04-16
Add global and channel rate limits

neonobjclash:
Modify the @rate() decorator to support a second and third argument for channel and global rate limits.

### DIFF
--- a/sopel/bot.py
+++ b/sopel/bot.py
@@ -428,13 +428,13 @@ class Sopel(irc.Bot):
 
         if not trigger.admin and not func.unblockable:
             usertimediff = current_time - self._times[nick][func]
-            if func.user_rate > 0 and usertimediff < func.user_rate and \
+            if func.rate > 0 and usertimediff < func.rate and \
                     func in self._times[nick]:
                 self._times[nick][func] = current_time
                 LOGGER.info(
                     "%s prevented from using %s in %s due to user limit: %d < %d",
                     trigger.nick, func.__name__, trigger.sender, usertimediff,
-                    func.user_rate
+                    func.rate
                 )
                 return
             globaltimediff = current_time - self._times[self.nick][func]
--- a/sopel/bot.py
+++ b/sopel/bot.py
@@ -432,7 +432,7 @@ class Sopel(irc.Bot):
             if func in self._times[nick]:
                 usertimediff = current_time - self._times[nick][func]
                 if func.rate > 0 and usertimediff < func.rate:
-                    self._times[nick][func] = current_time
+                    #self._times[nick][func] = current_time
                     LOGGER.info(
                         "%s prevented from using %s in %s due to user limit: %d < %d",
                         trigger.nick, func.__name__, trigger.sender, usertimediff,
@@ -442,7 +442,7 @@ class Sopel(irc.Bot):
             if func in self._times[self.nick]:
                 globaltimediff = current_time - self._times[self.nick][func]
                 if func.global_rate > 0 and globaltimediff < func.global_rate:
-                    self._times[self.nick][func] = current_time
+                    #self._times[self.nick][func] = current_time
                     LOGGER.info(
                         "%s prevented from using %s in %s due to global limit: %d < %d",
                         trigger.nick, func.__name__, trigger.sender, globaltimediff,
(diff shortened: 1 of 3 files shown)


sopel/module.py
```diff
@@ -248,8 +248,10 @@ def add_attribute(function):
     return add_attribute
 
 
-def rate(value):
-    """Decorate a function to limit how often a single user may trigger it.
+def rate(user_rate, channel_rate=0, global_rate=0):
```
dgw: Any particular reason not to also specify a default value for `user_rate`? It's the first positional argument, so it should remain backward-compatible without requiring any particular limit to be specified in new modules.

dgw:
Should close #718 if/when merged.

I've been wanting these for a while. :+1:

Would be nice to be able to specify a message for the bot to send via `NOTICE` to the user if their command is rate-limited, but that should probably be discussed and implemented in a future PR.

sopel/module.py
neonobjclash: No particular reason, other than that I don't know what a good default would be. If zero, that's the same as having no @rate decorator at all.

dgw: It would, however, allow using the decorator like `@rate(channel_rate=300)` to limit a command per-channel without having to specify a `0` for the first argument.

I'd bet whoever owns the username "rate" on Github is going to get mighty tired of this thread.

sopel/loader.py
```diff
@@ -158,7 +158,9 @@ def clean_callable(func, config):
     func.unblockable = getattr(func, 'unblockable', False)
     func.priority = getattr(func, 'priority', 'medium')
     func.thread = getattr(func, 'thread', True)
-    func.rate = getattr(func, 'rate', 0)
```
maxpowa: Remove `rate` will break backwards compatibility with jenni modules. Maybe alias `rate` to `user_rate` in the getattr?

neonobjclash:
I was trying to make the function rate attributes unambiguous, but apparently that breaks backwards-compatibility with jenni. I restored the func.rate name and added a default so non-user rates can be specified alone with a kwarg.

sopel/module.py
neonobjclash: Didn't think of that; It was 3 in the morning, haha. Fixed now.

maxpowa:
Hmm, not really a huge thing but writing `@rate(user_rate=5, channel_rate=10)` is a bit repetitive, maybe we could drop the `_rate` suffixes on the kwargs. `@rate(user=5, channel=10)` feels a bit more smoother IMO.

neonobjclash:
Sadly, `global` is a keyword in Python. If non-uniformity is fine, I can make the change.

maxpowa:
Maybe `server` would work instead of `global`?

neonobjclash:
I literally edited my post to say that just before you commented, kek. Done.

dgw:
I just had the thought that it might make more sense to use `network` instead of `server`, but at this point it's splitting hairs.

maxpowa:
Found something: last command usage is reset every time the user uses the command, even if they were prevented from using. If you have verbose logging on you'll clearly see this reflected in stdio, simply run a rate limited command within the rate period from a non-admin nick.

neonobjclash:
That behavior was present before I added this, I believe. I'll fix it

sopel/bot.py
```diff
@@ -453,7 +453,7 @@ def call(self, func, sopel, trigger):
             if not trigger.is_privmsg and func in self._times[trigger.sender]:
                 chantimediff = current_time - self._times[trigger.sender][func]
                 if func.channel_rate > 0 and chantimediff < func.channel_rate:
-                        self._times[trigger.sender][func] = current_time
+                        #self._times[trigger.sender][func] = current_time
                         LOGGER.info(
                             "%s prevented from using %s in %s due to channel limit: %d < %d",
                             trigger.nick, func.__name__, trigger.sender, timediff,
```
maxpowa: `timediff` is undefined, should be `chantimediff`

sopel/bot.py
```diff
@@ -420,22 +420,46 @@ def reply(self, message, destination=None, reply_to=None, notice=False):
...
+                    )
+                    return
+
+            if not trigger.is_privmsg and func in self._times[trigger.sender]:
+                chantimediff = current_time - self._times[trigger.sender][func]
+                if func.channel_rate > 0 and chantimediff < func.channel_rate:
+                        #self._times[trigger.sender][func] = current_time
+                        LOGGER.info(
+                            "%s prevented from using %s in %s due to channel limit: %d < %d",
+                            trigger.nick, func.__name__, trigger.sender, timediff,
```
maxpowa: `timediff` is undefined, should be `chantimediff`

maxpowa:
Indentation is fudged from L456-462 in bot.py, looks like its just an extra indent.

neonobjclash:
Derp. Fixed now

PR #1070, opened 2016-04-20
slight change to avoid ad results

dasu:
A slight regex change to avoid yahoo ad results from duck duck go if it ends up using the HTML search

### DIFF
--- a/sopel/modules/search.py
+++ b/sopel/modules/search.py
@@ -33,7 +33,7 @@ def bing_search(query, lang='en-GB'):
     if m:
         return m.group(1)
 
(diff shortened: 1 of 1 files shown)


dasu:
Forgot to mention: since sopel.web is deprecated, I can rewrite this to use requests instead.

maxpowa:
Might want to `https?` because I'm not sure if `r.search.yahoo.com` URLs are always http

dasu:
From what I've seen I've never ran into an https, but doesn't hurt to have a check for it.  Thanks.

PR #1071, opened 2016-04-24
Made it so Sopel doesn't automatically message who triggered an error.

j-crit:
Also added core config settings for toggling whether to message the logging channel (if it is set) or to use the original behavior of messaging the sender.

I'm not sure why Sopel automatically messages the sender (as it's a channel most of the time), since it can get pretty disruptive pretty quickly.

### DIFF
--- a/sopel/config/core_section.py
+++ b/sopel/config/core_section.py
@@ -178,7 +178,7 @@ class CoreSection(StaticSection):
     It is a regular expression (so the default, ``\.``, means commands start
     with a period), though using capturing groups will create problems."""
 
-    reply_errors = ValidatedAttribute('reply_errors',bool,default=False)
+    reply_errors = ValidatedAttribute('reply_errors', bool, default=True)
     """Whether to message the sender of a message that triggered an error with the exception."""
 
     throttle_join = ValidatedAttribute('throttle_join', int)
--- a/sopel/config/core_section.py
+++ b/sopel/config/core_section.py
@@ -178,7 +178,7 @@ class CoreSection(StaticSection):
     It is a regular expression (so the default, ``\.``, means commands start
     with a period), though using capturing groups will create problems."""
 
-    reply_errors = ValidatedAttribute('message_logs_sender',bool,default=False)
+    reply_errors = ValidatedAttribute('reply_errors',bool,default=False)
     """Whether to message the sender of a message that triggered an error with the exception."""
 
     throttle_join = ValidatedAttribute('throttle_join', int)
--- a/sopel/config/core_section.py
+++ b/sopel/config/core_section.py
@@ -131,6 +131,12 @@ class CoreSection(StaticSection):
                                      'DEBUG'],
                                     'WARNING')
     """The lowest severity of logs to display."""
+    
+    message_logs = ValidatedAttribute('log_to_channel',bool,default=True)
+    """Whether to message the logging channel with thrown exceptions."""
+    
+    message_logs_sender = ValidatedAttribute('message_logs_sender',bool,default=False)
+    """Whether to message the sender of a message that triggered an error with the exception."""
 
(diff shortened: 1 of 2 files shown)


sopel/irc.py
```diff
@@ -367,12 +367,18 @@ def error(self, trigger=None):
...
+            if trigger and self.config.core.message_logs and self.config.core.logging_channel is not None:
+                self.msg(self.config.logging_channel, signature)
+                if self.config.core.message_logs_sender:
+                    self.msg(trigger.sender,signature)
         except Exception as e:
-            if trigger and trigger.sender is not None:
-                self.msg(trigger.sender, "Got an error.")
+            if trigger:
+                if self.config.core.message_logs and self.config.core.logging_channel is not None:
+                    self.msg(self.config.core.logging_channel,"Got an error.")
```
maxpowa: No need for this line, `LOGGER.error` is sent to the `logging_channel` just below and that's more verbose than this is.

sopel/irc.py
```diff
@@ -367,12 +367,18 @@ def error(self, trigger=None):
             except Exception as e:
                 stderr("Could not save full traceback!")
                 LOGGER.error("Could not save traceback from %s to file: %s", trigger.sender, str(e))
-
-            if trigger and trigger.sender is not None:
-                self.msg(trigger.sender, signature)
+            
+            # Errors will pass silently otherwise, not sure how to handle this
+            if trigger and self.config.core.message_logs and self.config.core.logging_channel is not None:
+                self.msg(self.config.logging_channel, signature)
```
maxpowa: Same thing here, this should be done via `LOGGER`.

sopel/config/core_section.py
```diff
@@ -131,6 +131,12 @@ def homedir(self):
                                      'DEBUG'],
                                     'WARNING')
     """The lowest severity of logs to display."""
+    
+    message_logs = ValidatedAttribute('log_to_channel',bool,default=True)
```
maxpowa: This doesn't appear to be used at all.

j-crit:
Removed redundant "message_logs_sender" lines and the relevant config function, so the only thing affected now is whether logs get printed to message senders, without having to change the Sopel's entire  global logging setting.

sopel/config/core_section.py
```diff
@@ -178,6 +178,9 @@ def homedir(self):
     It is a regular expression (so the default, ``\.``, means commands start
     with a period), though using capturing groups will create problems."""
 
+    reply_errors = ValidatedAttribute('message_logs_sender',bool,default=False)
```
maxpowa: why `message_logs_sender` if you're calling it `reply_errors` internally?

j-crit: I thought I changed that, too, dang. Thanks!

maxpowa: What I meant is that the name should be consistent across the section - if the variable name is `reply_errors`, the ValidatedAttribute should use the same for the key.

j-crit: Ah, I see it now. Thanks.

maxpowa:
There's some spacing issues you can see in the diff, please resolve those and then this should be good.

sopel/config/core_section.py
```diff
@@ -178,6 +178,9 @@ def homedir(self):
     It is a regular expression (so the default, ``\.``, means commands start
     with a period), though using capturing groups will create problems."""
 
+    reply_errors = ValidatedAttribute('reply_errors',bool,default=False)
```
maxpowa: As much as I don't like it either, the default for this should be `True`. I only say this to stay consistent with existing configurations.

j-crit: That makes sense.

Ah, I didn't realize that was the convention until rereading the other definitions. Thanks.

PR #1072, opened 2016-05-02
irc/bot: implement echo-message support

maxpowa:
Allows modules to listen in on what's being sent by Sopel. Compatible with all the other decorators, as it goes through all the same logic. If `echo-message` is supported on the server and is toggled on by a module, only functions with the `@sopel.module.echo(True)` decorator will receive those echo messages. If `echo-message` isn't enabled, the same behavior still exists but the hostmask will not be as accurate.

### DIFF
--- a/sopel/bot.py
+++ b/sopel/bot.py
@@ -484,6 +484,9 @@ class Sopel(irc.Bot):
                     if (hasattr(func, 'intents') and
                             trigger.tags.get('intent') not in func.intents):
                         continue
+                    if (not (hasattr(func, 'echo') and func.echo is True) and
+                            trigger.nick.lower() == self.nick.lower()):
+                        continue
                     if func.thread:
                         targs = (func, wrapper, trigger)
                         t = threading.Thread(target=self.call, args=targs)
--- a/sopel/irc.py
+++ b/sopel/irc.py
@@ -148,6 +148,16 @@ class Bot(asynchat.async_chat):
             self.send(temp.encode('utf-8'))
         finally:
             self.writing_lock.release()
+        # Simulate echo-message
+        if 'echo-message' not in self.enabled_capabilities:
+            # Since we have no way of knowing the hostmask the IRC server uses
+            # for us, we'll just use something reasonable
+            host = 'localhost'
+            if self.config.core.bind_host:
+                host = self.config.core.bind_host
+            pretrigger = PreTrigger(self.nick, ':{0}!{1}@{2} {3}'
+                .format(self.nick, self.user, host, temp))
+            self.dispatch(pretrigger)
 
     def run(self, host, port=6667):
         try:
--- a/sopel/module.py
+++ b/sopel/module.py
@@ -115,6 +115,23 @@ def thread(value):
     return add_attribute
 
 
+def echo(value=True):
+    """Decorate a function to specify if it should receive echo messages.
+
+    This decorator can be used to listen in on the messages that Sopel is
+    sending and react accordingly.
+
+    Args:
+        value: Either True or False. If True the function will receive echo
+            messages, and if False only messages received from other users.
+
+    """
+    def add_attribute(function):
+        function.echo = value
+        return function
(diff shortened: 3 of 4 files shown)


dgw:
Awesome! My only thought is a concern that there will be increased risk of modules generating loops if this particular decorator defaults to `True`. Would it not be better to default `False` and let only modules that depend on observing the bot's output (loggers, for example) override that?

maxpowa:
Close, but it's default `True` when decorated, so if you have an undecorated function it won't get echos. Its just when you decorate with `@sopel.module.echo()` it assumes true.

dgw:
Ah, OK. Still possible for module authors to write confusing code that doesn't explicitly specify `True`, but modules that don't use the decorator won't get echos. Just me being an idiot. :+1:

maxpowa:
I just realized that this isn't really correct. As it is right now, it will echo every type of message, but the echo message spec only includes `NOTICE` and `PRIVMSG`. I will update this PR when I get time.

embolalia:
I like the idea, like seriously. This is awesome and has been desired for ages. But I don't think the `@echo` decorator should take a value. The default should be definitely not to get echoes, meaning the decorator only needs to mean "yup this thing can get echo messages". But other than that, this is the perfect solution.

sopel/bot.py
```diff
@@ -484,6 +484,9 @@ def dispatch(self, pretrigger):
                     if (hasattr(func, 'intents') and
                             trigger.tags.get('intent') not in func.intents):
                         continue
+                    if (not (hasattr(func, 'echo') and func.echo is True) and
+                            trigger.nick.lower() == self.nick.lower()):
+                        continue
```
dgw: I'd put this check first in the enclosing `for` loop, but the order isn't really an issue tbh.

sopel/irc.py
```diff
@@ -148,6 +148,16 @@ def write(self, args, text=None):
...
             self.writing_lock.release()
+        # Simulate echo-message
+        if 'echo-message' not in self.enabled_capabilities:
+            # Since we have no way of knowing the hostmask the IRC server uses
+            # for us, we'll just use something reasonable
+            host = 'localhost'
+            if self.config.core.bind_host:
+                host = self.config.core.bind_host
+            pretrigger = PreTrigger(self.nick, ':{0}!{1}@{2} {3}'
+                .format(self.nick, self.user, host, temp))
```
dgw: This line made me notice the overuse of a variable named `temp` in this function. Renaming that to something sensible should happen after this is merged.

sopel/module.py
```diff
@@ -115,6 +115,23 @@ def add_attribute(function):
     return add_attribute
 
 
+def echo(value):
```
dgw: I agree [with @embolalia](https://github.com/sopel-irc/sopel/pull/1072#issuecomment-258612306):

> I don't think the `@echo` decorator should take a value. The default should be definitely not to get echoes, meaning the decorator only needs to mean "yup this thing can get echo messages".

This decorator should just set the `echo` function attribute to `True` if called. That's all it needs to do.

sopel/irc.py
```diff
@@ -148,6 +148,16 @@ def write(self, args, text=None):
             self.send(temp.encode('utf-8'))
         finally:
             self.writing_lock.release()
+        # Simulate echo-message
+        if 'echo-message' not in self.enabled_capabilities:
```
dgw: As [@maxpowa mentioned](https://github.com/sopel-irc/sopel/pull/1072#issuecomment-221168631) in the main PR comments, "As it is right now, it will echo every type of message, but the echo message spec only includes `NOTICE` and `PRIVMSG`." Therefore, this block should only fire if the message is one of those types.

dgw:
@maxpowa Would you be willing to allow edits from maintainers on this PR? I just rebased it, but can't push the updated history back here (probably because the PR is so old).

dgw:
Closing in favor of rebased/updated version (#1470).

PR #1083, opened 2016-05-25
Fixed an issue with extended-join and recognizing accounts

andir:
Sopel would only add information about the user to the channel and not
the "global" memory which is being used to populate the trigger object
in command handlers.

I'm not very sure this is the right approach to this issue. Would it be better to just check the channel's user list for a valid user/account info?


maxpowa:
Duplicate of #1014

PR #1084, opened 2016-05-26
Added op, deop, voice and devoice functions.

tom29739:
A quick addition to add those functions to Sopel.

### DIFF
--- a/sopel/modules/adminchannel.py
+++ b/sopel/modules/adminchannel.py
@@ -287,7 +287,7 @@ def op(bot, trigger):
             return bot.reply("I'm not a channel operator!")
         elif trigger.group(2) == None:
             return bot.reply("You need to specify a nickname/channel")
-        elif is_privmsg:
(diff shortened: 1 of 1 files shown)


maxpowa:
These had been removed in a previous commit as they were mostly useless. https://github.com/sopel-irc/sopel/commit/902e715b308b7f3a663c961a64a3481af67f6ebf

tom29739:
How are they mostly useless? I would use them, as would a few other users judging from #sopel.

embolalia:
The problem with this is that someone with admin on the bot is not necessarily supposed to be an op in the channel. Obviously the owner can make it do anything, but opening up to admins means a larger group that needs to be trusted. As such, I don't think this is a good candidate for inclusion. However, as has been said, there is definitely demand for it, and I think making this installable through PyPI would be a great idea. It's definitely a thing people want bots to do, it's just not something the bot should be able to do by default.

PR #1095, opened 2016-06-28
Allow environment variables to replace configuration options

Koshatul:
I added in checks to allow environment variables to be defined to override core configuration options.
so CORE_AUTH_PASSWORD=blah could be used instead of

```
[core]
auth_password=blah
```


Koshatul:
My bad github skills meant I did a merge from my master instead of from a branch, closed and opened a new PR #1096

PR #1096, opened 2016-06-28
Allow environment variables to replace configuration options

Koshatul:
I added in checks to allow environment variables to be defined to override core configuration options.
so CORE_AUTH_PASSWORD=blah could be used instead of

```
[core]
auth_password=blah
```

### DIFF
--- a/sopel/config/types.py
+++ b/sopel/config/types.py
@@ -56,6 +56,9 @@ class StaticSection(object):
         self._section_name = section_name
         for value in dir(self):
             try:
+                env_name = '%s_%s' % (self._section_name.upper(), value.upper())
+                if os.environ.get(env_name) is not None:
+                    setattr(self, value, os.environ.get(env_name))
                 getattr(self, value)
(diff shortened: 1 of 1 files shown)


sopel/config/types.py
```diff
@@ -143,7 +143,10 @@ def __get__(self, instance, owner=None):
             # instance here.
             return self
 
-        if instance._parser.has_option(instance._section_name, self.name):
+        env_name = '%s_%s' % (instance._section_name.upper(), self.name.upper())
+        if os.environ.get(env_name) is not None:
```
maxpowa: Should probably just do `if env_name in os.environ:` instead here.

dgw:
Thanks!

> Sorry for the slow reply

On a 2½-year-old pull request, I consider a response in three hours to be pretty fast, actually. :grin:

koshatul:
I'll have to check my notifications, I think someone else has been waiting a couple of days for me to do something :)

Issue #1101, opened 2016-07-15
bot.db.unalias_nick() behavior does not match docs


dgw:
>     Raises ValueError if there is not at least one other nick in the group.

The ValueError is never raised:

````
<dgw> .nickunmerge ThisIsATestNick
<Sopel> Removed ThisIsATestNick from nick group 1497.
<dgw> that's not supposed to work
<ThisIsATestNick> blerg
<dgw> .nickunmerge ThisIsATestNick
<Sopel> Removed ThisIsATestNick from nick group 1498.
````

I have traced this issue to a logic error, and will submit a proposed fix when I'm back in front of my laptop later tonight.

If it's more desirable to change the documentation than the behavior, I'd be happy to rewrite that instead—but I believe that the described behavior is correct, and the observed behavior is a bug.

PR #1102, opened 2016-07-15
[db] Fix group nick count check in unalias_nick()

dgw:
As described in #1101, this DB API call will currently remove even a nickname that is the only member of its group, contrary to what the documentation string says. This fixes the problem.


maxpowa:
I think this should be `<=` rather than `==` for clarity. A ValueError is raised earlier in the flow if the nick doesn't exist in the DB but it's not exactly obvious because we don't catch it or anything.

Nick grouping probably needs to be refactored to be done automatically via coretasks and the new account tracking stuff.

dgw:
I initially planned to change `==` to `<=` but decided against changing the actual logic without discussion. Now that there's established support for using `<=` I'll do that in a force push momentarily.

maxpowa:
Thanks

PR #1103, opened 2016-07-15
Cleaned up code slightly and implemented a bot restart command.

pachuco:
Very small cleanup, implemented a bot restart command in Admin module.

Changed strftime() param %T with its equivalent. %T is linux specific.

Removed exit from root/Administrator check and changed it from Error to Warning.

### DIFF
--- a/sopel/__init__.py
+++ b/sopel/__init__.py
@@ -54,7 +54,7 @@ def _version_info(version=__version__):
 version_info = _version_info()
 
 
-def run(config, pid_file, daemon=False):
+def run(config, daemon=False):
     import sopel.bot as bot
     import sopel.logger
     from sopel.tools import stderr
@@ -90,14 +90,14 @@ def run(config, pid_file, daemon=False):
             logfile.write(trace)
             logfile.write('----------------------------------------\n\n')
             logfile.close()
-            os.unlink(pid_file)
-            os._exit(1)
+            return 1
 
         if not isinstance(delay, int):
             break
+        if p.restart:
+            return -1
         if p.hasquit:
             break
         stderr('Warning: Disconnected. Reconnecting in %s seconds...' % delay)
         time.sleep(delay)
-    os.unlink(pid_file)
-    os._exit(0)
+    return 0
--- a/sopel/irc.py
+++ b/sopel/irc.py
@@ -62,6 +62,7 @@ class Bot(asynchat.async_chat):
         self.ca_certs = ca_certs
         self.enabled_capabilities = set()
         self.hasquit = False
+        self.restart = False
 
         self.sending = threading.RLock()
         self.writing_lock = threading.Lock()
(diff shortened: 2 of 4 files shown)


pachuco:
Also, sorry, I thought I had to make merge request for each individual commit. Derp.

dgw:
I can almost guarantee you that this will not be merged as long as it includes changes that remove the exit calls from the root/administrator checks.

The restart functionality has a good chance at making it in—I've wanted it myself—though it might need some tweaking. But the root/administrator checks have a pretty long history. They're considered fatal errors because Sopel is in no way guaranteed to be safe against exploits that could abuse root/administrator privileges to ruin the system Sopel is running on.

pachuco:
Then you can discard root/admin check and take what you like.

dgw:
Looks like the rebase was successful. Keep on learning those git commands, @pachuco! 😉

maxpowa:
Just couple minor nickpicks
- I think negative exit code usually indicates an error. Restarts aren't technically an error.
- `run_script.py` [L#204](https://github.com/sopel-irc/sopel/pull/1103/files#diff-5dae56ced8258dad1db1e9f926dc21adR204) doesn't seem to cleanly exit if it goes down the restart path (or at least it doesn't give an exit code indicative of a restart rather than normal exit)
- Not sure what the `time.py/config_section.py` changes have to do with restarting (if it's windows compat that's probably better in a separate PR)
- Restarts should probably be logged to stdout/stderr somewhere, like disconnections & quits are.

pachuco:
I removed the time.py/config_section.py changes as suggested and will move them to another PR.

dgw:
Still kind of confused what is required for the restart command, what is just cleanup (but can be left in), and what relates to the running-as-root check (and can either stay or go depending on opinions from @maxpowa et al).

pachuco:
The cleanup consists of replacing 
```
os.unlink(pid_file)
os._exit(x)
```
Structures with return values and moving the above in a single place in the calling function. It felt needed for what I wanted to implement.

Any commit relating to rootcheck has already been already squashed, OP text is what remains. Same goes for strftime().

pachuco:
I've met some difficulties with run_script.py L#204 
I've run the bot on a remote linux host(as opposed to my usual windows environment).
I would launch the bot via 'python3' and !restarting it would launch it via 'python' instead, which seems to default to 2.5.

That line needs a bit of work.

converge:
the restart idea is great !

dgw:
@pachuco If you're still around, this will get looked at sometime for the next major release, perhaps. It'll be a bit of a road from 6.5.2 (upcoming bug fixes) to the 6.6.x series (minor new stuff) to possible additional 6.x.x versions before 7.0.0, when this would likely be included. But the ability for Sopel to restart itself would be a nice addition.

dgw:
Since the source repo has been deleted, and there's no way to update this PR and make it run tests, I'm going to rebase this patch on current `master` and open a new pull request. (I might also try to add a test for the new functionality, but no promises. Sopel's test suite still feels kind of hodgepodge…)

PR #1113, opened 2016-07-28
Fixed request.exceptions.SSLError

psachin:
Bot throw SSLerror
---
requests.packages.urllib3.exceptions.SSLError: [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:645)
---

even when 'verify_ssl = False'
The core.verify_ssl was not passed to url.find_title()

Signed-off-by: Sachin Patil <psachin@redhat.com>

PR #1114, opened 2016-07-28
meetbot/url: fix SSLError

maxpowa:
The `core.verify_ssl` was not passed to `url.find_title()`, resulting in SSL errors on sites with invalid certs when `verify_ssl = False`

Slight refactor of @psachin's original code for backwards compatibility. Resolves #1113

### DIFF
--- a/sopel/modules/meetbot.py
+++ b/sopel/modules/meetbot.py
@@ -342,7 +342,7 @@ def meetinglink(bot, trigger):
     if not link.startswith("http"):
         link = "http://" + link
     try:
-        title = find_title(link)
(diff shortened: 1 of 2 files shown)


psachin:
I see few more module where we need to take care of `verify_ssl`

```
find_updates.py:    info = requests.get(version_url).json()
movie.py:    data = requests.get(uri, params={'t': word}, timeout=30).json()
translate.py:    result = requests.get(url, params=query, timeout=40, headers=headers).text
xkcd.py:    data = requests.get(url).json()
```

PR #1120, opened 2016-08-01
[New Module|WIP] Karma module for Sopel

psachin:
- Removed rate() <- time-delay
- Handles string such as 'psachin++ great!'

Todo:
- Should handle string such as 'Cool, psachin++ great!'

### DIFF
--- a/sopel/modules/karma.py
+++ b/sopel/modules/karma.py
@@ -1,0 +1,132 @@
+# coding=utf-8
+"""
+karma.py - Sopel karma module
+Copyright © 2015-2016, Alex Newman <anewman@redhat.com>
+Copyright © 2016, Sachin Patil <psachin@redhat.com>
+Licensed under the Eiffel Forum License 2.
+
+This module is based on https://github.com/anewmanRH/sopel-karma
+"""
+from __future__ import (unicode_literals, absolute_import,
+                        print_function, division)
+import re
+
+from sopel.module import (rule, commands, example,
+                          require_privilege, OP)
+from sopel.tools import Identifier
+from sopel.formatting import bold as bold_text
+
+
+@rule(r'([\S]+?)\+\+')
+def promote_karma(bot, trigger):
+    """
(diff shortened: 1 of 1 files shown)


psachin:
@maxpowa Now this is somthing where I need you help/suggestion. As you see, string like 'Cool, psachin++ great!' doesn't get parse.

I have following cases to test with.
```
foo bar sometext psachin++ asa
tas++toona++
sopel
eck++ great!
hooya++ anande++
sansa hooya++ anande++
foobar hooya++ anande++ sometext
psachin+-
psachin--
psachin++sansa++
```
Also, if see that last line, both niks should be in .group(). But I think handling multiple nicks should not be a problem.

dasu:
'c++ is the worst' will trigger it, and notepad++ will too.
maybe have a check to see that the nick actually exists in the channel using bot.channels[trigger.sender.lower()].users to get a list of users and checking if the trigger is in the list?

psachin:
@dasu 
Even `bot.channels[trigger.sender.lower()].users` don't output anything for strings like,

```
foo bar sometext psachin++ asa
foobar hooya++ anande++ sometext
```
 
I don't think this has to do anything with existence of nick

maxpowa:
@psachin You'd probably have to do a second regex search on the matched string to get multiple nicks as Sopel's `@rule` only grabs the first match in the string.

psachin:
@maxpowa Adding multiple `@rule` will cause it to trigger multiple times. That should add more complexity. The latest commit can handle multiple nicks. ;)

Now only thing remains is, why string like
```
'Cool, psachin++ great!'
```
is not handled. Any idea?

maxpowa:
It's likely because the ++ bit is in the middle of the string and your regex doesn't have fuzzy matches around it. 

I'm going to close this though because we aren't looking to add new default modules at the moment. If you restructure this module to a [sopel-cookiecutter](https://github.com/sopel-irc/sopel-cookiecutter) structure you can distribute it on pypi.

PR #1121, opened 2016-08-04
[xkcd] Join sites query without prefix

xZise:
The sites query can be joined where each element already uses the prefix. That way the prefix doesn't need to be added manually for the first element.

### DIFF
--- a/sopel/modules/xkcd.py
+++ b/sopel/modules/xkcd.py
@@ -25,7 +25,8 @@ ignored_sites = [
     'wiki.xkcd.com',
     'what-if.xkcd.com',
 ]
-sites_query = ' site:xkcd.com -site:' + ' -site:'.join(ignored_sites)
+sites_query = ' site:xkcd.com ' + ' '.join('-site:{}'.format(ignored_site)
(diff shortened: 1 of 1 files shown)


xZise:
I have to admit that I didn't test the patch. I noticed this and thought I just change it and make a pull request (as I don't really use it anymore).

sopel/modules/xkcd.py
```diff
@@ -25,7 +25,8 @@
     'wiki.xkcd.com',
     'what-if.xkcd.com',
 ]
-sites_query = ' site:xkcd.com -site:' + ' -site:'.join(ignored_sites)
+sites_query = ' site:xkcd.com ' + ''.join('-site:{}'.format(ignored_site)
```
maxpowa: Pretty sure this won't work as it will join to form `-site:a-site:b-site:c` without any spaces.

xZise:
Sorry for this mistake. I updated and rebased it accordingly.

maxpowa:
Not sure if this is any more clear than the current method of building the query, sorry.

PR #1126, opened 2016-08-19
catch also UnicodeError which is raised by requests

freeboson:


maxpowa:
Use `requests=2.10.0` https://github.com/kennethreitz/requests/issues/3481#issuecomment-238749507

I pinned 2.10.0 in commit https://github.com/sopel-irc/sopel/commit/f28daeaea1908f1c3025ccf4c443719bd3ea58d0

freeboson:
OK, but why not just catch both exceptions? Then you don't have to be tied to that version of requests.

maxpowa:
Because they're reverting the behavior for the next release anyways, as per the comments on the linked issue.

PR #1127, opened 2016-08-19
sopel/trigger.py: fix intent_regex

RebelCodeBase:
The intent_regex in sopel/trigger.py does not separate "intent" from "args" correctly. The way it was there needed to be a blank after "intent" which caused test_ctcp_intent_pretrigger to fail:

```
test/test_trigger.py:118: in test_ctcp_intent_pretrigger
    assert pretrigger.tags == {'intent': 'VERSION'}
E   assert {} == {'intent': 'VERSION'}
E     Right contains more items:
E     {u'intent': u'VERSION'}
E     Use -v to get the full diff
```

Fixed it by adding a question mark to the regexp after the blank in order to make the blank optional.

### DIFF
--- a/sopel/trigger.py
+++ b/sopel/trigger.py
@@ -16,7 +16,7 @@ class PreTrigger(object):
     """A parsed message from the server, which has not been matched against
     any rules."""
     component_regex = re.compile(r'([^!]*)!?([^@]*)@?(.*)')
-    intent_regex = re.compile('\x01(\\S+) (.*)\x01')
+    intent_regex = re.compile('\x01(\\S+) ?(.*)\x01')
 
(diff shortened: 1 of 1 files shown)


maxpowa:
@embolalia had reverted a previous version of this in https://github.com/sopel-irc/sopel/commit/c78571b85eb4fd07f3e44bf32309b598dfeab1e0, which worked the same way, and was slightly faster to match in the case of a single word CTCP. They claimed it didn't work but I've been running it on my personal bot for more than a year now without any issues.

embolalia:
I'm not sure why I said the other one wouldn't work, but this looks fine to me.

PR #1132, opened 2016-09-07
url: make find_title more robust

hdhoang:
Previously, each 512-byte chunk is prone to decoding mishap when a UTF-8 sequence is incomplete. Now we decode all of content at once, ignoring errors.

The old problem appears reliably for pages with many high codepoints:

```
<user> http://www3.nhk.or.jp/news/easy/k10010665021000/k10010665021000.html
<bot-old> [ NEWS WEB EASY|æ±åã¢ã¸ã¢ã§ã¸ã«ç±ã«ãªãäººãå¢ãã ] - www3.nhk.or.jp
```

### DIFF
--- a/sopel/modules/url.py
+++ b/sopel/modules/url.py
@@ -186,16 +186,12 @@ def find_title(url, verify=True):
     """Return the title for the given URL."""
     response = requests.get(url, stream=True, verify=verify)
     try:
-        content = ''
-        for byte in response.iter_content(chunk_size=512, decode_unicode=True):
-            if not isinstance(byte, bytes):
-                content += byte
-            else:
+        content = b''
+        for byte in response.iter_content(chunk_size=512):
+            content += byte
+            if b'</title>' in content or len(content) > max_bytes:
                 break
-            if '</title>' in content or len(content) > max_bytes:
-                break
-    except UnicodeDecodeError:
-        return  # Fail silently when data can't be decoded
+        content = content.decode('utf-8', errors='ignore')
     finally:
         # need to close the connexion because we have not read all the data
         response.close()


sopel/modules/url.py
```diff
@@ -186,16 +186,12 @@ def find_title(url, verify=True):
...
+        content = b''
+        for byte in response.iter_content(chunk_size=512):
+            content += byte
+            if b'</title>' in content or len(content) > max_bytes:
                 break
-            if '</title>' in content or len(content) > max_bytes:
-                break
-    except UnicodeDecodeError:
-        return  # Fail silently when data can't be decoded
+        content = content.decode('utf-8', errors='ignore')
```
maxpowa: This may raise a https://docs.python.org/3/library/exceptions.html#UnicodeError, should probably keep the except statement.

hdhoang: AIUI, errors=ignore removes that possibility. This way, when content is chopped off at a partial character, it is still searched for a title later on.

maxpowa: Ah right, cool.

sopel/modules/url.py
```diff
@@ -186,16 +186,12 @@ def find_title(url, verify=True):
     """Return the title for the given URL."""
     response = requests.get(url, stream=True, verify=verify)
     try:
```
anarcat: just curious: is the try/finally still necessary, or could this just be simplified a bit further?

embolalia:
Great, thanks!

sopel/modules/url.py
hdhoang: I think `with requests.get(url, stream=True, verify=verify) as response:` is appropriate, seeing that py2.7 is required

PR #1140, opened 2016-09-15
Create wunder.py (optional weather module replacing weather.py)

ktwilson86:
Add wunder.py Weather Underground weather module (to replace yahoo weather API weather.py). Requires user-specified API key.

### DIFF
--- a/sopel/modules/wunder.py
+++ b/sopel/modules/wunder.py
@@ -1,0 +1,89 @@
+"""
+weather.py - sopel Weather Underground Module
+Copyright 2008, Sean B. Palmer, inamidst.com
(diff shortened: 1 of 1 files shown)


dgw:
If an API key is required, this will most likely not be merged as a core module. Maintaining it in a separate repository (as sopel-wunderground or similar) either as a simple python file or packaged up on PyPI with the cookiecutter template would be best.

maxpowa:
Sorry, we aren't looking to replace existing modules at the moment. If you restructure this module to a [sopel-cookiecutter](https://github.com/sopel-irc/sopel-cookiecutter) structure you can distribute it on pypi.

PR #1156, opened 2016-12-12
Fixes for pronouns.py

sofoxiya:
Fixes setpronouns error on lack of trigger.group(2), fixes autocomplete of nicks with a space so that it's stripped out automatically, fixes that it will say the wrong username if you request someone else's pronouns.


sofoxiya:
To clarify, what the strip of " " is supposed to do is make it so if you do .setpronouns sofoxiya<space> it'll not know that person. That strips out the space so that won't happen if you're using like, weechat.

PR #1123, opened 2016-08-09
Fixed missing `verify_ssl` param

psachin:
- `verify_ssl` param was missing in few function calls
- Fixed few linting errors. I think, linting should be addressed in a  separate commit

Closes-Bug: #1118

Signed-off-by: Sachin Patil psachin@redhat.com

Issue #1118, opened 2016-07-31
Missing `verify_ssl` 


psachin:
I see few more module where we need to take care of `verify_ssl`

```
find_updates.py:    info = requests.get(version_url).json()
movie.py:    data = requests.get(uri, params={'t': word}, timeout=30).json()
translate.py:    result = requests.get(url, params=query, timeout=40, headers=headers).text
xkcd.py:    data = requests.get(url).json()
```

Also, if `verify_ssl=False`, we can disable warning.

[https://urllib3.readthedocs.io/en/latest/advanced-usage.html#ssl-warnings](https://urllib3.readthedocs.io/en/latest/advanced-usage.html#ssl-warnings)

elad661:
This makes no sense. At all. 

There's absolutely no good reason to disable TLS verification in these modules or any other modules. verify_ssl should only be for core, for people who, for some weird reason, wish to connect to IRC servers with self signed certificates over TLS.

On the long term, it'd be better to implement certificate pinning for this usecase, and *not* provide an option to disable TLS verification completely.

anarcat:
this is still happening in the xkcd module:

```
14:49:23 <@anarcat> http://xkcd.com/1779/
14:49:23 <peoplesmic> TypeError: numbered_result() missing 1 required positional argument: 'verify_ssl' (file "/usr/lib/python3/dist-packages/sopel/modules/xkcd.py", line 124, in get_url)
```

PR coming up.

antdude:
When will this fix be released?

tylerhoot:
kinda still needed and issue still lives in xkcd AND in url

antdude:
Still no fix? :(

psachin:
@anarcat  The fixed is merged -- https://github.com/sopel-irc/sopel/pull/1123

@tylerhoot It is fixed for xkcd in -- https://github.com/sopel-irc/sopel/pull/1123, if you still encounter this in url can you raise an issue?

PR #1158, opened 2016-12-19
xkcd: supply verify_ssl to numbered_result

hdhoang:


anarcat:
this is related to #1118 - which was incomplete and should be merged.

anarcat:
this seems to duplicate #1123, which was merged. not sure why they don't conflict with each other...

hdhoang:
Thanks for the heads up! I will close this one.

That PR supplies the parameter via a different route, so we don't show up as conflicting.

PR #1160, opened 2016-12-29
version: remove debug(?) print statement

alanhuang122:
Yeah, I don't think that's supposed to be there.


alanhuang122:
Oops. Looks like I evidently need more practise working with branches.

PR #1161, opened 2016-12-29
version: fix CTCP VERSION reply

alanhuang122:
Removed 'wat' statement and fixed discrepancies.


alanhuang122:
This PR (which doesn't fix anything) has been superseded by #1241.

PR #1163, opened 2016-12-29
wikipedia: retrieve section summaries

alanhuang122:
Module will now retrieve text from the relevant section
instead of the beginning of the article irrespective of
section designations in the URL.

### DIFF
--- a/sopel/modules/wikipedia.py
+++ b/sopel/modules/wikipedia.py
@@ -23,6 +23,43 @@ else:
 REDIRECT = re.compile(r'^REDIRECT (.*)')
 
 
+class WikiParser(HTMLParser):
+    def __init__(self):
+        HTMLParser.__init__(self)
+        self.consume = True
+        self.citations = False
+        self.span_depth = 0
+        self.result = ''
+    
+    def handle_starttag(self, tag, attrs):
+        if tag == 'h3' or tag == 'sup': # don't consume the header or anything in superscript (citation-related tags)
+            self.consume = False
+        elif tag == 'span':
+            if self.span_depth:
+                self.span_depth += 1
+            else:
+                for attr in attrs:  # remove 'edit' tags, and keep track of depth
+                    if attr[0] == 'class' and 'edit' in attr[1]:
+                        self.span_depth += 1
+        elif tag == 'ol':
+            for attr in attrs:
+                if attr[0] == 'class' and 'references' in attr[1]:
+                    self.citations = True   # once we hit citations, we can stop
+    
+    def handle_endtag(self, tag):
+        if not self.consume and (tag == 'h3' or tag == 'sup'):
+            self.consume = True
+        if self.span_depth and tag == 'span':
+            self.span_depth -= 1
+    
+    def handle_data(self, data):
+        if self.consume and not self.citations and not self.span_depth:
+            self.result += data
+    
+    def get_result(self):
+        return self.result
+
+
 class WikipediaSection(StaticSection):
     default_lang = ValidatedAttribute('default_lang', default='en')
     """The default language to find articles from."""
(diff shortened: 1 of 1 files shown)


sopel/modules/wikipedia.py
```diff
@@ -2,11 +2,13 @@
...
 # Licensed under the Eiffel Forum License 2.
 from __future__ import unicode_literals, absolute_import, print_function, division
-from sopel import web, tools
+import requests
+from sopel import tools
 from sopel.config.types import StaticSection, ValidatedAttribute
 from sopel.module import NOLIMIT, commands, example, rule
 import json
 import re
+import bs4
```
dgw: `bs4` needs to be added to `requirements.txt`, it's not part of the Python standard library.

But ideally, another parsing method that doesn't add another requirement would be great.

alanhuang122:
I've pushed a very rough initial feature. It's not suitable for merging yet, and fails style checks. I'll continue working on it soon.

alanhuang122:
This is now a more complete ("complete") function. There will almost certainly be stylistic issues (at the very least) that would be simple to correct.

One known issue is that the snippet feature does not recognize section titles that are italicized (e.g., [this page](https://en.wikipedia.org/wiki/Procopius#Secret_History) [comparing `<i>Secret History</i>` to `Secret History`]). I could special-case the removal of HTML tags, but that assumes that HTML tags will never be part of any section title.

sopel/modules/wikipedia.py
```diff
@@ -8,19 +8,79 @@
...
 import re
-
 import sys
+
 if sys.version_info.major < 3:
     from urllib import quote as _quote
     from urlparse import unquote as _unquote
     quote = lambda s: _quote(s.encode('utf-8')).decode('utf-8')
     unquote = lambda s: _unquote(s.encode('utf-8')).decode('utf-8')
+    import HTMLParser
```
dgw: I think this will fix py2:

```suggestion
    from HTMLParser import HTMLParser
```

But I also haven't tested it, so it might just create a different error :joy_cat:

dgw:
Happy to see activity on your PRs whenever you have the time and motivation! As I said on IRC, the milestones I assigned are just when I think it'll be ready; if something's not done in time, I don't have any problem with bumping it to the next release unless it's a really important bugfix. Take your time! :smile_cat:

> One known issue is that the snippet feature does not recognize section titles that are italicized (e.g., [this page](https://en.wikipedia.org/wiki/Procopius#Secret_History) [comparing `<i>Secret History</i>` to `Secret History`]). I could special-case the removal of HTML tags, but that assumes that HTML tags will never be part of any section title.

The JSON pulled down to find the section also includes a field named `anchor`, which should exactly match the trigger group that captures the section anchor off the URL. I doubt it would introduce any new issues to rewrite the code to match based on that instead, and it would fix the known issue you described.

sopel/modules/wikipedia.py
alanhuang122: Not testing with python2; guilty as charged. Thanks.

sopel/modules/wikipedia.py
```diff
@@ -161,14 +163,13 @@ def mw_snippet(server, query):
...
     query = quote(query.replace(' ', '_'))
-    section = section.replace('_', ' ')
 
     snippet = mw_section(server, query, section)
     if not snippet:
         bot.say("[WIKIPEDIA] Error fetching section \"{}\" for page \"{}\".".format(section, page_name))
         return
 
-    msg = '[WIKIPEDIA] {} - {} | "{}"'.format(page_name, section, snippet)
+    msg = '[WIKIPEDIA] {} - {} | "{}"'.format(page_name, section.replace('_', ' '), snippet)
```
dgw: Had a couple thoughts while looking at this before bed. Cursory stuff, not a formal review (and not that this PR needs another until it's "really ready"), but things that might inform the evolution as it's getting close to done.

I wonder how many users would be confused if that separating `-` turned into a `§`… Somehow the hyphen doesn't seem right. Section symbol, some Unicode arrow, `>`… one of those maybe?

Also: What do you think about returning a tuple from `mw_section()`, say `(section_title, snippet)`? Have to think about the possibility of HTML, character entities, etc. in the section title, though. The `replace` here is certainly the simplest way to go!

alanhuang122: I find the hyphen to fit better than the alternatives, but that's wholly a personal opinion.

Wrt. the tuple suggestion, I presume you're referring to taking the entry from the MediaWiki sections listing. That'd involve at least some HTML parsing in some cases (replacing `<i></i>` with a call to `formatting.italic()`, e.g.) which I'd remove from my implementation, but could be done.

dgw: I don't think parsing HTML in the section title is worth it. Stripping tags is fine. It's just that the anchor name is sanitized of more than just HTML tags, and might be significantly different from the "real" section title that `mw_section()` has access to.

`§` seems appropriate because that's literally the reason that symbol exists—to denote a document section—but I'm not going to push. It can always be changed later. :smile_cat:

Exirel:
I feel like this one is either stale or need to be kept for when we extract some plugins out of the core—this module would be a great candidate to be an external module.

dgw:
@Exirel Why do you say this is a great candidate for an external module? It requires no API keys or authentication of any kind, and is broadly useful (everyone links Wikipedia on IRC eventually). Grabbing the relevant section when someone links directly to one will only make it more broadly useful.

Exirel:
> @Exirel Why do you say this is a great candidate for an external module?

Because of its increasing complexity, it becomes more and more subject to issues, bugs, that will ultimately require a faster release cycle than the bot itself. All I see in this PR looks good, but also more complex, more prone to errors and bugs.

(That being said... it also lacks some unit-test.)

dgw:
I'm far more comfortable with the output format of Wikipedia remaining the same than with many of the APIs that we use in other core modules, honestly.

All of this discussion could be moot, though, if @alanhuang122 no longer wants to work on this and we can't find anyone to take over the work. I'm willing, but there are so many more important things to do…

PR #1164, opened 2016-12-29
spellcheck: complete rewrite

alanhuang122:
I had experimented with this module a bit, and had found that I was completely unsatisfied with its performance (namely, the corrections it suggested). As a result, I rewrote the module to use aspell instead of pyenchant. The resulting performance satisfied me.

### DIFF
--- a/sopel/modules/spellcheck.py
+++ b/sopel/modules/spellcheck.py
@@ -5,27 +5,21 @@ spellcheck.py - Sopel spelling checker module
 Copyright © 2016, Alan Huang
 Licensed under the Eiffel Forum License 2.
 http://sopel.chat
-
-This module relies on aspell-python-py2 or py3, available through pypi or
-https://github.com/WojciechMula/aspell-python
 """
 from __future__ import unicode_literals, absolute_import, print_function, division
 import aspell
-from sopel.module import commands
-
+from sopel.module import commands, example, require_admin
 
 @commands('add')
+@require_admin('I only trust admins to add words.')
 def add_command(bot, trigger):
     """
     Add words to the bot's personal dictionary
     """
-    if(trigger.owner):
-        c = aspell.Speller('lang', 'en')
-        c.addtoPersonal(trigger.group(2))
-        c.saveAllwords()
-        bot.say('Added {0}.'.format(trigger.group(2)))
-    else:
-        bot.say('I only trust {0} to add words >:c'.format(bot.config.core.owner))
+    c = aspell.Speller('lang', 'en')
+    c.addtoPersonal(trigger.group(2))
+    c.saveAllwords()
+    bot.say('Added {0}.'.format(trigger.group(2)))
 
 def check_multiple(bot, words):
     mistakes = []
@@ -54,6 +48,7 @@ def check_one(bot, word):
         bot.say("That doesn't seem to be correct. Try {0}.".format(', '.join(['"{0}"'.format(s) for s in suggestions])))
 
 @commands('spell')
+@example('.spell stuff')
 def spellchecker(bot, trigger):
     """
     Checks if the given word is spelled correctly, and suggests corrections.
--- a/sopel/modules/spellcheck.py
+++ b/sopel/modules/spellcheck.py
@@ -1,10 +1,24 @@
 # coding=utf-8
 
+"""
+spellcheck.py - Sopel spelling checker module
+Copyright © 2016, Alan Huang
+Licensed under the Eiffel Forum License 2.
+http://sopel.chat
+
+This module relies on aspell-python-py2 or py3, available through pypi or
+https://github.com/WojciechMula/aspell-python
+"""
+from __future__ import unicode_literals, absolute_import, print_function, division
 import aspell
 from sopel.module import commands
 
+
 @commands('add')
 def add_command(bot, trigger):
+    """
+    Add words to the bot's personal dictionary
+    """
     if(trigger.owner):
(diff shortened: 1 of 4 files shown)


dgw:
@alanhuang122 When you next have time to do code-y stuff, rebasing this would be much appreciated!

alanhuang122:
@dgw Rebased.

dgw:
@alanhuang122 I got excited, but there's still a conflict. Did you pull master before rebasing?

Also why did AppVeyor try to build this? I haven't set that up yet lmao

dgw:
That'd do it.

Self-assigned to kick the tires one last time before merging. :)

dgw:
Hmm… Upon kicking the tires, and looking over the code once more, I have a few questions.

  1. Why remove `spellcheck` as an alternate command name? If length is the concern, something shorter like `sc` can be added as an additional option (`@commands('spellcheck', 'spell', 'sc')`.
  2. Can we get a more specific command name than `add` for adding words to the dictionary? Something like `scadd` would go with the short-form I suggested in point 1.
  3. Is there a way to remove custom words from the dictionary? I get that adding words is admin-only, but even a bot admin can fat-finger something and have no way to undo the mistake.

alanhuang122:
Acknowledged, agreed, will implement.

dgw:
#<span></span>1 is what this rewrite currently does, and we seem to agree that _that_ isn't the solution.

I don't like #<span></span>3 either. Sopel shouldn't be messing with that file (or any file outside its home directory, ideally), and I have no idea if the location is predictable in all cases. What if Sopel is run as a system service user with no home directory, for example? Best to let aspell do its thing and butt out.

So… yeah, #<span></span>2 is the only not-awful way to handle aspell's missing functionality.

What we'll end up with, as I understand it so far:

* `spellcheck`, `spell`, and `sc` as command names for checking one or more words
* `scadd` to add a word to the staging list
* `scsave` (or a better idea?) to actually call aspell's save function on each word in the staging list

Sound right? Good?

dgw:
@alanhuang122 Did you fall off the edge of the ~~planet~~ world? :mag_right:

dgw:
Implemented everything I asked for and more, since @alanhuang122 appears to have gone quiet.

In my opinion, I added or rewrote too much of the code to be an objective reviewer any more, so this should get a look from @Exirel, @HumorBaby, @kwaaak, and/or anyone else who's contributed recently.

sopel/modules/spellcheck.py
```diff
@@ -1,55 +1,151 @@
...
+
+
+@commands('scclear')
+@require_admin('Only admins may clear the pending word list.')
+def clear_command(bot, trigger):
+    """
+    Clear the list of words pending addition to the bot's personal dictionary.
+    """
+    count = len(bot.memory['spellcheck_pending_adds'])
+    bot.memory['spellcheck_pending_adds'].clear()
```
HumorBaby: `.clear()` is not a valid method for `list`s in Python 2, so this, and the one in `save_command()` will need to be changed. 

Something like `del bot.memory['spellcheck_pending_adds'][:]` or `bot.memory['spellcheck_pending_adds'][:] = []` should suffice if you want to maintain the same list object (as opposed to replacing it with a new `[]`).

sopel/modules/spellcheck.py
```diff
@@ -1,55 +1,151 @@
...
+                    '{1}scdel/{1}scadd, or force saving anyway with {1}scsave force.'
+                    .format(word, bot.config.core.help_prefix))
+            return
+    c = aspell.Speller('lang', 'en')
+    for word in bot.memory['spellcheck_pending_adds']:
+        c.addtoPersonal(word)
+    c.saveAllwords()
+    bot.say('Saved {0} pending words to my word list.'
+            .format(len(bot.memory['spellcheck_pending_adds'])))
+    bot.memory['spellcheck_pending_adds'].clear()
```
HumorBaby: See previous comment...

sopel/modules/spellcheck.py
```diff
@@ -1,55 +1,151 @@
...
-    enchant = None
-from sopel.module import commands, example
+
+@commands('scpending')
+def pending_command(bot, trigger):
+    """
+    List words that are waiting to be saved to the bot's personal dictionary.
+    """
+    bot.say('Ready to save: "{0}". (To remove a word before saving, use {1}scdel '
+            'word, or clear the list with {1}scclear.)'
```
HumorBaby: Can there be something here to indicate that `word` in `..., use .scdel word, or ...` is an argument, and not literally the word `word`?

```suggestion
            '<word>, or clear the list with {1}scclear.)'
```

sopel/modules/spellcheck.py
dgw: Bleh, the documentation I read didn't mention when `.clear()` was added to lists. I'll fix that (and below).

sopel/modules/spellcheck.py
dgw: There can, certainly. I was going for consistency with the rest of the core modules, which generally don't delineate arguments in any way.

HumorBaby: Hmm, no problem them. I'd say consistency trumps, then.

dgw: If it's worth fixing, let's fix them all in another PR. 😸

HumorBaby:
@dgw, A+

dgw:
OK, now this is really "Done in Feature Rewrites" (project board status since November 2018, wow).

PR #1166, opened 2017-01-08
pronouns: Ignore whitespace

nfearnley:
Changed trigger.group(2) to trigger.group(2).strip() in order to ignore trailing whitespace. (Occurs when user uses tab completion). Also fixed issue where wrong nick is given when pronouns are requested.

### DIFF
--- a/sopel/modules/pronouns.py
+++ b/sopel/modules/pronouns.py
@@ -42,10 +42,10 @@ def pronouns(bot, trigger):
             bot.reply("I don't know your pronouns! You can set them with "
                       ".setpronouns")
     else:
-        pronouns = bot.db.get_nick_value(trigger.group(2), 'pronouns')
+        pronouns = bot.db.get_nick_value(trigger.group(2).strip(), 'pronouns')
         if pronouns:
             say_pronouns(bot, trigger.nick, pronouns)
-        elif trigger.group(2) == bot.nick:
+        elif trigger.group(2).strip() == bot.nick:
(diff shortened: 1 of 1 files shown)


dgw:
Can also use `trigger.group(3)`.

nfearnley:
That was my initial thought, but I decided .strip() was more explicit and
easier to understand. Either method would work.

dgw:
The reason I suggest using group 3 is because it can't be more than one word. `trigger.group(2).strip()` can still contain multiple words, but nicks can't have spaces in them.

nfearnley:
Works for me.

dgw:
What I mean is this:

````
<dgw> .pronouns
<Sopel> dgw: I don't know your pronouns! You can set them with .setpronouns
<dgw> .setpronouns he
<Sopel> dgw: Thanks for telling me!
<dgw> .pronouns
<Sopel> dgw's pronouns are he/him/his/his/himself. See http://pronoun.is/he for examples.
<dgw> .pronouns dgw
<Sopel> dgw's pronouns are he/him/his/his/himself. See http://pronoun.is/he for examples.
<dgw> .pronouns dgw dgw
<Sopel> I don't know dgw dgw's pronouns. They can set them with .setpronouns
````

Any extraneous word(s) after the nickname breaks it. May or may not be of concern here, but all (most?) modules I wrote for myself use groups 3+ to avoid this when taking single-words arguments.

maxpowa:
Another solution would be to `split()` group 2, and then list off all
matches from that array.

Issue #1091, opened 2016-06-21
web: broken header behavior


maxpowa:
Yes, I know its deprecated, I'm going to fix it anyways.

headers bleed between web requests because we're using `update` on `default_headers`.

```python
web.get('http://google.com/')
# Headers: User-Agent: Sopel/6.3.1 (http://sopel.chat)
web.get('http://google.com/', headers={'some-header': 'value'})
# Headers: User-Agent: Sopel/6.3.1 (http://sopel.chat), some-header: value
web.get('http://google.com/')
# Headers: User-Agent: Sopel/6.3.1 (http://sopel.chat), some-header: value
```

PR #1171, opened 2017-01-23
Fix .g giving broken links

dasu:
This is a two part fix.
1. Fix in web.py for headers being set as blank if they were set in web.get()
2. Use a common user-agent string for duck_search function only.  
Replacing the Sopel version user-string for all web.get calls via web.py would have been a bit too extreme, in my opinion.

### DIFF
--- a/sopel/web.py
+++ b/sopel/web.py
@@ -179,7 +179,8 @@ def get_urllib_object(uri, timeout, headers=None, verify_ssl=True, data=None):
         headers = default_headers
     else:
         tmp = default_headers.copy()
-        headers = tmp.update(headers)
+        tmp.update(headers)
+        headers = tmp
     if data is not None:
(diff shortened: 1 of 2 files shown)


sopel/web.py
```diff
@@ -86,7 +86,7 @@ def get(uri, timeout=20, headers=None, return_headers=False,
         headers = default_headers
     else:
         tmp = default_headers.copy()
-        headers = tmp.update(headers)
+        tmp.update(headers)
```
dasu: So tmp.update(headers) returns nothing, so previously headers was being set to nothing.
Just realized that I probably should have added a headers = tmp then, but I'm not sure why there is a copying of the headers to a tmp variable in the first place...
This does occur again in another spot in web.py which I neglected to update.

dasu: Just saw the reason why: #1091 .  Just set the same functionality for all instances of tmp.update(headers)

PR #1178, opened 2017-02-20
Move etymology lookup to wiktionary

dgw:
This spelling error prevents etymology lines from matching, because the code sets the `mode` to `etymology` but then later looks for a `mode` equal to `etmyology`. Fixed.


dgw:
This actually does nothing, because the returned etymology from `wikt()` is never used. Perhaps this PR will morph into a pseudo-fix for #1248 (moving the `.ety` command into wiktionary.py)?

dgw:
Built a proof-of-concept replacement `.ety` command on this branch, since the original PR was useless.

The `wiktionary` module only parses definitions for the first sense of a word, and the new `.ety` command implementation only looks for the first etymology (if it exists) to match that. A future rewrite (either in this PR or another, not yet decided) will add handling of multiple senses to both commands.

The Coveralls check failure for -0.9% test coverage tells me that my rewrite should also add example tests to this module. 12% coverage is downright awful. (The large drop must be from removing `etymology.py`, which was technically "30% covered" just from defining functions & initializing a ton of global variables.)

<details><summary>(Future implementation notes)</summary>
Good words for testing this future rewrite include "[sake](https://en.wiktionary.org/wiki/sake)" and "[forte](https://en.wiktionary.org/wiki/forte)". As of writing this comment, both include two separate senses with independent etymologies and their own sets of definitions.</details>

dgw:
6.6.0 will have #1432 to fix #1248, and this rewrite will maybe happen later—or not.

dgw:
Shelving this until 7.1, unless `etymology` breaks again between now and 7.0's release.

dgw:
I have no idea why I left this as WIP. It clearly works:

```
<dgw> ,ety bailiwick
<SopelTest> bailiwick: From bailie (“bailiff”) and wick (“dwelling”), from Old
            English wīc.
```

The main difference is obviously how detailed the etymologies are, but I think Wiktionary is a better bet over the long term. WT's editors will continue to flesh it out over time.

I have a conflicts-resolved merge already done locally (it's not that bad) for testing, so don't worry about GitHub showing that this isn't mergeable @ reviewers.

PR #1180, opened 2017-03-09
Add 'UTC' as a timezone abbreviation for UTC timestamps

Merovius:
Currently, if no timezone is given, we use None. The default of datetime
in that case is to use a UTC timezone *without a name*. This is
confusing, as it's not clear from the printed timestamp, that it's in
UTC. This instead uses an actual implementation of UTC, with a name, as
a default.

### DIFF
--- a/sopel/tools/time.py
+++ b/sopel/tools/time.py
@@ -5,8 +5,26 @@ from __future__ import unicode_literals, absolute_import, print_function, divisi
 import datetime
 try:
     import pytz
+    utc = pytz.timezone('UTC')
 except:
     pytz = False
+
+    class _UTC(datetime.tzinfo):
+        """UTC default implementation.
+
+        Using this as a tzinfo over None has the advantage, that it has the correct
+        timezone abbreviation"""
+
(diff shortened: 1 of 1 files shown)


sopel/tools/time.py
```diff
@@ -132,13 +150,10 @@ def format_time(db=None, config=None, zone=None, nick=None, channel=None,
...
-    else:
-        if not time.tzinfo:
-            utc = pytz.timezone('UTC')
-            time = utc.localize(time)
+    if pytz and zone:
         zone = pytz.timezone(zone)
-        return time.astimezone(zone).strftime(tformat)
+    if not zone:
+        zone = utc
+    return time.astimezone(zone).strftime(tformat)
```
dgw: What happens `if zone and not pytz`? I'm not clear on this restructured logic, but I do want to have this feature.

Merovius:
I don't remember what I did here and the details of why, so let's just close it.

dgw:
Fair enough. This could still serve as a starting point for redoing the change.

PR #1184, opened 2017-03-30
add physical limit to temperature conversion (units module)

normcyr:
There is a bug in the units module (temp sub-module): the user is allowed to convert temperatures below the absolute 0 Kelvin, which is not physically possible. This small code modification allows to return the correct answer to the conversion only if the result if 0 Kelvin or above and will return an error to the user if the resulting temperature for the conversion is below 0 K.

### DIFF
--- a/sopel/modules/units.py
+++ b/sopel/modules/units.py
@@ -60,7 +60,7 @@ def temperature(bot, trigger):
     if kelvin >= 0:
         bot.reply("{:.2f}°C = {:.2f}°F = {:.2f}K".format(celsius, fahrenheit, kelvin))
     else:
-        bot.reply("Physically impossible temperature. Do you live in an imaginary world?")
+        bot.reply("Physically impossible temperature.")
 
 
 @commands('length', 'distance')
--- a/sopel/modules/units.py
+++ b/sopel/modules/units.py
@@ -56,7 +56,11 @@ def temperature(bot, trigger):
 
     kelvin = c_to_k(celsius)
     fahrenheit = c_to_f(celsius)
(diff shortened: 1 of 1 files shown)


sopel/modules/units.py
```diff
@@ -56,7 +56,11 @@ def temperature(bot, trigger):
 
     kelvin = c_to_k(celsius)
     fahrenheit = c_to_f(celsius)
-    bot.reply("{:.2f}°C = {:.2f}°F = {:.2f}K".format(celsius, fahrenheit, kelvin))
+
+    if kelvin >= 0:
+        bot.reply("{:.2f}°C = {:.2f}°F = {:.2f}K".format(celsius, fahrenheit, kelvin))
+    else:
+        bot.reply("Physically impossible temperature. Do you live in an imaginary world?")
```
dgw: I'm happy with everything except the high level of sass in this error.

kwaaak: Sopel is sassy by default

normcyr:
@dgw You're right. I removed the comment.

dgw:
@normcyr If you could just amend the commit messages this will be :ok_hand:

The original commit just needs the "units:" prefix, so the log shows what module it is, and the commit amending the error message needs the "units:" prefix and a reworded message that just says it's lowering the sass (not adding the limit again).

I *could* just change the commit messages myself, but that takes away the committer credit from you :)

normcyr:
Done (hopefully!). 

Still learning git and its (numerous) subtleties. Thanks for bearing with me.

dgw:
It's this project's commit message style guide, nothing to do with git. (Git doesn't care what the commit messages say, generally.) We just prefer to have the module being changed in the commit message for ease of reviewing history when it comes time to write the changelog—which I'll be doing soon.

PR #1195, opened 2017-05-11
Make the movie module more useful

freeboson:
The `modules/movie.py` module provided one command: `.imdb` which would interpret the arguments as a movie title and query OMDb API. This PR includes the following changes:

1. Allow specification of year
  - Originally, if you did `.imdb Get Carter` the only result would be for [Get Carter (2000)](http://www.imdb.com/title/tt0208988/)
  - You can now do `.imdb Get Carter (1971)` and instead get results for the far superior Michael Caine classic [Get Carter (1971)](http://www.imdb.com/title/tt0067128/)
2. Include plot summary
  - The updated command will also fetch the available plot summary (clipped to stay within one message length)
3. Add URL call back
  - This was probably the strangest thing omitted from the module
  - Any IMDb link will trigger extraction of the IMDb ID (e.g. `tt0067128`) for a direct IMDb-ID-based query of OMDb
  - When the module is triggered via URL call back, it will not re-print the URL

Some example usage:
```
12:45 <suj> ,imdb Get Carter
12:45 <euclidtst> [MOVIE] Title: Get Carter | Year: 2000 | Rating: 5.1 | Genre: Action, Crime, Drama | Plot: Years ago, Jack Carter left his Seattle home to become a Las Vegas mob casino financial enforcer. He returns for the funeral of his brother Richard 'Richie' afte[…] | IMDB Link: http://imdb.com/title/tt0208988
12:45 <suj> ,imdb Get Carter (1971)
12:45 <euclidtst> [MOVIE] Title: Get Carter | Year: 1971 | Rating: 7.5 | Genre: Crime, Thriller | Plot: When his brother dies under mysterious circumstances in a car accident, London gangster Jack Carter travels to Newcastle to investigate. | IMDB Link: http://imdb.com/title/tt0067128
12:46 <suj> http://imdb.com/title/tt0067128
12:46 <euclidtst> [MOVIE] Title: Get Carter | Year: 1971 | Rating: 7.5 | Genre: Crime, Thriller | Plot: When his brother dies under mysterious circumstances in a car accident, London gangster Jack Carter travels to Newcastle to investigate.
```

In the above, note that the plot summary for Get Carter (2000) is clipped to keep the length of the output at 300 characters.

### DIFF
--- a/sopel/modules/movie.py
+++ b/sopel/modules/movie.py
@@ -11,12 +11,14 @@ from __future__ import unicode_literals, absolute_import, print_function, divisi
 import requests
 import sopel.module
 from sopel.logger import get_logger
+from sopel import tools
 import re
-from urllib.parse import quote 
+from urllib.parse import quote
 
 LOGGER = get_logger(__name__)
 
-yearfmt = re.compile('\((\d{4})\)')
+yearfmt = re.compile(r'\((\d{4})\)')
+
 
 def setup(bot):
     if not bot.memory.contains('url_callbacks'):
@@ -24,8 +26,10 @@ def setup(bot):
     imdb_re = re.compile(r'(imdb\.com\/title\/)(tt[0-9]+)')
     bot.memory['url_callbacks'][imdb_re] = imdb_url
 
+
 def shutdown(bot):
     del bot.memory['url_callbacks'][imdb_url]
+
 
 @sopel.module.commands('movie', 'imdb')
 @sopel.module.example('.movie ThisTitleDoesNotExist', '[MOVIE] Movie not found!')
@@ -37,7 +41,7 @@ def movie(bot, trigger):
     if not trigger.group(2):
         return
     word = trigger.group(2).rstrip()
-    params={}
+    params = {}
 
     # check to see if there is a year e.g. (2017) at the end
     last = word.split()[-1]
@@ -49,10 +53,11 @@ def movie(bot, trigger):
     params['t'] = word
     bot.say(run_omdb_query(params, bot.config.core.verify_ssl, True))
 
+
 def run_omdb_query(params, verify_ssl, add_url=True):
     uri = "https://theimdbapi.org/api/find/movie?"
     if 'i' in params:
-        uri = uri.replace("/find","")
+        uri = uri.replace("/find", "")
         data = requests.get(uri + "movie_id={}".format(params['i']), timeout=30,
                         verify=verify_ssl)
     elif 'y' in params:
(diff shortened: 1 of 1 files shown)


dgw:
@freeboson and/or @kwaaak are you available to rebase this on current `master` and make sure it still works as expected? Along with the rebase, it would be great to re-enable and update the tests for this module as part of this update.

If this PR doesn't get updated, the `movie` module very well might just get dropped from core.

freeboson:
@dgw yes I'd like to fix this up. There's something about the status of the omdb or tmdb api access that I need to check. When are you planning to get 6.5.2 out by?

dgw:
@freeboson The main motivation for me taking over maintenance was the help module breaking. @hekel is working on that, so 6.5.2 will probably go out whenever that work is finished. Since this PR adds functionality, it will probably go in 6.6.0.

kwaaak:
Free keys available at http://www.omdbapi.com/apikey.aspx

freeboson:
@kwaaak yeah I just saw that. I would prefer switching back to omdb instead of using imdbapi. What do you think?

dgw:
Whether keys are available for free or not, it's been a long-standing point for Sopel not to require any API keys for core modules. Functionality that changes to require an API key (e.g. YouTube, Wolfram|Alpha) gets split out into a third-party module that isn't included by default when you install Sopel.

I'm not at all against using OMDB, but I am against requiring an API key for a module in Sopel's core.

So, either `movie.py` continues to work without an API key (which is what this PR does, I believe), or it starts requiring an OMDB API key and becomes its own project. Thoughts?

kwaaak:
I am running a working module with omdbapi, only downside is that the API key lives in the module file instead of the config: https://github.com/freeboson/sopel/pull/3

dgw:
@kwaaak That wouldn't be merged into Sopel's main branch even if core modules requiring API keys were desired, because the user should never have to edit a core module file to configure it.

kwaaak:
I know - just for freeboson to have a look - maybe it finds its way into a separate module.
Just found https://github.com/theapache64/movie_db though, maybe we can find the time to spin up a module without a key.

freeboson:
@kwaaak as you might have guessed, that guy's endpoint url is down, and has been for at least 8 days.

@dgw is it possible to have a module override a command? I'm thinking we get this working with "imdbapi" (no-key) but offer a separate module that replaces it with key-based omdb.

dgw:
@freeboson Multiple modules defining the same command will simply call both functions. An external "sopel-omdb" module could provide the same functionality, but users would have to exclude the built-in movie.py module to use it without conflicts, assuming both modules define the same commands.

freeboson:
@dgw can you disable a module from within another module? I also note that e.g. `safety.py` uses an API key and is in the main sopel.

dgw:
@freeboson I don't know how I missed this, sorry!

Modules can't affect each other, by design. They can only add triggers.

WRT `safety.py`, the API key is optional. Core modules should not _require_ API keys to work, but they _may_ support enhanced functionality if configured with an API key.

dgw:
I rebased this to squash out the fixup commits. Up next is proper testing & code review.

dgw:
Well, I was going to test this but the API endpoint it uses doesn't exist any more. `theimdbapi.org` is now parked, and redirects to ads.

Moving to "Work In Progress", and removing from the 6.6.0 milestone for now.

dgw:
Without an API to call on (that doesn't require a key), this rewrite is dead in the water. I don't think it's worth keeping stale code around in an open PR, so I'll close it to pare down the list.

@freeboson I hope you'll return to work on other things in the future—maybe you'll even find another keyless movie API and revamp this PR. Feel free to reopen it if you do! :smile_cat:

PR #1205, opened 2017-06-10
pronouns: Ignore whitespace

nfearnley:
Changed trigger.group(2) to trigger.group(3) in order to ignore trailing whitespace. (Occurs when user uses tab completion). Also fixed issue where wrong nick is given when pronouns are requested.

### DIFF
--- a/sopel/modules/pronouns.py
+++ b/sopel/modules/pronouns.py
@@ -42,10 +42,10 @@ def pronouns(bot, trigger):
             bot.reply("I don't know your pronouns! You can set them with "
                       ".setpronouns")
     else:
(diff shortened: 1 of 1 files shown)


sopel/modules/pronouns.py
```diff
@@ -55,7 +55,7 @@ def pronouns(bot, trigger):
             )
         else:
             bot.say("I don't know {}'s pronouns. They can set them with "
-                    ".setpronouns".format(trigger.group(2)))
+                    ".setpronouns".format(trigger.group(2).strip()))
```
dgw: Why is this still `trigger.group(2)`?

dgw:
Replaced by #1288

PR #1208, opened 2017-06-24
Currency module: fix API urls for Bank of Canada and BitcoinAverage

kwaaak:


dgw:
This is much, much more than just "Fix API urls"… Good to see the switch away from using the deprecated `web` tools though!

PR #1225, opened 2017-09-20
remove wat from version.py module

lenisko:
wat?... https://github.com/sopel-irc/sopel/commit/c676d1829ab07bdf9485d2baa7ed36446fba35c0


lenisko:
oh, it's there https://github.com/sopel-irc/sopel/pull/1161

PR #1228, opened 2017-09-22
Support to disable modules modules / modules methods per channel

lenisko:
Ability to disable modules or even methods inside chosen modules for channel name using default configuration file.

In this example we are disabling modules / modules methods on `#example` channel.
`disable_modules` is used to completely disable provided modules on channel for example `choice, weather` won't respond at all.

`disable_commands` is used to disable methods from provided modules for example `event_watcher` from `url` module which means all other methods should work (unless disabled are needed).

Adding other channel excludes is as simple as adding another section named like channel.
```
...

[#example]
disable_modules = choice,weather
disable_commands = {"url": ["event_watcher"]}
```

### DIFF
--- a/sopel/modules/help.py
+++ b/sopel/modules/help.py
@@ -60,6 +60,7 @@ def help(bot, trigger):
             name_length = max(6, max(len(k) for k in bot.command_groups.keys()))
             for category, cmds in collections.OrderedDict(sorted(bot.command_groups.items())).items():
                 category = category.upper().ljust(name_length)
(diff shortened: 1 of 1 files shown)


lenisko:
ah sorry, by mistake I have also added `help.py` change in this PR

dgw:
That'll happen if you open pull requests from your default branch. :)

lenisko:
I'll push it one more time

dgw:
Note from the future: This was replaced by #1235.

PR #1149, opened 2016-11-12
[weather] Use help_prefix in hint text when no location given

dgw:
Takes care of #1019 for the weather module. One of my users bumped into this tonight, so I decided to just fix it.

PR #1231, opened 2017-12-02
weather: update from deprecated sopel.web to requests

RustyBower:


dgw:
This undid my `help_prefix` fix from #1149, sigh.

PR #1236, opened 2017-12-04
help module - remove duplicates

lenisko:
Remove duplicates in help output. They are showing itself after plugin reload/unload... I don't really remember but worth taking care as well ;-)

### DIFF
--- a/sopel/modules/help.py
+++ b/sopel/modules/help.py
@@ -60,6 +60,7 @@ def help(bot, trigger):
             name_length = max(6, max(len(k) for k in bot.command_groups.keys()))
             for category, cmds in collections.OrderedDict(sorted(bot.command_groups.items())).items():
                 category = category.upper().ljust(name_length)
+                cmds = set(cmds)  # remove duplicates
                 cmds = '  '.join(cmds)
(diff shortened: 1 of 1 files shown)


dgw:
@hekel I can't formally request a review from you using GitHub's UI, but can you look at this?

lenisko:
@dgw nothing to test I guess

```python
>>> cmds = ['1', '2', '3', '3']
>>> '  '.join(cmds)
'1 2 3 3'
>>> cmds = set(cmds)
>>> '  '.join(cmds)
'2 1 3'
```

unless we care about order

dgw:
@lenisko The commands aren't sorted anyway, so nbd about the order. What I am worried about is the duplicated output being a small symptom of a bigger issue. Do duplicated commands run multiple times? If so, please do open an issue for that specifically (or search for one; I haven't finished going through all the open stuff yet).

lenisko:
@dgw Tbh I'm on my own fork so don't had a chance to test it atm. 
If you want to check it, try logging `cmds` to console and make few `.plugin reload x`, `.help x` I think you should see commands `y` times where `y` is amount of reloads.

Issue #1246, opened 2018-01-31
FreeBSD [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:777)


kitten77:
sopel/config/core_section.py
#  python3.6 [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:777)
```
from sys import platform

if 'freebsd' in platform:
    certs = '/etc/ssl/cert.pem'

def _find_certs():
    certs = '/etc/pki/tls/cert.pem'
    if not os.path.isfile(certs):
        certs = '/etc/ssl/certs/ca-certificates.crt'
        if not os.path.isfile(certs):
            return None
    return certs

```
Seems like there is missing a good validation on what platform it is running on and where the certs are stored.

diveyez:
Still got this issue
Connection error: [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:661)

diveyez:
That is with a certbot letsencrypt certificate

kitten77:
the path for letsencrypt certificates need to be set, on freebsd the certificates are
/usr/local/etc/letsencrypt/live/domain/fullchain.pem

dgw:
@kitten77 If #1249 gets merged, should that resolve (close) this issue?

@diveyez Can you also try with the changes from #1249 and see if you still have the issue?

kitten77:
Connection error: [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:661) 
seems to point to the cert file not found error.
@diveyez if he is using a different certificate than the standard os openssl he will need to add that part to the list inside #1249 before he tries to connect. I would recommend pushing the fix cause it is a cleaner way to fix this issue, and there should be a error handler placed on this pointing to the list for easy find and fixing these issues on a later point

alanhuang122:
@kitten77 That's not correct, unless you're assuming the server in question is serving an incomplete chain. In order for a Let's Encrypt certificate to be verified, a certificate bundle with DST Root CA X3 in it needs to be trusted, and the Let's Encrypt Authority X3 certificate needs to be presented by the server.

You cannot use `fullchain.pem` on the client side to accomplish anything, unless you explicitly add Let's Encrypt Authority X3 to a trust store, which you really shouldn't be doing.

kitten77:
[...]

PR #1249, opened 2018-02-20
TLS root cert fixes

tpltnt:
Hi there,

I documented and rewrote `_find_certs()` to support more platforms and be easier to extend. This PR was inspired by issue #1246.

Cheers
tpltnt

### DIFF
--- a/sopel/config/core_section.py
+++ b/sopel/config/core_section.py
@@ -17,12 +17,15 @@ def _find_certs():
 
     :returns: str (path to file)
     """
-    certs = '/etc/pki/tls/cert.pem'
-    if not os.path.isfile(certs):
-        certs = '/etc/ssl/certs/ca-certificates.crt'
-        if not os.path.isfile(certs):
-            return None
-    return certs
+    # check if the root CA store is at a known location
+    locations = [
+        '/etc/pki/tls/cert.pem',  # best first guess
+        '/etc/ssl/certs/ca-certificates.crt',  # Debian
+        ]
+    for certs in locations:
+        if os.path.isfile(certs):
+            return certs
+    return None
 
 
 def configure(config):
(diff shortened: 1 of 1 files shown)


kitten77:
good fix :)

dgw:
@kitten77 Did applying this fix #1246 for you?

anarcat:
there's a python package just for that which we might consider instead, [certifi](https://pypi.python.org/pypi/certifi/).

dgw:
We can consider it, but the certifi documentation site is unavailable at the moment (and has been since November), which doesn't exactly make me want to jump on it.

alanhuang122:
`certifi` is a dependency for `requests`, so it's already available, FYI - it's not a separate package

anarcat:
> We can consider it, but the certifi documentation site is unavailable at the moment (and has been since November), which doesn't exactly make me want to jump on it.

This is just a domain registration issue, documented here: https://github.com/certifi/python-certifi/issues/78

I wouldn't block on this: certifi is widely distributed, and rolling our own is a serious challenge that I don't think is within the scope of this project.

alanhuang122:
So, my two cents - what happens if (when) there are changes to the root certificate store? Python packages don't get updated automatically, so even though `certifi` tracks changes to the Mozilla trust store, it doesn't get updated when the system `ca-certificates` file does.

anarcat:
On 2018-03-24 15:28:45, Alan Huang wrote:
> So, my two cents - what happens if (when) there are changes to the root certificate store? Python packages don't get updated automatically, so even though `certifi` tracks changes to the Mozilla trust store, it doesn't get updated when the system `ca-certificates` file does.

I believe certifi uses the system trust store when available.

a.

kitten77:
@dgw yes it looks good for me thanks :)

PR #1251, opened 2018-02-20
adjust test to missing API key

tpltnt:
Hi there,

the imdb API key seems to be missing an makes two tests fail. This PR adjusts the tests for that. Feel free to reject as soon as an API key is provided for the tests.

Cheers
tpltnt


dgw:
I'm working on a test-fixing branch that will simply ignore the `movie` module. Because it requires an API key now, the module will probably be moved out of core anyway.

Issue #1080, opened 2016-05-20
[help] Bot can be DoS'd by spamming .help / .commands


dgw:
Perhaps it's at least partially a symptom of running my bot behind ZNC, but I recently had to remove the `help` module from active use because users were sending many `.help` commands in succession and causing the bot to appear unresponsive for long periods of time while the sending queue worked through the excessive number of lines generated when listing commands.

I don't recall seeing any other module generate anywhere near that much output in response to a single command, so I think it's worth bringing up. In 1252d5c, the output was converted from a simple alphabetical list of commands to a list of modules, each with their own command list. This increased the output of `.commands` on my instance from 4 lines to 44 lines. In addition, `.commands` is no longer a separate trigger; it was merged into `.help`, so now both trigger an absurdly long response.

Even if my bot wasn't behind a bouncer, Sopel's internal flood prevention still appears insufficient to prevent issues. It appears to be a trade-off between sending too slowly (and appearing to be unresponsive—this issue) or sending too quickly (and getting killed for excess flood—see #952).

While I love the idea of listing modules separately, giving a clearer indication of what commands are from which modules, and thus making it more obvious what they might do, producing nearly 15 times the output by line count is crippling; even when only one user is listing commands, it takes more than 30 seconds for the bot to output everything. Even taking Espernet's limits (10 messages per 5 seconds) and removing ZNC's admittedly cautious defaults from the picture, it would take around 20 seconds to send a command list to a single user.

One possible solution (which I and one of my users both thought of) is to dump the command list to a text or HTML file that the bot could link to. Obviously that's not possible for all Sopel instances, but it would be great to have as an option (and I just might prototype it for my own bot, since I already set up HTTP support for `meetbot`). The other major idea was to rate-limit use of the command-list function, but that gets messy. Is it limited per-user? per-channel? globally?

Anyone have thoughts? I'm happy to just leave `help` unloaded until I get around to implementing text/HTML dumping, but this is probably going to affect other bots too to some degree. (And it would be nice to know if I should PR the dump feature, too.)

dasu:
I have also ran into the same issue when 2 users decide to run .help at the same time.  Made worse if you have a lot of custom modules with help commands.  
In my situation a text/HTML dump file that I can host on my web server would be ideal.  Although that may not be the case for everyone.

mossarelli:
Use `@rate(seconds)` to limit the repetition of a command.
Also add rate to the `from sopel.module import commands, example`

[...]

PR #1257, opened 2018-03-20
Replace deprecated gist API in help

### LINKED ISSUE #1254: Anonymous gists are going away. Help module will break.
GitHub [announced](https://blog.github.com/2018-02-18-deprecation-notice-removing-anonymous-gist-creation/) last month that anonymous gist creation will be disabled on March 19, 2018 (UTC).

The `help` module was modified in fa89eb7 to essentially paste its output to a gist and simply return the link, and this functionality will break (as of writing) tomorrow. `.help` (or `.commands`) will return an error until (and unless) the module is updated to paste its contents elsewhere and/or create a static HTML file instead as I originally proposed in #1080.

(As an aside, I've been meaning to implement that HTML mode for ages. Maybe this is the kick I needed to finally get off my arse and write it. We'll see.)

hekel:
Resolves #1254

### DIFF
--- a/sopel/modules/help.py
+++ b/sopel/modules/help.py
@@ -3,6 +3,7 @@
 help.py - Sopel Help Module
 Copyright 2008, Sean B. Palmer, inamidst.com
 Copyright © 2013, Elad Alfassa, <elad@fedoraproject.org>
+Copyright © 2018, Adam Erdman, pandorah.org
 Licensed under the Eiffel Forum License 2.
 
 http://sopel.chat
@@ -11,8 +12,6 @@ from __future__ import unicode_literals, absolute_import, print_function, divisi
 
 import textwrap
 import collections
-import json
-
 import requests
 
 from sopel.logger import get_logger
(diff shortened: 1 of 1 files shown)


dgw:
Just as a general comment, replacing gists with another service that supports HTTPS would be ideal. Something like https://ptpb.pw/ perhaps?

kwaaak:
I like Hastebin (https://hastebin.com/about.md), a service with some ephemerality would be a good choice. Expiration in well under 24 hours should be more than enough.

dgw:
@anarcat See #1080 for one reason why Sopel no longer simply dumps the command list _en masse_. That said, changing from Gist to another pastebin provider (ix.io is not final) is merely one part of the solution. Eventually the plan is to have a local HTML/text option that can be hosted from the server Sopel runs on, and perhaps also have an option to turn the "deluge of PMs" method back on if it's really wanted.

anarcat:
hmm... yeah okay, that makes sense - thanks!

dgw:
@anarcat, want to take a second look before I merge this?

anarcat:
LGTM.

PR #1261, opened 2018-03-22
currency: limit precision to two decimal places

### LINKED ISSUE #1253: Results from currency conversion should be rounded to 2 decimals
At the moment, the results of a currency conversion query is reported as a float:

    < mynick> .cur 100 SEK to USD
    < mybot> 100.0 SEK (Swedish krona) = 12.202017114914426 USD (US dollar)

As far as I know, no currency reports more than 2 decimals, and some even less (JPY comes to mind). Rounding the result to 2 decimals will result in more readable output.

    < mynick> .cur 100 SEK to USD
    < mybot> 100.0 SEK (Swedish krona) = 12.20 USD (US dollar)

alanhuang122:
This PR fixes #1253.

Should it be possible to set a custom precision? It'd be an easy addition.

### DIFF
--- a/sopel/modules/currency.py
+++ b/sopel/modules/currency.py
@@ -74,7 +74,7 @@ def display(bot, amount, of, to):
         return NOLIMIT
 
     result = amount / of_rate * to_rate
-    bot.say("{} {} ({}) = {} {} ({})".format(amount, of.upper(), of_name,
+    bot.say("{:.2f} {} ({}) = {:.2f} {} ({})".format(amount, of.upper(), of_name,
                                              result, to.upper(), to_name))
 
 


dgw:
The vast majority of currencies use two decimal places, so a custom precision would be a "neat" feature rather than a must-have.

What could be nice, but is well beyond the scope of this PR (which should remain just a tweak that closes #1253 and gets into 6.5.2), is output precision based on the currency. For example, JPY would round up to the next integer, because you can't have a fraction of a yen. Dinar-based currencies (JOD, LYD, etc.) tend to use three decimal places. [ISO 4217](https://en.wikipedia.org/wiki/ISO_4217) for more information if you'd like to work on that for a future PR, @alanhuang122. :)

PR #1263, opened 2018-03-22
Fix IPython requirement

### LINKED ISSUE #1194: Cannot install dev dependencies on Python 2.7
This caused the failure in my PR #1193:
```
IPython 6.0+ does not support Python 2.6, 2.7, 3.0, 3.1, or 3.2.
    When using Python 2.7, please install IPython 5.x LTS Long Term Support version.
    Beginning with IPython 6.0, Python 3.3 and above is required.
    See IPython `README.rst` file for more information:
        https://github.com/ipython/ipython/blob/master/README.rst
    Python sys.version_info(major=2, minor=7, micro=9, releaselevel='final', serial=0) detected.
```
The dev-requirements therefore would need to be modified (or have a separate dev-requirements for python 2.7), pinning IPython to 5.x so that people could develop for that version?

alanhuang122:
IPython 6.0+ requires Python 3.3+, so only install that version if we're using Python 3.3+.

This fixes #1194.

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -1,2 +1,3 @@
 pytest
-ipython
+ipython<6.0; python_version < '3.3'
+ipython>=6.0; python_version >= '3.3'
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -1,3 +1,1 @@
 pytest
-ipython<6.0; python_version < '3.3'
-ipython>=6.0; python_version >= '3.3'
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,4 +3,5 @@ pytz
 praw
 pyenchant
 pygeoip
+ipython
 requests>=2.0.0,<2.11.0


dgw:
Looks good, though the `ipython` module is actually excluded from tests so… lol.

That said, while this shouldn't hurt anything, I don't think it's actually necessary. Installing IPython with my local `pip2` installed IPython 5.5.0, just as it should, because 6.x doesn't support py2. `pip` is generally smart enough to avoid installing incompatible package versions.

alanhuang122:
Yeah, I'd realized after opening this that builds were still succeeding even without this explicit specification. Not sure, in that case, what's going on with the issue.

There is no IPython requirement currently in requirements.txt, unless I'm blind.

dgw:
You're not blind, I'm just GitHubbing tired (which one should never do, but I do anyway on a regular basis). It's not there. Which is strange, because it's a runtime dependency, not a dev dependency.

Maybe take this patch and put the requirement where it belongs, and remove it from dev-requirements?

At some point it might be worth considering making niche stuff like IPython (both the requirement and the module, if possible) only install if specifically requested using the extra requirements feature of python (e.g. you'd need to `pip install sopel[with_ipython]` to get it), but that's a bigger project.

Issue #1210, opened 2017-07-27
.g (DuckDuckGo) showing yahoo ads again


dasu:
.g is starting to show yahoo ads again instead of relevant links.

To resolve, open search.py replace line 36 with:
`r_duck = re.compile(r'nofollow" class="[^"]+" href="(?!(?:https?:\/\/r\.search\.yahoo)|(?:https?:\/\/duckduckgo\.com\/y\.js))(.*?)">')` 

~~and in line 47, change `m.group(1)` to `m.group(3)`~~

~~(If anyone is really good at regex and know how to make it the first group, feel free to share)~~

antdude:
Which search.py do I edit to fix this issue?

$ locate search.py
/home/ant/SopelPython3/lib/python3.4/site-packages/pip/commands/search.py
/home/ant/SopelPython3/lib/python3.4/site-packages/sopel/modules/search.py
/usr/lib/python2.7/dist-packages/pip/commands/search.py
/usr/lib/python2.7/dist-packages/pip/commands/search.pyc
/usr/share/doc/python-xapian/examples/simplesearch.py

dasu:
Your second one ( /home/ant/SopelPython3/lib/python3.4/site-packages/sopel/modules/search.py )
Hopefully it's line 36... otherwise whatever starts with r_duck = ...

antdude:
Thanks Dasu, but it didn't seem to fix mine:
[08:56pm] <Ant> .g test
08:56PM <URL> Ant: /l/?kh=-1&uddg=https%3A%2F%2Fwww.speedtest.net%2F
[08:56pm] <Ant> .g ant
08:56PM <URL> Ant: /l/?kh=-1&uddg=http%3A%2F%2Fwww.youtube.com%2Fchannel%2FUCEfdlFLFtqXDIrrbWy78peQ

I see:
Function get is deprecated.
  File "/usr/lib/python3.4/threading.py", line 888, in _bootstrap
    self._bootstrap_inner()
  File "/usr/lib/python3.4/threading.py", line 920, in _bootstrap_inner
    self.run()
  File "/usr/lib/python3.4/threading.py", line 868, in run
    self._target(*self._args, **self._kwargs)
  File "/home/ant/SopelPython3/lib/python3.4/site-packages/sopel/bot.py", line 470, in call
    exit_code = func(sopel, trigger)
  File "/home/ant/SopelPython3/lib/python3.4/site-packages/sopel/modules/search.py", line 78, in duck
[...]

PR #1265, opened 2018-03-23
fix .g giving broken result

dasu:
Requires importing unquote and some regex changes.

The change I propose in #1210 doesn't fix the UK-en version, but does for us-en :thinking: 

Tested with 3.5, untested with 2.7

### DIFF
--- a/sopel/modules/search.py
+++ b/sopel/modules/search.py
@@ -11,9 +11,9 @@ import json
 import sys
 
 if sys.version_info.major < 3:
-    from urllib import quote_plus
+    from urllib import quote_plus, unquote
 else:
-    from urllib.parse import quote_plus
+    from urllib.parse import quote_plus, unquote
 
 
 def formatnumber(n):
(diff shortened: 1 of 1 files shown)


sopel/modules/search.py
```diff
@@ -12,9 +12,10 @@
 
 if sys.version_info.major < 3:
     from urllib import quote_plus
+    from urllib import unquote
 else:
     from urllib.parse import quote_plus
-
+    from urllib.parse import unquote
```
dgw: Can you collapse these imports like `from urllib.parse import quote_plus, unquote`?

And flake8 will complain about removing the second blank line before `def formatnumber`.

dasu:
Sorry, I can't do it (using github web, and not git).  You should have the option to if you click the drop down next to merge pull request and select 'Squash and merge'.

Squash & Merge

dgw:
"Squash and merge" is not a workflow I like to use, but since you allowed edits to your branch from maintainers I was able to do the squash/rebase myself.

Issue #570, opened 2014-07-13
[Core]Connecting with SSL when CA Certs are unavailable needs better error message


tyrope:
Found by @Connor22

```
Could not open CA certificates file. SSL will not work properly.
Could not open CA certificates file. SSL will not work properly.

Welcome to Willie. Loading modules...
[snip]
Registered 33 modules,
9 modules failed to load


Connecting to irc.twitch.tv:6667...
Connection error: [Errno 2] No such file or directory
```

dgw:
@tpltnt You recently submitted a TLS-related PR, so would you mind giving this a look at some point?

tpltnt:
I am using the following configuration:
```
[core]
nick = Sopel
host = irc.twitch.tv
use_ssl = true
port = 6697
owner = foo
channels = #test
```
Running sopel yields:
```
./sopel.py               

Welcome to Sopel. Loading modules...




Registered 42 modules,
0 modules failed to load


Connecting to irc.twitch.tv:6697...
Invalid certficate, hostname mismatch!
Traceback (most recent call last):
  File "/home/foobar/code/sopel/sopel/irc.py", line 215, in handle_connect
    ssl.match_hostname(self.ssl.getpeercert(), self.config.core.host)
  File "/usr/lib/python3.5/ssl.py", line 305, in match_hostname
    % (hostname, ', '.join(map(repr, dnsnames))))
ssl.CertificateError: hostname 'irc.twitch.tv' doesn't match either of '*.chat.twitch.tv', 'chat.twitch.tv'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/foobar/code/sopel/sopel/__init__.py", line 81, in run
    p.run(config.core.host, int(config.core.port))
  File "/home/foobar/code/sopel/sopel/irc.py", line 155, in run
    self.initiate_connect(host, port)
  File "/home/foobar/code/sopel/sopel/irc.py", line 171, in initiate_connect
    self.connect((host, port))
  File "/usr/lib/python3.5/asyncore.py", line 341, in connect
    self.handle_connect_event()
  File "/usr/lib/python3.5/asyncore.py", line 429, in handle_connect_event
    self.handle_connect()
  File "/home/foobar/code/sopel/sopel/irc.py", line 218, in handle_connect
    os.unlink(self.config.core.pid_file_path)
AttributeError: 'CoreSection' object has no attribute 'pid_file_path'
```
[...]

PR #1266, opened 2018-03-24
handle TLS and CNAMEs gracefully

tpltnt:
Hi there,

after looking at #570 I decided to handle CNAMEs gracefully. This PR looks if CNAMEs match the TLS certificate in case the configured host does not match.

Cheers,
tpltnt


sopel/irc.py
```diff
@@ -214,9 +214,14 @@ def handle_connect(self):
...
-                    os.unlink(self.config.core.pid_file_path)
-                    os._exit(1)
+                    # try to recover if CNAMEs are the issue
+                    for cname in self._get_cnames(self.config.core.host):
+                        try:
+                            ssl.match_hostname(self.ssl.getpeercert(), cname)
+                        except ssl.CertificateError:
+                            stderr("Invalid certficate, hostname mismatch!")
+                            os.unlink(self.config.core.pid_file_path)
+                            os._exit(1)
```
dgw: The way this is structured, it will exit if the first CNAME doesn't match and never try the others if they exist.

````
dgw@Kleiner $ python -i
Python 3.6.2 (default, Sep  7 2017, 13:16:50)
[GCC 6.3.0] on msys
>>> import os
>>> listme = [0, 1, 2]
>>> for i in listme:
...     try:
...         raise IndexError
...     except IndexError:
...         print("Tried {}".format(i))
...         os._exit(1)
...
Tried 0
dgw@Kleiner $ 
````

Granted, having more than one CNAME is probably rare, but if the code _looks_ like it's trying to handle multiple CNAMEs, it _should_ handle them.

Comment (tpltnt) [reply sopel/irc.py]: You are right. I was quite tired when I hacked it together. I fixed it now.

File: sopel/irc.py [hunk 2]
```diff
@@ -211,30 +216,35 @@ def handle_connect(self):
...
-                            stderr("Invalid certficate, hostname mismatch!")
-                            os.unlink(self.config.core.pid_file_path)
-                            os._exit(1)
+                # try to handle both names in config, certificate and CNAMEs
+                hosts_to_match = [self.config.core.host] + self._get_cnames(self.config.core.host)
+                has_matched = False
+                for hostname in hosts_to_match:
+                    try:
+                        ssl.match_hostname(self.ssl.getpeercert(), hostname)
+                        has_matched = True
```
dgw: This should break out of the for loop once a match is found.

requirements.txt
```diff
@@ -4,3 +4,4 @@ praw
 pyenchant
 pygeoip
 requests>=2.0.0,<2.11.0
+dnspython
```
dgw: Rebase needed because requirements were updated on `master`.

tpltnt:
rebased and new commit messages

dgw:
@tpltnt Thanks. :) Any thoughts on breaking out of that loop immediately if a CNAME matches?

tpltnt:
I thought about it and decided against it. This code is slightly inefficient, but makes error handling easier to understand and follow if the certificate does not match at all.

dgw:
Wait, can't you just add `break` after `has_matched = True`?

tpltnt:
done that

ekohl:
This feels like a weird construct like no other SSL/TLS client that I know. Normally you must ensure the correct name is on the certificate. You can include multiple names via subjectAltName. This makes me wonder if it opens a security risk.

tpltnt:
@ekohl I hear you. The cert I saw (if I remember correctly) did not have any subjectAltName values, hence my background check for CNAMEs. The security implication arises from the ability of an attacker to add CNAME entries on the nameserver(s) responsible for the domain of the IRC server. I assume the administrators had their reasons to end up with their configuration.
Whitelisting certain domains implies maintainance work for sopel (which in turn gives rise to the same potential security risk). The proper solution (socially) is to establish a feddback channel to the server administrators about the state of their TLS config (preferably programatically, since very few people are going to debug TLS issues when their IRC client is failing). I am not sure if that is a task that is on sopel.

ekohl:
> The proper solution (socially) is to establish a feddback channel to the server administrators about the state of their TLS config (preferably programatically, since very few people are going to debug TLS issues when their IRC client is failing).

You're right. Supporting incorrectly configured servers often leads to subtle bugs and weird interactions. If I was a maintainer of this project I wouldn't want to maintain that burden.

The service in #570 tells you you should be using `chat.twitch.tv` rather than `irc.twitch.tv` for TLS connections to that particular server:

```
X509v3 Subject Alternative Name: 
    DNS:*.chat.twitch.tv, DNS:chat.twitch.tv
```

Could you verify that works as well?

dgw:
I have absolutely no problem removing this feature if it turns out to be problematic.

It might be prudent to include a warning in Sopel's output if the matched hostname is not the one entered in the config file, though. Preferably, try to match the configured hostname first before even checking CNAMEs, then query for CNAMEs and print a warning that a potentially insecure fallback mode is in use. @tpltnt, would you be up for adding that to this PR?

I'm not _too_ terribly worried about security here because ultimately, Sopel is more of a toy than anything. If security is important, one should make sure to use the correct hostname in the config. (I'm also not against making this behavior off by default and enabling it with a config flag.)

tpltnt:
I modified the code to check for the hostname in the config first and only fall back to CNAMEs if there is no match. The code is also peppered with more verbose logging. If you think this is not a good way to handle the situation I am cool with dropping this PR.

sopel/irc.py
```diff
@@ -211,25 +216,45 @@ def handle_connect(self):
...
+                            LOGGER.warning("using {0} instead of {1} for TLS connection".format(hostname,
+                                                                                             self.config.core.host))
+                            has_matched = True
+                            break
+                        except ssl.CertificateError:
+                            pass
+                    if not has_matched:
+                        # everything is broken
+                        stderr("Invalid certficate, hostname mismatch!")
+                        LOGGER.error("invalid certficate, no hostname matches")
```
dgw: Can you amend the tip commit to fix these typos? `certficate` -> `certificate`

sopel/irc.py
```diff
@@ -211,25 +216,45 @@ def handle_connect(self):
...
-                    os._exit(1)
+                    # the host in config and certificate don't match
+                    LOGGER.error("hostname mismatch between configuration and certificate")
+                    # check (via exception) if a CNAME matches as a fallback
+                    has_matched = False
+                    for hostname in self._get_cnames(self.config.core.host):
+                        try:
+                            ssl.match_hostname(self.ssl.getpeercert(), hostname)
+                            LOGGER.warning("using {0} instead of {1} for TLS connection".format(hostname,
+                                                                                             self.config.core.host))
```
dgw: Along with the typos below, please move the `.format(hostname,` to the next line and align it with the containing function call, so this looks less weird.

````
                            LOGGER.warning("using {0} instead of {1} for TLS connection"
                                            .format(hostname, self.config.core.host))
````

PR #1267, opened 2018-03-24
Allow any 2.x version of requests

dgw:
Testing with the latest `requests` (2.18.4 as of writing) showed no issues with all the Sopel modules and functions that I was able to try out.

Newer versions of PRAW (which the `reddit.py` module needs) prefer to be installed with more recent versions of `requests`, and I see no reason not to keep up with `requests` releases unless something breaks.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,4 +4,4 @@ praw
 pyenchant
 pygeoip
 ipython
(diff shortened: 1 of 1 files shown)


dgw:
Treating @kwaaak's :+1: reaction as an approval.

PR #1268, opened 2018-03-24
Require PRAW >= 4.0.0

dgw:
The new `reddit.py` module (updated in December) will not work with PRAW older than 4.0.0. Sopel must therefore require a compatible version.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,6 @@
 xmltodict
 pytz
-praw
+praw>=4.0.0,<6.0.0
 pyenchant
 pygeoip
 requests>=2.0.0,<2.11.0


Helldembez:
A small thing to consider is that from what I can find the latest version of PRAW is 5.3.0. Possibly breaking changes can be introduced between 5.3.0 and 6.0.0. This is unlikely since it are minors, but possible. Whenever PRAW releases a new version its easier to test it and then support it instead of supporting non existing future versions.

https://github.com/praw-dev/praw/releases

dgw:
@Helldembez Worry not, I learned that lesson from working on the Wolfram module. :grinning: This version bump requires 4.0.0 up to (but not including) the future 6.0.0.

PR #1271, opened 2018-03-25
Clean up badges in README

dgw:
PyPI download count information is no longer available, so that badge is patently useless now (it literally shows "no longer available" as the download count).

I understand the sense of whimsy that comes from saying "works: usually" or "ages: 12+" in badges, but at the same time the badge real estate at the top of a README is for practical information that potential contributors/users need to know. I'm also not sure there's any value in bragging about how many stars/forks the project has in the README. That information isn't really relevant on PyPI, where the README gets the most exposure, and is readily available from GitHub's UI for users visiting the project there.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1,4 +1,4 @@
-|version| |downloads| |license| |build| |issues| |forks| |stars| |ages| |works| |badges|
+|version| |build| |issues| |coverage-status| |license|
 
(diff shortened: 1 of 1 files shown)


dgw:
Considering the :+1: from @alanhuang122 as an approval, more or less.

PR #1273, opened 2018-03-25
updated wiki submodule

nfearnley:
I've updated the wiki submodule reference to point to the correct url.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,4 +1,4 @@
 [submodule "docs/wiki"]
 	path = docs/wiki
-	url = git://github.com/embolalia/willie.wiki.git
+	url = git://github.com/sopel-irc/sopel.wiki.git
 	branch = master


nfearnley:
The travis build failure appears to be related to the py module service being down, and unrelated to this patch.

dgw:
@nfearnley I kicked Travis and the check passed. Sometimes the `.py` service goes over quota.

nfearnley:
It looks like we have to manually pull and recommit every time the wiki changes, if we want the main repo to keep up to date. Not sure if we'd want to make a standard time for doing this, or maybe just remove the submodule altogether.

dgw:
Yep, that's how git submodules work. They point to a specific commit in the foreign repository, so when you check out an old version of the repo you get all the right versions of its submodule dependencies.

Realistically, what's the benefit of having the wiki as a submodule? I have never seen Sopel install with wiki documentation, and searching the code for "docs" or "wiki" mostly turns up links in comments. Does the submodule actually do anything in particular?

dgw:
Still curious about the utility of this submodule, but didn't want to hold off merging.

nfearnley:
It's just something for future thought. The submodule was already in place,
but I didn't really see the point in having it in the repo. I just fixed it
and updated it. I guess we can consider removing it in the future if it's
just a hassle to maintain with little benefit.

PR #1274, opened 2018-03-26
Added sphinx command line doc generation

nfearnley:
Sphinx now automatically generates documentation for the sopel command. It uses the [Sphinxcontrib-Autoprogram](https://pythonhosted.org/sphinxcontrib-autoprogram/) extension to do so.
I'm not sure where we should be putting dev requirements for document generation. Sphinx doesn't seem to be referenced anywhere in the existing requirements documents.

The man page is now split into two separate sections: `sopel.1` (command) and `sopel.3` (api library).

I had to move the argparse parser declaration outside of the run_script.py's main() in order to allow the documentation tool to read it.


nfearnley:
The build failure seems to be related to the py module, like my other pull request, and not related to this PR.

dgw:
@nfearnley Kicked Travis, built fine. Just like the other PR.

Looks like I have some stuff to learn about. Sphinx, specifically. It won't be a big deal for 6.5.2, but with 6.6.0 and beyond there will be documentation updates and I have to learn how to build those. Since you seem to have figured out a fair bit of this, I might ping you on IRC some time about it.

dgw:
I'm touching a couple things related to docs as I make final preparations to release tomorrow, but they shouldn't cause conflicts. (I did notice a couple whitespace errors in this PR, though. Check for trailing spaces. :smile_cat: Some of them will be caught by `checkstyle.sh` when it checks the Python files, but the script doesn't look at rST documents.)

You might as well put the Sphinx stuff in dev-requirements.txt. That's where they should be. I can only guess why they're not included already (probably just forgotten).

Especially with these changes to have `make` build suitable manpages already, is there any reason to keep `docs/willie.man`? It hasn't been updated in five years (the file date in the footer is January 15, 2013), and obviously refers to a very old version of the bot. What it documents, your changes to auto-generate the command docs in `sopel.1` will take care of documenting. Feel free to add a commit to this PR that deletes the existing (very outdated) file, since any future versions will be built with Sphinx.

nfearnley:
I'll fix up those space formatting issues, and add to dev-requirements. My concern about deleting willie.man at this point is that the man file generation only generates the bare minimum man file, while the willie.man file contains some extra details we might want to add later.

dgw:
While building the manpages from this PR to compare them, I noticed that `sopel.1` and `sopel.3` both contain the command-line arguments. The idea of splitting into two manpages was to separate the CLI stuff and the API stuff, right? Fixing that means specifying the sections to include in `sopel.3` explicitly instead of just building `index`. I think... haven't looked too much into Sphinx yet (but I did look enough to see that `man_show_urls` should probably be set to `True`).

Now that I think about it, the `config` section probably belongs in `sopel.1` too, because it's directly related to running the bot. (And a tiny nitpick: Put the manpages in numeric order in the conf.py file, however many sections you end up using.) I honestly don't know if what I'm asking for is possible—again, I'm just spitballing right now without having read any Sphinx docs. I don't know if changing the manpage stuff around like this will mess up the HTML build or anything. But that's what testing is for…

As far as `willie.man` goes, I expect many (most? all?) of the extra details present in it can be added to the ArgumentParser, with the added benefit that they'll also be displayed at runtime if the user calls `sopel --help` as well as in the generated manpage.

Definitely feel free to kind of spam commits at this, keeping things separated, since the commits (and their messages) can be fixed up into larger changesets with an interactive rebase before merging.

nfearnley:
I noticed that duplication in willie.3. That's because I include cmd.rst in the index.rst (which willie.3 is based on). I thought including the command line documentation in the HTML output outweighed the duplication in the man output. We could fix this by properly creating a subsection specifically for api documentation, so willie.3 could reference that.

I considered adding the details to the argparse parameters but didn't want to make too many changes or spam the output. Though if you're good with it, I'll go ahead and add that in.

dgw:
If you can build a special subsection for the manpage and keep the HTML output unchanged (except for useful side effects, like including the command-line options in the HTML), i.e. don't output a `man3.html` (or whatever the file is called), that's dandy. I think it should be fine, because the HTML build just uses `index.rst`? Having multiple indexes is just how it has to work if you want to split up the documentation differently based on what format is being built, right? No big deal, it is what it is.

Technically, yes, the changes to the _content_ (rather than location) of the argparse help info should probably go in a separate PR. But I'll leave it up to you whether you want to juggle multiple branches like that, since you'd have to rebase one PR after the other gets merged because they'd touch the same lines.

nfearnley:
I'm looking through the documentation, and I think I can divide it like this:

```
index
    sopel command
        command line parameters
        config file
        The [core] configuration section
    sopel modules
        plugin structure
        The bot and it's state
        Triggers
        Configuration functionality
        The bot's database
        Additional API features
```

The index used for html document generation will still cover all the pages. We can have the sopel.1 generated from the "sopel command" page, and the sopel.3 generated from the "sopel modules" page. I can split the "Configuration functionality" into two pages: one covering how to access the configuration from a module, and one covering the [core] section.

dgw:
Ooh, I like it. Run with that idea, would you? Sounds great!

dev-requirements.txt
```diff
@@ -1,2 +1,4 @@
 pytest
 ipython
+Sphinx
```
dgw: Sphinx just has to be different and capitalize its package name, huh? smh lol

dgw:
@nfearnley Besides needing a rebase, did you do everything you wanted to do in this PR?

dgw:
Considering this stalled and punting to 7.0.0

Exirel:
I'm very interested in this PR. However, a lot has changed since it was first published—a year ago now!

If you don't mind, I'll reuse this work and adapt it for our current codebase. It's a great idea, and it would be a shame if we didn't reuse it. That being said, I'll give all the credits to @nfearnley because I'll mostly adapt her work.

Exirel:
Closing here since #1564 has been published. I don't forget the conversation about restructuring the documentation to prevent duplication in man page 1 and 3, nor the idea to look at the old Willie's man page.

PR #1277, opened 2018-03-27
[help] use help_prefix instead of assuming "."

hekel:


sopel/modules/help.py
```diff
@@ -19,10 +19,13 @@
...
 logger = get_logger(__name__)
 
+def setup(bot):
+    global help_prefix
+    help_prefix = bot.config.core.help_prefix
 
 @rule('$nick' '(?i)(help|doc) +([A-Za-z]+)(?:\?+)?$')
 @example('.help tell')
-@commands('help', 'commands')
+@commands('help', 'commands', 'halp')
```
dgw: This is unrelated to the prefix.

PR #1278, opened 2018-03-27
Start failing builds on style checks

dgw:
With all the little style errors cleaned up in this commit, flake8 will now pass. That means checkstyle.sh can be allowed to fail builds in PRs and enforce all these little things going forward.

No doubt there will be conflicts with some open PRs, but a healthy number of those are either awaiting changes anyway (and can be rebased as part of that) or likely to be closed unmerged for various reasons.

None of these changes should affect functionality, but in a few places I did correct "bare except" with something more specific than `except Exception:` and a `TODO` comment, where I was reasonably sure what exception type might be raised. But with only ~40% test coverage, there's a lot of code that can't be vetted by Travis just yet, and so this is flagged "Needs Review".

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,7 @@ install:
   - pip install coveralls
   - pip install flake8
 script:
-  - ./checkstyle.sh || true
+  - ./checkstyle.sh
   - coverage run --source sopel -m py.test -v .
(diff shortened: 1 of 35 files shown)


dgw:
Thanks for the sanity check, @nfearnley!

Issue #1248, opened 2018-02-18
[etymology] Site HTML changes broke etymology module


dgw:
etymonline no longer uses nice, clean `dt`/`dd` markup, sadly. There will need to be a little work done to rewrite the parser in `etymology.py` so it can pull etymologies from the new HTML structure.

PR #1279, opened 2018-03-27
Etymology with lxml

kwaaak:
https://github.com/sopel-irc/sopel/issues/1248


dgw:
OK, so here's the thing... `lxml` used to be a requirement, and was removed in 21bbd98. I'm not sure how I feel about putting it back. Getting rid of it as a dependency in the first place was a sizable task.

Sopel does still have the `xmltodict` requirement. Does that help you at all with this change?

PR #1286, opened 2018-03-29
[calc] Switch to Sopel's own tumbolia instance

dgw:
tumbolia-two is shared with a number of jenni instances and forks, so it is often over quota for hours at a time. That's inconvenient for users of Sopel, and inconvenient for the development process (because there are tests that rely on the service).

If we run into the same quota issues with our own instance, we can make the URL configurable and provide documentation on setting up your own.


dgw:
Thank you, Travis. :shipit:

PR #1288, opened 2018-03-29
Fixed space stripping in pronouns.py

nfearnley:


sopel/modules/pronouns.py
```diff
@@ -54,7 +54,7 @@ def pronouns(bot, trigger):
             )
         else:
             bot.say("I don't know {}'s pronouns. They can set them with "
-                    ".setpronouns".format(trigger.group(2).replace(" ", "")))
+                    ".setpronouns".format(trigger.group(2).strip()))
```
dgw: Same question I asked in #1205: Why isn't this also `trigger.group(3)`?

PR #1293, opened 2018-04-01
Update to testing process

nfearnley:
Moved coveralls and flake8 install to dev-requirements.txt
Moved coverage parameters to .coveragerc config file
Added .pytest_cache to .gitignore
Updated "willie" to "sopel" in comments
Added dev extras_require to setup.py

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -54,5 +54,6 @@ setup(
     license='Eiffel Forum License, version 2',
     platforms='Linux x86, x86-64',
     install_requires=requires,
+    extras_require={'dev': dev_requires},
     entry_points={'console_scripts': ['sopel = sopel.run_script:main']},
 )
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,5 +1,6 @@
 # Sample conf file from http://nedbatchelder.com/code/coverage/config.html
 [run]
+source = sopel
 branch = True
 
 [report]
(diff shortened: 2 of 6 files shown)


.gitignore
```diff
@@ -5,6 +5,7 @@ MANIFEST
 doc/build/*
 logs/*
 tests
+.pytest_cache
```
dgw: This is already included below, next to `.cache` and `.coverage`.

setup.py
```diff
@@ -54,5 +54,6 @@ def read_reqs(path):
     license='Eiffel Forum License, version 2',
     platforms='Linux x86, x86-64',
     install_requires=requires,
+    extras_require={'dev': dev_requires},
```
dgw: This allows doing `pip install sopel[dev]`, right? Nice.

nfearnley:
Yeah. I was trying to look into the best way to separate dev and production
requirements. In the future we might want to do more fine grain
separations, such as docs, test, etc. Or not, depending on needs.

nfearnley:
I noticed that it wasn't ignoring my .pytest-cache, but I was likely
working on a commit before it was added.

nfearnley:
Yes, I see you added it 4 days ago, and the branch I was working in was a
hit older.

dgw:
@nfearnley FYI when you reply by email, Github just dumps your comments into the top-level PR thread instead of whatever line note you responded to. (That's something they should probably work on, tbh. Replying by email is really limited, I've learned.) I know what you're responding to anyway, though. This is good to go once the duplicate gitignore entry is amended out.

PR #1295, opened 2018-04-02
Added test for pytest module before inserting example test functions

nfearnley:


nfearnley:
flake8 is failing on me because I'm importing pytest without actually using it. Not sure what best approach to this is. Any suggestions?

dgw:
There's always `imp.find_module()`, but we'd be using a library that's already deprecated in Python 3. The py3 replacement, `importlib`, has multiple ways of checking the availability of a module too, depending on which version of py3 is installed. Basically, it's kind of a messy route.

The simplest solution is just adding `del pytest` after the import. I tested it just now, and flake8 is happy. The tests appear to run and pass as always. Maybe do that as a quick fix for now.

Later, ideally the test functions shouldn't be inserted into modules if tests aren't being run. That can be a separate project for one of us, for a future release (6.6 or 7).

dgw:
You know what? I just found this: https://stackoverflow.com/a/44595269/5991

> A simpler solution I came to:
>
> ````
> import sys
>
> if "pytest" in sys.modules:
>     ...
> ````
>
> Pytest runner will always load the `pytest` module, making it available in `sys.modules`.
>
> Of course, this solution only works if the code you're trying to test does not use `pytest` itself.

The `@example` decorator doesn't use `pytest` except in this new check you've added, so maybe this is the way to go?

nfearnley:
sys.modules only lists modules that have loaded, right? So we'd still have to attempt to load the module anyways.

dgw:
Why would we have to attempt to load it? If Sopel is running in a Pytest session, `pytest` is loaded by the test framework. It will be loaded already. Run #1296 if you don't believe me, lol.

nfearnley:
Sorry, I missed the part where the pytest runner would load it.

dgw:
Went with #1296 for now.

PR #1297, opened 2018-04-02
Prune tests from release distribution

dgw:
I'd like to exclude `sopel.test_tools` as well, to have as clean a build as possible at release, but since that module is imported in several places that change will take longer to make and validate.

These files aren't even installed, so why are they included in the sdist?

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,1 +1,2 @@
 include *requirements.txt
+prune test/


dgw:
That DuckDuckGo, being dumb. Test failure unrelated to this PR, all it changes is comments and little things like print statements.

PR #1299, opened 2018-04-02
Additional config tests

dgw:
I spent some time this morning figuring out more of how the tests actually work, and expanding the test suite to cover all attribute types in `sopel.config.types`. It was a fun exercise, with the added benefit of catching any future breakage due to changes in how the config stuff works.

`test_config` is the only test module that uses `unittest`, apparently. Maybe I'll work on that later.


dgw:
Build fails on py2, interesting. I probably won't get around to fixing it until tonight or tomorrow.

PR #1300, opened 2018-04-03
formatting: add italics

dgw:
I never noticed before that italics weren't a thing already in `sopel.formatting`, but indeed, the only existing options were colors, bold, and underline. I'm glad I decided to go through and make sure the existing test modules cover as much as possible, because that's how I noticed this.


dgw:
Merging to ease my own further work on formatting stuff. Dependent branches are kinda hell.

PR #1301, opened 2018-04-03
formatting: strikethrough & monospace

dgw:
This PR currently shows extra changes because the branch is based on the italics branch (#1300) to avoid merge conflicts later. By necessity, both PRs have to touch some of the same lines.

Italics are a definite "go" for 6.6.0, but I'm not sure if these newer formatting options will be included in that release. These could be held back for another version or two to see about uptake in the client ecosystem. From personal experience I know that https://modern.ircdocs.horse/formatting.html isn't quite accurate when it says that strikethrough and monospace only work in Textual and IRCCloud (respectively)—for example, The Lounge has supported both for some time now. But they are still new, and not as widely supported as the other formatting characters Sopel supports (including italics).


dgw:
I decided to include all the new formatting stuff in 6.6.0, because why not. It won't break anything, after all.

The color stuff left to do in #1302 will come separately, and soon™ (meaning whenever I get around to it).

Issue #1200, opened 2017-06-07
.help only shows 1 example


hemna:
Looking at the code for the calc command, it has declared several examples.  When I run ".help calc",  I always get one and only one example.

```
 <hemna> .help calc
 <pewp> hemna: Evaluate some calculation.
 <pewp> hemna: e.g. .c 5 / 2
```

The calc code is here: 
https://github.com/sopel-irc/sopel/blob/master/sopel/modules/calc.py#L22-L29

I'm writing my own module, and I am trying to get all of the examples to show, but it functions the same as the calc example above.  No matter how many `@example` decorators I have on the method, it only shows one.

dgw:
The `@example` decorator is mostly for auto-generating tests, and the `help` module just kind of uses it too. I'm putting in a fair amount of work on `@example` for the next version, and might try to tackle this. (If I don't, @hekel might. :grin:)

dgw:
Did some work on this in 3fe610c. Not final, subject to change, etc. Most importantly it kind of relies on #1303, so if that gets postponed so will this.

dgw:
Pushed to Sopel 7, as my WIP solution requires a new argument to `@sopel.module.example()`, which I do not want to add in a point release.

dgw:
Merged a fix for this that modules will have to opt into. See #1403.

PR #1303, opened 2018-04-03
loader, module: Fix help_prefix cleanup & example docs

### LINKED ISSUE #1283: Example "e.g." text for help commands omits first character and includes escape character
Let there be these lines in the configuration:
`help_prefix = \-`
`prefix = \-`

Using the help command to get detailed help for something produces this output:
`<kwaaak>	-help time`
` <Sopel>	kwaaak: Returns the current time.`
` <Sopel>	kwaaak: e.g. \- America/New_York`

This is generated by:
https://github.com/sopel-irc/sopel/blob/master/sopel/modules/help.py#L52

I have looked briefly into figuring out bot.doc, but did not manage to do it so far.

dgw:
Callable cleanup in loader now always uses help_prefix's default value to replace. The documentation for module.example() reflects this, and I also moved the constructor documentation up to where it would be included in the auto-generated API docs (so the parameters are now visible in the public documentation without reading the code).

This should resolve #1283. I might add sanity checking for `help_prefix` (e.g. checking its length and printing a warning on startup if it's longer than one character), based on feedback from @kwaaak.

### DIFF
--- a/sopel/loader.py
+++ b/sopel/loader.py
@@ -7,6 +7,10 @@ import re
 import sys
 
 from sopel.tools import compile_rule, itervalues, get_command_regexp, get_nickname_command_regexp
+from sopel.config import core_section
+
+default_prefix = core_section.CoreSection.help_prefix.default
+del core_section
 
 if sys.version_info.major >= 3:
     basestring = (str, bytes)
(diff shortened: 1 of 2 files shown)


dgw:
Merging this will let me clean up my branch that addresses #1200. It's the oldest "ready-to-go" PR at present, and I do prefer merging in FIFO order when possible.

PR #1312, opened 2018-04-05
reddit: Cross-compatibility for praw

kwaaak:
Work with a wide range of praw versions

### DIFF
--- a/sopel/modules/reddit.py
+++ b/sopel/modules/reddit.py
@@ -47,13 +47,17 @@ def shutdown(bot):
 
 @rule('.*%s.*' % post_url)
 def rpost_info(bot, trigger, match=None):
-    r = praw.Reddit(
-        user_agent=USER_AGENT,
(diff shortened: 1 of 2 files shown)


sopel/modules/reddit.py
```diff
@@ -47,13 +47,17 @@ def shutdown(bot):
...
     match = match or trigger
-    s = r.submission(id=match.group(2))
+    try:
+        r = praw.Reddit(
+            user_agent=USER_AGENT,
+            client_id='6EiphT6SSQq7FQ',
+            client_secret=None,
+        )
+        s = r.submission(id=match.group(2))
+    except Exception:
```
dgw: I don't like this, but there's no practical way of catching the real exception because it's moved locations in `praw`. I'm not a fan of having a try/except tree of imports just for this, so I'll let it slide. :stuck_out_tongue:

Issue #1316, opened 2018-04-10
Remove remaining uses of sopel.web in core


dgw:
All remaining core modules that still use `sopel.web` should be updated to use `requests` directly instead. I'll be doing this in bits of spare time when I don't feel like working on anything substantial.

Modules I found in my search (and migration status):

- [x] bugzilla
- [x] etymology
- [x] isup
- [ ] ~~meetbot (still using `quote`, which should be replaced or refactored)~~ NAK for now, pending #1318
- [ ] ~~reddit (using `sopel.web.USER_AGENT`, which should probably be defined somewhere else)~~ NAK for now, pending #1318
- [x] safety
- [x] search
- [x] tld
- [ ] ~~translate (imports _both_ `requests` and `sopel.web`...because it needs `web.decode()`)~~ NAK for now, pending #1318
- [ ] ~~url (uses `web.decode()` and `web.iri_to_uri()`)~~ NAK for now, pending #1318
- [x] wiktionary (left `web.quote()` in for now)

dasu:
isup: web.get is used, but i'm wondering if we should do requests.head instead.  Would save on downloading the whole site to get an answer the headers would have provided anyways.

PS: etymology doesn't seem to be working

dgw:
web didn't have a head method, so switching is totally on the table.

etymology not working is a separate issue, #1248

dgw:
Closing this, as everything that was planned has been done. Can be reopened and updated if anything happens with #1318 (or a new tracking issue created for the moved functions).

PR #1319, opened 2018-04-30
Remove deprecated sopel.web calls from core modules

dgw:
All calls to `sopel.web.get()`, `sopel.web.post()`, and `sopel.web.head()` should be gone now.

This is part of #1316, but does not close the issue as there is still further work to do for Sopel 7.


dgw:
What do I get for not running `checkstyle.sh` before pushing? Travis errors, that's what. (Fixed.)

PR #1321, opened 2018-05-01
Add missing docstring to 'set_channel_value' function

ebarrier:
I was reading [Sopel's documentation](https://sopel.chat/docs/db.html) when I noticed that an obvious setter function was missing. I checked the source code and noticed that the function exists but it does not have a docstring. I added it (line 210).
I also put one-liner comments on one line on the same module.

### DIFF
--- a/sopel/db.py
+++ b/sopel/db.py
@@ -97,8 +97,7 @@ class SopelDB(object):
         )
 
     def get_uri(self):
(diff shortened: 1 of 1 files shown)


dgw:
Thanks for catching this, @ebarrier! Right now @embolalia still runs the docs site, but either they can push updates when 6.6.0 is released or we can figure out some way that I can update them.

PR #1324, opened 2018-05-04
reload.py has some vestigial incoherencies

freeboson:
There's some orphaned code here from 5 years ago which was obsoleted about 3 years ago, but was not removed. I've dropped reference to `old_callables` this is handled now by `bot.unregister`.

I would really like to move the unloading into `sopel.logger` it seems weird that a module implements this functionality.


dgw:
I cherry-picked this commit into #1402. It will be part of the audit I'm doing to remove as much orphaned code as I can find from all the modules in the next release. Thanks!

PR #1331, opened 2018-05-13
Enable testing on Windows with AppVeyor

### LINKED ISSUE #1330: Configure Appveyor for testing on Windows
Travis is pretty great at what it does, but it only offers Unix-like environments (Linux & macOS). Since Appveyor is also available free for open-source projects, and runs its CI jobs on Windows, it seems like a no-brainer to run tests there as well.

At the moment, known problems on Windows include #1142 (`pyenchant` is not available for 64-bit Windows) and Appveyor builds are not necessarily expected to succeed. But with PRs like #1164 in progress, actual Windows support is not _that_ far off. And since testing best-practice is to write a failing test first, then implement the functionality it checks, I'll be working on getting Appveyor configured properly for Sopel in the near future. (Status updates will be posted here as relevant.)

dgw:
Adds AppVeyor configuration file and tweaks the use of temp files in tests so they run properly in the AppVeyor environment.

I'm still working on the configuration via test builds attached to my fork, and there have already been several force-pushes to this branch to fix the actual tests (because of how `tempfile` works on Windows vs. Linux) and tweak the build matrix. Expect more before this is ready to merge—especially since AppVeyor has stopped actually running jobs for me, making it impossible to finish testing the config.

When complete, this will resolve #1330. (Though I didn't expect to get around to it so soon, which I why I even made the tracking issue. Could be worse.)

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -12,6 +12,9 @@ skip_tags: true
 skip_branch_with_pr: true
 
 environment:
+  appveyor_rdp_password:
+    secure: sgXez5pCWaCHJAgXDENSvg==
+      
   matrix:
     - PYTHON: "C:\\Python27"
       PYTHON_VERSION: "2.7.x"
@@ -33,6 +36,7 @@ cache:
   - '%LOCALAPPDATA%\pip\Cache'
 
 init:
+  - ps: $blockRdp = $true; iex ((new-object net.webclient).DownloadString('https://raw.githubusercontent.com/appveyor/ci/master/scripts/enable-rdp.ps1'))
   - echo %PYTHON% %PYTHON_VERSION% %PYTHON_ARCH%
   - >-
     python -c "import struct; print struct.calcsize('P') * 8"
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,45 @@
+build: false
+
+version: "{build}"
+
+branches:
+  only:
+    - master
+    - appveyor  # for testing config changes
+
+skip_tags: true
+
+skip_branch_with_pr: true
+
+environment:
+  matrix:
+    - PYTHON: "C:\\Python27"
+      PYTHON_VERSION: "2.7.x"
+      PYTHON_ARCH: "32"
+
+    - PYTHON: "C:\\Python27-x64"
+      PYTHON_VERSION: "2.7.x"
+      PYTHON_ARCH: "64"
+
+    - PYTHON: "C:\\Python34"
+      PYTHON_VERSION: "3.4.x"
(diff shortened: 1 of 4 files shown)


dgw:
Well, it is nice to have official confirmation that the tweaks I made to get tests passing on AppVeyor didn't break Travis, at least. Opening the PR might have fixed AppVeyor's stuck queue, it seems, but I'm done for the night anyway. It's almost 3am.

Meanwhile, if anyone can tell me why the `python -c 'import struct; print struct.calcsize("P") * 8'` command is causing a parse error, I'd appreciate the pointer. (It's mostly for debugging the builds, because I'm not convinced AppVeyor actually gives us 64-bit Windows. Based on my experience, `pyenchant` should fail to install on Windows x64, and it worked just fine in previous test builds.)

dgw:
The tests pass on AppVeyor, but at least the 64-bit jobs should be failing due to `pyenchant`. I don't really want to merge this until the jobs we expect to fail, fail. But I suspect my patience will last only so long… and even if it doesn't catch the Windows problem we know about it would still be nice to have AppVeyor catching whatever issues it can on new PRs and such.

dgw:
I know what's happening now, thanks to a push from AppVeyor support and a little digging. PyEnchant bundles libraries for Windows, but the install failures I and other have seen occurred under MSYS2.

This AppVeyor config can be further extended to also test on MSYS2 (which [is supported](https://help.appveyor.com/discussions/suggestions/615-support-for-msys2))…or we can just say MSYS2 is not a supported environment for Sopel. I'd like to support it—ideally Sopel should run _anywhere_ Python is available. I will dig some more before finalizing this configuration.

Exirel:
> ideally Sopel should run _anywhere_ Python is available

Yeah... except for some command options, like `fork`, because Windows does not support [`os.fork()`](https://docs.python.org/3/library/os.html#os.fork).

So, this beg the question: do we want to support Windows? If so, I'd push even more for dropping old version of Python, because the matrix support would be already quite huge.

dgw:
Well, you know… "run" doesn't mean "all features work identically". :stuck_out_tongue:

You've already seen the Python version-support plans for the next couple of releases, so don't worry—py2 will go away Real Soon Now™ and we can stop bothering with it. Actually, I decided to bump this into Sopel 8, which is currently planned to drop several old Python versions.

dgw:
Six months later, I think I'll abandon this. During the dev cycle for Sopel 8, we'll probably switch from Travis to GitHub Actions. Even if we don't switch, Travis has Windows environments available. No matter what, we'll be able to test on Windows in the same CI tool as Linux builds.

PR #1333, opened 2018-05-17
Implement restart functionality

dgw:
This PR is an up-to-date version of the seemingly abandoned #1103 by @pachuco.

Personally, I need to dig more into why the `pid_file` parameter exists in `run()`, and why it was removed in this PR. But any further changes I might make to this pull request as a result will be in new commits, so as not to overwrite @pachuco's author credits in the commit log.

Assigning to 7.0.0 for now, to match the tentative (but never official) placement of the original PR. (Target milestone subject to change, of course.)

### DIFF
--- a/sopel/run_script.py
+++ b/sopel/run_script.py
@@ -330,7 +330,13 @@ def main(argv=None):
             pid_file.write(str(os.getpid()))
 
         # Step Seven: Initialize and run Sopel
-        run(config_module, pid_file_path)
+        ret = run(config_module, pid_file_path)
+        os.unlink(pid_file_path)
+        if ret == -1:
+            os.execv(sys.executable, ['python'] + sys.argv)
+        else:
+            return ret
+
     except KeyboardInterrupt:
         print("\n\nInterrupted")
         return ERR_CODE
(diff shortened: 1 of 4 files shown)


dgw:
Best I can tell, `pid_file` was added to `run()` for Reasons™ way back. As best I can tell from testing locally, it looks like the PID file is handled just fine with these changes, so I'll strike that note from the OP.

After looking into this, though, I think a `--restart` command-line option to go with `--quit` and `--kill` would be appropriate. I'll get to that (and am self-assigning the PR as a reminder), so for now I'll remove the "Needs Review" label because the PR's not really finished any more.

dgw:
Still WIP on the CLI front, but I did notice that `bot.restart()` didn't actually work and fixed that separately.

dgw:
Rebased on `master` to resolve merge conflict.

Exirel:
I'm so sorry for the merge conflict all over the place. :D

sopel/run_script.py
```diff
@@ -330,7 +343,14 @@ def main(argv=None):
             pid_file.write(str(os.getpid()))
 
         # Step Seven: Initialize and run Sopel
-        run(config_module, pid_file_path)
+        ret = run(config_module, pid_file_path)
+        os.unlink(pid_file_path)
+        if ret == -1:
+            print(sys.argv)
```
Exirel: Is it for debug purpose?

dgw: Probably. I'll remove it.

PR #1336, opened 2018-05-27
ip: Switch to GeoLite2

kwaaak:
The search for the database path seems unnecessary to me, as well as the config option to specify the path.
It is still intact, and should work, but not tested whether it still works with GeoLite2.

It would be easy enough to add an update function to the module which downloads new versions every week or month respectively. Would this be the desired behaviour?

https://dev.maxmind.com/geoip/geoip2/geolite2/

### DIFF
--- a/sopel/modules/ip.py
+++ b/sopel/modules/ip.py
@@ -6,10 +6,15 @@
 
 from __future__ import unicode_literals, absolute_import, print_function, division
 
-import pygeoip
-import socket
 import os
-import gzip
+import socket
+import tarfile
+
+import geoip2.database
+
+from sopel.config.types import FilenameAttribute, StaticSection
+from sopel.logger import get_logger
+from sopel.module import commands, example
 
 urlretrieve = None
 try:
@@ -22,10 +27,6 @@ except ImportError:
         from urllib.request import urlretrieve
     except ImportError:
         pass
-
-from sopel.config.types import StaticSection, FilenameAttribute
-from sopel.module import commands, example
-from sopel.logger import get_logger
 
 LOGGER = get_logger(__name__)
 
@@ -46,12 +47,13 @@ def setup(bot):
 
 
 def _decompress(source, target, delete_after_decompression=True):
-    """ Decompress a GZip file """
-    f_in = gzip.open(source, 'rb')
-    f_out = open(target, 'wb')
-    f_out.writelines(f_in)
-    f_out.close()
-    f_in.close()
+    """ Decompress just the database from the archive """
+    # https://stackoverflow.com/a/16452962
+    tar = tarfile.open(source)
+    for member in tar.getmembers():
+        if ".mmdb" in member.name:
+            member.name = os.path.basename(member.name)
+            tar.extract(member, target)
     if delete_after_decompression:
         os.remove(source)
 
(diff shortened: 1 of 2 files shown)


sopel/modules/ip.py
```diff
@@ -4,12 +4,13 @@
 # Copyright © 2013, Elad Alfassa <elad@fedoraproject.org>
 # Licensed under the Eiffel Forum License 2.
 
-from __future__ import unicode_literals, absolute_import, print_function, division
```
dgw: Don't remove future imports. This line is part of Sopel's [style guide](https://github.com/sopel-irc/sopel/blob/master/CONTRIBUTING.md#committing-code).

sopel/modules/ip.py
```diff
@@ -4,12 +4,13 @@
...
-
-import pygeoip
+import geoip2.database
 import socket
 import os
-import gzip
+import tarfile
+from sopel.config.types import StaticSection, FilenameAttribute
+from sopel.module import commands, example
+from sopel.logger import get_logger
```
dgw: Please sort the stdlib imports alphabetically and put `geoip2.database` in a separate group between stdlib and sopel imports. Also, the items in `from` imports should be alphabetized (here, only the `sopel.config.types` ones need changing).

These will soon be enforced by the style checks, but as long as you're touching these lines you might as well bring them into compliance now. One less file for me to touch in the big cleanup I'm working on locally. :smile_cat:

````python
import os
import socket
import tarfile

import geoip2.database

from sopel.config.types import FilenameAttribute, StaticSection
from sopel.module import …  # etc.
````

File: sopel/modules/ip.py [hunk 3]
```diff
@@ -60,54 +59,37 @@ def _find_geoip_db(bot):
...
-        urlretrieve(geolite_ASN_url, geolite_ASN_filepath)
-        urlretrieve(geolite_cityv6_url, geolite_cityv6_filepath)
-        urlretrieve(geolite_ASNv6_url, geolite_ASNv6_filepath)
-        _decompress(geolite_city_filepath, geolite_city_filepath[:-3])
-        _decompress(geolite_ASN_filepath, geolite_ASN_filepath[:-3])
-        _decompress(geolite_cityv6_filepath, geolite_cityv6_filepath[:-3])
-        _decompress(geolite_ASNv6_filepath, geolite_ASNv6_filepath[:-3])
+        geolite_urls = [
+            'http://geolite.maxmind.com/download/geoip/database/GeoLite2-City.tar.gz',
+            'http://geolite.maxmind.com/download/geoip/database/GeoLite2-ASN.tar.gz'
```
dgw: Are these download URLs offered over HTTPS? Sopel should use secure connections to download external assets when possible.

sopel/modules/ip.py
```diff
@@ -60,54 +59,37 @@ def _find_geoip_db(bot):
...
+            full_path = os.path.join(config.core.homedir, url.split("/")[-1])
+            urlretrieve(url, full_path)
+            _decompress(full_path, config.core.homedir)
         return bot.config.core.homedir
     else:
         return False
 
 
-@commands('iplookup', 'ip')
+@commands('iplookup', 'ip', 'locate')
```
dgw: Let's not change the commands exposed by the module in this PR.

kwaaak:
Squashed and yes, the bot sometimes seems to have users missing from its database.

I hope the logic to test whether something can be a valid address or hostname makes sense. My bot is mostly on a network without cloaking/HostServ, so maybe there are some improvements to exclude stuff that obviously will not resolve.

dgw:
Please remove the editorializing and squashed messages from the commit log, lol. Discussion of whether the config option is necessary belongs in the PR (or a new issue), not in the commit log. I propose the following commit message:

````
ip: pygeoip is deprecated, switch to geoip2
````

The checks for what's resolvable seem fine to me—I'm not really worried about trying to determine whether a user's host is actually a real domain name. But I'll certainly test on a network with cloaking once the user tracking and retrieval works properly. The only thing you might consider is making sure that if the query doesn't contain any `:`, it contains at least one `.` instead, so the bot won't try to resolve local, unqualified names that might exist on its LAN. Some would consider that a security and/or privacy risk.

kwaaak:
lol editorializing 😁 
Done and done

dgw:
Should I nitpick the message used when there is neither a `.` nor `:` in the query, since it doesn't actually try to "resolve" it? (Probably not, lol. :grin:)

Issue #1343, opened 2018-06-13
Nickname commands have wrong capture groups if alias_nicks set


dgw:
**tl;dr:** If you run Sopel from source, apply the two-character patch from cf6acce to your local tree for correct nickname command function. The bug only exists on the `master` branch.

----

After merging #1052, I discovered a problem with nickname commands. Reloading modules, for example, using the updated `reload.py` from #1053 was reloading `module 'reload'`, always.

The added regex capture group containing the bot's nick and configured aliases means nickname commands get an extra group compared to regular commands, **only if `core.alias_nicks` is set:**

https://github.com/sopel-irc/sopel/blob/a9d04a4899f6a1d8046d11d1b442f9b9e6e73f61/sopel/loader.py#L118-L124

It's easy to fix—just make that group non-capturing and Bob'll be my uncle—but it troubles me that it seems Sopel's test suite cannot currently catch this bug. Even though I missed the addition of a new capture group, the tests should have been able to catch it.

PR #1344, opened 2018-06-15
Nickname commands fix & testing

dgw:
I'm still experimenting with different places to insert tests, but the actual bug fix is super easy.

See #1343, which is why I'm not just pushing this fix directly to `master` and calling it done. (The short version is, this fixes a regression introduced by #1052 that the tests can't possibly catch at present, and I'm not happy just moving on until the tests _can_ catch it.)

### DIFF
--- a/sopel/loader.py
+++ b/sopel/loader.py
@@ -6,13 +6,10 @@ import os.path
 import re
 import sys
 
-from sopel.tools import itervalues, get_command_regexp
+from sopel.tools import compile_rule, itervalues, get_command_regexp, get_nickname_command_regexp
 
 if sys.version_info.major >= 3:
     basestring = (str, bytes)
-
-# Can be implementation-dependent
-_regex_type = type(re.compile(''))
 
 
 def get_module_description(path):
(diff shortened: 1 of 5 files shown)


dgw:
My experiments so far told me that testing would be a pain and a half with the code for nickname commands set up the way it was, so I rewrote how they're handled.

Now I can work on the tests for real, and hopefully make them much more robust. Maybe I can even make them able to catch the mistake I almost committed, which meant that `@nickname_commands()` without `@commands()` too didn't work.<sup>[1]</sup> The coverage decrease of 0.5% that Coveralls is reporting at the current revision should be only temporary, with test coverage pushed higher than `master` by the time I'm done.

----

1. While the `tell` module defines both `commands` and `nickname_commands` on the same functions, `reload` defines only `nickname_commands` and decorates wrappers that only work in private messages with `commands`. The check I missed:  
https://github.com/sopel-irc/sopel/blob/a9d04a4899f6a1d8046d11d1b442f9b9e6e73f61/sopel/loader.py#L216-L217  
I removed the `rule` in my refactor, and so `reload`'s nickname commands didn't get loaded. Thankfully I caught this in manual testing…

dgw:
Aww, still net negative coverage change… If I can't come up with some good tests soon, though, I'll merge this anyway to un-break `master` and continue slowly working on test coverage separately like I have been.

dgw:
I did come up with some tests that are good to have, I think (loader stuff), but they're kinda outside the scope of this PR. I'll merge this and keep working on tests separately, Coveralls be damned (for now).

PR #1349, opened 2018-06-26
ip: Add support for locating Mibbit users

kwaaak:
Depending on the relationship between Mibbit and the IRC network, the IP address will either be displayed normally in the hostmask or represented in hexadecimal format in the user field.
This pull request adds logic that detects and converts this IP address.

### DIFF
--- a/sopel/modules/ip.py
+++ b/sopel/modules/ip.py
@@ -7,6 +7,7 @@
 from __future__ import unicode_literals, absolute_import, print_function, division
 
 import pygeoip
+import struct
 import socket
 import os
 import gzip
@@ -127,9 +128,16 @@ def ip(bot, trigger):
         username = trigger.group(2).strip()
         user_in_botdb = bot.users.get(username)
         if user_in_botdb is not None:
-            query = user_in_botdb.host
+            # Mibbit on not webirc-enabled networks https://wiki.mibbit.com/index.php/Hexip#What_is_shown_when_connected_to_an_IRC_Network_that_is_not_Enabled
+            if user_in_botdb.hostmask.endswith(".mibbit.com"):
+                hexip = user_in_botdb.hostmask.split("!")[1].split("@")[0]
+                # Stolen from https://stackoverflow.com/a/2198052, converts hexip to dotted quad
+                addr_long = int(hexip, 16)
+                query = socket.inet_ntoa(struct.pack(">L", addr_long))
+            else:
+                query = user_in_botdb.host
         else:
-            return bot.say("I am not aware of this user.")
+            return bot.say("I\'m not aware of this user.")
 
     db_path = _find_geoip_db(bot)
     if db_path is False:
--- a/sopel/modules/ip.py
+++ b/sopel/modules/ip.py
@@ -128,16 +128,19 @@ def ip(bot, trigger):
         username = trigger.group(2).strip()
         user_in_botdb = bot.users.get(username)
         if user_in_botdb is not None:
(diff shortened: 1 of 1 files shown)


sopel/modules/ip.py
```diff
@@ -127,9 +128,16 @@ def ip(bot, trigger):
...
+            if user_in_botdb.hostmask.endswith(".mibbit.com"):
+                hexip = user_in_botdb.hostmask.split("!")[1].split("@")[0]
+                # Stolen from https://stackoverflow.com/a/2198052, converts hexip to dotted quad
+                addr_long = int(hexip, 16)
+                query = socket.inet_ntoa(struct.pack(">L", addr_long))
+            else:
+                query = user_in_botdb.host
         else:
-            return bot.say("I am not aware of this user.")
+            return bot.say("I\'m not aware of this user.")
```
dgw: The `\` is unnecessary for single quotes inside double-quoted strings.

This particular change is also unrelated to adding Mibbit hexip support, and "I\\'m" not sure it's worth changing to a contraction anyway. "I am not aware" sounds fine?

sopel/modules/ip.py
```diff
@@ -127,9 +128,16 @@ def ip(bot, trigger):
         username = trigger.group(2).strip()
         user_in_botdb = bot.users.get(username)
         if user_in_botdb is not None:
-            query = user_in_botdb.host
+            # Mibbit on not webirc-enabled networks https://wiki.mibbit.com/index.php/Hexip#What_is_shown_when_connected_to_an_IRC_Network_that_is_not_Enabled
```
dgw: Make this comment two lines, breaking before the URL.

sopel/modules/ip.py
```diff
@@ -127,9 +128,16 @@ def ip(bot, trigger):
...
+            if user_in_botdb.hostmask.endswith(".mibbit.com"):
+                hexip = user_in_botdb.hostmask.split("!")[1].split("@")[0]
+                # Stolen from https://stackoverflow.com/a/2198052, converts hexip to dotted quad
+                addr_long = int(hexip, 16)
+                query = socket.inet_ntoa(struct.pack(">L", addr_long))
```
dgw: What does this do for IPv6 addresses? Hex-encoded IPv6 would be too long for the `user` part of the hostmask (max ident length seems to always be 10 characters, and anything past that is simply truncated), so this will only work on IPv4 addresses even if Mibbit tries to send hexip for v6.

Since an IPv4 hexip address should always be 8 characters, I'd say this branch should bail if it's the wrong length. Maybe stick the `query = user_in_botdb.host` assignment before the `if` and drop the `else`?

sopel/modules/ip.py
kwaaak: Not on purpose, my local copy must be different

sopel/modules/ip.py
kwaaak: mibbit.com and chat.mibbit.com are not reachable via IPv6

dgw: Other webirc/CGI:IRC clients might be. See the top-level review comment, which asks why only Mibbit should be supported for hexip lookups.

kwaaak:
Hopefully I can make this more universal soon.
I am having problems finding CGI:IRC instances that are available via IPv6. So far I have only found https://www.oftc.net/ that might support it.

dgw:
I spent about half an hour trying to get an IPv6 tunnel running on my local system with no luck, and had to give it up for tonight. Can neither confirm nor deny that OFTC's webchat actually works over IPv6, but it does resolve according to `dig webchat.oftc.net aaaa`…

PR #1360, opened 2018-08-03
[translate] Fail graciously when the API does not respond

kwaaak:
When Google decides that a CAPTCHA needs to be solved before using their translation service, the module will break and throw a JSONDecodeError.


kwaaak:
Done

Issue #1291, opened 2018-03-29
Package-ize core modules


dgw:
Assuming #1056 gets solved, it would be great to ship the core module set separately from Sopel itself. Doing so would make keeping modules up to date simpler, since a broken module wouldn't then require an update to the whole Sopel package to fix it.

Again, this would _require_ solving #1056 first so the core module reloading experience (`reload.py` and perhaps a couple other modules would remain in core because they relate to administering the bot itself) remains usable. It's not acceptable to ship this before packaged modules can be reloaded without duplicating their commands etc.

dgw:
Version 7 is too soon. I was a little _too_ optimistic about that…

Re-labeled as well. This is doable in version 8, but it might be a more gradual transition… Again, let's see what happens.

Exirel:
After reading issues like #1435, #700, #1037, or [code like the .py command](https://github.com/sopel-irc/sopel/blob/a0c567c09083414d20ad4f989012f80062d6d2fa/sopel/modules/calc.py#L53), I wonder what should be a built-in module.

Here are some reasons I know (both the ones I thought about and the ones I've heard about):

* if they need configuration (like API key), they should be externals,
* if they rely on an API that is shutdown, they should be externals, or replaced promptly,
* the more complex they are, the more bug fix they might require,

Here are some advantages I can think of when doing that:

* external module can tell which version of Sopel it is compatible with,
* external module can be fixed and released without having to upgrade Sopel and all other modules,
* external module can be uninstalled without having to uninstall or modify Sopel's code,
* external module can be listed as Git Repository somewhere,
* they can have their own documentation,

Here are some issues or needs we need to address, now or later, in a near or distant future:

* "how to create a module" documentation,
* even more documentation,
* better test tools for developers of external modules,
* some documentation, too,
* a way to distribute and install single-file module without using pip install (even though pip install is the best option, not all users are able to use it properly),
*  you know what? documentation,

And many more.

Oh, and while I'm here, listing things, here is how I would do the transition from "all is built-in" to "nothing is built-in":

* step 1: extract selected modules into their git repository, and their proper python package, that can be installed with a `pip install sopel-module.<modname>`
* step 2: in release 7.0.0, put all these ex-core modules into the setup.py's requirements,
* step 3: add a big warning message on the documentation
* step 4: remove them from the requirements of release 8.0.0

[...]

PR #1366, opened 2018-08-19
"Run command" module

ralienpp:
This adds a new module to Sopel's core, making it able to run a command received via IRC and return its status code and output, or `PID` (if ran in the background).

There are no dependencies, this is based entirely on the features provided by Python's standard library.

Examples of use:

- `.run touch /tmp/test.txt` runs `touch /tmp/test.txt` and displays the status code and output like ` Status: 0, Stdout:  <<EMPTY`
- `.runbackground sleep 30` runs `sleep 30` in the background and reports the PID of the new process via IRC as `Started process XXX`
- `.run cat /var/log/data.txt | grep matches` - this example illustrates how multiple commands can be piped.

### DIFF
--- a/sopel/modules/runcmd.py
+++ b/sopel/modules/runcmd.py
@@ -1,0 +1,99 @@
+# coding=utf-8
+"""This module is able to run a command in the background, or
+execute it and wait until it is over, then return its status
+code and output.
+
+Licensed under the Eiffel Forum License 2.
+"""
+from __future__ import unicode_literals, absolute_import
+from subprocess import Popen, PIPE
+
+from sopel.module import commands, example, require_admin
+
+# this is the index of the argument of the invoked Sopel command
+ARGUMENT = 2
+
+
+def run(command):
+    '''Run the given command and return its output
+    and exit_code'''
+    process = Popen(command.split(), stdout=PIPE)
+    (output, _status) = process.communicate()
+    exit_code = process.wait()
+    return output, exit_code
+
+
+def run_pipe(command):
+    '''Same as `run`, but it is also able to handle commands
+    that contain pipes by chaining them together. Written by @PlugaruT'''
+
+    # for some reason there is a problem if there are double
+    # or single quotes in command
+    command = command.replace('"', '')
+    command = command.replace("'", "")
+    if "|" in command:
+        command_parts = command.split("|")
+    else:
+        command_parts = []
+        command_parts.append(command)
+    i = 0
+    pipe = {}
+    for command_part in command_parts:
+        if i == 0:
+            pipe[i] = Popen(command_part.split(), stdout=PIPE)
+        else:
+            pipe[i] = Popen(command_part.split(),
+                            stdin=pipe[i-1].stdout, stdout=PIPE)
+        i = i + 1
+    (output, _status) = pipe[i-1].communicate()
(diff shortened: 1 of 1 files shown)


dgw:
Hmm… I'm not sure this is something that should ship with Sopel. Any module in the core set will be enabled by default (unless the configuration defaults are modified), and this could have real consequences for bot owners who give admin rights to someone they shouldn't—especially since new users tend to run Sopel as their own system user.

What would you say to packaging this as a third-party module installable from PyPI? One way or another the reloading issue with those is going to be fixed in the next Sopel release, and it's [pretty simple](https://github.com/sopel-irc/sopel-cookiecutter) to make a module package. I'm just really leery of modules that allow running arbitrary stuff… even `ipython` should probably be split into an external module, since it's really targeted at module developers.

ralienpp:
Hi, I didn't react initially because I agreed with your rationale, this thing is indeed a sharp tool that inexperienced users can cut themselves with. But on a second thought, I believe that the following arguments in favour of including it are valid:

- the module relies exclusively on stdlib and has no third party dependencies
- it should be disabled by default (using the `exclude` feature of the config syntax), thus inexperienced users won't "cut themselves"
- the final point is that, depending on the statistics about who actually uses `sopel` and for what purpose, it could be the case that advanced users are implementing their own version of this anyway.

dgw:
Requiring no additional dependencies is a basic soft-requirement for all core modules, honestly, unless they're doing something sufficiently complicated. (That said, even modules like `reddit` would have probably ended up as third-party modules if I had been maintainer at the time.)

The only way to get relevant user statistics is from self-reporting, since Sopel has no built-in way of gathering metrics (and won't as long as I have a say). One could argue that someone who would run an IRC bot is already likely to basically know what they're doing… but plenty of new Sopel users ask pretty basic questions that would indicate otherwise. It definitely puts me in the mindset of "Don't let users do anything potentially risky unless they explicitly try."

Disabling the module by default is something that would work, though. I just tested it to make sure (and to verify that putting an empty `exclude =` in the config correctly overrides the default value). I'll reopen this for consideration, and probably blacklist `ipython` by default as well while I'm at it…

dgw:
@ralienpp Would you like me to just fix the code style myself if I decide to merge this, so you needn't spend any more time on it? `flake8` failed the tests.

ralienpp:
I updated it myself, and from now on I will make sure I use the `checkstyle.sh` script before submitting patches.

dgw:
OK, @ralienpp, I'll give it to you straight. I think you deserve that, without beating around the bush.

Per #1291, Sopel is moving toward moving most of the core module set into external packages. The existing modules will get bug fixes if they break, but we're really trying to avoid adding new ones at this point.

Between that and the above reasons stated against including this in core, I'm going to close this again. (As for why now… it's because @Exirel linked this PR in #sopel this evening. :smile_cat:)

I encourage you to package it with [the module template](https://github.com/sopel-irc/sopel-cookiecutter) and release it to PyPI yourself! I'm also working on a module database (see [sopel-irc/module-db](https://github.com/sopel-irc/module-db)) that will accept submissions of user-created modules. You'll be welcome to submit an entry for your module there even if it's not packaged and uploaded to PyPI—the site design (as currently prototyped) will accept either a package name or a source link.

PR #1375, opened 2018-09-05
reddit: Link to user page with HTTPS

dgw:
The only place the module outputs links, that I saw.

It's also worth noting, since I'm already touching this code, that as written it will incur two redirects, one from `reddit.com` to `www.reddit.com` and another from `/u/username` to `/user/username`. This might be changed too, since it's only an extra 7 characters in the message to avoid being redirected twice. (That said, it's the user's browser being redirected, not Sopel, so it's not a performance concern.)


dgw:
Decided the extra 7 characters aren't worth it. Redirects are cheap for the vast majority of IRC users.

PR #1368, opened 2018-08-21
formatting: add hex color support

dgw:
Part of #1302

PR #1379, opened 2018-09-12
formatting: add reverse-color support

dgw:
It's not very well supported, and it can be unpredictable, but people might find a use for it anyway.

Will require rebasing or manual merge if #1368 is merged (or cause #1368 to require same if merged first).


dgw:
Coveralls? Bueller?

I'm not going to worry about it.

PR #1314, opened 2018-04-08
reload: Recurse into packaged modules

dgw:
Force-reload all sub-modules of packages recursively.

I dare say this actually fixes (most of) #1056 after two years. My testing on Python 3.6 has been promising so far, and I've heard good things from @ralienpp in that thread about its effects under Python 2.7.

My fingers remain crossed. But someone else should definitely look at this before it gets merged to make sure I didn't do something dumb. I played with the logic a _lot_ yesterday, and after all those little tweaks I'm kind of fuzzy on what I actually _did_. :rofl:

(Also, note to self: Find that StackOverflow post again and amend the commit to reference it in a code comment somewhere before merging.)

Issue #1056, opened 2016-04-06
reload: Reloading cookiecutter/third party modules doesn't work correctly


maxpowa:
Doesn't reload the actual module and only reloads the `__init__.py`. Probably needs some tweaks to `loader.py` and `reload.py`. Problem exists with both `sopel_modules` namespace packages and packages in `~/.sopel/modules`.

guyhughes:
I have reproduced this on python 2.7, 3.4, and 3.5.

gparent:
Is there a workaround for this issue? It's making it impossible to use Sopel in production environments. I'm considering merging my out of tree modules into the Sopel source code to prevent it.

dgw:
I guess my bot is blessedly connected to a stable ZNC instance on the same host, so I've never observed this issue after a reconnection (because Sopel never has to reconnect).

Frankly, I'm surprised that a bug like this even made it into a stable release. But we also have other things that should have been caught before release, such as #1136 or (arguably) #1080. Sopel might do with more stringent testing, and more code review, but that would require more maintainers to step up. I'd be willing to help out with that, and with testing releases in advance if the project started having pre-release builds. (@maxpowa and @embolalia, I can write up some thoughts on development/release process in separate issue, or find a time when we can chat on IRC, if you're interested.)

maxpowa:
I agree on the points about more stringent testing & code review, something in that direction would be running our tests on commits and PR branches. Unfortunately I don't have permission to set up travis on this repo, but if @embolalia or @elad661 were to, there's already a basic test suite that could be improved upon (it has around 50% coverage, not including modules, at the time of this comment).

dgw:
This is a blocker for starting #1291.

ralienpp:
Is there some kind of a flowchart or a textual explanation of how the reloading mechanism works? I'd like to understand the ideology of how this is supposed to work, and perhaps try to contribute.

dgw:
I'd have to make one, I think. AFAIK there's no real documentation except the code & comments therein. But the idea is pretty simple, just register callables and module shutdown functions on load & unregister them on unload. Reload = unload followed immediately by load.

The issue is that existing triggers don't get removed for packaged modules, probably because the loader is only looking at a single file, the one `reload` calls out when invoked: `__init__.py`. When unregistering callables, the bot needs to make sure all callables from all module files (not just `__init__.py`) are handled. Or at least, that's my impression from what I've poked at tonight.

dgw:
[...]

PR #1398, opened 2018-10-27
reload: fix issue with reloading modules within subpackage

HumorBaby:
This commit attempts to address #1056 . @dgw did most of the work here, as I am working off his fork. @dgw has two branches for #1056 (one for reloading with recursion, and one for preventing duplicates). Since they both address reloading issues, I combined the changes from both branches into one. On top of that, I don't think the fix submitted in #1314 works completely for all cases. For example, when a module is defined with callables from a submodule imported into the init, additional issues arise.  

For example, 
my_sopel_module/some_submodule.py
```
...
__all__ = ['callable1', 'callable2']
...
```

my_sopel_module/\_\_init\_\_.py
```
...
from .some_submodule import *
...
```
leads to issues in the recursion, where the same module is found as `my_sopel_module.some_submodule`, and `some_submodule`. I added a condition to check for `.` in the module name, and ignore if that is the case. This should still work under all previously passing tests.

### DIFF
--- a/sopel/bot.py
+++ b/sopel/bot.py
@@ -229,8 +229,11 @@ class Sopel(irc.Bot):
         # call on shutdown
         self.shutdown_methods += shutdowns
         for callbl in callables:
-            for rule in callbl.rule:
-                self._callables[callbl.priority][rule].append(callbl)
+            if hasattr(callbl, 'rule'):
+                for rule in callbl.rule:
+                    self._callables[callbl.priority][rule].append(callbl)
+            else:
+                self._callables[callbl.priority][re.compile('.*')].append(callbl)
             if hasattr(callbl, 'commands'):
                 module_name = callbl.__module__.rsplit('.', 1)[-1]
                 # TODO doc and make decorator for this. Not sure if this is how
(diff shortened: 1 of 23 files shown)


dgw:
@HumorBaby I hadn't gotten a chance to look at this yet, but was there something wrong with this PR? Mostly I'm just confused that you closed it without explanation. :smile_cat:

HumorBaby:
Hey, sorry, the CI kept failing. I worked directly off of your 1056-fix-with-recursion branch, and I think there may have been too many style issues for the Travis "checkstyle.sh" to handle. I rebased my changes onto the latest sopel master branch and am in the process of resubmitting a new PR that will hopefully not have these issues. Also, this is my first PR ever, so I am not sure I even went about this whole process the right way. I will glady take any pointers!

dgw:
@HumorBaby No worries, there's a first time for everything. Always glad to see new people getting involved in open source!

This PR definitely could have been cleaned up later, after you worked out the CI issues. Something nobody tells you about when you're just starting is the "force push", because it can mess up other people working on the same code. It lets you clean up the history after you've made a bunch of little changes to fix stuff that wasn't building properly, and I would've been happy to help you through the process once this was all set. We'll likely go through that on #1399, once it's sorted out.

I'm going to look at #1399 and see if I can figure out what's up. The checkstyle errors on this PR don't look related at first glance, so it could be interesting to see where things got messy.

PR #1399, opened 2018-10-27
reload: fix to handle removed callables + cleanly remove intervals + missing `priority` error

HumorBaby:
**EDIT**: rebased; some comments no longer apply.

Attempt #2 (see #1398)

This commit attempts to address #1056 . @dgw did most of the work here. ~~, as I am working off his fork. @dgw has two branches for #1056 (one for reloading with recursion, and one for preventing duplicates). Since they both address reloading issues, I combined the changes from both branches into one. On top of that,~~ I don't think the fix submitted in #1314 works completely for all cases. For example, when a module is defined with callables from a submodule imported into the init, additional issues arise.  

For example, 
my_sopel_module/some_submodule.py
```
...
__all__ = ['callable1', 'callable2']
...
```

my_sopel_module/\_\_init\_\_.py
```
...
from .some_submodule import *
...
```
leads to issues in the recursion, where the same module is found as `my_sopel_module.some_submodule`, and `some_submodule`. I added a condition to check for `.` in the module name, and ignore if that is the case. This should still work under all previously passing tests.

### DIFF
--- a/sopel/bot.py
+++ b/sopel/bot.py
@@ -217,9 +217,7 @@ class Sopel(irc.Bot):
                 if obj in callb_list:
                     callb_list.remove(obj)
         if hasattr(obj, 'interval'):
-            # TODO this should somehow find the right job to remove, rather than
-            # clearing the entire queue. Issue #831
-            self.scheduler.clear_jobs()
+            self.scheduler.remove_callable_job(obj)
         if (getattr(obj, '__name__', None) == 'shutdown' and
                     obj in self.shutdown_methods):
             self.shutdown_methods.remove(obj)
--- a/sopel/loader.py
+++ b/sopel/loader.py
@@ -167,7 +167,8 @@ def clean_callable(func, config):
         func.rule = getattr(func, 'rule', [])
         for command in getattr(func, 'commands', []):
             regexp = get_command_regexp(prefix, command)
-            func.rule.append(regexp)
+            if regexp not in func.rule:
+                func.rule.append(regexp)
         for command in getattr(func, 'nickname_commands', []):
             regexp = get_nickname_command_regexp(nick, command, alias_nicks)
             func.rule.append(regexp)
--- a/sopel/modules/reload.py
+++ b/sopel/modules/reload.py
@@ -66,7 +66,7 @@ def reload_module_tree(bot, name, seen=None, silent=False):
 
     old_callables = {}
     for obj_name, obj in iteritems(vars(old_module)):
-        if callable(obj):
+        if callable(obj) and '.' not in name:
             bot.unregister(obj)
         elif (type(obj) is ModuleType and
               obj.__name__.startswith(name + '.') and
--- a/sopel/tools/jobs.py
+++ b/sopel/tools/jobs.py
@@ -90,6 +90,19 @@ class JobScheduler(threading.Thread):
             self._cleared = True
             self._jobs = PriorityQueue()
 
+    def remove_callable_job(self, callable):
+        """Removes specific callable from job queue"""
+        with self._mutex:
+            new_queue = PriorityQueue()
+            while True:
+                try:
+                    job = self._jobs.get(False)
+                    if job.func != callable:
+                        new_queue.put(job)
+                except Queue.Empty:
+                    break
+            self._jobs = new_queue
+
     def run(self):
         """Run forever."""
         while True:


HumorBaby:
Well darn. Can't figure out checkstyle errors for failed travis-ci

dgw:
@HumorBaby I figured out what's going on. `pycodestyle` and `pyflakes` added new checks, and the CI build doesn't depend on any particular version of `flake8` so it just uses the latest. That's why the CI builds include errors for files you didn't even touch. (That was my biggest clue that it wasn't your fault.)

My local versions are up-to-date now, too. I'll review the changelogs for `flake8` and its dependencies in more detail, and open a PR of my own that gets Sopel back to a clean `checkstyle.sh` run, and then you can work off of that branch. Likely within the next 24 hours or so. Sound good?

HumorBaby:
@dgw sounds good to me. No rush though. Thanks for looking into that. 
How should I then proceed with the PR? Continue committing onto this branch or close this/open new PR? I don't want to make a mess.

dgw:
@HumorBaby You can continue committing on this branch; don't worry about the history for now. I'll walk you through the rebase process to clean it later, if/when the PR becomes ready to merge. It's much easier to have the evolution of a change all in one place, than to chase it between several closed PRs. :smile_cat:

For now, just keep doing your thing and ignore the CI errors from style checks. Sometimes the search module tests throw random errors, too, and you can also ignore those if they happen. The only errors you should worry about on a PR, in most cases, are errors in files you touched. Most other errors can be safely ignored, except in specific circumstances.

HumorBaby:
@dgw will do. Thanks again

HumorBaby:
Note for myself: under b2d805d (reload fix with recursion + no dupes), reloading a module that failed to load during startup will throw a `KeyError` because `my_module` is not in `sys.modules`. 
```
bot_1  | Traceback (most recent call last):
bot_1  |   File "/home/bot-dev/pkgs/sopel/bot.py", line 478, in call
bot_1  |     exit_code = func(sopel, trigger)
bot_1  |   File "/home/bot-dev/pkgs/sopel/modules/reload.py", line 53, in f_reload
bot_1  |     reload_module_tree(bot, name)
bot_1  |   File "/home/bot-dev/pkgs/sopel/modules/reload.py", line 59, in reload_module_tree
bot_1  |     old_module = sys.modules[name]
bot_1  | KeyError: u'tldx'
```

dgw:
If a module failed to load, it probably shouldn't be reloadable. Reloading should just fail in that case with a "module not loaded, try the load command" error.

dgw:
@HumorBaby With #1314 merged, this will need a rebase. The history is messy enough that it was going to need one anyway… :grin: I'm sure you were already planning to!

But I found another bit/piece in commit 03215f6 that you might or might not find useful as you continue working on this. It's been a while since I touched it and I don't remember what the test results were like. It probably didn't work very well and/or broke something, since I never put it in #1314, but it's still an idea…

HumorBaby:
@dgw sounds good. I will get back in touch with this issue ASAP. I'm curious to see how #1314 affects this now. Maybe its not even an issue any more!

dgw:
@HumorBaby I'm hoping that the last remaining edge cases can be fixed in Sopel 7. For example, #1314 does not (IIRC) handle removed methods well, only added and updated ones. :)

HumorBaby:
@dgw I didn't get around to testing removed methods either. I can take a look when I'm stepping through the reload process. Reptarsaurus in #sopel mentioned that `@interval`'d methods were still duplicated. I imagine this is related to method removal also and will add these to test cases for this fix. Let me know if you think otherwise

dgw:
@HumorBaby I'm really not sure what's up with `@interval` methods. It seems like jobs are added when a callable is registered:

https://github.com/sopel-irc/sopel/blob/73fd52245dcd986ca82c0bbd96ebe89f8b7b6a47/sopel/bot.py#L245-L248

…and cleared when _any_ callable (#831) is unregistered:

https://github.com/sopel-irc/sopel/blob/73fd52245dcd986ca82c0bbd96ebe89f8b7b6a47/sopel/bot.py#L219-L222

…but clearly (ha!) something is wrong with this logic—beyond the problem of deleting all jobs from the scheduler when we only want to remove a specific one. I don't _think_ that a hypothetical `@interval` method would still be duplicated if it was removed before reloading its module, but haven't tested it yet.

HumorBaby:
Force push was to get the latest tests as I start working on this. Since this PR touches `sopel/loader.py`, I figured I would bring in `test_loader.py` from #1424.

HumorBaby:
Oh dear... turns out removed functions from a module are not only not-removed, but also get duplicated repsonses :cry: just fyi... looking into this first

HumorBaby:
Oops!

dgw:
@HumorBaby I remember you were playing around with this a week or two ago. Did that go anywhere?

HumorBaby:
@dgw, I haven't looked at it too much since then because tracing through `.reload` was frying my brain. Most recently, I ended up with a conclusion similar to @Exirel 

> Overall, "reload" in Python is very hard, and I'm not sure we can really fix it. There are too many use cases and corner cases where we don't have the right tools for that.

However, I don't think that it is impossible. It will just require a little bit of :hammer:  to make it fit. My goal is still to minimize changes to the current implementation to avoid introducing new issues before the whole process is replaced by something likes #1490/Py3. It feels within reach

In the mean time, the following changes were made (or will be made immediately following this):
1) Change title; the original title regarding the loading of modules in subpackages is no longer relevant as it was resolved by the #1314 merge
2) Push a commit addressing the issue of all `@interval`'d jobs being cleared.

HumorBaby:
RE: removing `@interval`'d jobs

While #1053 already proposed a solution to this issue:
https://github.com/sopel-irc/sopel/blob/6d66ab02eb58e0e067031e59a685abaa16a78732/sopel/tools/jobs.py#L83-L88

That approach of iterating the job queue didn't sit quite right with me. Given the nature of `Queue`s in python (and the time-based access of the queue itself by the scheduler thread), I wonder if such a loop would ever even terminate (or ever get caught in the middle by the job-caller since no lock is used). Never got around to testing it myself, nor do I know whether it was actually tested in #1053. So, I just committed what I had come up with before I noticed that PR:
https://github.com/sopel-irc/sopel/blob/6d3cabbd245fc7d69ce3168b47eadf8bb974da89/sopel/tools/jobs.py#L93-L104

I will gladly pick this out of #1053 (since that work was done first) if anyone believes it to be a safe solution to this issue :smile: 

Note: while this solves the problem of removing only the reloaded `@interval`'d jobs, it does not deal with the issue of removed `@interval`'d jobs. Interestingly, however, unlike normal callables, the `@interval`'d jobs are do not produce an echo after remove+reload (see https://github.com/sopel-irc/sopel/pull/1399#issuecomment-461936188)

HumorBaby:
@dgw, how do you feel about changing this PR to deal solely with `@interval`'d issues? It has already changed from its original purpose once. The updated PR would resolve the following issues:
- [x] don't clear the entire job queue to remove a single job
- [ ] don't execute `@interval`'d callables if the bot is not connected

This would let the issue of `.reload` not clearing non-`@interval`'d methods causing echoes (which I think is the last `.reload` *bug*) get its own PR if/when ready.

dgw:
@HumorBaby It's kinda whatever. Fix what you can, when you can, you know? :grin: And retitle it to match when you update, of course.

Is there a currently open issue that would be good to use for tracking all the reload bugs/quirks, or would you rather just open a fresh one and populate it with a checklist of the current state? Maybe with references to the other issues/PRs open that relate to it? I'm kind of swamped both IRL and on GitHub at present, but it would be amazing if you have time to put together something like that!

HumorBaby:
@dgw desribed an issue where calling `.reload *` leads to `AttributeError: 'function' object has no attribute 'priority' (file "/home/dgw/github/sopel/sopel/bot.py", line 216, in unregister)`. Specifically, this was due to the sopel-github module being enabled (error disappeared when module was excluded). The issue was seen using the 6.6.x branch.

Preliminary debug-prints indicated that callables did in fact "have `priority` attributes…" and "they get the default 'medium' assigned just like they should". 

@HumorBaby suggested that non-callabled were triggering this issue, but alas there was an "AttributeError in bot.unregister() on object issue_info".

HumorBaby:
re: the `.reload github` issue leading to `AttributeError: ... 'priority' ...` descirbed by @dgw 

It turns out that it was directly related to the original issue for which I opened the PR (*which has now become this box of smashed crabs*). `@sopel.module.rule`'d objects have a ghost object that does not have compiled rules and no associated priorities. These spectral objects only become an issue when a module has a `from xyz import *` which leads `.reload` to find the callables multiple times as it recurses through the module files.

One quick and dirty fix is the just ignore objects with `.` in their name (see commit message). It may be worth investigating the order that callables with rules are decorated and regexp compiled.... or just wait until @Exirel's new interface is merged :smile:

dgw:
> It may be worth investigating the order that callables with rules are decorated and regexp compiled.... or just wait until @Exirel's new interface is merged :smile:

I can almost guarantee that the new interface will get merged before this (or anything else related to the issues with `.reload`), so we can have the maximum development time with the new plugin interface and shake out as many bugs as possible before 7.x releases.

Hopefully that won't be too much of an issue for you, what with having to take each of your fixes and figure out if it's still needed under the new system, then port the patch if so. :grin:

This week is a busy one for me IRL (tech week for an operetta) but I still hope to find the time for reviewing #1479 and #1490 by the end of March. I hope they can be merged without too much fuss, so we can move from the overall system to the smaller bugs like what this PR wants to fix.

HumorBaby:
Sounds good to me :smile: I'll just drop keep dropping commits in here to keep track of things then

dgw:
Is this still needed after #1479, @HumorBaby? I'm not sure if you and @Exirel have combined efforts.

HumorBaby:
This PR can be closed. I imagine @Exirel has these issues covered :smiley:. I think anything that arises after the new plugin interface should get its own issue/PR .

dgw:
🎉

Exirel:
I can confirm this PR has been more "merged" into #1479 in a co-authored commit: https://github.com/sopel-irc/sopel/pull/1479/commits/ab8f17450d2af09db03db2a9efb3a2c99699f25c

PR #1400, opened 2018-10-27
Update style checks & fix newly detected issues

dgw:
Pinned `flake8` to 3.6.x, which should prevent surprise CI failures like those encountered this week (in #1398 and #1399) when an update introduced new checks.

W504 "line break after binary operator" is ignored by default, and had to be added in `checkstyle.sh`'s list of ignored error codes. (Using `--ignore=list,of,codes` overrides the default list.)

Two places where `raise NotImplemented()` was used incorrectly (instead of `raise NotImplementedError()`) were corrected.

Most of the changes fix invalid escape sequences in regex strings all over the place—strings that likely should have already used `r''` but didn't for whatever reason. A few (mostly docstrings) were fixed by just escaping the backslash(es) properly.

I hope that the CI build of this PR goes off without a hitch… because local testing still isn't quite like running on Travis. Then all that's left is to finish testing all the functions I touched to make sure converting rules etc. to `r'string'` didn't break something unexpectedly. (It shouldn't, but still… surprises suck.)


dgw:
Decided to leave the `version` module cleanup for a future "module cruft" branch instead. This PR now contains only the `checkstyle`-related updates.

PR #1404, opened 2018-11-09
config: don't assume ~/.sopel as dotdir

### LINKED ISSUE #1243: Option to specify alternate config directory on first run
When the bot is first run, it executes ```sopel/run_script.py``` at startup.
Said script is hardcoded to create ```~/.sopel``` if it doesn't already exist.
Should there be an option to allow the user to specify an alternate directory to use?

This was observed on d9cfb41 running ```python sopel.py``` without prior configuration.

dgw:
Infer config directory from given config filename. I found this branch kicking around while looking for something else… how I forgot about it for six months, I have no idea.

Resolves #1243 (@alanhuang122 please confirm :sweat_smile:)

### DIFF
--- a/sopel/config/__init__.py
+++ b/sopel/config/__init__.py
@@ -214,12 +214,12 @@ class Config(object):
 
 
 def _wizard(section, config=None):
-    dotdir = os.path.expanduser('~/.sopel')
+    dotdir = os.path.dirname(config) if config is not None else os.path.expanduser('~/.sopel')
     configpath = os.path.join(dotdir, ((config or 'default.cfg') + ('.cfg' if config and not config.endswith('.cfg') else '')))
     if section == 'all':
         _create_config(configpath)
     elif section == 'mod':
-        _check_dir(False)
+        _check_dir(dotdir, False)
         if not os.path.isfile(configpath):
             print("No config file found." +
                   " Please make one before configuring these options.")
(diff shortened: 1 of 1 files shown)


alanhuang122:
I think I was reporting for someone in the channel. I don't really remember the justification anymore. If I'm not misreading the code, this would work by specifying `-c <file>` (as opposed to being interactive), which doesn't sound 100% correct, but I wasn't the one having the issue :v

dgw:
Yeah, you were. I'm glad I keep my IRC logs, because it means https://github.com/sopel-irc/sopel/issues/1243#issuecomment-445119525 now has the whole conversation that led to you opening the issue in the first place.

I don't know if this is actually enough, based on that added context, but there's certainly time to add a new argument in Sopel 7 (`--home` or something). Sopel's probably always going to try for `~/.sopel` if run without arguments that change the config dir, though. _Something_ has to be the default…

Exirel:
In  #1429 I used a `sopel.config.DEFAULT_HOMEDIR` variable for the default home directory. I think it can be used here too.

dgw:
@Exirel Want to drop a line note or two with suggested changes to show what you mean? Obviously I haven't touched this one in a while. :sweat_smile:

Exirel:
Over here: https://github.com/sopel-irc/sopel/pull/1429/files#diff-7fa6048ad26c4e738f57cc6ed66b156aR37

From my point of view, one of the issue faced by Sopel's code is that default variable and business rules (or behavior rules/features) are all over the places. So I thought it could be a good idea to put this default variable (here, the default location to Sopel's homedir) where it belongs, ie. in `sopel.config.DEFAULT_HOMEDIR`.

We recently discussed on IRC about reorganizing the code without any change in feature, and I think this is one example of how I would change things: making sure each feature has one and only place to define their rules.

Exirel:
It also means that this:

```
dotdir = os.path.dirname(config) if config is not None else os.path.expanduser('~/.sopel')
```

Would become this:

```
dotdir = os.path.dirname(config) if config is not None else DEFAULT_HOMEDIR
```

dgw:
Rebased and tweaked to use `DEFAULT_HOMEDIR` in the spots where it seemed appropriate.

@Exirel does this look better now?

Issue #1177, opened 2017-02-19
[wiktionary] Cannot look up affixes


dgw:
When passed an affix (prefix/suffix), the module incorrectly reports that it can't find any definitions, even though the pages exist on Wiktionary.

I guess it's just not handling those cases. Looks like the code is specifically looking for a whitelist of terms, and prefix types aren't in there, so it filters them all out.

Examples: `-ic`, `-ish`, `pre-`, `post-`

dgw:
Fixed by #1405.

PR #1405, opened 2018-11-09
wiktionary: fix inability to look up affixes

HumorBaby:
Directly addresses sopel-irc/sopel#1177 by adding "Prefix" and "Suffix"
types to parsing loop.

### DIFF
--- a/sopel/modules/wiktionary.py
+++ b/sopel/modules/wiktionary.py
@@ -52,6 +52,10 @@ def wikt(word):
             mode = 'particle'
         elif 'id="Preposition"' in line:
             mode = 'preposition'
+        elif 'id="Prefix"' in line:
+            mode = 'prefix'
+        elif 'id="Suffix"' in line:
+            mode = 'suffix'
         elif 'id="' in line:
             mode = None
 
@@ -66,7 +70,8 @@ def wikt(word):
 
 
 parts = ('preposition', 'particle', 'noun', 'verb',
-    'adjective', 'adverb', 'interjection')
+         'adjective', 'adverb', 'interjection',
+         'prefix', 'suffix')
 
 
 def format(result, definitions, number=2):


HumorBaby:
Realistically, this module should be able to handle all parts of speech headers defined in the Wiktionary entry layout (see [Wiktionary:Entry layout#Part of speech](https://en.wiktionary.org/wiki/Wiktionary:Entry_layout#Part_of_speech))... but that's probably more a feature request than bug?? I'll gladly work on it, but will most likely be a large refactor of the module.

dgw:
You'd be welcome to work on a more substantial change! I'm fully in favor of a single PR being able to close multiple related issues in the same module (even if not all of them have numbers).

Though if it would really require a _refactor_, this particular one would probably be left for Sopel 7. I've been letting a _lot_ of stuff slip into the 6.6.0 milestone, and it's time to stop that. :sweat_smile:

sopel/modules/wiktionary.py
```diff
@@ -6,7 +6,8 @@
 
 https://sopel.chat
 """
-from __future__ import unicode_literals, absolute_import, print_function, division
+from __future__ import unicode_literals, absolute_import, print_function, \
+    division
```
dgw: Please remove this style-only change. No other files in Sopel's codebase have this line-break. (The style checker specifically ignore line length, for now.)

HumorBaby: Sorry about all the line breaks. My IDE insists on auto-cleaning my code that way :/ I have convinced it not to anymore. Requested change coming right up

dgw:
@HumorBaby Thanks for teaching your IDE not to mess with stuff without being asked! :) I'll ask you to squash your second commit into the first by using `rebase -i` to mark it as a `fixup`, before merging, but hold off in case I see more changes needed when I have a chance to test.

HumorBaby:
Hmmm... As of right now, all 5 of those sub-definitions you showed are retrieved, but only 2 are shown due to message length thresholds. If a minimum length threshold (currently 150 chars) is met with just a subset of all "definition `parts` (main definition + sub-definitions), then that is all that is returned, despite more being available. So here, the initial default of 2 definition `parts` is below the sending threshold, so 3 definition `parts` are returned (main definition + 2 subs). There is a hard upper limit of 300 chars before truncation with `...`.

So a quick fix could be to just increase the default number of definition `parts`, and make the minor changes to the message length logic.

However, I'm leaning towards another PR. You may have noticed that in the returned definition: 
```
 <Sopel> -able — suffix: 1. An adjectival suffix; forms adjectives meaning:, 2. Able to be
        done; fit to be done, 3. Relevant to or suitable to, in accordance with
```
the sub-definitions get their own definition number.  Definition 1 ends with a `:`, which indicates the sub-definitions on the Wiktionary page, but is irrelevant/incorrect here. 2 and 3 in the example above need to be shown as sub-definitions (1a, 1b, etc...?), and not separate definitions they are currently. This would require a relatively large change to the parsing and message length logics, and may be better off done as a part of a separate PR that handles the additional parts of speech as well.

HumorBaby:
Yeah hoping to make some headway with those :smiley:

PR #1406, opened 2018-11-09
tld: add argument check

HumorBaby:
No arguments makes bot say exception. Argument check prevents this.


HumorBaby:
I pushed an amended commit without any extraneous changes, as requested @dgw . Let me know if I shouldn't do something like this in the future please (or if this is actually a proper way?) Thanks!

dgw:
It's easier to see if review comments have been addressed if you push a new commit, but for cases like this (quick turnaround, removing stuff from the PR without evolving it at all) I ask to squash anyway.

Amending the commit is easier than pushing a new one, waiting for approval, then rebasing, anyway. :) You did good!

Issue #1361, opened 2018-08-07
wiktionary: Some words just don't work (?)


dgw:
I _think_ this is a bug, but haven't done a whole lot of digging. Just noticed earlier today that `.wt abecedarian` spit out "Couldn't get any definitions" despite the page existing on `en.wiktionary.org` and containing multiple definitions in English.

Could be a small irregularity in the structure of that page, or a logic problem in the module. Or both. Or neither, I suppose, since I'm just speculating at this point.

Of note, other random words I tested with similar page structure, like `pleb`, work just fine. It's curious.

PR #1409, opened 2018-11-11
wiktionary: fix failed lookup for certain words

HumorBaby:
Fixes the issue sopel-irc/sopel#1361 where certain words do not return
a definition when one is available.

Words that include references in definitions expose a missed condition
check in parsing that is not relevant for most cases. References use the
`<sup>` tag, which in reference has its own `'id="'` attribute, causing
the `mode` variable in the loop to be reset, even though a definition
should be parsed from that line. These superscript references are also
now parsed out of the definition line.

### DIFF
--- a/sopel/modules/wiktionary.py
+++ b/sopel/modules/wiktionary.py
@@ -14,12 +14,14 @@ from sopel import web
 from sopel.module import commands, example
 
 uri = 'https://en.wiktionary.org/w/index.php?title=%s&printable=yes'
+r_sup = re.compile(r'<sup[^>]+>.+</sup>')  # Superscripts that are references only, not ordinal indicators, etc...
 r_tag = re.compile(r'<[^>]+>')
 r_ul = re.compile(r'(?ims)<ul>.*?</ul>')
 
 
 def text(html):
-    text = r_tag.sub('', html).strip()
+    text = r_sup.sub('', html)  # Remove superscripts that are references from definition
+    text = r_tag.sub('', text).strip()
     text = text.replace('\n', ' ')
     text = text.replace('\r', '')
     text = text.replace('(intransitive', '(intr.')
@@ -56,7 +58,9 @@ def wikt(word):
             mode = 'prefix'
         elif 'id="Suffix"' in line:
             mode = 'suffix'
-        elif 'id="' in line:
+        # 'id="' can occur in definition lines <li> when <sup> tag is used for references;
+        # make sure those are not excluded (see e.g., abecedarian).
+        elif ('id="' in line) and ('<li>' not in line):
             mode = None
 
         elif (mode == 'etmyology') and ('<p>' in line):


sopel/modules/wiktionary.py
```diff
@@ -52,7 +54,9 @@ def wikt(word):
             mode = 'particle'
         elif 'id="Preposition"' in line:
             mode = 'preposition'
-        elif 'id="' in line:
+        # 'id="' can occur in definition lines <li> when <sup> tag is used for references;
+        # make sure those are not excluded (see e.g., abecedarian).
+        elif ('id="' in line) and ('<li>' not in line):
```
dgw: Could some `<li>` elements contain classes, IDs, etc. that would make this match fail? Is there any reason you've found in testing not to use `<li` without the closing `>`?

HumorBaby: Good question; I was really just building off the previous logic, which, as implemented, assumes that a definition line will contain a plain `<li>` tag with no attributes or properties. However, I don't know whether this is always true  for definitions but seems to have worked for this long and I didn't want to change too much.

It is possible that `<li>` tags with classes, IDs, etc. are found elsewhere in the page, but are not part of the definition, and expanding the match to allow those may introduce more problems in the current parsing logic. 

Granted, my list of test words is very small and likely doesn't even come close to edge cases, it may be best to leave as is until another failing case is found, because this match seems to work for now (and has been). I think words are more likely to fail because of whats between the `<li>` tags (like the `<sup...>` in this case) than the `<li>` match itself.

dgw: Fair enough. Let's let this sleeping dog lie. :grinning:

dgw:
@HumorBaby Want to rebase this yourself or should I do it? It's just conflicted because I merged #1373; probably the easiest rebase ever.

HumorBaby:
I can do it, if you don't mind; get some git practice in.... rebase to follow in a few mins.

HumorBaby:
Ah snap, forgot to change the commit message! force-push coming right up

dgw:
> I can do it, if you don't mind

Actually I prefer when PR authors do their own rebases, because it avoids me getting my info plastered onto commits that I didn't create. Git still maintains the "Author" details as-is, but I wind up being the "Committer" and my avatar/name gets displayed in the history when I rebase other people's work.

Sopel just seems to have a lot of contributors who only use GitHub's web UI, which doesn't support rebasing, so I've gotten used to asking. :smile_cat:

HumorBaby:
New conflict from #1405 merge. Rebasing to resolve...

sopel/modules/wiktionary.py
```diff
@@ -13,13 +13,15 @@
 from sopel import web
 from sopel.module import commands, example
 
-uri = 'https://en.wiktionary.org/w/index.php?title=%s&printable=yes'
+uri = 'http://en.wiktionary.org/w/index.php?title=%s&printable=yes'
```
dgw: This reverts #1373… Can we not do that? :smile_cat:

HumorBaby: Gotta remember to check my diffs for unintended changes :confused:

PR #1410, opened 2018-11-12
xkcd: Change search engine from Duckduckgo to Bing

kwaaak:
For whatever reason Duckduckgo (which is used when it says google_search in this module) does not return results anymore. A quick switch to Bing restores the search function.

### DIFF
--- a/sopel/modules/xkcd.py
+++ b/sopel/modules/xkcd.py
@@ -8,7 +8,7 @@ from __future__ import unicode_literals, absolute_import, print_function, divisi
 import random
 import re
 import requests
-from sopel.modules.search import google_search
+from sopel.modules.search import bing_search
 from sopel.module import commands, url
 
 ignored_sites = [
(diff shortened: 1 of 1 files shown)


dgw:
DuckDuckGo no longer returns results when multiple `site:` operators are present in the query, _on its plain HTML version_ (`https://duckduckgo.com/html/`). The `xkcd` module uses a bunch of negative `-site:` operators to try and get comics only, and for whatever reason DDG stopped working for that case.

I reached out to DuckDuckGo via feedback and social media; we'll see if they address it. I consider this ultimately a "bug" in the `search` module (even though it's not Sopel's fault), but if all else fails and DDG won't address the issue then this patch is a good fallback for the next release.

Thanks for bringing the problem to my attention!

dgw:
DuckDuckGo isn't showing any signs of even acknowledging the `site:` bug's existence. Unless they suddenly decide to fix it, this workaround will be released with the next version of Sopel.

Issue #1411, opened 2018-11-13
reddit: support all reddit subdomains


dgw:
#1397 adds support for links to `old.reddit.com`, but any two-letter subdomain of `reddit.com` is valid as well. Reddit uses these for internationalization (e.g. `it.reddit.com` -> Italian UI) and also to allow subreddits to add custom styles (a common example is using `np.reddit.com` to trigger a "no participation" stylesheet that hides voting arrows).

Sopel's reddit module should support these subdomains.

There are also really old three-letter subdomains (`pay`, `ssl`) that still work. Most three-letter combos I tried redirect to `www.reddit.com/r/combohere` though.

PR #1397, opened 2018-10-15
add support for old.reddit.com links

dasu:

PR #1417, opened 2018-12-01
reddit: match more subdomains

Srinivas11789:
Fix for #1411 

- **Current Fix:** Wildcard subdomain for reddit.com
- **Changes:** sopel/modules/reddit.py --> regexp for url with wildcard subdomain

**Background:**

I think it would be better to have wildcards rather than calling out specific subdomains in the regex. (unless we scope the subdomains to support)

Based on the examples in the issue. I initially preferred to allow 2 char or 3 char length subdomain wildcards as `https?://(?:[a-z0-9]{2}\.|[a-z0-9]{3}\.)?reddit\.com`

Digging further, based on a initial vague subdomain scanning (summary below)
```
park1.reddit.com
glee.reddit.com
devel.reddit.com
edin.reddit.com
divms.reddit.com
eriogerg.reddit.com
gogo0.reddit.com
hofmystery.reddit.com
hq.reddit.com
maoliworld.reddit.com
aniwthoi.reddit.com
portableappz.reddit.com
uspeh-v-tebe.reddit.com
iheartfaces.reddit.com
glitterbox.reddit.com
iwanbanaran.reddit.com
realestalker.reddit.com
planet-social.reddit.com
server101.reddit.com
optimalmediagroup.reddit.com
statsperso.reddit.com
avtosaltanat.reddit.com
filmekaneten.reddit.com
mga.reddit.com
oviclub.reddit.comR
tweetdeck.reddit.com
secretebase.reddit.com
indiabooks.reddit.com
hoffmann.reddit.com
movidreams.reddit.com
to-gamato.reddit.com
permaculturaportugal.reddit.com
pica.reddit.com
atlantic-pacific.reddit.com
thetannerbrigade.reddit.com
metin58.reddit.com
creationsci.reddit.com
psychiatriinfirmiere.reddit.com
kinlb.reddit.com
primmoroz.reddit.com
plaqueta.reddit.com
...
...
```
Also, Most of the subdomains of reddit redirects to subreddits. 

I think it would be better to use a wildcard match `*.reddit.com` or to limit particular length of the subdomain. Suggestions?

### DIFF
--- a/sopel/modules/reddit.py
+++ b/sopel/modules/reddit.py
@@ -22,7 +22,7 @@ else:
     unescape = HTMLParser().unescape
 
 
-domain = r'https?://(?:www\.|np\.|old\.)?reddit\.com'
+domain = r'https?://(?:www\.|old\.|pay\.|ssl\.|[a-z]{2}\.)?reddit\.com'
 post_url = r'%s/r/(.*?)/comments/([\w-]+)' % domain
 user_url = r'%s/u(ser)?/([\w-]+)' % domain
 post_regex = re.compile(post_url)


dgw:
I think the best things to add are any two-letter subdomain and the specific three-letter subs mentioned in #1411. The rest of the subdomains are vanishingly unlikely ever to get linked in a channel, and it's not worth complicating the regex further to support them.

The only other thing I'd say about this PR is to tweak the commit message header into something like "reddit: expand subdomain matching (resolves #1411)". That solves two things: 1) it makes the short commit log much more informative, and 2) it will make GitHub close the issue when this PR is merged.

Srinivas11789:
Makes sense.  Fixed. 
* Wildcard two letter subdomain + subdomains mentioned in the issue.

dgw:
@Srinivas11789 Actually, this needs rebasing now that #1397 was merged. All you need to do is add `old` as one of the three-letter subdomains matched by the regex.

PR #1424, opened 2018-12-27
loader: test coverage

Exirel:
The `sopel.loader` module is responsible for loading Sopel Module (as in, "plugins" for Sopel), so it's quite critical.

So I added unit-tests, and change as few code as possible.

There is one function that I didn't test, because I want to rework it later: `enumerate_modules`. This function does too many things, and relies on too many global or hard-coded variables. This must be changed.

So, for now, here is a good increase in coverage for the project.

---

**Initial description**:

> Hi! It's a WIP and I don't expect it to be merged "as-is". I wanted to dive into Sopel's code, just to give it a try, and found that I could help on the test side of thing.
>
> I think I can do plenty more on the `sopel.loader` module, but I wanted to share this as soon as possible, to see if it would help and/or was in the right direction, code-style wise in particular.
>
> I'll work on this in the following days, but don't hesitate to stop me, ask question, make remarks, etc. feedback is very welcome!

### DIFF
--- a/test/test_loader.py
+++ b/test/test_loader.py
@@ -1,10 +1,50 @@
 # coding=utf-8
 """Test for the ``sopel.loader`` module."""
 import imp
+import os
 
 import pytest
 
 from sopel import loader, config
+
+
+MOCK_MODULE_CONTENT = """# coding=utf-8
+import sopel.module
+
+
+@sopel.module.commands("first")
+def first_command(bot, trigger):
+    pass
+
+
+@sopel.module.commands("second")
+def second_command(bot, trigger):
+    pass
+
+
+@sopel.module.interval(5)
+def interval5s(bot):
+    pass
+
+
+@sopel.module.interval(10)
+def interval10s(bot):
+    pass
+
+
+@sopel.module.url(r'.\\.example\\.com')
+def example_url(bot):
+    pass
+
+
+def shutdown():
+    pass
+
+
+def ignored():
+    pass
+
+"""
 
 
 @pytest.fixture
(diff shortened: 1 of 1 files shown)


Exirel:
I'll fix CI build tomorrow, I forgot to check flake8, my bad!

dgw:
Just took a peek because I was curious (and noticed I hadn't labeled this yet). The most exciting thing I see is Coveralls' current check status: "Coverage increased (+1.06%) to 43.248%" :tada:

dgw:
@Exirel I just found a really old (last edited in June 2018) `test_loader.py` in one of my local repos. You've probably already moved beyond it, but in case any of it is useful: [here](https://gist.github.com/dgw/1c21ce936353fc7377157bd754942c41). :smile_cat:

Exirel:
@dgw I need to check that. There are a lot of interesting things in there!

Exirel:
@dgw you can remove the WIP tag and have a look. I bet you can merge that one eyes closed. :wink:

test/test_loader.py
```diff
@@ -0,0 +1,477 @@
...
+
+def ignored():
+    pass
+
+"""
+
+
+@pytest.fixture
+def func():
+    """Pytest fixture to get a function that will returns True all the time"""
```
dgw: Since I don't expect to find any big things to change in this PR, I'll point out the little tiny ones like this grammatical mistake. :grin:

```suggestion
    """Pytest fixture to get a function that will return True all the time"""
```

test/test_loader.py
```diff
@@ -0,0 +1,477 @@
...
+    # Test the regex is compiled properly
+    regex = func.rule[0]
+    assert regex.match('abc')
+    assert regex.match('abcd')
+    assert not regex.match('efg')
+
+
+def test_clean_callable_rule_nick(tmpconfig, func):
+    """Assert ``$nick`` in a rule will match ``Sopel: `` or ``Sopel, ``."""
+    setattr(func, 'rule', [r'$nickhello'])
```
dgw: I feel like this shouldn't work without a space following `$nick`, but… Travis doesn't lie.

Not sure I'd want to _change_ it because it might affect backward compatibility with phenny/jenni modules, but… It certainly looks weird.

test/test_loader.py
```diff
@@ -0,0 +1,477 @@
...
+
+    # Test the regex is compiled properly
+    regex = func.rule[0]
+    assert regex.match('abc')
+    assert regex.match('abcd')
+    assert not regex.match('efg')
+
+
+def test_clean_callable_rule_nick(tmpconfig, func):
+    """Assert ``$nick`` in a rule will match ``Sopel: `` or ``Sopel, ``."""
```
dgw: Probably worth defining the bot's nick in `tmpconfig`, just in case the default value changes at some point. The tests don't need to start failing because of that.

Same for the `_nickname` test below, and others that assume this default.

sopel/loader.py
```diff
@@ -140,7 +120,7 @@ def clean_callable(func, config):
     prefix = config.core.prefix
     help_prefix = config.core.help_prefix
     func._docs = {}
-    doc = trim_docstring(func.__doc__)
+    doc = inspect.cleandoc(func.__doc__ or '').splitlines()
```
dgw: I appreciate what this replacement (and associated function deletion) is doing, but I don't think it belongs in this PR. Would you mind splitting it out? :smile_cat:

Unlike the added tests, this change really requires messing with the resulting output interactively via the `help` module to see if it affects anything. It shouldn't, but… I'm careful.

Exirel:
> It's really thorough overall, and I like where it's at. It's really _almost_ there!

Should be better now @dgw and thanks for the review again!

PR #1421, opened 2018-12-09
dice: strip whitespace from choices

dgw:

PR #1425, opened 2018-12-28
dice: clean up the dice "choose" command output (fix #1420)

### LINKED ISSUE #1420: dice: .choose output is messy
I started this thought process after noticing that `.choose thing1, thing2, thing3` will yield choices that include the leading space (`thing1`, ` thing2`, and ` thing3`), but it's not that simple.

Since `.choose` just uses a simple `str.split()` as-is, without cleaning the values at all, it's already starting with messy data. Then it swaps the delimiter used for no particular reason ([comment claims](https://github.com/sopel-irc/sopel/blob/d73a410717d6cb2b164bf5411f20a995210bbc22/sopel/modules/dice.py#L248) it's "to prevent ambiguity", but I think it actually does the opposite).

What comes out the other end:

```
<dgw> .choose thing1, thing2, thing3
<Sopel> dgw: Your options: thing1|  thing2|  thing3. My choice:  thing2
```

Extra spaces everywhere, delimiters that don't really make sense with the spacing around them… Ugly.

Current plans to sort this out:

  - [x] **In 6.6.0:** Clean the input to avoid extra spaces in the output. ~~Maybe switch to always use `, `&nbsp;as delimiter in output.~~ [#1421]
  - [x] **In 7.0.0:** ~~Drop support for everything but `,` (and maybe `/`) in the input, and~~ always use `, `&nbsp;in output ~~(if not already changed for 6.6.0)~~. [#1425]

There's considerable lead time before Sopel 7, so any user objections to dropping `\` and `|` as supported delimiters may be dropped in the comments here, or in the comments of the PR I will open with the change some time in early 2019, after 6.6.0 is out.

Exirel:
See issue #1420 and PR #1421 that I used as my starting point.

Following my comment on #1420 I decided to go with a PR that:

* strip the space from the input,
* properly format delimiter in the ouput,
* add some test example while I was at it,
* and use the space `\s` as a last-resort delimiter, because why not,

which improve coverage and make for a better output.

**EDIT**: we discussed another option, which I implemented afterward, to use comma as an output delimiter in all cases and to add quotation marks when needed.

Example with comma `,`:

    [nick] .choose a, b, c
    [Sopel] Your options: a, b, c. My choice: b

Example with comma `,` (without spaces):

    [nick] .choose a,b,c
    [Sopel] Your options: a, b, c. My choice: b

Example with pipe `|`:

    [nick] .choose a | b | c
    [Sopel] Your options: a, b, c. My choice: b

Example with pipe `|` (without spaces):

    [nick] .choose a|b|c
    [Sopel] Your options: a, b, c. My choice: b

Example with space `\s`:

    [nick] .choose a b c
    [Sopel] Your options: a, b, c. My choice: b

Example with pipe `|` but there are comma in some options:

    [nick] .choose a, b and c | a, b | just a
    [Sopel] Your options: "a, b and c", "a, b", just a. My choice: a, b

### DIFF
--- a/sopel/modules/dice.py
+++ b/sopel/modules/dice.py
@@ -242,6 +242,14 @@ def roll(bot, trigger):
 @sopel.module.commands("ch")
 @sopel.module.commands("choose")
 @sopel.module.priority("medium")
+@sopel.module.example(".choose a, b, c", r'Your options: a, b, c. My choice: (a|b|c)', re=True)
+@sopel.module.example(".choose a | b | c", r'Your options: a, b, c. My choice: (a|b|c)', re=True)
+@sopel.module.example(".choose a,b,c", r'Your options: a, b, c. My choice: (a|b|c)', re=True)
+@sopel.module.example(".choose a|b|c", r'Your options: a, b, c. My choice: (a|b|c)', re=True)
+@sopel.module.example(".choose a, b | just a",
+                      r'Your options: "a, b", just a. My choice: ((a, b)|(just a))',
+                      re=True)
+@sopel.module.example(".choose a", 'Your options: a. My choice: a')
 def choose(bot, trigger):
     """
     .choice option1|option2|option3 - Makes a difficult choice easy.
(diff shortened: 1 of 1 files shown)


dgw:
This looks good so far! I'm just going to leave a note, for now, but it doesn't need to be addressed immediately. It can wait until after Sopel 6.6.0 is (finally) released.

Commit fb4eb44 duplicates what #1421 does, just in a slightly different way. This PR should be amended (eventually) to make sure only one `strip()` operation survives into Sopel 7.

Exirel:
Yes indeed! It'll be done.

Exirel:
First rebase, just to clean up the `split` thing. The second one will change the output delimiter to a comma, with option quoted when needed.

PR #1430, opened 2019-01-02
currency: Moved to different API, support multiple target currencies, cache rates

Duckle29:
Hey there. Since I was moving from an old bot to sopel, I wanted to have it mimic what I and others were used to from my old bot. Because of this, I made it support multiple target currencies like:

`.cur 100 usd in dkk eur usd aud cad btc`

additionally, I moved to https://exchangeratesapi.io/ as the bank of Canada apparently doesn't offer rates for the Danish krona. Exchangeratesapi doesn't have a whole ton of currencies either, but the output is fixer.io compatible, so if it becomes a problem, a simple change would be to add a config section, allowing people to choose between API-key-less, exchangeratesapi.io or getting a free API key from fixer.io

A free API key should be more than enough as well, as the module caches rates for 24h in a global variable, and the free fixer.io plan offers 1000 requests / month

### DIFF
--- a/sopel/modules/currency.py
+++ b/sopel/modules/currency.py
@@ -2,6 +2,7 @@
 """
 currency.py - Sopel Currency Conversion Module
 Copyright 2013, Elsie Powell, embolalia.com
+Copyright 2019, Mikkel Jeppesen
 Licensed under the Eiffel Forum License 2.
 
 https://sopel.chat
@@ -9,102 +10,215 @@ https://sopel.chat
 from __future__ import unicode_literals, absolute_import, print_function, division
 
 import re
+import time
 
-from requests import get
+import requests
 
-from sopel.module import commands, example, NOLIMIT
-
-
-# The Canadian central bank has better exchange rate data than the Fed, the
-# Bank of England, or the European Central Bank. Who knew?
-base_url = 'https://www.bankofcanada.ca/valet/observations/FX{}CAD/json'
-regex = re.compile(r'''
-    (\d+(?:\.\d+)?)        # Decimal number
-    \s*([a-zA-Z]{3})       # 3-letter currency code
-    \s+(?:in|as|of|to)\s+  # preposition
-    ([a-zA-Z]{3})          # 3-letter currency code
-    ''', re.VERBOSE)
+from sopel.config.types import StaticSection, ValidatedAttribute
+from sopel.logger import get_logger
+from sopel.module import commands, example, NOLIMIT, rule
 
 
-def get_rate(code):
-    code = code.upper()
-    if code == 'CAD':
-        return 1, 'Canadian Dollar'
-    elif code == 'BTC':
-        btc_rate = get('https://apiv2.bitcoinaverage.com/indices/global/ticker/BTCCAD')
-        rates = btc_rate.json()
-        return 1 / rates['averages']['day'], 'Bitcoin—24hr average'
+FIAT_URL = 'https://api.exchangeratesapi.io/latest?base=EUR'
+FIXER_URL = 'http://data.fixer.io/api/latest?base=EUR&access_key={}'
+CRYPTO_URL = 'https://apiv2.bitcoinaverage.com/indices/global/ticker/short?crypto=BTC'
+EXCHANGE_REGEX = re.compile(r'''
+    ^(\d+(?:\.\d+)?)                                            # Decimal number
+    \s*([a-zA-Z]{3})                                            # 3-letter currency code
+    \s+(?:in|as|of|to)\s+                                       # preposition
+    (([a-zA-Z]{3}$)|([a-zA-Z]{3})\s)+$                          # one or more 3-letter currency code
+''', re.VERBOSE)
+LOGGER = get_logger(__name__)
+UNSUPPORTED_CURRENCY = "Sorry, {} isn't currently supported."
+UNRECOGNIZED_INPUT = "Sorry, I didn't understand the input."
 
-    data = get(base_url.format(code))
-    name = data.json()['seriesDetail']['FX{}CAD'.format(code)]['description']
-    name = name.split(" to Canadian")[0]
-    json = data.json()['observations']
-    for element in reversed(json):
-        if 'v' in element['FX{}CAD'.format(code)]:
-            return 1 / float(element['FX{}CAD'.format(code)]['v']), name
+rates_fiat_json = {}
+rates_btc_json = {}
+rates_updated = 0.0
+
+
+class CurrencySection(StaticSection):
+    fixer_io_key = ValidatedAttribute('fixer_io_key', default=None)
+    """Optional API key for Fixer.io (increases currency support)"""
+    auto_convert = ValidatedAttribute('auto_convert', parse=bool, default=False)
+    """Whether to convert currencies without an explicit command"""
+
+
+def configure(config):
+    """
+    | name | example | purpose |
+    | ---- | ------- | ------- |
+    | auto\\_convert | False | Whether to convert currencies without an explicit command |
+    | fixer\\_io\\_key | 0123456789abcdef0123456789abcdef | Optional API key for Fixer.io (increases currency support) |
+    """
+    config.define_section('currency', CurrencySection, validate=False)
+    config.currency.configure_setting('fixer_io_key', 'Optional API key for Fixer.io (leave blank to use exchangeratesapi.io):')
+    config.currency.configure_setting('auto_convert', 'Whether to convert currencies without an explicit command?')
+
+
+def setup(bot):
+    bot.config.define_section('currency', CurrencySection)
+
+
+class FixerError(Exception):
+    """A Fixer.io API Error Exception"""
+    def __init__(self, status):
+        super(FixerError, self).__init__("FixerError: {}".format(status))
+
+
+class UnsupportedCurrencyError(Exception):
+    """A currency is currently not supported by the API"""
+    def __init__(self, currency):
+        super(UnsupportedCurrencyError, self).__init__(currency)
+
+
+def update_rates(bot):
+    global rates_fiat_json, rates_btc_json, rates_updated
+
+    # If we have data that is less than 24h old, return
+    if time.time() - rates_updated < 24 * 60 * 60:
+        return
+
+    # Update crypto rates
+    response = requests.get(CRYPTO_URL)
+    response.raise_for_status()
+    rates_btc_json = response.json()
+
+    # Update fiat rates
+    if bot.config.currency.fixer_io_key is not None:
+        response = requests.get(FIXER_URL.format(bot.config.currency.fixer_io_key))
+        if not response.json()['success']:
+            raise FixerError('Fixer.io request failed with error: {}'.format(response.json()['error']))
+    else:
+        response = requests.get(FIAT_URL)
+
+    response.raise_for_status()
+    rates_fiat_json = response.json()
+    rates_updated = time.time()
+    rates_fiat_json['rates']['EUR'] = 1.0  # Put this here to make logic easier
+
+
+def btc_rate(code, reverse=False):
+    search = 'BTC{}'.format(code)
+
+    if search in rates_btc_json:
+        rate = rates_btc_json[search]['averages']['day']
+    else:
+        raise UnsupportedCurrencyError(code)
+
+    if reverse:
+        return 1 / rate
+    else:
+        return rate
+
+
+def get_rate(of, to):
+    of = of.upper()
+    to = to.upper()
+
+    if of == 'BTC':
+        return btc_rate(to, False)
+    elif to == 'BTC':
+        return btc_rate(of, True)
+
+    if of not in rates_fiat_json['rates']:
+        raise UnsupportedCurrencyError(of)
+
+    if to not in rates_fiat_json['rates']:
+        raise UnsupportedCurrencyError(to)
+
+    return (1 / rates_fiat_json['rates'][of]) * rates_fiat_json['rates'][to]
+
+
+def exchange(bot, match):
+    """Show the exchange rate between two currencies"""
+    if not match:
+        bot.reply(UNRECOGNIZED_INPUT)
+        return NOLIMIT
+
+    try:
+        update_rates(bot)  # Try and update rates. Rate-limiting is done in update_rates()
+    except requests.exceptions.RequestException as err:
+        bot.reply("Something went wrong while I was getting the exchange rate.")
+        LOGGER.error("Error in GET request: {}".format(err))
+        return NOLIMIT
+    except ValueError:
+        bot.reply("Error: Got malformed data.")
+        LOGGER.error("Invalid json on update_rates")
+        return NOLIMIT
+    except FixerError as err:
+        bot.reply('Sorry, something went wrong with Fixer')
+        LOGGER.error(err)
+        return NOLIMIT
+
+    query = match.string
+
+    others = query.split()
+    amount = others.pop(0)
+    of = others.pop(0)
+    others.pop(0)
+
+    # TODO: Use this instead after dropping Python 2 support
+    # amount, of, _, *others = query.split()
+
+    try:
+        amount = float(amount)
+    except ValueError:
+        bot.reply(UNRECOGNIZED_INPUT)
+    except OverflowError:
+        bot.reply("Sorry, input amount was out of range.")
+
+    if not amount:
+        bot.reply("Zero is zero, no matter what country you're in.")
+        return NOLIMIT
+
+    out_string = '{} {} is'.format(amount, of.upper())
+
+    for to in others:
+        try:
+            out_string = build_reply(amount, of.upper(), to.upper(), out_string)
+        except ValueError:
+            LOGGER.error("Raw rate wasn't a float")
+            return NOLIMIT
+        except KeyError as err:
+            bot.reply("Error: Invalid rates")
+            LOGGER.error("No key: {} in json".format(err))
+            return NOLIMIT
+        except UnsupportedCurrencyError as cur:
+            bot.reply(UNSUPPORTED_CURRENCY.format(cur))
+            return NOLIMIT
+
+    bot.reply(out_string[0:-1])
 
 
 @commands('cur', 'currency', 'exchange')
-@example('.cur 20 EUR in USD')
-def exchange(bot, trigger):
-    """Show the exchange rate between two currencies"""
+@example('.cur 100 usd in btc cad eur',
+         r'100\.0 USD is [\d\.]+ BTC, [\d\.]+ CAD, [\d\.]+ EUR',
+         re=True)
+@example('.cur 3 can in one day', 'Sorry, CAN isn\'t currently supported.')
+def exchange_cmd(bot, trigger):
     if not trigger.group(2):
-        return bot.reply("No search term. An example: .cur 20 EUR in USD")
-    match = regex.match(trigger.group(2))
-    if not match:
-        # It's apologetic, because it's using Canadian data.
-        bot.reply("Sorry, I didn't understand the input.")
-        return NOLIMIT
+        return bot.reply("No search term. Usage: {}cur 100 usd in btc cad eur"
+                         .format(bot.config.core.help_prefix))
 
-    amount, of, to = match.groups()
-    try:
-        amount = float(amount)
-    except ValueError:
-        bot.reply("Sorry, I didn't understand the input.")
-    except OverflowError:
-        bot.reply("Sorry, input amount was out of range.")
-    display(bot, amount, of, to)
+    match = EXCHANGE_REGEX.match(trigger.group(2))
+    exchange(bot, match)
 
 
-def display(bot, amount, of, to):
-    if not amount:
-        bot.reply("Zero is zero, no matter what country you're in.")
-    try:
-        of_rate, of_name = get_rate(of)
-        if not of_name:
-            bot.reply("Unknown currency: %s" % of)
-            return
-        to_rate, to_name = get_rate(to)
-        if not to_name:
-            bot.reply("Unknown currency: %s" % to)
-            return
-    except Exception:  # TODO: Be specific
-        bot.reply("Something went wrong while I was getting the exchange rate.")
-        return NOLIMIT
-
-    result = amount / of_rate * to_rate
-    bot.say("{:.2f} {} ({}) = {:.2f} {} ({})".format(amount, of.upper(), of_name,
-                                             result, to.upper(), to_name))
+@rule(EXCHANGE_REGEX)
+@example('100 usd in btc cad eur')
+def exchange_re(bot, trigger):
+    if bot.config.currency.auto_convert:
+        match = EXCHANGE_REGEX.match(trigger)
+        exchange(bot, match)
 
 
-@commands('btc', 'bitcoin')
-@example('.btc 20 EUR')
-def bitcoin(bot, trigger):
-    """Convert between Bitcoin and a fiat currency"""
-    # if 2 args, 1st is number and 2nd is currency. If 1 arg, it's either the number or the currency.
-    to = trigger.group(4)
-    amount = trigger.group(3)
-    if not to:
-        to = trigger.group(3) or 'USD'
-        amount = 1
+def build_reply(amount, of, to, out_string):
+    rate_raw = get_rate(of, to)
+    rate = float(rate_raw)
+    result = float(rate * amount)
 
-    try:
-        amount = float(amount)
-    except ValueError:
-        bot.reply("Sorry, I didn't understand the input.")
-        return NOLIMIT
-    except OverflowError:
-        bot.reply("Sorry, input amount was out of range.")
-        return NOLIMIT
+    if to == 'BTC':
+        return out_string + ' {:.5f} {},'.format(result, to)
 
-    display(bot, amount, 'BTC', to)
+    return out_string + ' {:.2f} {},'.format(result, to)


Duckle29:
I added the option to configure the module with a fixer.io API key, as fixer.io has 167 currencies while exchangeratesapi.io has 32

additionally, the module can be configured to also respond directly to regex matches.

If no configuration is done, the module defaults to exchangerates.io and no regex.

dgw:
Look at that title-edit mess I just made, ugh. :grin:

@Duckle29 Are you OK with leaving your `master` branch untouched for possibly several months? As long as this PR is open, you won't be able to make any other changes on your `master` without messing up this pull request. It's fine if you are—I don't care what branch name shows up in the merge commit—but it could be inconvenient for you if you forget that `master` is part of a PR and push something. :smile_cat:

Duckle29:
@dgw Yeah that's okay. If I have to work on something in the future I'll make sure to move to a different branch, but for now it works as it should :)

sopel/modules/currency.py
```diff
@@ -1,102 +1,189 @@
...
-    match = regex.match(trigger.group(2))
+
+    # match = regex.match(trigger)                                                      # that
     if not match:
-        # It's apologetic, because it's using Canadian data.
         bot.reply("Sorry, I didn't understand the input.")
         return NOLIMIT
 
-    amount, of, to = match.groups()
+    update_rates(bot)  # Try and update rates. Rate-limiting is done in update_rates()
```
HumorBaby: This call needs to have a `try`\`catch` to handle the various errors that could bubble up from `update_rates`. A quick scan shows at least these are currently unhandled:
- `request.raise_for_status()`
- `request.json()` will raise `ValueError` if it gets a malformed string

Without handling these exceptions to say `Hey! Something bad happen'd`, the user has no way something went wrong during their command.

sopel/modules/currency.py
```diff
@@ -1,102 +1,189 @@
...
-def exchange(bot, trigger):
+    ^(\d+(?:\.\d+)?)                                            # Decimal number
+    \s*([a-zA-Z]{3})                                            # 3-letter currency code
+    \s+(?:in|as|of|to)\s+                                       # preposition
+    (([a-zA-Z]{3}$)|([a-zA-Z]{3})\s)+$                          # one or more 3-letter currency code
+''', re.VERBOSE)
+
+
+def btc_rate(code, reverse=False):
+    global rates_btc_json
```
HumorBaby: This line `global ...` is not needed since rates_btc_json is not being modified here.

sopel/modules/currency.py
```diff
@@ -1,102 +1,189 @@
...
         bot.reply("Sorry, I didn't understand the input.")
     except OverflowError:
         bot.reply("Sorry, input amount was out of range.")
-    display(bot, amount, of, to)
+
+    out_string = '{} {} is'.format(amount, of.upper())
+    for to in others:
+        try:
+            out_string = build_reply(bot, amount, of.upper(), to.upper(), out_string)
+        except (requests.exceptions.HTTPError, requests.exceptions.ConnectionError) as e:
```
HumorBaby: `build_reply` doesn't seem to throw any of these exceptions as it doesn't actually make any requests in the call stack.

This block would better serve in `exchange` where I left another comment about error handling.

sopel/modules/currency.py
```diff
@@ -1,102 +1,189 @@
...
-    # if 2 args, 1st is number and 2nd is currency. If 1 arg, it's either the number or the currency.
-    to = trigger.group(4)
-    amount = trigger.group(3)
-    if not to:
-        to = trigger.group(3) or 'USD'
-        amount = 1
+    return out_string + ' {:.2f} {},'.format(result, to)
+
+
+class CurrencySection(StaticSection):
```
HumorBaby: Most modules seem to define the config at the top. Can this be moved up top for consistency since it is a built-in plugin?

sopel/modules/currency.py
```diff
@@ -1,102 +1,189 @@
...
+    fixer_io_key = ValidatedAttribute('fixer_io_key', default=None)
+    enable_regex = ValidatedAttribute('enable_regex', parse=bool, default=False)
+
+
+def configure(config):
+    config.define_section('currency', CurrencySection, validate=False)
+    config.currency.configure_setting('fixer_io_key', 'API key for fixer IO. Leave blank to use exchangeratesapi.io:')
+    config.currency.configure_setting('enable_regex', 'automatically respond to regex matches:')
```
HumorBaby:
```suggestion
    config.currency.configure_setting('enable_regex', 'Automatically respond to regex matches?')   
```

Most other settings like this phrase the configuration prompt as a question; e.g., https://github.com/sopel-irc/sopel/blob/eb23031ea7729836de6eaece61b42434a8b44fd5/sopel/modules/admin.py#L37-L40

sopel/modules/currency.py
```diff
@@ -1,102 +1,189 @@
...
+from sopel.config.types import StaticSection, ValidatedAttribute
 
-# The Canadian central bank has better exchange rate data than the Fed, the
-# Bank of England, or the European Central Bank. Who knew?
-base_url = 'https://www.bankofcanada.ca/valet/observations/FX{}CAD/json'
+rates_fiat_json = {}
+rates_btc_json = {}
+
+
+fiat_url = 'https://api.exchangeratesapi.io/latest?base=EUR'
```
HumorBaby: How do you feel about capitalized constants? Should the URLs be changed? e.g., `fiat_url` to `FIAT_URL`?

sopel/modules/currency.py
```diff
@@ -1,102 +1,189 @@
...
+    # Update crypto rates
+    request = requests.get(crypto_url)
+    request.raise_for_status()
+    rates_btc_json = request.json()
+
+    # Update fiat rates
+    if bot.config.currency.fixer_io_key is not None:
+        request = requests.get(fixer_url.format(bot.config.currency.fixer_io_key))
+        if not request.json()['success']:
+            stderr(str(request.json()['error']))
```
HumorBaby: This should be a `LOGGER.error` call instead of `stderr`, which IMO should be reserved for internal bot usage only. Modules should log with `LOGGER`.

```python
LOGGER = get_logger(__name__)
```

sopel/modules/currency.py
dgw: It's a nice style guideline to follow, yeah. :+1:

I think in addition to being capitalized, `regex` should become something a bit more specific, like `EXCHANGE_REGEX` or similar. (Yes, I know that name is a holdover from the original code. It was kinda bad then, too, lol. I didn't touch the variables in my cruft cleanup commit [1cfe818] because, well, I forgot—but I can *totally* pretend that it's because I wanted to minimize merge conflicts with this PR! :stuck_out_tongue:)

sopel/modules/currency.py
dgw: There are also more potential exceptions that could happen than just these two types. See [this `help` code](https://github.com/sopel-irc/sopel/blob/eb23031ea7729836de6eaece61b42434a8b44fd5/sopel/modules/help.py#L40-L48) for one solid way to handle everything at once. (I don't want to have `sopel.web` go back to wrapping `requests`, but… they could really use it. The error-catching boilerplate can be tiresome.)

sopel/modules/currency.py
dgw: No reason it can't be. Shouldn't cause any `NameError`s (unlike when I tried to move the `help` config stuff higher in #1451).

The `CurrencySection` class, `configure`, and `setup` should all be moved up to just before the first `def` (`btc_rate()`).

sopel/modules/currency.py
```diff
@@ -1,102 +1,189 @@
...
 
+import time
 import re
+import requests
 
-from requests import get
-from sopel.module import commands, example, NOLIMIT
+from sopel.module import commands, example, NOLIMIT, rule
+from sopel.tools import stderr
+from sopel.config.types import StaticSection, ValidatedAttribute
```
dgw: Should be first in this "internal" imports block. Eventually we'll start enforcing import order, and the more stuff that's already sorted correctly, the easier it will be to turn that on. :smile_cat:

sopel/modules/currency.py
```diff
@@ -1,102 +1,189 @@
 # coding=utf-8
 # Copyright 2013 Elsie Powell, embolalia.com
+# Copyright 2019 Mikkel Jeppesen
 # Licensed under the Eiffel Forum License 2
 from __future__ import unicode_literals, absolute_import, print_function, division
 
+import time
```
dgw: Should be below `re`. (See below comment.)

sopel/modules/currency.py
```diff
@@ -1,102 +1,189 @@
...
 import re
+import requests
```
dgw: Another import-sorting prep change.

```suggestion

import requests
```

sopel/modules/currency.py
```diff
@@ -1,102 +1,189 @@
...
         bot.reply("Sorry, input amount was out of range.")
-    display(bot, amount, of, to)
+
+    out_string = '{} {} is'.format(amount, of.upper())
+    for to in others:
+        try:
+            out_string = build_reply(bot, amount, of.upper(), to.upper(), out_string)
+        except (requests.exceptions.HTTPError, requests.exceptions.ConnectionError) as e:
+            bot.reply("Something went wrong while I was getting the exchange rate.")
+            stderr("Error in GET request: {}".format(e))
```
dgw: Another `stderr` call that should be converted to use the logger.

sopel/modules/currency.py
```diff
@@ -1,102 +1,189 @@
...
+            return NOLIMIT
+        except ValueError:
+            return NOLIMIT
+
+    bot.reply(out_string[0:-1])
+
+
+@commands('cur', 'currency', 'exchange')  # this
```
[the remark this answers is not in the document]
Duckle29: hah. no clue

sopel/modules/currency.py
Duckle29: Moved to a LOGGER.error()

sopel/modules/currency.py
```diff
@@ -1,102 +1,189 @@
...
-    amount = trigger.group(3)
-    if not to:
-        to = trigger.group(3) or 'USD'
-        amount = 1
+    return out_string + ' {:.2f} {},'.format(result, to)
+
+
+class CurrencySection(StaticSection):
+    fixer_io_key = ValidatedAttribute('fixer_io_key', default=None)
+    enable_regex = ValidatedAttribute('enable_regex', parse=bool, default=False)
```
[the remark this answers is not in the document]
Duckle29: Hmm how about the name auto_convert?

sopel/modules/currency.py
Duckle29: I'm not quite sure what you mean. I sorted imports according to what I think was meant, is that the correct sorting?

sopel/modules/currency.py
Duckle29: Should've fixed this.

sopel/modules/currency.py
Duckle29: reordered imports to be alphabetical, native python first, sopel modules after.

sopel/modules/currency.py
Duckle29: Done :)

sopel/modules/currency.py
Duckle29: Done :)

sopel/modules/currency.py
Duckle29: Done :)

sopel/modules/currency.py
Duckle29: I looked at the request module and some discussions about it. it seems `requests.exceptions.RequestException` wraps up all of the mentioned exceptions, so I just catch this now.

Duckle29: Also, I moved this to the exchange function.

sopel/modules/currency.py
Duckle29: removed :)

sopel/modules/currency.py
Duckle29: Added a try-catch. let me know if it catches what it should. 

`requests.exceptions.RequestException` should catch all request exceptions.

sopel/modules/currency.py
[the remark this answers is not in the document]
Duckle29: I'll get around to rebasing once all the requested changes have been worked on :)

sopel/modules/currency.py
dgw: Yep, this section is :ok_hand:

sopel/modules/currency.py
dgw: `requests` is not a standard import. It comes from PyPI. This block should actually be two blocks:

```python
import re
import time

import requests
```

Note the blank line.

sopel/modules/currency.py
```diff
@@ -4,43 +4,43 @@
...
 
-rates_fiat_json = {}
-rates_btc_json = {}
-
-
-fiat_url = 'https://api.exchangeratesapi.io/latest?base=EUR'
-fixer_url = 'http://data.fixer.io/api/latest?base=EUR&access_key={}'
-crypto_url = 'https://apiv2.bitcoinaverage.com/indices/global/ticker/short?crypto=BTC'
-regex = re.compile(r'''
+FIAT_URL = 'https://api.exchangeratesapi.io/latest?base=EUR'
```
dgw: Two blank lines before constants (relatively new convention, standardized to the best of my ability in #1402).

sopel/modules/currency.py
```diff
@@ -4,43 +4,43 @@
...
+FIXER_URL = 'http://data.fixer.io/api/latest?base=EUR&access_key={}'
+CRYPTO_URL = 'https://apiv2.bitcoinaverage.com/indices/global/ticker/short?crypto=BTC'
+EXCHANGE_REGEX = re.compile(r'''
     ^(\d+(?:\.\d+)?)                                            # Decimal number
     \s*([a-zA-Z]{3})                                            # 3-letter currency code
     \s+(?:in|as|of|to)\s+                                       # preposition
     (([a-zA-Z]{3}$)|([a-zA-Z]{3})\s)+$                          # one or more 3-letter currency code
 ''', re.VERBOSE)
```
dgw: This blank line is for…? If anything, put a blank between `LOGGER` (a "constant") and the `rates_*_json` cache dicts.

sopel/modules/currency.py
```diff
@@ -4,43 +4,43 @@
...
+    config.currency.configure_setting('auto_convert', 'Automatically respond to regex matches?')
 
-    if reverse:
-        return 1 / rate
-    else:
-        return rate
+
+def setup(bot):
+    bot.config.define_section('currency', CurrencySection)
+# End config
```
dgw: This isn't really part of the config. Both this and the `# Config` comment above are not needed.

sopel/modules/currency.py
```diff
@@ -4,43 +4,43 @@
...
 
-    search = 'BTC{}'.format(code)
 
-    if search in rates_btc_json:
-        rate = rates_btc_json[search]['averages']['day']
-    else:
-        return "Sorry {} isn't currently supported".format(code)
+def configure(config):
+    config.define_section('currency', CurrencySection, validate=False)
+    config.currency.configure_setting('fixer_io_key', 'API key for fixer IO. Leave blank to use exchangeratesapi.io:')
```
dgw: This prompt line would read better if it was consistent about how it calls each site. They are `fixer.io` and `exchangeratesapi.io`, but one is called `fixer IO` instead.

sopel/modules/currency.py
```diff
@@ -53,27 +53,41 @@ def update_rates(bot):
...
+        request = requests.get(FIAT_URL)
 
     request.raise_for_status()
     rates_fiat_json = request.json()
     rates_fiat_json['date'] = time.time()
     rates_fiat_json['rates']['EUR'] = 1.0  # Put this here to make logic easier
 
 
+def btc_rate(code, reverse=False):
+
```
dgw: No blank needed here

sopel/modules/currency.py
```diff
@@ -99,12 +113,19 @@ def get_rate(of, to):
 def exchange(bot, match):
     """Show the exchange rate between two currencies"""
```
dgw: No blank needed here

sopel/modules/currency.py
```diff
@@ -128,29 +149,30 @@ def exchange(bot, match):
             out_string = build_reply(bot, amount, of.upper(), to.upper(), out_string)
         except (requests.exceptions.HTTPError, requests.exceptions.ConnectionError) as e:
             bot.reply("Something went wrong while I was getting the exchange rate.")
-            stderr("Error in GET request: {}".format(e))
+            LOGGER.error("Error in GET request: {}".format(e))
+
```
dgw: Unnecessary blank line added

sopel/modules/currency.py
```diff
@@ -1,102 +1,196 @@
...
+# Config
+class CurrencySection(StaticSection):
+    fixer_io_key = ValidatedAttribute('fixer_io_key', default=None)
+    auto_convert = ValidatedAttribute('auto_convert', parse=bool, default=False)
+
+
+def configure(config):
+    config.define_section('currency', CurrencySection, validate=False)
+    config.currency.configure_setting('fixer_io_key', 'API key for fixer IO. Leave blank to use exchangeratesapi.io:')
+    config.currency.configure_setting('auto_convert', 'Automatically respond to regex matches?')
```
dgw: The new setting name works for me, good idea!

Now, can you perform that same magic again and come up with a more descriptive prompt text than this? :smile_cat: This doesn't really get across "regex matches" of what.

sopel/modules/currency.py
dgw: You "moved" this? Then why is it still here? :laughing:

sopel/modules/currency.py
```diff
@@ -1,102 +1,189 @@
...
+
+
+class CurrencySection(StaticSection):
+    fixer_io_key = ValidatedAttribute('fixer_io_key', default=None)
+    enable_regex = ValidatedAttribute('enable_regex', parse=bool, default=False)
+
+
+def configure(config):
+    config.define_section('currency', CurrencySection, validate=False)
+    config.currency.configure_setting('fixer_io_key', 'API key for fixer IO. Leave blank to use exchangeratesapi.io:')
```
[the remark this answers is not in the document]
dgw: This change was marked as resolved, but not addressed.

sopel/modules/currency.py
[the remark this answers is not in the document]
dgw: New setting name works for me. Rebasing later is good :+1:

sopel/modules/currency.py
HumorBaby: Your latest changes seem to cover it. I believe the idea was that imports should not only be separated (into blocks by stdlib, internal, third-party), but also that they should be alphabetical wherever possible. So for example, `sopel.config..` should be before `sopel.module...` which is before `sopel.tools...`. I think that's what @dgw meant. ...?

HumorBaby: > Your latest changes seem to cover it. I believe the idea was that imports should not only be separated (into blocks by stdlib, internal, third-party), but also that they should be alphabetical wherever possible within those blocks. So for example, `sopel.config..` should be before `sopel.module...` which is before `sopel.tools...`. I think that's what @dgw meant. ...?

Oops, sorry for the late (and unnecessary) comment :grimacing: 

That's what I get for leaving the tab open all day and not refreshing... :hushed:

sopel/modules/currency.py
Duckle29: That is a great question. must've forgotten to save the file before committing or something :laughing:

sopel/modules/currency.py
Duckle29: Should be resolved if I understood you correctly

sopel/modules/currency.py
Duckle29: I changed it to: "Automatically respond to currency conversion lines?"
alternatively I considered : "Automatically respond to conversions (10 usd in dkk eur)?"

sopel/modules/currency.py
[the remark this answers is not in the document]
Duckle29: Again. Must've forgotten to save a file or something.

sopel/modules/currency.py
```diff
@@ -2,109 +2,196 @@
 """
 currency.py - Sopel Currency Conversion Module
 Copyright 2013, Elsie Powell, embolalia.com
+Copyright 2019 Mikkel Jeppesen
```
dgw: For consistency with (almost) all other module headers:

```suggestion
Copyright 2019, Mikkel Jeppesen
```

sopel/modules/currency.py
```diff
@@ -2,109 +2,196 @@
...
 Licensed under the Eiffel Forum License 2.
 
 https://sopel.chat
 """
+
```
dgw: We pretty well never put a blank line between the module docstring and future imports.
```suggestion
```

sopel/modules/currency.py
```diff
@@ -2,109 +2,196 @@
...
+    config.currency.configure_setting('auto_convert', 'Automatically respond to currency conversion lines?')
+
+
+def setup(bot):
+    bot.config.define_section('currency', CurrencySection)
+
+
+def update_rates(bot):
+    global rates_fiat_json
+    global rates_btc_json
```
dgw: You [can combine these](https://docs.python.org/3/reference/simple_stmts.html#the-global-statement) into one line.

```py
    global rates_fiat_json, rates_btc_json
```

sopel/modules/currency.py
```diff
@@ -2,109 +2,196 @@
...
+        request = requests.get(FIXER_URL.format(bot.config.currency.fixer_io_key))
+        if not request.json()['success']:
+            LOGGER.error(str(request.json()['error']))
+            return bot.reply('Sorry, something went wrong')
+    else:
+        request = requests.get(FIAT_URL)
+
+    request.raise_for_status()
+    rates_fiat_json = request.json()
+    rates_fiat_json['date'] = time.time()
```
dgw: Seems odd that this is a property of the cached fiat currency data, when it actually applies to both/everything. I'm no fan of globals, but maybe this deserves to be its very own value, e.g. `rates_updated`?

sopel/modules/currency.py
```diff
@@ -2,109 +2,196 @@
...
+    rates_fiat_json['rates']['EUR'] = 1.0  # Put this here to make logic easier
+
+
+def btc_rate(code, reverse=False):
+    search = 'BTC{}'.format(code)
+
+    if search in rates_btc_json:
+        rate = rates_btc_json[search]['averages']['day']
+    else:
+        return "Sorry {} isn't currently supported".format(code)
```
dgw: Grammar nitpick:

```suggestion
        return "Sorry, {} isn't currently supported.".format(code)
```

sopel/modules/currency.py
```diff
@@ -2,109 +2,196 @@
...
+            code = to
+            reverse = False
+        else:
+            code = of
+            reverse = True
+
+        return btc_rate(code, reverse)
+
+    if of not in rates_fiat_json['rates']:
+        return "Sorry {} isn't currently supported".format(of)
```
dgw: Grammar nitpick:

```suggestion
        return "Sorry, {} isn't currently supported.".format(of)
```

sopel/modules/currency.py
```diff
@@ -2,109 +2,196 @@
...
+
+    if to not in rates_fiat_json['rates']:
+        return "Sorry {} isn't currently supported".format(to)
```
dgw: Grammar nitpick:

```suggestion
        return "Sorry, {} isn't currently supported.".format(to)
```

(This is three times I made the same change to the same string… Might deserve to be a constant, e.g. `UNSUPPORTED_CURRENCY = "Sorry, {} isn't currently supported."`)

sopel/modules/currency.py
```diff
@@ -2,109 +2,196 @@
...
+
+
+def exchange(bot, match):
     """Show the exchange rate between two currencies"""
-    if not trigger.group(2):
-        return bot.reply("No search term. An example: .cur 20 EUR in USD")
-    match = regex.match(trigger.group(2))
     if not match:
-        # It's apologetic, because it's using Canadian data.
         bot.reply("Sorry, I didn't understand the input.")
```
dgw: You didn't touch this, but like the other "Sorry" error string, this could be a constant too. It's used multiple times.

sopel/modules/currency.py
```diff
@@ -2,109 +2,196 @@
...
-        bot.reply("Something went wrong while I was getting the exchange rate.")
-        return NOLIMIT
+    bot.reply(out_string[0:-1])
+
+
+@commands('cur', 'currency', 'exchange')
+@example('.cur 100 usd in btc cad eur')
+def exchange_cmd(bot, trigger):
+    if not trigger.group(2):
+        return bot.reply("No search term. An example: .cur 100 usd in btc cad eur")
```
dgw: We're trying _really_ hard to fix the hard-coded `.`s in usage messages all through the core modules!

```suggestion
        return bot.reply("No search term. An example: {}cur 100 usd in btc cad eur"
                         .format(bot.config.core.help_prefix))
```

Optionally, also replace "An example" with "Usage" to align with most of the other core command set.

sopel/modules/currency.py
```diff
@@ -2,109 +2,196 @@
...
+        return NOLIMIT
+
+    query = match.string
+
+    others = query.split()
+    amount = others.pop(0)
+    of = others.pop(0)
+    others.pop(0)
+
+#    amount, of, _, *others = query.split() # I'd much rather use this, but it's not python 2.7 compatible
```
dgw: Fix comment indent and convert it to a TODO marker:

```suggestion
    # TODO: Use this instead after dropping Python 2 support
    # amount, of, _, *others = query.split()
```

sopel/modules/currency.py
```diff
@@ -2,109 +2,196 @@
...
+    global rates_fiat_json
+    global rates_btc_json
+
+    # If we have data that are less than 24h old, return
+    if 'date' in rates_fiat_json:
+        if time.time() - rates_fiat_json['date'] < 24 * 60 * 60:
+            return
+
+    # Update crypto rates
+    request = requests.get(CRYPTO_URL)
```
dgw: When naming with a full word like this, the word chosen should make sense. `request` isn't really a request, it's a response (or result)—which is why this pattern often uses the name `res` (or, for the truly terse, `r`).

This is a real nitpick, though, and isn't required for PR approval. :smile_cat:

Duckle29: But now that you've brought my attention to it I can't simply leave it. That'd hurt :P

sopel/modules/currency.py
```diff
@@ -2,109 +2,199 @@
...
-        bot.reply("Something went wrong while I was getting the exchange rate.")
-        return NOLIMIT
+    bot.reply(out_string[0:-1])
 
-    result = amount / of_rate * to_rate
-    bot.say("{:.2f} {} ({}) = {:.2f} {} ({})".format(amount, of.upper(), of_name,
-                                             result, to.upper(), to_name))
 
+@commands('cur', 'currency', 'exchange')
+@example('.cur 100 usd in btc cad eur')
```
HumorBaby: You can test success and failure even if the responses are not fixed with `re=True` :grinning: 

```suggestion
@example('.cur 100 usd in btc cad eur', '100\.0 USD is [\d\.]+ BTC, [\d\.]+ CAD, [\d\.]+ EUR', re=True)
@example('.cur 3 can in one day', 'Sorry, CAN isn\'t currently supported.')
```

Depending on the order of this and #1555, you may need to add an online=True argument to the `@example`.

dgw: **me:** quietly edits @HumorBaby's suggestion to be less ridiculous, line length–wise

sopel/modules/currency.py
```diff
@@ -30,9 +29,12 @@
     (([a-zA-Z]{3}$)|([a-zA-Z]{3})\s)+$                          # one or more 3-letter currency code
 ''', re.VERBOSE)
 LOGGER = get_logger(__name__)
+UNSUPPORTED_CURRENCY = "Sorry, {} isn't currently supported."
+GENERIC_SORRY = "Sorry, I didn't understand the input."
```
dgw: I should have suggested a name for this constant, haha. Maybe `UNRECOGNIZED_INPUT`?

sopel/modules/currency.py
HumorBaby: u wot m8?! :rage2: 

JK :smile: Gotta remember that line length, @HumorBaby!!

sopel/modules/currency.py
```diff
@@ -2,109 +2,201 @@
...
+    response = requests.get(CRYPTO_URL)
+    response.raise_for_status()
+    rates_btc_json = response.json()
+
+    # Update fiat rates
+    if bot.config.currency.fixer_io_key is not None:
+        response = requests.get(FIXER_URL.format(bot.config.currency.fixer_io_key))
+        if not response.json()['success']:
+            LOGGER.error(str(response.json()['error']))
+            return bot.reply('Sorry, something went wrong')
```
HumorBaby: Have this raise an exception so you can differentiate this `return` (en error) from the `return` when data less than 24h old already exist (a normal continuation).

sopel/modules/currency.py
```diff
@@ -2,109 +2,201 @@
...
-    if not trigger.group(2):
-        return bot.reply("No search term. An example: .cur 20 EUR in USD")
-    match = regex.match(trigger.group(2))
     if not match:
-        # It's apologetic, because it's using Canadian data.
-        bot.reply("Sorry, I didn't understand the input.")
+        bot.reply(UNRECOGNIZED_INPUT)
+        return NOLIMIT
+
+    try:
```
HumorBaby: Catch the exception from `update_rates()` here, and display the message from there, here instead.

sopel/modules/currency.py
```diff
@@ -2,109 +2,201 @@
...
+        if of == 'BTC':
+            code = to
+            reverse = False
+        else:
+            code = of
+            reverse = True
+
+        return btc_rate(code, reverse)
+
+    if of not in rates_fiat_json['rates']:
```
HumorBaby: This condition and the next one will both raise `ValueError` if the `rates` key is not defined.

sopel/modules/currency.py
```diff
@@ -2,109 +2,201 @@
...
+    ^(\d+(?:\.\d+)?)                                            # Decimal number
+    \s*([a-zA-Z]{3})                                            # 3-letter currency code
+    \s+(?:in|as|of|to)\s+                                       # preposition
+    (([a-zA-Z]{3}$)|([a-zA-Z]{3})\s)+$                          # one or more 3-letter currency code
+''', re.VERBOSE)
+LOGGER = get_logger(__name__)
+UNSUPPORTED_CURRENCY = "Sorry, {} isn't currently supported."
+UNRECOGNIZED_INPUT = "Sorry, I didn't understand the input."
 
+rates_fiat_json = {}
```
HumorBaby: Initialize with the expected structure to prevent the issues I pointed out (`if of in rates_fiat_json['rates']...`). Just a suggestion:
```suggestion
rates_fiat_json = {'rates':{}}
```

**OR**

Change the conditions below to first check if `'rates'` is in this dictionary... this may be the better option since then you can also detect errors.

Duckle29:
Just a heads up to re-clone my fork if testing it as I squashed a few commits together again.

sopel/modules/currency.py
```diff
@@ -125,14 +128,18 @@ def exchange(bot, match):
...
         bot.reply("Something went wrong while I was getting the exchange rate.")
-        LOGGER.error("Error in GET request: {}".format(e))
+        LOGGER.error("Error in GET request: {}".format(err))
         return NOLIMIT
     except ValueError:
         bot.reply("Error: Got malformed data.")
         LOGGER.error("Invalid json on update_rates")
         return NOLIMIT
+    except APIError as err:
+        bot.reply('Sorry, something went wrong with fixer')
```
dgw: It's not necessarily Fixer, though, is it? Could be exchangeratesapi

Duckle29: Exchangeratesapi doesn't have the success field. Looking through the source it seems there's no special way for it to fail, other than standard http return codes. As such the raise_for_status() will catch it.

As discussed in IRC I'll rename the exception to better reflect that it's Fixer specific.

sopel/modules/currency.py
```diff
@@ -2,109 +2,209 @@
...
+    ^(\d+(?:\.\d+)?)                                            # Decimal number
+    \s*([a-zA-Z]{3})                                            # 3-letter currency code
+    \s+(?:in|as|of|to)\s+                                       # preposition
+    (([a-zA-Z]{3}$)|([a-zA-Z]{3})\s)+$                          # one or more 3-letter currency code
+''', re.VERBOSE)
+LOGGER = get_logger(__name__)
+UNSUPPORTED_CURRENCY = "Sorry, {} isn't currently supported."
+UNRECOGNIZED_INPUT = "Sorry, I didn't understand the input."
 
+rates_fiat_json = {'rates': {}}
```
HumorBaby: Given the way you handle errors not, the `'rates': {}` is not needed now. In fact, leaving this in may end up masking errors that lead to a missing `rates` key. While not likely a breaking point, it may be for the best to make `rates_fiat_json` an empty dict as you had before.

sopel/modules/currency.py
```diff
@@ -2,109 +2,209 @@
...
+    config.define_section('currency', CurrencySection, validate=False)
+    config.currency.configure_setting('fixer_io_key', 'Optional API key for Fixer.io (leave blank to use exchangeratesapi.io):')
+    config.currency.configure_setting('auto_convert', 'Automatically respond to currency conversion lines?')
+
+
+def setup(bot):
+    bot.config.define_section('currency', CurrencySection)
+
+
+class FixerError(Exception):
```
HumorBaby: Since you don't need to reference the status attribute of the exception anywhere else, you can simplify this to:
```
class FixerError(Exception):
    """An API Error Exception"""

    def __init__(self, status):
        super(FixerError, self).__init__(self, "APIError: {}".format(status))
````
which just passes your message to the base `Exception` constructor.

File: sopel/modules/currency.py [hunk 43]
```diff
@@ -2,109 +2,209 @@
...
     try:
         amount = float(amount)
     except ValueError:
-        bot.reply("Sorry, I didn't understand the input.")
+        bot.reply(UNRECOGNIZED_INPUT)
     except OverflowError:
         bot.reply("Sorry, input amount was out of range.")
-    display(bot, amount, of, to)
 
+    out_string = '{} {} is'.format(amount, of.upper())
```
kwaaak: Always printing the input with two decimal places would look better.

dgw: Unless the input currency doesn't use decimals, like JPY.

dgw:
For the record, I'm perfectly happy to merge this as soon as @HumorBaby's small request is taken care of. We can continue tweaking the actual output format (when to use decimals, how many places, etc.) in a future PR, maybe for Sopel 7.1.

sopel/modules/currency.py
```diff
@@ -2,109 +2,215 @@
...
-            return 1 / float(element['FX{}CAD'.format(code)]['v']), name
+class CurrencySection(StaticSection):
+    fixer_io_key = ValidatedAttribute('fixer_io_key', default=None)
+    auto_convert = ValidatedAttribute('auto_convert', parse=bool, default=False)
 
 
-@commands('cur', 'currency', 'exchange')
-@example('.cur 20 EUR in USD')
-def exchange(bot, trigger):
+def configure(config):
```
dgw: I hate that we still have to do this, but ain't no way am I maintaining the website's module config page manually.

```suggestion
def configure(config):
    """
    | name | example | purpose |
    | ---- | ------- | ------- |
    | auto\\_convert | False | Whether to convert currencies without an explicit command |
    | fixer\\_io\\_key | 0123456789abcdef0123456789abcdef | Optional API key for Fixer.io (increases currency support) |
    """
```

Also going to suggest (via IRC) using these as docstrings in the `CurrencySection`.

sopel/modules/currency.py
```diff
@@ -2,109 +2,223 @@
...
+        bot.reply("Zero is zero, no matter what country you're in.")
         return NOLIMIT
 
-    display(bot, amount, 'BTC', to)
+    out_string = '{} {} is'.format(amount, of.upper())
+
+    for to in others:
+        try:
+            out_string = build_reply(amount, of.upper(), to.upper(), out_string)
+        except ValueError:
```
kwaaak: This looks like it is being handled already in line 166

kwaaak:
I've been using it, and it works! :rocket:

PR #1436, opened 2019-01-05
Removing weather module because of deprecated API

RustyBower:
Per #1435, removing this file from sopel core

PR #1438, opened 2019-01-07
weather: Change API to metaweather

kwaaak:


dgw:
Some time ago, I played with this API and determined that it returns WOEIDs for relatively few cities. I tried looking up a number of reasonably large cities that this API can't find. It can't return weather data for all WOEIDs it does find, either. Many WOEIDs I tried (both in the past and just now) return empty JSON.

At some point, with improvements to the API, this might become a viable alternative to #1436, but I don't think it is one at this time.

I'll leave this PR open for reconsideration when Sopel 7 is nearing release. :smile_cat:

sopel/modules/weather.py
```diff
@@ -17,52 +17,50 @@ def woeid_search(query):
     node for the result, so that location data can still be retrieved. Returns
     None if there is no result, or the woeid field is empty.
     """
-    query = 'q=select * from geo.places where text="%s"' % query
-    body = requests.get('https://query.yahooapis.com/v1/public/yql?' + query)
-    parsed = xmltodict.parse(body.text).get('query')
-    results = parsed.get('results')
-    if results is None or results.get('place') is None:
+    query = query.replace(" ", "+")
+    body = requests.get('https://www.metaweather.com/api/location/search/?query=' + query)
```
Exirel: You should let `requests` do its job with the query string:

```
>>> r = requests.get('https://www.metaweather.com/api/location/search/', {'query': 'Los Angeles'})
>>> r.json()
[{'latt_long': '34.053490,-118.245323', 'woeid': 2442047, 'location_type': 'City', 'title': 'Los Angeles'}]
>>> r.url
'https://www.metaweather.com/api/location/search/?query=Los+Angeles'
```

You don't need to escape it yourself, *and* it'll take care itself of any special char.

sopel/modules/weather.py
```diff
@@ -17,52 +17,50 @@ def woeid_search(query):
...
+    results = body.json()
```
Exirel: This might fail if the status is not 200 or if the content is not JSON. You may want to add at least a try/except block around this.

sopel/modules/weather.py
```diff
@@ -161,18 +159,17 @@ def say_info(bot, trigger, mode):
...
 
     if not woeid:
         return bot.reply("I don't know where that is.")
 
-    query = 'q=select * from weather.forecast where woeid="%s" and u=\'c\'' % woeid
-    body = requests.get('https://query.yahooapis.com/v1/public/yql?' + query)
-    parsed = xmltodict.parse(body.text).get('query')
-    results = parsed.get('results')
+    query = woeid
+    body = requests.get('https://www.metaweather.com/api/location/' + str(query))
```
Exirel: I think you should need to encode the `query` as an URL first. I might be wrong, but that seem unsafe as is.

dgw:
It's unlikely that this is going to happen, at this point. The weather module has already been removed from core, and we're trying to move as many core modules out as possible in the coming releases.

Issue #1219, opened 2017-09-07
help command should not include the IRC server address.


antdude:
Even though help command uploaded its results to https://gist.github.com/ web server anonymously, it should not include the IRC server name. There should be an option to include it. Not all IRC servers are meant to be public.

dgw:
@hekel Would you like to work this in alongside updating the help module to use another pastebin service?

dgw:
Cross-reference: #1459

Assigning **7.0.0** milestone to match.

Exirel:
This might need documentation, but it looks fixed in master.

dgw:
Fixed by new option added in #1459. Thanks for the nudge, @Exirel!

PR #1439, opened 2019-01-08
url: fixing private URL lookups

RustyBower:
Prevents private IPs being used for URL lookups. This doesn't work for resolved IPs, but prevents folks from actively using http://192.168.1.1 to access potentially private sites.

### DIFF
--- a/sopel/modules/url.py
+++ b/sopel/modules/url.py
@@ -13,10 +13,18 @@ from __future__ import unicode_literals, absolute_import, print_function, divisi
 
 import re
 
+import dns.resolver
+import ipaddress
 import requests
 
 from sopel import __version__, module, tools, web
 from sopel.config.types import ListAttribute, StaticSection, ValidatedAttribute
+
+# Python3 vs Python2
+try:
+    from urllib.parse import urlparse
+except ImportError:
+    from urlparse import urlparse
 
 USER_AGENT = 'Sopel/{} (https://sopel.chat)'.format(__version__)
 default_headers = {'User-Agent': USER_AGENT}
@@ -42,6 +50,10 @@ class UrlSection(StaticSection):
     shorten_url_length = ValidatedAttribute(
         'shorten_url_length', int, default=0)
     """If greater than 0, the title fetcher will include a TinyURL version of links longer than this many characters."""
+    enable_private_resolution = ValidatedAttribute(
+        'enable_private_resolution', bool, default=False)
+    enable_dns_resolution = ValidatedAttribute(
+        'enable_dns_resolution', bool, default=False)
 
 
 def configure(config):
@@ -66,6 +78,14 @@ def configure(config):
         'Enter how many characters a URL should be before the bot puts a'
         ' shorter version of the URL in the title as a TinyURL link'
         ' (0 to disable)'
+    )
+    config.url.configure_setting(
+        'enable_private_resolution',
+        'Enable URL lookups for RFC1918 addresses?'
+    )
+    config.url.configure_setting(
+        'enable_dns_resolution',
+        'Enable DNS resolution for all domains to validate if there are RFC1918 resolutions?'
     )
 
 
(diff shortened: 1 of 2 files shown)


RustyBower:
This should fix https://github.com/sopel-irc/sopel/issues/1219

RustyBower:
Are you sure? It looks like it was either backported or new features were added, because that module installs on my 2.7 instance.

root@d4aa7d892682:/# pip -V
pip 18.1 from /usr/local/lib/python2.7/site-packages/pip (python 2.7)
root@d4aa7d892682:/# pip search ipaddress
ipaddress (1.0.22)             - IPv4/IPv6 manipulation library
  INSTALLED: 1.0.22 (latest)

I'll triple check with my 2.7 virtualenv, but it _should_ be ok

RustyBower:
I just tested with a Python 2.7 install of Sopel and it seems to be playing nice

/shrug

dgw:
I checked on my own 2.7 install, without touching `pip`, and was greeted with this:

```
$ python -i
Python 2.7.15rc1 (default, Nov 12 2018, 14:31:15) 
[GCC 7.3.0] on linux2
>>> import ipaddress
ImportError: No module named ipaddress
>>> 
```

`ipaddress` was added to the Python standard library in 3.3. It's still technically a new dependency if it requires installing a PyPI package on certain Python versions.

So, if this patch is to be compatible with all the Python releases that Sopel officially supports, it needs to add `ipaddress<2.0; python_version < '3.3'` in `requirements.txt`. (The `<2.0` part is for safety, in case there is eventually a version 2 that happens to modify the API. Unlikely, but still.)

----

Also, it shouldn't be too annoying to look up hostnames with the existing `dnspython` dependency. That's probably not something that should be on by default, as it might cause a lot of extra DNS lookups if Sopel is run in a channel where lots of links get sent, but if you feel like implementing that function and an option to turn it on (either in this or a later PR) I'd be cool with it. Extra cool, if it implements caching!

RustyBower:
Adding in 2 configuration variables:
- enable_private_resolution (default: False): enables the querying of RFC1918 IP addresses (http://192.168.1.1)
- enable_dns_resolution (default: False): enables dns query to determine if dns records contain an RFC1918 address

sopel/modules/url.py
```diff
@@ -54,6 +66,14 @@ def configure(config):
         ' shorter version of the URL in the title as a TinyURL link'
         ' (0 to disable)'
     )
+    config.url.configure_setting(
+        'enable_private_resolution',
+        'Enables URL lookups for RFC1918 addresses'
```
HumorBaby: I've noticed other `boolean` settings in the config wizard phrase these as questions, with an explicit `?` at the end (e.g., `admin.py`). What do you think about changing this prompt (and the next) to something like `Enable URL lookups for RFC1918 address?` (remove extra `s`, and add that `?` in)

dgw:
@RustyBower This needs a rebase (back up your branch first! :stuck_out_tongue_closed_eyes:). Unfortunately I think you'll be fighting with @Exirel and #1510, so you might want to wait for that to get finished (it's a much larger, API-related change, so it has priority).

I would also love it if you reworded the config wizard questions like @HumorBaby suggested! :rocket:

RustyBower:
Should be in a good state now, I'll wait on #1510 and the rebase appropriately

sopel/modules/url.py
```diff
@@ -64,6 +76,14 @@ def configure(config):
         ' shorter version of the URL in the title as a TinyURL link'
         ' (0 to disable)'
     )
+    config.url.configure_setting(
+        'enable_private_resolution',
+        'Enables URL lookups for RFC1918 addresses?'
```
HumorBaby: `'Enable**s** URL lookups for RFC1918 addresses?'` :scream: but dat ess tho!

RustyBower: Lol. Fixed.

Exirel:
PR #1510 is almost ready! Just waiting on @dgw to review/merge and this PR can move forward.

dgw:
@RustyBower Merged #1510 yesterday, so you can come back to this whenever you're ready. :smile_cat:

Exirel:
The merge conflict looks... interesting.

Sorry about that. 😞

sopel/modules/url.py
```diff
@@ -165,6 +185,26 @@ def process_urls(bot, trigger, urls):
     """
     shorten_url_length = bot.config.url.shorten_url_length
     for url in urls:
+        parsed = urlparse(url)
+
+        # Prevent private addresses form being queried if enable_private_resolution is False
+        if not bot.config.url.enable_private_resolution:
```
Exirel: I think this check can be performed *after* the exclusion char check: if the URL is excluded, I don't think we need to check if it's private, as it'll be discarded right away.

Exirel: ... and probably *after* the `check_callback` call too.

RustyBower: Moved the code after check_callback

Exirel:
Yay, conflicting files...

Exirel:
Sensible default values, clean code, useful feature, configuration is simple and optional...

... :rocket: :ship: now!

sopel/modules/url.py
```diff
@@ -173,6 +193,26 @@ def process_urls(bot, trigger, urls):
         if check_callbacks(bot, url):
             continue
 
+        # Prevent private addresses form being queried if enable_private_resolution is False
```
dgw: Oh look, I missed a typo. Ah well.

PR #1440, opened 2019-01-08
Removing weather module

RustyBower:


RustyBower:
Updated the PR description to be in line with style :)

dgw:
Nice, but I meant the actual commit message. Sorry if I wasn't clear earlier; was on my phone! :)

PR #1441, opened 2019-01-09
movie: removing broken module

### LINKED ISSUE #1198: imdb module not working anymore
I just noticed that anytime you make a call to imdb now the bot responds: 

> [MOVIE] No API key provided.

I know it used to work, not sure how recently. Maybe it can be switched to a different database that doesn't require an API key?

RustyBower:
movie: removing since no keyless API exists

standalone module can be found https://pypi.org/project/sopel-modules.imdb/

This should fix #1198


dgw:
Two comments:

  1. The commit message would be better if it mentioned the reason for the removal (same as #1440). Even just `movie: remove broken module (see #1198)`.
  2. Removing the `movie.py` file leaves a dangling reference in `conftest.py` that should be removed, too:
     https://github.com/sopel-irc/sopel/blob/6a675ac8178d58c2d7083d6c32c0454acec8c89f/conftest.py#L1-L2

dgw:
@RustyBower Just to make sure you're aware, the current release of your replacement on PyPI still lists `enchant` as a requirement in the readme/description. :)

RustyBower:
Ack. I'll get that fixed :D

PR #1442, opened 2019-01-09
wiktionary: fix query logic

### LINKED ISSUE #1214: [wiktionary] Relax case sensitivity a bit?
Looking up anything other than the exact capitalization in Wiktionary's database appears to fail. It might be worthwhile, for user experience, to always try the all-lowercase version of a query if no results are returned for it as entered.

That said, the MediaWiki API would allow querying automatically for several variants (using `action=query&titles=variant1|Variant2|etc.`) and then fetching the first one that exists for display, if making multiple requests is OK in the plugin.

RustyBower:
Looking at the wiktionary results so far, it appears that all (or nearly all) of the words are lowercase when querying. I believe it is safe to cast our query to lower().

Should fix #1214

### DIFF
--- a/sopel/modules/wiktionary.py
+++ b/sopel/modules/wiktionary.py
@@ -99,8 +99,11 @@ def wiktionary(bot, trigger):
 
     _etymology, definitions = wikt(word)
     if not definitions:
-        bot.say("Couldn't get any definitions for %s." % word)
-        return
+        # Cast word to lower to check in case of mismatched user input
+        _etymology, definitions = wikt(word.lower())
+        if not definitions:
+            bot.say("Couldn't get any definitions for %s." % word)
+            return
 
     result = format(word, definitions)
     if len(result) < 150:


HumorBaby:
Just chiming in because I'm working on updating the wiktionary module to: 1) parse all parts of speech specifiec by wiktionary, and 2) manage sub-defintions; see #1405.

Casting everything to lower by default will break queries like `Paul` vs `paul`, where `Paul` should return the Proper noun defintion, and `paul` should return the noun defintion.

I don't imagine there being too many searches for Proper nouns, but I'd hate to break such a function on purpose :hushed:.. Thoughts?

RustyBower:
I'm testing on master and `.define Paul` isn't returning results, so I'm not entirely convinced that this was ever working...

RustyBower:
I did clean up the logic so it'll only .lower() if the first definition fails to find anything

HumorBaby:
Yeah, you're right! I realize now that Proper nouns are not parsed in the current version of wiktionary. It was part of an idea that had come up in #1405, and is implemented in #1443.

`wiktionary.py` as is, however, would likely benefit from casting to lower as you mentioned :smile: I'm all on-board with the fallback implementation you've added as the ideal middle-ground

sopel/modules/wiktionary.py
```diff
@@ -32,6 +32,7 @@ def text(html):
 
 def wikt(word):
     bytes = requests.get(uri % web.quote(word)).text
+    print(bytes)
```
HumorBaby: Ooops! Want to leave in this print?

RustyBower: Nope! Fixed :D

dgw:
Since commit e2f0104 is entirely reverted by the second commit's change of approach, @RustyBower wanna squash these?

RustyBower:
All squashed

dgw:
Right, so Codacy has been stuck on this all day and I've given up. It should be turned off for future pull requests now. It gets stuck too much.

PR #1443, opened 2019-01-10
wiktionary: refactor/bugfix parsing all possible parts of speech

HumorBaby:
The _refactor_ comes from changing the control flow, and the _bugfix_ is the ability of the module to parse more parts of speech. While this may not be a _bug_ per se, I'd think not being able to get definitions for a word/term that is actually in Wiktionary is not a preferred behavior (pardon the double negative :sweat_smile:).

Regardless, this PR is opened to address the points that came up in #1405:
- [x] parse all parts of speech
- ~~give subdefinitions like those for [`-able`](https://en.wiktionary.org/wiki/-able) their own sub-number (e.g., 1a., 1b., ...)~~

- ~~update formatting algo to be smarter about sub-definitions and such~~
- ~~re-order parts of speech list so that definitions can be prioritized... it would be a shame if some obscure _circumposition_ (A circumposition is a discontinuous adposition... _um wat?_) definition led to the more common _verb_ definition being truncated.~~

I'll put the commit messages as comments below to get a prettified output of what was done (and since I'll probably overwrite/summarize them in a squash, if/when it is time)

**EDIT**: The removed (strikethrough) enhancements will be moved to another issue.

### DIFF
--- a/sopel/modules/wiktionary.py
+++ b/sopel/modules/wiktionary.py
@@ -18,6 +18,25 @@ r_sup = re.compile(r'<sup[^>]+>.+</sup>')  # Superscripts that are references on
 r_tag = re.compile(r'<[^>]+>')
 r_ul = re.compile(r'(?ims)<ul>.*?</ul>')
 
+# From https://en.wiktionary.org/wiki/Wiktionary:Entry_layout#Part_of_speech
+PARTS_OF_SPEECH = [
+    # Parts of speech
+    'Adjective', 'Adverb', 'Ambiposition', 'Article', 'Circumposition',
+    'Classifier', 'Conjunction', 'Contraction', 'Counter', 'Determiner',
+    'Ideophone', 'Interjection', 'Noun', 'Numeral', 'Participle', 'Particle',
+    'Postposition', 'Preposition', 'Pronoun', 'Proper noun', 'Verb',
+    # Morphemes
+    'Circumfix', 'Combining form', 'Infix', 'Interfix', 'Prefix', 'Root', 'Suffix',
+    # Symbols and characters
+    'Diacritical mark', 'Letter', 'Ligature', 'Number', 'Punctuation mark', 'Syllable', 'Symbol',
+    # Phrases
+    'Phrase', 'Proverb', 'Prepositional phrase',
+    # Han characters and language-specific varieties
+    'Han character', 'Hanzi', 'Kanji', 'Hanja',
+    # Other
+    'Romanization',
+]
+
 
 def text(html):
     text = r_sup.sub('', html)  # Remove superscripts that are references from definition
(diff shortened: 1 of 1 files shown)


HumorBaby:
### refactor(wiktionary): improve control flow
Move all parts of speech to constant to make iteration easier; this helps avoid the messy `if-elif-else` chain that results when parsing for defintions, and makes it easier to change acceptable parts of speech if/when Wiktionary changes specifications. 

 _Note: not all parts of speech are actually parts of speech, but they are listed in the specification under the "Parts of speech" heading, so just easier to call them that. See the comments in the_ `PARTS_OF_SPEECH` _constant for details._ 

 For now, this refactor does not break any old features, but also does not yet correctly parse the newly added parts of speech (e.g., phrases, proper nouns).

HumorBaby:
### fix(wiktionary): parse definitions for all parts of speech
With this commit, the wiktionary module now provides defintions for _almost_ all "parts of speech" listed on the [Wiktionary specifications page](https://en.wiktionary.org/wiki/Wiktionary:Entry_layout#Part_of_speech).

I say _almost_ and quote "parts of speech" above because things that are not really parts of speech do not work well, e.g., the "Han character" type, which is likely due to a unicode translation issue that could happen anywhere from the client to the web request; have not specifically pinned it down. Symbols like Δ and ⰴ, and diacritical marks like ῀ work, however (at least from weechat).

Fun new definitions for things like "Prepositional phrases", "Proper nouns", "Articles", and more are now provided. Try some: 
```
<HumorBaby> .wt Paul
<ro-bot-o> Paul — proper noun: 1. In the New Testament, Apostle to the Gentiles and author of fourteen epistles, 2. A male given name of biblical origin, 3. A patronymic surname​

<HumorBaby> .wt out of hand
<ro-bot-o> out of hand — adjective: 1. Not under control — adverb: 1. Without (further) thought or consideration, 2. (now  rare) Immediately, forthwith, or incontinently

<HumorBaby>.wt on the other hand
<ro-bot-o> on the other hand — prepositional phrase: 1. (sequence, idiomatic) From another point of view

<HumorBaby> .wt in time
<ro-bot-o> in time — prepositional phrase: 1. At or before the time assigned, 2. (with for) Sufficiently early, 3. As time passes, 4. In rhythm, 5. (in time with) At the same rhythm as

<HumorBaby> .wt the
<ro-bot-o> the — adverb: 1. With a comparative or more and a verb phrase, establishes a parallel with one or more other such comparatives, 2. With a comparative, and often with for it, indicates a result more like said comparative. This can be negated with none — article: 1. Definite grammatical article t[...]

<HumorBaby> .wt A
<ro-bot-o> A — letter: 1. The first letter of the basic modern Latin alphabet — symbol: 1. Marks the first item in a list, 2. A hypothetical item or person designated the first, usually when there are more than one

<HumorBaby> .wt ⰴ
 <ro-bot-o> ⰴ — letter: 1. The fifth letter of the Glagolitic alphabet

<HumorBaby> .wt !
<ro-bot-o> ! — punctuation mark: 1. Denoting excitement, surprise, or shock, 2. Expresses excitement, anger or other emotion — symbol: 1. (mathematics) The factorial: the result of a multiplication of all integers from 1 to the given number, 2. (chess) Indicates a good move
```

dgw:
@HumorBaby This looks WIP, 'cause of the task list showing "1 of 4" in the PR list. Is it ready or did you intend to add more stuff before review?

HumorBaby:
This final result of this PR is to address an issue related to #1214, where definitions were not being returned unless the exact case was provided. While the relaxation of the case requirement has been addressed in #1442, the case where a word has different cases for different definitions was not directly addressed. 

One such example is `Paul` vs `paul`. One is a proper noun, while the other is a common noun. Before this PR, only the definition of `paul` would be returned despite a query like `.wt Paul` (which before #1442 would have failed outright). Despite having a definition in Wiktionary, `Paul` would never return the desired definition because all the cataloged parts of speech were not parsed (proper nouns, being one). Now, all parts of speech tracked by Wiktionary are parsed (which not only solves the issue but has the side-effect of providing definitions for a whole new set of queries!)

So, how about a review @sopel-irc/rockstars??

HumorBaby:
Boy, I can't spell ~~defintions~~ definitions at all... fixed in _both_ commit messages.

Exirel:
@dgw I'm shamelessly invoking you. No reason. At all. Nah just kidding: as I'm scanning through oldest PR available for 7.x.

You can probably merge that "tomorrow" for your github streak. :wink:

PR #1434, opened 2019-01-03
cli: sopel-module

Exirel:
**Disclaimer**:  this is an experimental command line tool. I suggest to merge it as-is, and to iterate in further PR if we need to. I've been working on that for few days, and I think now it is in a good enough shape to work properly.

---

So last day I was working on #1424 and #1429 and I was wondering: how can I get the list of available/enabled modules for Sopel? So I look up the CLI (`sopel` or `python sopel.py`) with no result.

So... maybe I could do something about that? After all, there is an old issue about module enabled/excluded (#244), and maybe I can suggest a solution.

And here it is: a Pull Request that adds a new command line interface: `sopel-module`.

---

## Entry point

At the moment, I added a `sopel-module` entry point in the `setup.py` configuration.
Maybe later we'll need to add a `.py` file at the root of this repository, like there is to launch `python sopel.py` from the source.

That'll be for later.

## Code organization

I don't like having a `sopel.run_script` module. I think, if we want to add multiple CLI for sopel, we should have a proper sub-package for that.

So I put all my code into `sopel.cli`, and the function used as entry point is `main`.

I think this structure could be use for the `run_script` code too, but I don't want to change that in *that* PR: it'll be for another one, once this is merged.

## CLI

To test it from the source:

    python setup.py develop

This will install both entry point `sopel` and `sopel-module` in your virtualenv, and then you can do:

    sopel-module list
    sopel-module show xkcd
    sopel-module disable xkcd
    sopel-module enable xkcd

Good luck with the review!

### list

The command `sopel-module list` enumerate modules, as Sopel does when looking for modules.

```
usage: sopel-module list [-h] [-c filename] [-p] [-t] [-a | -e]

List availables sopel modules

optional arguments:
  -h, --help            show this help message and exit
  -c filename, --config filename
                        Use a specific configuration file
  -p, --path            Show the path to the module file
  -t, --type            Show the type to the module file: `p` for package
                        directory, `m` for module file, `?` for unknown
  -a, --all             Show all available module, enabled or not
  -e, --excluded        Show only excluded module
```

### show

The `sopel-module show <module>` command display various information about the module (if found): it tries to list commands, rules pattern, events and intents, that will be loaded by the module.

```
usage: sopel-module show [-h] [-c filename] module

Show a sopel module's details

positional arguments:
  module

optional arguments:
  -h, --help            show this help message and exit
  -c filename, --config filename
                        Use a specific configuration file
```

### enable

The `sopel-module enable <module>` tries to enable a module:

[...]

PR #1445, opened 2019-01-10
cli: use ENV variable to load configuration

Exirel:
Remember #1434 I did to work on a new cli interface for Sopel? Well, it gets better and better!

**TL;DR**: I use `SOPEL_CONFIG` env var to load the configuration, using a `sopel.cli.utils` module

## The core of this PR

The core of this PR is to add a `SOPEL_CONFIG` env var to load configuration.

What happens is that:

* first, we try to get the config from the command line argument, like `-c filename`
* then, if the `-c` is not used, we try to look at `SOPEL_CONFIG` env var,
* if nothing is set, we fall back on the `default` config file,

If the `-c` option is used or the `SOPEL_CONFIG` is defined, then it tries to load it, and it won't fall back to the default, ie. if the user decided to use a wrong config file, then Sopel will warn them.

## About the state of the project

To really understand what's going on here, you need to have a look at #1434 because I based this branch from it. Then, here are the good bits of this PR:

* In 614a730 I added `sopel.cli.utils` to store cli-related tools,
* In 3324ca3 I introduce the usage of a `SOPEL_CONFIG` env var,
* In eea0863 I use the 2 previous commit's work to improve the `sopel.run_script` module.

Note: both #1434 and this PR should be rebased after #1429 is merged into master, because the clean-up and the unit-test are very valuable!

**The end goal is to move** both `sopel` and `sopel-module` command line `main` functions into `sopel.cli`.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -72,5 +72,10 @@ setup(
     platforms='Linux x86, x86-64',
     install_requires=requires,
     extras_require={'dev': dev_requires},
-    entry_points={'console_scripts': ['sopel = sopel.run_script:main']},
+    entry_points={
+        'console_scripts': [
+            'sopel = sopel.run_script:main',
+            'sopel-module = sopel.cli.modules:main',
+        ],
+    },
 )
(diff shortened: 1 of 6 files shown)


Exirel:
For the same reason #1434 is close for now, I close this one. Let's rework things first, then we'll be able to improve the situation here.

Exirel:
Replaced by #1473

PR #1446, opened 2019-01-13
db: adding multi database support

### LINKED ISSUE #736: OperationalError: database is locked
This started popping up today, the bot is the only thing using that database!

```
  File "/home/TrollOP/willie/willie/bot.py", line 634, in call
    exit_code = func(willie, trigger)
  File "/home/TrollOP/willie/willie/modules/seen.py", line 50, in note
    bot.db.set_nick_value(trigger.nick, 'seen_timestamp', time.time())
  File "/home/TrollOP/willie/willie/db.py", line 150, in set_nick_value
    nick_id = self.get_nick_id(nick)
  File "/home/TrollOP/willie/willie/db.py", line 114, in get_nick_id
    [slug]).fetchone()
  File "/home/TrollOP/willie/willie/db.py", line 64, in execute
    return cur.execute(*args, **kwargs)
OperationalError: database is locked

  File "/home/TrollOP/willie/willie/bot.py", line 634, in call
    exit_code = func(willie, trigger)
  File "/home/TrollOP/willie/willie/modules/weather.py", line 129, in weather
    woeid = bot.db.get_nick_value(trigger.nick, 'woeid')
  File "/home/TrollOP/willie/willie/db.py", line 161, in get_nick_value
    [nick.lower(), key]
  File "/home/TrollOP/willie/willie/db.py", line 64, in execute
    return cur.execute(*args, **kwargs)
OperationalError: database is locked

  File "/home/TrollOP/willie/willie/bot.py", line 634, in call
    exit_code = func(willie, trigger)
  File "/home/TrollOP/willie/willie/modules/seen.py", line 50, in note
    bot.db.set_nick_value(trigger.nick, 'seen_timestamp', time.time())
  File "/home/TrollOP/willie/willie/db.py", line 152, in set_nick_value
    [nick_id, key, value])
  File "/home/TrollOP/willie/willie/db.py", line 64, in execute
    return cur.execute(*args, **kwargs)
OperationalError: database is locked
```

RustyBower:
This should fix #736.

This PR migrates from direct SQLite SQL statements to using the SQLAlchemy ORM. This will allow users to choose whether they want to stick with SQLite (Default), or leverage another database engine.

I've added a few more variables to core to handle non-SQLite databases: db_type, db_user, db_pass, db_host, and db_name. If db_type is not specified, or is specified with sqlite, the other values do not need to be specified. If db_type is specified with MySQL, users will need to specify the other parameters.

We could probably move the Models into their own model.py to clean things up a bit, but I will leave that to your discretion @dgw

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -44,6 +44,16 @@ In the source directory (whether cloned or from the tarball) run
 bot. Alternately, you can just run the ``sopel.py`` file in the source
 directory.
 
+Database Support
+----------------
+Sopel leverages SQLAlchemy to support the following database types: SQLite,
+MySQL, PostgreSQL, MSSQL, Oracle, Firebird, and Sybase. By default Sopel will
+use a SQLite database in the current configuration directory, but alternative
+databases can be configured with the following config options: ``db_type``,
+``db_filename`` (SQLite only), ``db_driver``, ``db_user``, ``db_pass``,
+``db_host``, ``db_port``, and ``db_name``. You will need to manually install
+any packages (system or ``pip``) needed to make your chosen database work.
+
 Adding modules
 --------------
 The easiest place to put new modules is in ``~/.sopel/modules``. Some newer
--- a/sopel/config/core_section.py
+++ b/sopel/config/core_section.py
@@ -108,8 +108,38 @@ class CoreSection(StaticSection):
     channels = ListAttribute('channels')
     """List of channels for the bot to join when it connects"""
 
+    db_type = ChoiceAttribute('db_type', choices=[
+        'sqlite', 'mysql', 'postgres', 'mssql', 'oracle', 'firebird', 'sybase'], default='sqlite')
+    """The type of database to use for Sopel's database.
+
+    mysql - pip install mysql-python (Python 2) or pip install mysqlclient (Python 3)
+    postgres - pip install psycopg2
+    mssql - pip install pymssql
+
+    See https://docs.sqlalchemy.org/en/latest/dialects/ for a full list of dialects"""
+
     db_filename = ValidatedAttribute('db_filename')
-    """The filename for Sopel's database."""
+    """The filename for Sopel's database. (SQLite only)"""
+
+    db_driver = ValidatedAttribute('db_driver')
+    """The driver for Sopel's database.
+    This is optional, but can be specified if user wants to use a different driver
+    https://docs.sqlalchemy.org/en/latest/core/engines.html"""
+
+    db_user = ValidatedAttribute('db_user')
+    """The user for Sopel's database."""
+
+    db_pass = ValidatedAttribute('db_pass')
+    """The password for Sopel's database."""
+
+    db_host = ValidatedAttribute('db_host')
+    """The host for Sopel's database."""
+
+    db_port = ValidatedAttribute('db_port')
+    """The port for Sopel's database."""
+
+    db_name = ValidatedAttribute('db_name')
+    """The name of Sopel's database."""
 
     default_time_format = ValidatedAttribute('default_time_format',
                                              default='%Y-%m-%d - %T%Z')
--- a/sopel/db.py
+++ b/sopel/db.py
@@ -4,9 +4,14 @@ from __future__ import unicode_literals, absolute_import, print_function, divisi
 import json
 import os.path
 import sys
-import sqlite3
 
 from sopel.tools import Identifier
+
+from sqlalchemy import create_engine, Column, ForeignKey, Integer, String
+from sqlalchemy.engine.url import URL
+from sqlalchemy.exc import OperationalError, SQLAlchemyError
+from sqlalchemy.ext.declarative import declarative_base
+from sqlalchemy.orm import scoped_session, sessionmaker
 
 if sys.version_info.major >= 3:
     unicode = str
@@ -28,6 +33,47 @@ def _deserialize(value):
     return value
 
 
+BASE = declarative_base()
+
+
+class NickIDs(BASE):
+    """
+    NickIDs SQLAlchemy Class
+    """
+    __tablename__ = 'nick_ids'
+    nick_id = Column(Integer, primary_key=True)
+
+
+class Nicknames(BASE):
+    """
+    Nicknames SQLAlchemy Class
+    """
+    __tablename__ = 'nicknames'
+    nick_id = Column(Integer, ForeignKey('nick_ids.nick_id'), primary_key=True)
+    slug = Column(String, primary_key=True)
+    canonical = Column(String)
+
+
+class NickValues(BASE):
+    """
+    NickValues SQLAlchemy Class
+    """
+    __tablename__ = 'nick_values'
+    nick_id = Column(Integer, ForeignKey('nick_ids.nick_id'), primary_key=True)
+    key = Column(String(255), primary_key=True)
+    value = Column(String(255))
+
+
+class ChannelValues(BASE):
+    """
+    ChannelValues SQLAlchemy Class
+    """
+    __tablename__ = 'channel_values'
+    channel = Column(String(255), primary_key=True)
+    key = Column(String(255), primary_key=True)
+    value = Column(String(255))
+
+
 class SopelDB(object):
     """*Availability: 5.0+*
 
@@ -39,20 +85,70 @@ class SopelDB(object):
     directory as the config."""
 
     def __init__(self, config):
-        path = config.core.db_filename
-        config_dir, config_file = os.path.split(config.filename)
-        config_name, _ = os.path.splitext(config_file)
-        if path is None:
-            path = os.path.join(config_dir, config_name + '.db')
-        path = os.path.expanduser(path)
-        if not os.path.isabs(path):
-            path = os.path.normpath(os.path.join(config_dir, path))
-        self.filename = path
-        self._create()
+        # MySQL - mysql://username:password@localhost/db
+        # SQLite - sqlite:////home/sopel/.sopel/default.db
+        db_type = config.core.db_type
+
+        # Handle SQLite explicitly as a default
+        if db_type == 'sqlite':
+            path = config.core.db_filename
+            config_dir, config_file = os.path.split(config.filename)
+            config_name, _ = os.path.splitext(config_file)
+            if path is None:
+                path = os.path.join(config_dir, config_name + '.db')
+            path = os.path.expanduser(path)
+            if not os.path.isabs(path):
+                path = os.path.normpath(os.path.join(config_dir, path))
+            self.filename = path
+            self.url = 'sqlite:///%s' % path
+        # Otherwise, handle all other database engines
+        else:
+            if db_type == 'mysql':
+                drivername = config.core.db_driver or 'mysql'
+            elif db_type == 'postgres':
+                drivername = config.core.db_driver or 'postgresql'
+            elif db_type == 'oracle':
+                drivername = config.core.db_driver or 'oracle'
+            elif db_type == 'mssql':
+                drivername = config.core.db_driver or 'mssql+pymssql'
+            elif db_type == 'firebird':
+                drivername = config.core.db_driver or 'firebird+fdb'
+            elif db_type == 'sybase':
+                drivername = config.core.db_driver or 'sybase+pysybase'
+            else:
+                raise Exception('Unknown db_type')
+
+            db_user = config.core.db_user
+            db_pass = config.core.db_pass
+            db_host = config.core.db_host
+            db_port = config.core.db_port  # Optional
+            db_name = config.core.db_name  # Optional, depending on DB
+
+            # Ensure we have all our variables defined
+            if db_user is None or db_pass is None or db_host is None:
+                raise Exception('Please make sure the following core '
+                                'configuration values are defined: '
+                                'db_user, db_pass, db_host')
+            self.url = URL(drivername=drivername, username=db_user, password=db_pass,
+                           host=db_host, port=db_port, database=db_name)
+
+        self.engine = create_engine(self.url)
+
+        # Catch any errors connecting to database
+        try:
+            self.engine.connect()
+        except OperationalError:
+            print("OperationalError: Unable to connect to database.")
+            raise
+
+        # Create our tables
+        BASE.metadata.create_all(self.engine)
+
+        self.ssession = scoped_session(sessionmaker(bind=self.engine))
 
     def connect(self):
         """Return a raw database connection object."""
-        return sqlite3.connect(self.filename, timeout=10)
+        return self.engine.connect()
 
     def execute(self, *args, **kwargs):
         """Execute an arbitrary SQL query against the database.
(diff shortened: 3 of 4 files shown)


sopel/db.py
```diff
@@ -39,51 +84,65 @@ class SopelDB(object):
...
-        path = os.path.expanduser(path)
-        if not os.path.isabs(path):
-            path = os.path.normpath(os.path.join(config_dir, path))
-        self.filename = path
-        self._create()
+        # MySQL - mysql://username:password@localhost/db
+        # SQLite - sqlite:////home/sopel/.sopel/default.db
+        db_type = config.core.db_type
+
+        if db_type and db_type == 'mysql':
```
cerealcable: This ties the database to mysql and sqlite only. Can we first check for sqlite and otherwise build the connection string from the other args?

Also, shouldn't db_type be required? This should never be None in my opinion. Does the configuration support required fields?
```python
# some imports first, obviously not intended for literally right here...
from sqlalchemy.engine.url import URL

if not db_type:
    raise Exception("Please make sure core.db_type is defined")
if db_type == 'sqlite':
    # sqlite section here...
    pass
else:
    db_user = config.core.db_user
    db_pass = config.core.db_pass
    db_host = config.core.db_host
    db_port = config.core.db_port  # this is missing, this should totes exist, and a default None value is fine
    db_database = config.core.db_database
    db_query = config.core.db_query  # this is an optional set of options to be passed, probably not important enough)

    self.engine = create_engine(URL(db_type, db_user, db_pass, db_host, db_port, db_database, db_query))
```

sopel/db.py
```diff
@@ -39,51 +84,65 @@ class SopelDB(object):
...
         """Execute an arbitrary SQL query against the database.
 
         Returns a cursor object, on which things like `.fetchall()` can be
         called per PEP 249."""
         with self.connect() as conn:
-            cur = conn.cursor()
-            return cur.execute(*args, **kwargs)
+            return conn.execute(*args, **kwargs)
+            #cur = conn.cursor()
+            #return cur.execute(*args, **kwargs)
```
cerealcable: Should remove these two lines, they're no longer relevant.

sopel/config/core_section.py
```diff
@@ -96,8 +96,23 @@ class CoreSection(StaticSection):
     channels = ListAttribute('channels')
     """List of channels for the bot to join when it connects"""
 
+    db_type = ValidatedAttribute('db_type')
+    """The type of database to use for Sopel's database. (SQLite or MySQL)"""
```
cerealcable: Can we simply refer to sqlalchemy docs instead of limiting choices?

sopel/db.py
dgw: Configuration settings that have no default set will throw an error when code tries to use them:

https://github.com/sopel-irc/sopel/blob/8a6df765072e44ecbfb9acdcc3275fcd2cd4284f/sopel/config/types.py#L145-L151

sopel/config/core_section.py
cerealcable: That's probably going to be infrequent though and a minor PR to address, that seems like a reasonable choice. Downside to this is that it limits usability in non-standard databases. There are some that ship supported in SQLAlchemy and others that are through external packages.

The docs for what Dialects (databases) SQLAlchemy supports (as in listed on their website): https://docs.sqlalchemy.org/en/latest/dialects/index.html

After thinking about it, given that SQLAlchemy could have any database plugin written and then consumed since anyone could write a new Dialect for SQLAlchemy, it'd probably be best to not limit this by using a ChoiceAttribute. I think warning when a connection couldn't be made instead is best.

sopel/db.py
cerealcable: > `db_type` should have `default='sqlite'` in the config class. There's no reason to do a hard require if a sensible default will do (and SQLite will remain the default engine, so this is fine).

And a defaulted filename? This would just get dropped/ignored with using anything other than sqlite.

dgw: I haven't looked through this code yet (in the middle of a busy few days), but the current state is a default filename based on the config name if not set:

https://github.com/sopel-irc/sopel/blob/a0c567c09083414d20ad4f989012f80062d6d2fa/sopel/db.py#L42-L46

sopel/config/core_section.py
RustyBower: I've made this a ChoiceAttribute for now. It looks like there aren't _that_ many dialects to deal with (7), and honestly people only probably use 4 of those (sqlite, mysql, postgres, and mssql).

The biggest question is how we are going to handle dependencies for these, since mysql requires pymysql, etc. Is that something that should be documented or should the PyMySQL dependency be included in requirements (which would also need to include the dependencies for _all_ other database dialects, and that feels bad)

dgw: Dependencies for the various dialects should be left to documentation, IMO. Adding them all to requirements would partially defeat the purpose of making the backend a choice, because the point is to use SQLite (from Python's standard library) unless the user specifically sets it up otherwise. And to do that, they should need to read some documentation. :smile_cat:

RustyBower:
Just finished testing. MSSQL, MySQL, Postgres, and SQLite all work.

I am unable to test Oracle because apparently you need to pay for a license. I haven't tested the last 2 (firebird or sybase).

sopel/db.py
```diff
@@ -160,18 +274,40 @@ def unalias_nick(self, alias):
...
         nick_id = self.get_nick_id(nick, False)
-        self.execute('DELETE FROM nicknames WHERE nick_id = ?', [nick_id])
-        self.execute('DELETE FROM nick_values WHERE nick_id = ?', [nick_id])
+        session = self.ssession()
+        try:
+            count = session.query(Nicknames) \
+                .filter(Nicknames.nick_id == nick_id) \
+                .count()
+            if count <= 1:
+                raise ValueError('Given alias is the only entry in its group.')
```
dgw: A nick group is allowed to have only one member/alias. This would prevent deleting ungrouped nicks.

sopel/config/core_section.py
```diff
@@ -96,8 +96,38 @@ class CoreSection(StaticSection):
...
+    """The password for Sopel's database."""
+
+    db_host = ValidatedAttribute('db_host')
+    """The host for Sopel's database."""
+
+    db_port = ValidatedAttribute('db_port')
+    """The port for Sopel's database."""
+
+    db_database = ValidatedAttribute('db_database')
+    """The database for Sopel's database."""
```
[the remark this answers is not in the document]
RustyBower: Thoughts on changing it to db_type? Which is more in line with what we say in db.py

sopel/db.py
RustyBower: Are you sure? If you look at test_db.py#178, there's multiple associated with a group.

RustyBower: I might be misunderstanding what you're getting at, I tried to reimplement the exact logic as before

sopel/config/core_section.py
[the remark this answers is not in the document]
dgw: `db_type` is already used above, to select between sqlite/mysql/pgsql/etc. This value is the "database name" on mysql.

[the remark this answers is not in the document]
RustyBower: Ah! Misread which section you were referring to when I looked at the code. Fixed!

sopel/db.py
dgw: What was here before included no such check for the group size. It looks like you copied a chunk of `unalias_nick()`, which _does_ care about whether the group will be empty, into here.

dgw:
@RustyBower I think you messed up this PR the other day, like you did with the reddit plugin one. Review conversations that were marked as resolved are missing their resolutions in the current PR state.

Looks like 2c0b884 was the last good state.

dgw:
@RustyBower Still broken compared to before. A lot of issues that were fixed as of 2c0b884 are present again in the current state, because some commits are missing.

Merging upstream into a feature branch is a bad idea. I suggest doing a `git reset --hard 2c0b884` and rebasing that onto `master`, fixing the conflicts manually. (Something I can obviously do myself, but if you learn about rebasing yourself it's a helpful skill to have. :smile_cat:)

RustyBower:
This should be all set. I just triple checked with various databases, and it all works 👍

sopel/db.py
```diff
@@ -124,33 +204,67 @@ def alias_nick(self, nick, alias):
...
-        sql = 'INSERT INTO nicknames (nick_id, slug, canonical) VALUES (?, ?, ?)'
-        values = [nick_id, alias.lower(), alias]
+        session = self.ssession()
         try:
-            self.execute(sql, values)
-        except sqlite3.IntegrityError:
-            raise ValueError('Alias already exists.')
+            result = session.query(Nicknames) \
+                .filter(Nicknames.slug == alias.lower()) \
+                .filter(Nicknames.canonical == alias) \
```
Exirel: For this case, I prefer to do that:

```
session.query(Nicknames).filter(
    Nicknames.slug == alias.lower(),
    Nicknames.canonical == alias
).one_or_none()
```

which prevents the usage of `\`. I feel like it's more readable, but I didn't include that in my review, because I don't want to request a change I'm not sure is in-line with what @dgw think about it.

dgw: @dgw thinks using `\` is kludgy and we should avoid it where possible. :smile_cat:

Exirel:
@RustyBower almost done! There is this generic Exception raised that I would like to see replaced by a more-specific one.

sopel/db.py
```diff
@@ -39,51 +77,78 @@ class SopelDB(object):
...
+
+            db_user = config.core.db_user
+            db_pass = config.core.db_pass
+            db_host = config.core.db_host
+            db_port = config.core.db_port  # Optional
+            db_name = config.core.db_name  # Optional, depending on DB
+
+            # Ensure we have all our variables defined
+            if db_user is None or db_pass is None or db_host is None:
+                raise sopel.config.ConfigurationError('Please make sure the following core '
```
Exirel: I think you forgot to import `sopel.config` first.

sopel/db.py
```diff
@@ -39,51 +77,78 @@ class SopelDB(object):
...
+            elif db_type == 'oracle':
+                drivername = config.core.db_driver or 'oracle'
+            elif db_type == 'mssql':
+                drivername = config.core.db_driver or 'mssql+pymssql'
+            elif db_type == 'firebird':
+                drivername = config.core.db_driver or 'firebird+fdb'
+            elif db_type == 'sybase':
+                drivername = config.core.db_driver or 'sybase+pysybase'
+            else:
+                raise sopel.config.ConfigurationError('Unknown db_type')
```
Exirel: I think you forgot to import `sopel.config` first.

dgw:
Tsk tsk, @Exirel, approving a PR with failed checks:

```
$ ./checkstyle.sh
./sopel/db.py:8:1: F401 'sopel.config' imported but unused
./sopel/db.py:79:5: F811 redefinition of unused 'config' from line 8
```

Exirel:
> Tsk tsk, @Exirel, approving a PR with failed checks:

Well, I approuve, as long as everything's green. 😛

RustyBower:
Squashed all the commits, should be ready for the big show now

dgw:
> Well, I approuve, as long as everything's green. 😛

Too bad everything's not green. There's an undefined name, that `DatabaseConfigurationError` you wanted so much… (Not that I disagree about making the exception specific!)

sopel/db.py
```diff
@@ -4,9 +4,14 @@
 import json
 import os.path
 import sys
-import sqlite3
 
-from sopel.tools import Identifier
+from sopel import tools
```
dgw: Changing this import kind of cluttered the diff unnecessarily, since nothing else from `sopel.tools` is used in this file.

`Identifier()` as used throughout this module is basically type-casting. None of Python's built-in casting operations contain a dot (think `str()`, `int()`, `list()`, etc.), and I don't like putting one in ours.

sopel/config/core_section.py
```diff
@@ -96,8 +96,41 @@ class CoreSection(StaticSection):
     channels = ListAttribute('channels')
     """List of channels for the bot to join when it connects"""
 
+    db_type = ChoiceAttribute('db_type', choices=[
+        'sqlite', 'mysql', 'postgres', 'mssql', 'oracle', 'firebird', 'sybase'], default='sqlite')
+    """The type of database to use for Sopel's database.
+
+    mysql - pip install mysql-python (Python 2) or pip install mysqlclient (Python 3)
+    postgres - pip install psycopg2
+    mssql - pip install pymssql
```
dgw: Why not add the required additional package(s) for each dialect in this `ChoiceAttribute` as [setuptools extras](https://setuptools.readthedocs.io/en/latest/setuptools.html#declaring-extras-optional-features-with-their-own-dependencies) (each extra named for its corresponding choice here)?

Can be done in a separate PR, though. Just wanted to put the bug in your ear!

sopel/config/core_section.py
```diff
@@ -96,8 +96,41 @@ class CoreSection(StaticSection):
     channels = ListAttribute('channels')
     """List of channels for the bot to join when it connects"""
 
+    db_type = ChoiceAttribute('db_type', choices=[
+        'sqlite', 'mysql', 'postgres', 'mssql', 'oracle', 'firebird', 'sybase'], default='sqlite')
```
dgw: I'd put `choices=[` down on this line, and move the `default` parameter down to its own line. Purely stylistic preference.

sopel/db.py
RustyBower: This is legacy from the old db module. If you have a better solution for it, I am happy to implement.

dgw: What's legacy? Are we talking about different changes, maybe?

sopel/config/core_section.py
RustyBower: I'll throw this in a separate PR

sopel/db.py
RustyBower: I am specifically referring to the Identifier stuff (which has been in the codebase for like 5 years). I've updated the import to be less gross.

dgw: We covered this one on IRC, and figured out the disconnect. :heart:

deathbybandaid:
Very excited for this, will bot owners have any way to migrate existing sqlite content to another database?

RustyBower:
No currently, but there exist some tools like https://github.com/seanharr11/etlalchemy that can do the migrations between any of the now supported DBs

dgw:
> Not currently

Or ever, probably. I don't think it's Sopel's responsibility to offer migration from one database back-end to another. Existing users have numerous options from which to choose if they want to move away from SQLite, just like they had their choice of tools for [moving _to_ SQLite](https://sopel.chat/usage/installing/upgrading-to-willie-5/#database-rewrite) when Willie 5.0 came out.

If it "wasn't our problem" when dropping support for other database engines, it certainly "isn't our problem" when adding (back) support for engines other than SQLite _while still keeping SQLite support_.

deathbybandaid:
It's still good to note that there is a solution though.

dgw:
Assigning to @RustyBower as a reminder. We agreed on IRC that I'll merge this once some rudimentary documentation on the more popular drivers is in place.

Basic docs (in the readme here and in a section at https://sopel.chat/usage/installing/) should cover sqlite, mysql, and pgsql at minimum. Anything more obscure can wait, I think. Not many of us would want to buy, say, an Oracle license just to document it. :grin:

README.rst
```diff
@@ -44,6 +44,10 @@ In the source directory (whether cloned or from the tarball) run
 bot. Alternately, you can just run the ``sopel.py`` file in the source
 directory.
 
+Database Support
+----------------
+Sopel leverages SQLAlchemy to support the following database types: SQLite, MySQL, PostgreSQL, MSSQL, Oracle, Firebird, and Sybase. By default Sopel will use a SQLite database in the current configuration directory, but alternative databases can be configured with the following config options: db_type, db_filename (SQLite only), db_driver, db_user, db_pass, db_host, db_port, and db_name.
```
dgw: Some line-wrap and formatting would be helpful. I also added a bit about installing packages, since it doesn't necessarily go without saying. Feel free to use this as-is or tweak it further.

```suggestion
Sopel leverages SQLAlchemy to support the following database types: SQLite,
MySQL, PostgreSQL, MSSQL, Oracle, Firebird, and Sybase. By default Sopel will
use a SQLite database in the current configuration directory, but alternative
databases can be configured with the following config options: ``db_type``,
``db_filename`` (SQLite only), ``db_driver``, ``db_user``, ``db_pass``,
``db_host``, ``db_port``, and ``db_name``. You will need to manually install
any packages (system or ``pip``) needed to make your chosen database work.
```

RustyBower: Updated!

RustyBower:
Should be all set with README configuration

Issue #1328, opened 2018-05-07
A general way to truncate messages and stay within the maximum length


kwaaak:
After being rightfully scolded by @dgw for hijacking another issue, I am opening a separate one.

It would be nice to have a tool that automatically cuts off the bot responses which are going to be too long, adds a "[…]" at the end, and continues in private message or adds a web link.

There are a few modules which use their own logic to accomplish this.

dgw:
The bot needs to know the length of its hostmask to do this, so this should be implemented with #1322.

Issue #1322, opened 2018-05-03
Sopel should keep track of its full hostmask


freeboson:
Can we properly setup `Sopel.say()` to construct PRIVMSGs that will actually be at the 512 limit?

dgw:
This is actually harder to do than it sounds. I'd like for Sopel to use the full line length available, too, but external factors can change an IRC user's hostname without notice: server-side mode changes, HostServ or similar being turned on/off, cloaking applied after NickServ authentication, etc.

Is there a reliable way to monitor for changes in one's own hostname that I'm unaware of? Happy to look into doing this if it's technically possible without a kludge like polling. :smile_cat:

As an aside, I've used multiple other IRC clients and frameworks that conservatively truncate (split) lines because they don't want to deal with tracking hostname changes either.

dgw:
@kwaaak Why is that comment here? Make it in the relevant PR, make it on IRC, I don't care, just not on a random ticket that has nothing to do with any of what you're talking about. :confused:

dgw:
Sopel tracks its hostmask as of #1537, but this is left open for the "512 limit" piece—that is, making Sopel actually use the stored hostmask to send as much as possible.

deathbybandaid:
Just going to leave this here in regards to how many bytes the hostmask takes, and how many bytes it could portentially take as error handling.


````
        try:
            hostmaskbytes = len((bot.users.get(bot.nick).hostmask).encode('utf-8'))
        except AttributeError:
            hostmaskbytes = (len((bot.nick).encode('utf-8'))  # Bot's NICKLEN
                             + 1  # (! separator)
                             + len('~')  # (for the optional ~ in user)
                             + 9  # max username length
                             + 1  # (@ separator)
                             + 63  # <hostname> has a maximum length of 63 characters.
                             )
````

dgw:
@deathbybandaid In 7.x we have `bot.hostmask`, FYI.

PR #1448, opened 2019-01-13
core: properly truncate long messages with 2-byte characters

### LINKED ISSUE #467: [core]: Unicode issue in long messages
There is a bug concerning unicode characters when a message is long enough.

Here, where the message is short, everything works correctly:

<pre>
Erethon: !in 10s τεστ
BuildTest: Erethon: Okay, will remind in 10 secs
BuildTest: Erethon: τεστ
</pre>

But, when the message is long enough:

<pre>
Erethon: !in 10s 
αααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααα
BuildTest: Erethon: Okay, will remind in 10 secs
BuildTest: Erethon: 
Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î±Î
</pre>

This is marked as [core] because it happens in more than one module, I've confirmed it in ".py" too.
I'm using python 2.7.3 and latest Willie version from master branch and I've installed Willie by running 'python setup.py install'.

Exirel:
When a module wants to send a multi-line message, the `Bot.say` method can be used with a `max_messages` parameter to do so. The low-level `write` method on the other hand, will truncate and discard the excess no matter what.

In both case, we must not truncate unicode character composed of 2 bytes in half: we must remove the whole unicode character, then check the size again, until the encoded message's length is lower or equal to the maximum length.

It won't solve the case where the bot can send a valid message, but the server won't be able to propagate it properly to other clients because it has to add extra information (such as the bot's hostmask and all) and such a way that it generates a message too long.

But it's a good start to fix #467

### DIFF
--- a/sopel/bot.py
+++ b/sopel/bot.py
@@ -290,27 +290,15 @@ class Sopel(irc.Bot):
         message will contain the entire remainder, which may be truncated by
         the server.
         """
-        # We're arbitrarily saying that the max is 400 bytes of text when
-        # messages will be split. Otherwise, we'd have to acocunt for the bot's
-        # hostmask, which is hard.
-        max_text_length = 400
-        # Encode to bytes, for propper length calculation
-        if isinstance(text, unicode):
-            encoded_text = text.encode('utf-8')
-        else:
-            encoded_text = text
         excess = ''
-        if max_messages > 1 and len(encoded_text) > max_text_length:
-            last_space = encoded_text.rfind(' '.encode('utf-8'), 0, max_text_length)
-            if last_space == -1:
-                excess = encoded_text[max_text_length:]
-                encoded_text = encoded_text[:max_text_length]
-            else:
-                excess = encoded_text[last_space + 1:]
-                encoded_text = encoded_text[:last_space]
-        # We'll then send the excess at the end
-        # Back to unicode again, so we don't screw things up later.
-        text = encoded_text.decode('utf-8')
+        if not isinstance(text, unicode):
+            # Make sure we are dealing with unicode string
+            text = text.decode('utf-8')
+
+        if max_messages > 1:
+            # Manage multi-line only when needed
+            text, excess = tools.get_sendable_message(text)
+
         try:
             self.sending.acquire()
 
--- a/sopel/irc.py
+++ b/sopel/irc.py
@@ -138,13 +138,24 @@ class Bot(asynchat.async_chat):
             # CR-LF (Carriage Return - Line Feed) pair, and these messages SHALL
             # NOT exceed 512 characters in length, counting all characters
             # including the trailing CR-LF. Thus, there are 510 characters
-            # maximum allowed for the command and its parameters.  There is no
+            # maximum allowed for the command and its parameters. There is no
             # provision for continuation of message lines.
 
+            max_length = unicode_max_length = 510
             if text is not None:
-                temp = (' '.join(args) + ' :' + text)[:510] + '\r\n'
+                temp = (' '.join(args) + ' :' + text)
             else:
-                temp = ' '.join(args)[:510] + '\r\n'
+                temp = ' '.join(args)
+
+            # The max length of 512 is in bytes, not unicode
+            while len(temp.encode('utf-8')) > max_length:
+                temp = temp[:unicode_max_length]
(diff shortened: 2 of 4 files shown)


Exirel:
The checkstyle script didn't like my usage of `unicode`... in the name of a test function!

So I changed it from `test_something_something_unicode()` to something else. It's fine.

Exirel:
And now it pass. 👍

kwaaak:
Linking issues https://github.com/sopel-irc/sopel/issues/1328 and https://github.com/sopel-irc/sopel/issues/1322 for completeness

dgw:
> The checkstyle script didn't like my usage of `unicode`... in the name of a test function!

I was curious enough to dig out the previous build log…

````
./test/test_tools.py
ERROR: Above files use unicode() but do not make it safe for Python 3.
````

That's actually kind of hilarious, I have to say, and probably bad enough to be considered a bug in `checkstyle.sh` if you feel like fixing that in a separate PR:

https://github.com/sopel-irc/sopel/blob/cf4bccf4b98d429e9ef9b531e996be2ec92ba82b/checkstyle.sh#L46-L59

That "(heuristically)" is the problem. :laughing:

Exirel:
Ok, I found several issues here and there. The hostmask could be a good start, but there are way more things to consider.

For example, most module just use "say" without allowing for multi-line response. Which is bad because there is not much we can do here.

And, of course, there is this whole "hostmask" situation that is a total mess and oh my I don't know where to start...

... so this will wait a bit, the time for me to figure out better overall solution.

Exirel:
> For example, most module just use "say" without allowing for multi-line response. Which is bad because there is not much we can do here.

I was thinking about it. I think it's safe to merge this PR as-is. After all, we make sure here to never send a message that is longer than what the IRC protocol say to.

Because of the hostmask, after this PR, the problem will be only when the IRC server broadcast the bot's message to its clients. So that can be for later!

Issue #1212, opened 2017-08-04
help: using a pastebin should be optional


kkirsche:
Not all environments are public environments where it is appropriate to post commands  publicly. It would be extremely helpful if this behavior (print to channel or gist) was configurable.

dgw:
Also additional options, like outputting to a text or HTML file on the local filesystem instead, would be useful. Maybe I should just bite the bullet and implement all that.

dgw:
As an aside, this issue might be obsolete. See #1254.

dgw:
Switched to a different pastebin provider for the time being. This will stick around as a feature request.

Exirel:
I feel like this has been fixed by #1451

dgw:
@Exirel Nope. We still want to add locally-hosted HTML or something that doesn't touch a pastebin at all.

dgw:
Punting this. The implementation should add an option without changing the existing behavior if not configured, so no rush to get this into the next major version.

Exirel:
The more I think about that one, the more I'd like to create a separated package for the help plugin, and to have a dedicated `[help]` config section. I've plenty of ideas for that kind of plugins, and I think it would help (...) a lot to be able to update such plugin whenever a paste service become unavailable, or when they change their interface, and so on.

For instance, I'd love to allow third-party devs to implement their own help publishing backend, so whenever someone have a new use-case that is solely for them and their group, they could do it more easily.

Also, I'd use it as an excuse to push Sopel's internals to always be better - as one may say "eat your own food".

dgw:
Aren't we already pulling a bunch more stuff out of core for 8.0? `help` can be part of that.

Exirel:
Yes, but I'd love to start with `sopel-help` for Sopel 7.1: it's easy to do, and since almost everyone use it from time to time, if anything goes wrong it won't be silent.

Also, "help" is another name for "documentation". 😛

dgw:
Feel free to play around in the new `sopel-help` repo, @Exirel. Whatever you come up with in there, the only real question is whether it takes effect in 7.1 or 8.0.

dgw:
We'll consider this fixed by [sopel-irc/sopel-help](https://github.com/sopel-irc/sopel-help), which includes several pastebins, adds local HTML output (to be served by the user's chosen external HTTPd), _and_ resurrected direct-to-IRC output as well.

[...]

PR #1451, opened 2019-01-15
help: choose pastebin service for .help output

### LINKED ISSUE #1416: help: pastebin provider should be configurable
The `help` module used to use GitHub's Gist function, but anonymous Gists were deprecated (see #1254).

We switched to ptpb.pw as a pastebin provider (see #1257) for Sopel 6.5.2, but for a short while yesterday (November 30, 2018) its domain was in an expired state. While the domain has been restored, the notice @kwaaak sent me on IRC about it being gone gave me a good scare.

I think it's prudent to start thinking about how to make `help.py` more configurable so a more permanent outage of ptpb.pw won't break Sopel's help functionality until the next release. Our release cadence is slow enough that it's a valid concern. I want to keep trying to minimize the number of things that can break in a way that gives the user no recourse but to wait (like `etymology.py`—see #1248). New versions should come out when they're done, and not be rushed because something is broken in the latest release.

In #1212, it was suggested that the use of a pastebin should be configurable, with the option to use the old behavior of just sending the command list to IRC. Also in that issue, I brought up the idea of dumping to a local file on Sopel's disk and linking to it, kind of like how `meetbot.py`'s meeting notes work.

On top of those features, I propose a new `ChoiceAttribute` with several different supported pastebin providers. `ptpb` would be one, of course.

**Suggestions for additional pastebin services are welcome in the comments!** Any service like `pastebin`, `hastebin`, etc. is fine as long as it seems unlikely to vanish in the foreseeable future.

----

Here is the current list of possibilities:

  * ~~`ptpb.pw`~~ shut down (ptpb/pb#246)
  * ~~`pastebin.com`~~ requires API key
  * `hastebin.com`
  * `0x0.st`
  * `termbin.com`
  * `clbin.com` — default as of v6.6.5

uint:
# Description
I've created a new config section `help` with the `output` setting. `output` can currently be set to:
* `"ptpb"`
* `"0x0"`
* `"hastebin"`
* `"termbin"`

This resolves #1416.

# Example
```
19:42 <uint> .set help.output termbin
19:42 <uint> .help
19:43 <uint-bot> Hang on, I'm creating a list.
19:43 <uint-bot> I've posted a list of my commands at https://termbin.com/3tyx2 - You can see more
                 info about any of these commands by doing .help <command> (e.g. .help time)
```

# Potential problems
It seems that if I use `.help`, the bot will cache the url. If I then change the `output` option to something else, the bot won't let me use the new service since it still has the old link cached. This could lead to problems if the old link is erroneous and the user can't get a new one until they restart the bot.

@dgw Is it possible to make it so that changing the config option also triggers the cached data to be cleared? I think that would be the simplest solution.

# Next steps?
* We probably really should add an option for the bot to simply "say" the command list, either in the channel or a private message. Should I work on that too?

# Version
```
19:51 <uint> .version
19:51 <uint-bot> uint: Sopel v. 6.5.3
```

### DIFF
--- a/sopel/modules/help.py
+++ b/sopel/modules/help.py
@@ -22,8 +22,62 @@ from sopel.config.types import (
 )
 
 
+# Pastebin handlers
+
+
+def post_to_clbin(bot, msg):
+    result = requests.post('https://clbin.com/', data={'clbin': msg})
+    result = result.text
+
+    if 'https://clbin.com/' in result:
+        return result
+    else:
+        LOGGER.error("Invalid result %s", result)
+        raise PostingException()
+
+
+def post_to_0x0(bot, msg):
+    payload = {'file': msg}
+    result = requests.post('https://0x0.st', files=payload)
+    return result.text
+
+
+def post_to_hastebin(bot, msg):
+    result = requests.post('https://hastebin.com/documents', data=msg)
+    result = result.json()
+    if 'key' not in result:
+        LOGGER.error("Invalid result %s", result)
+        raise PostingException()
+    return "https://hastebin.com/" + result['key']
+
+
+def post_to_termbin(bot, msg):
+    s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+    s.connect(('termbin.com', 9999))
+    s.sendall(msg)
+    s.shutdown(socket.SHUT_WR)
+    response = ""
+    while 1:
+        data = s.recv(1024)
+        if data == "":
+            break
+        response += data
+    s.close()
+    return response.strip(u'\x00\n')
+
+
+_pastebin_providers = {
+    'clbin': post_to_clbin,
+    '0x0': post_to_0x0,
+    'hastebin': post_to_hastebin,
+    'termbin': post_to_termbin
+}
+
+
 class HelpSection(StaticSection):
-    output = ChoiceAttribute('output', ["clbin", "0x0", "hastebin", "termbin"], default='clbin')
+    output = ChoiceAttribute('output',
+                             list(_pastebin_providers),
+                             default='clbin')
 
 
 def configure(config):
@@ -100,53 +154,12 @@ def create_list(bot, msg):
     msg = 'Command listing for {}@{}\n\n'.format(bot.nick, bot.config.core.host) + msg
 
     try:
-        result = globals()['post_to_' + bot.config.help.output](bot, msg)
+        result = _pastebin_providers[bot.config.help.output](bot, msg)
     except (requests.RequestException, PostingException):
         bot.say("Sorry! Something went wrong.")
         LOGGER.exception("Error posting commands")
         return
     return result
-
-
-def post_to_clbin(bot, msg):
-    result = requests.post('https://clbin.com/', data={'clbin': msg})
-    result = result.text
-
-    if 'https://clbin.com/' in result:
-        return result
-    else:
-        LOGGER.error("Invalid result %s", result)
-        raise PostingException()
-
-
-def post_to_0x0(bot, msg):
-    payload = {'file': msg}
-    result = requests.post('https://0x0.st', files=payload)
-    return result.text
-
-
-def post_to_hastebin(bot, msg):
-    result = requests.post('https://hastebin.com/documents', data=msg)
-    result = result.json()
-    if 'key' not in result:
-        LOGGER.error("Invalid result %s", result)
-        raise PostingException()
-    return "https://hastebin.com/" + result['key']
-
-
-def post_to_termbin(bot, msg):
-    s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
-    s.connect(('termbin.com', 9999))
-    s.sendall(msg)
-    s.shutdown(socket.SHUT_WR)
-    response = ""
-    while 1:
-        data = s.recv(1024)
-        if data == "":
-            break
-        response += data
-    s.close()
-    return response.strip(u'\x00\n')
 
 
 class PostingException(Exception):
--- a/sopel/modules/help.py
+++ b/sopel/modules/help.py
@@ -130,7 +130,11 @@ def help(bot, trigger):
         # actually creating the list first. Maybe worth storing the link and a
         # heuristic in the DB, too, so it persists across restarts. Would need a
         # command to regenerate, too...
-        if 'command-list' in bot.memory and bot.memory['command-list'][0] == len(bot.command_groups):
+        if (
+                'command-list' in bot.memory and
+                bot.memory['command-list'][0] == len(bot.command_groups) and
+                bot.memory['command-list'][2] == bot.config.help.output
+        ):
             url = bot.memory['command-list'][1]
         else:
             bot.say("Hang on, I'm creating a list.")
@@ -149,7 +153,9 @@ def help(bot, trigger):
             url = create_list(bot, '\n\n'.join(msgs))
             if not url:
                 return
-            bot.memory['command-list'] = (len(bot.command_groups), url)
+            bot.memory['command-list'] = (len(bot.command_groups),
+                                          url,
+                                          bot.config.help.output)
         bot.say("I've posted a list of my commands at {0} - You can see "
                 "more info about any of these commands by doing {1}help "
                 "<command> (e.g. {1}help time)".format(url, help_prefix))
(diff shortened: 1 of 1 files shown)


HumorBaby:
In regards to the caching issue: Sopel won't re-read a config file after editing manually (_that I know of_). The main way to change the config on the fly is to use things like `.set` and `.save` from the `admin.py` module. This will update the configuration object used by the bot and allow you to save said changes to the config. Ideally this is how `.help` paste provides will be changed by the user also. 

One solution could be to store the current pastebin provider right next to the cached URL, and check if the current `config.help.output` matches the cached value you have stored before sending. This alternative would make it so the config load/update logic would not need to be changed, as this could be implemented purely in `help.py`.

uint:
@HumorBaby What I could do is change `help.py` so that it caches both the url and the string denoting the pastebin provider. Then I can very easily make sure the cache is only used if the cached provider matches the one currently configured. Does that sound sane?

HumorBaby:
@uint, yep :smile:

uint:
On rereading your post, that's exactly what you suggested ;) There goes my reading comprehension. But yeah. I'll add a commit soon.

kwaaak:
hastebin.com had an interruption a few days ago, posting was returning a 503 error.

I wonder if it would be a good idea to - if there is no service configured - randomly pick a service, and try another one if there is an error.

sopel/modules/help.py
```diff
@@ -82,23 +96,64 @@ def help(bot, trigger):
 
 def create_list(bot, msg):
     msg = 'Command listing for {}@{}\n\n'.format(bot.nick, bot.config.core.host) + msg
-    payload = {"content": msg}
-    headers = {'Content-type': 'application/json', 'Accept': 'application/json'}
 
     try:
-        result = requests.post('https://ptpb.pw/', json=payload, headers=headers)
-    except requests.RequestException:
+        result = globals()['post_to_' + bot.config.help.output](bot, msg)
```
Exirel: Using `globals()` sure is convenient, but that doesn't make me very comfortable: you never know what is in `globals()` and what can be exploited with that. Plus, the whole `'post_to_' + ...` is less readable than, say, this:

```
backends = {
    'ptpb': post_to_ptpb,
    '0x0': post_to_0x0,
    # and so on
}
try:
    result = backends[bot.config.help.output](bot, msg)
except KeyError:
    bot.say('Sorry! Wrong help configuration')
```

sopel/modules/help.py
```diff
@@ -13,16 +13,30 @@
...
+)
+
+
+class HelpSection(StaticSection):
+    output = ChoiceAttribute('output', ["ptpb", "0x0", "hastebin", "termbin"], default='ptpb')
+
+
+def configure(config):
+    config.define_section('help', HelpSection)
+
```
HumorBaby: Could be nice if there was a prompt to select the provide in the config wizard :smile: 
Maybe like so:

```
config.help.configure_setting(
    'output',
    'Pick a pastebin provider: ptpb, 0x0, hastebin, termbin: '
)
```
? You can likely use format string to put the choices in the prompt.

dgw:
After 6.6.4 is released and its changes merged to `master`, this should be rebased (it probably will gain a merge conflict or three because of #1495).

Also, ptpb is shut down (semi-?)permanently (ptpb/pb#246), so it would be good to remove that option. Maybe replace it with clbin?

dgw:
@uint \*poke\*

dgw:
Rebased and "squashed" the whitespace-fix commits (not actually squashed, since both batches of fixed applied to multiple commits each).

Next will be removing ptpb and adding… something else. Like clbin.

dgw:
And ptpb->clbin default change: done

sopel/modules/help.py
```diff
@@ -12,10 +12,81 @@
...
+def post_to_hastebin(bot, msg):
+    result = requests.post('https://hastebin.com/documents', data=msg)
+    result = result.json()
+    if 'key' not in result:
+        LOGGER.error("Invalid result %s", result)
+        raise PostingException()
+    return "https://hastebin.com/" + result['key']
+
+
+def post_to_termbin(bot, msg):
```
HumorBaby: This line in the file also needs to be updated:
https://github.com/sopel-irc/sopel/blob/38e30e67ad8dc5c90da87591d46f1f2f6167833e/sopel/modules/help.py#L169

If the socket cannot connect or is outright refused, it will not throw a `requests` error, but a [`socket.error`](https://docs.python.org/2/library/socket.html#socket.error)

The other (and IMO better) option is to just wrap the entire socket block (all socket related commands) in a `try`/`except`, catch and `socket` related exception, and raise `PostingError` instead. This may be best because there are at least four different non-nested `socket` error types that would otherwise have to be declared here individually; see next comment.

sopel/modules/help.py
```diff
@@ -12,10 +12,81 @@
...
+    s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
```
HumorBaby: Because `socket` is so low-level, theres a lot more things that need to be configured here to be presentable to the user; specifcally, a timeout [`s.settimeout`](https://docs.python.org/2/library/socket.html#socket.socket.settimeout) needs to be set so that the bot doesn't try to wait for a connection forever
```suggestion
    s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
    s.settimeout(10)  # raise `socket.timeout` if socket doesn't respond for >10 seconds
```

Don't forget to handle this exception nicely (see previous comment).

sopel/modules/help.py
```diff
@@ -12,10 +12,81 @@
...
+        return result
+    else:
+        LOGGER.error("Invalid result %s", result)
+        raise PostingException()
+
+
+def post_to_0x0(bot, msg):
+    payload = {'file': msg}
+    result = requests.post('https://0x0.st', files=payload)
+    return result.text
```
HumorBaby: The error handling is missing in this method. Following `post_to_clbin` (since the response from the server is a simple URL string):
```suggestion
    if 'https://0x0.st' in result:
        return result
    else:
        LOGGER.error('Invalid result %s', result)
        raise PostingException()
    return result.text
```

sopel/modules/help.py
```diff
@@ -12,10 +12,81 @@
...
+
+
+def post_to_hastebin(bot, msg):
+    result = requests.post('https://hastebin.com/documents', data=msg)
+    result = result.json()
```
HumorBaby: If an invalid response is returned, this line will throw a `ValueError: NO JSON...`, which is not handled nicely currently. See other comments, as it may just be better to do a larger change re: error handling.

dgw:
@HumorBaby Did a pretty big overhaul of error handling, and a few other little things while I was at it. :rocket:

sopel/modules/help.py
```diff
@@ -82,23 +96,64 @@ def help(bot, trigger):
...
+    while 1:
+        data = s.recv(1024)
+        if data == "":
+            break
+        response += data
+    s.close()
+    return response.strip(u'\x00\n')
+
+
+class PostingException(Exception):
```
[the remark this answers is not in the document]
dgw: Moved.

sopel/modules/help.py
```diff
@@ -13,16 +13,30 @@
 import textwrap
 import collections
 import requests
+import socket
 
 from sopel.logger import get_logger
 from sopel.module import commands, rule, example, priority
+from sopel.config.types import (
+    StaticSection, ChoiceAttribute
+)
```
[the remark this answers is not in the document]
dgw: Maybe don't need, but I can't be bothered to change it x)

uint:
Hey. Sorry for not having been responsive! Pretty busy and I... probably should configure e-mail notifications. Ahem.

Exirel:
@uint no worry! Our dear maintainer @dgw took it from where you left, co-authored the first commits and added some more. Also thank you for these first commits! :+1:

dgw:
@uint You got us 90% of the way there! Most of what I did was just sanding off rough edges. :raised_hands: Without your initial set of commits, this feature would still be "Patches Welcome", but it's done!

As for non-pastebin help output, we are tracking that in #1212. I think @RustyBower is working on something for that, unless he changed his mind. Feel free to join us in the #sopel IRC channel on freenode if you want to avoid duplicating effort on any future contributions. And thank you again for this PR!

Issue #1452, opened 2019-01-21
pip install sopel fails with sopel 6.6.0 and python 3.7 in fresh virtualenv


Corwind:
OS: Arch Linux
Python: 3.7.2
Sopel: 6.6.0

On an up-to-date Arch Linux, with python 3.7.2, in a fresh virtualenv the installation process of sopel fails because of one of its dependencies: pyenchant which is unmaintained since the 24th of february of 2018.
This dependency fails to install, leading to sopel to fail to install.

```
$ pip install sopel
Collecting sopel
  Using cached https://files.pythonhosted.org/packages/3b/cf/397ac0b33c8cccc6877acb434c96692eeb853333bacceff36bb0613bb05e/sopel-6.6.0.tar.gz
Collecting xmltodict (from sopel)
  Using cached https://files.pythonhosted.org/packages/42/a9/7e99652c6bc619d19d58cdd8c47560730eb5825d43a7e25db2e1d776ceb7/xmltodict-0.11.0-py2.py3-none-any.whl
Collecting pytz (from sopel)
  Using cached https://files.pythonhosted.org/packages/61/28/1d3920e4d1d50b19bc5d24398a7cd85cc7b9a75a490570d5a30c57622d34/pytz-2018.9-py2.py3-none-any.whl
Collecting praw<6.0.0 (from sopel)
  Using cached https://files.pythonhosted.org/packages/95/71/f3a208d68bf93709c52b4fc110a3c8aab2f68215dcd171ee42c7c3671645/praw-5.4.0-py2.py3-none-any.whl
Collecting pyenchant (from sopel)
  Using cached https://files.pythonhosted.org/packages/9e/54/04d88a59efa33fefb88133ceb638cdf754319030c28aadc5a379d82140ed/pyenchant-2.0.0.tar.gz
    Complete output from command python setup.py egg_info:
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "/tmp/pip-install-kdnf9ejs/pyenchant/setup.py", line 212, in <module>
        import enchant
      File "/tmp/pip-install-kdnf9ejs/pyenchant/enchant/__init__.py", line 92, in <module>
        from enchant import _enchant as _e
      File "/tmp/pip-install-kdnf9ejs/pyenchant/enchant/_enchant.py", line 145, in <module>
        raise ImportError(msg)
    ImportError: The 'enchant' C library was not found. Please install it via your OS package manager, or use a pre-built binary wheel from PyPI.

[...]

PR #1454, opened 2019-01-23
Don't require pyenchant on Python 3.7 or above

dgw:
pyenchant is broken on newer distributions, and has been unmaintained for quite a while. Since only the spellcheck module needs it, and a rewrite of that module to switch libraries is coming Soon™, we might as
well make Sopel installable in the mean time on systems that can't pull down a working pyenchant build.

See #1452. Alternative to #1453.

Note that this change is intended to make `master` installable, and there is no release planned between now and Sopel 7 (which will include the rewritten spellcheck module and drop pyenchant completely).

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,7 +1,7 @@
 xmltodict
 pytz
 praw<6.0.0
-pyenchant
+pyenchant; python_version < '3.7'
 geoip2
 ipython<6.0; python_version < '3.3'
 ipython>=6.0,<7.0; python_version >= '3.3'


dgw:
@Corwind Thoughts on this? Is it an acceptable stopgap until #1164 is ready to go?

Corwind:
Looks good to me, it will fix the installation using pip + git, which will enable devs to automate this process inside a virtualenv.

Thanks @dgw

dgw:
Thanks for the feedback, both of you!

PR #1456, opened 2019-01-23
modules: use @require_admin instead of if trigger.admin

Exirel:
There is a decorator in `sopel.module.require_admin` that check for `trigger.admin`, but it wasn't used everywhere!

So I fixed this, and added a `reply=False` optional argument, that, when `True`, uses `bot.reply` instead of `bot.say`.

It's a very small improvement.

### DIFF
--- a/sopel/modules/announce.py
+++ b/sopel/modules/announce.py
@@ -7,18 +7,16 @@ Licensed under the Eiffel Forum License 2.
 """
 from __future__ import unicode_literals, absolute_import, print_function, division
 
-from sopel.module import commands, example
+from sopel.module import commands, example, require_admin
 
 
 @commands('announce')
 @example('.announce Some important message here')
+@require_admin('Sorry, I can\'t let you do that', reply=True)
 def announce(bot, trigger):
     """
     Send an announcement to all channels the bot is in
     """
-    if not trigger.admin:
-        bot.reply('Sorry, I can\'t let you do that')
-        return
     for channel in bot.channels:
         bot.msg(channel, '[ANNOUNCEMENT] %s' % trigger.group(2))
     bot.reply('Announce complete.')
--- a/sopel/coretasks.py
+++ b/sopel/coretasks.py
@@ -596,15 +596,13 @@ def sasl_success(bot, trigger):
 @sopel.module.priority('low')
 @sopel.module.thread(False)
 @sopel.module.unblockable
+@sopel.module.require_admin
 def blocks(bot, trigger):
     """
     Manage Sopel's blocking features.\
     See [ignore system documentation]({% link _usage/ignoring-people.md %}).
 
     """
-    if not trigger.admin:
-        return
-
     STRINGS = {
         "success_del": "Successfully deleted block: %s",
         "success_add": "Successfully added block: %s",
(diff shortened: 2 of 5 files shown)


sopel/modules/announce.py
```diff
@@ -7,18 +7,16 @@
 """
 from __future__ import unicode_literals, absolute_import, print_function, division
 
-from sopel.module import commands, example
+from sopel.module import commands, example, require_admin
 
 
 @commands('announce')
 @example('.announce Some important message here')
+@require_admin('Sorry, I can\'t let you do that', reply=True)
```
rpatel: While you're in here, should this message be updated to say why "`[the bot] can't let you do that`"? i.e., you must be an admin

Thoughts?

HumorBaby: While you're in here: should this message be updated to say why "`[the bot] can't let you do that`"? i.e., tell user that only admins can do that...

Thoughts?

Exirel: Since I've 0 opinion bout that, I call @dgw like the coward I am. 😛

dgw: Update it if you want to, don't if you don't. The message has been that way for almost six years. :stuck_out_tongue:

Exirel:
Note: rebase to stay up-to-date with the master branch.

@dgw this is a fairly small change, so it's perfect for a quick & easy merge.

sopel/modules/announce.py
dgw: If you're doing "One change at a time", what's up with adding a new parameter to `sopel.module.require_admin` in a PR titled "use \@require_admin instead of if trigger.admin"? 😁

Exirel: @dgw I know. Optional parameter + same behavior for the end-user as before is what I went for. :sunglasses:

dgw: I'd kind of prefer that kind of change in its own PR, or at least its own commit… How much trouble would that be? Would you be mad? :fearful:

Exirel: I'll try that in a second commit then, because I can not replace `if not trigger.admin` check without this change, because some command want to reply something.

Exirel: Rebased & fix.

dgw: Cool! Though ipython is updated in the commit after adding the new arg but doesn't use it?

Exirel: Oh, I thought it was but... no, it doesn't. Why did I do that? I think I understand why you were confused the first time, because I sure am confused right now.

Exirel: Oh, ok, I found out:

* first commit: `@require_admin` without message
* second commit: add parameter
* third commit: `@require_admin` with message (and one with reply)

So it checks out. I think.

dgw: That's what I expected, but the "with reply" commit is where ipython gets updated. It doesn't use reply.

Exirel:
> _Maybe._ 😁

Maybe eventually. 😉

dgw:
In the distant future:

```Markdown
Changes between 8.8.3 and 9.0.0
===============================

[...]

API changes
-----------

* `module.require_admin()` now takes an optional boolean argument `reply` (default `False`)
  * This was added back in 7.0.0, but we forgot about it until now
```

🤣

Issue #1455, opened 2019-01-23
Sopel lacks the ability to send custom commands on connect


dgw:
Think ZNC's `perform` module, but without the ability to add/remove/rearrange lines from an IRC query.

I noted this down a while ago, but forgot about it until someone asked how to do this in #sopel tonight.

We want to keep this in the existing config file, but slipping this into the existing config value types could prove to be _interesting_. A custom login command might contain, say, a comma, because special characters in passwords are encouraged by modern recommendations. `ListAttribute` (the most logical option) doesn't escape commas at all, so it would split the values incorrectly upon reading them back.

(A separate PR implementing such escaping would be welcomed, if anyone's interested in working on it.)

dgw:
Milestoned to match the implementation, #1528. If the PR is deferred, this should be, too.

PR #1460, opened 2019-01-27
config: support comma in config ListAttribute item

HumorBaby:
This is a first step towards completing #1455, since, as @dgw mentioned, to do so would require accepting commas in a single entry of `ListAttribute`.

Items entered through the wizard do not need to be escaped manually. Items configured in the `.cfg` file manually require an escape sequence, like `\,` or `\\`. The prompt has also been changed such that the items are now surrounded by quotes so that one can differentiate items that may contain commas.

Relevant tests were added/updated (see commit messages).

### DIFF
--- a/sopel/config/types.py
+++ b/sopel/config/types.py
@@ -266,17 +266,22 @@ class ListAttribute(BaseValidated):
             prompt = prompt[0]
 
         if default is not NO_DEFAULT:
-            default = ','.join(default)
-            prompt = '{} [{}]'.format(prompt, default)
+            default_prompt = ','.join(['"{}"'.format(item) for item in default])
+            prompt = '{} [{}]'.format(prompt, default_prompt)
         else:
-            default = ''
+            default = []
         print(prompt)
         values = []
         value = get_input(each_prompt + ' ') or default
+        if (value == default) and not default:
+            value = ''
         while value:
-            values.append(value)
+            if value == default:
+                values.extend(value)
+            else:
+                values.append(value)
             value = get_input(each_prompt + ' ')
-        return self.parse(','.join(values))
+        return self.parse(self.serialize(values))
 
 
 class ChoiceAttribute(BaseValidated):
--- a/sopel/config/types.py
+++ b/sopel/config/types.py
@@ -212,13 +212,34 @@ class ListAttribute(BaseValidated):
     currently support commas within items in the list. By default, the spaces
     before and after each item are stripped; you can override this by passing
     ``strip=False``."""
+
+    ESCAPE_CHARACTER = '\\'
+    DELIMITER = ','
+
     def __init__(self, name, strip=True, default=None):
         default = default or []
         super(ListAttribute, self).__init__(name, default=default)
         self.strip = strip
 
     def parse(self, value):
-        value = list(filter(None, value.split(',')))
+        items = []
+        is_escape_on = False
+        current_token = []
+        for char in value:
+            if not is_escape_on:
+                if char == ListAttribute.ESCAPE_CHARACTER:
+                    is_escape_on = True
+                elif char == ListAttribute.DELIMITER:
+                    items.append(''.join(current_token))
+                    current_token = []
+                else:
+                    current_token.append(char)
+            else:
+                current_token.append(char)
+                is_escape_on = False
+        items.append(''.join(current_token))
+
+        value = list(filter(None, items))
         if self.strip:
             return [v.strip() for v in value]
         else:
@@ -227,7 +248,16 @@ class ListAttribute(BaseValidated):
     def serialize(self, value):
         if not isinstance(value, (list, set)):
             raise ValueError('ListAttribute value must be a list.')
-        return ','.join(value)
+
+        items = []
+        for item in value:
+            current_token = []
+            for char in item:
+                if char in [ListAttribute.ESCAPE_CHARACTER, ListAttribute.DELIMITER]:
+                    current_token.append(ListAttribute.ESCAPE_CHARACTER)
+                current_token.append(char)
+            items.append(''.join(current_token))
+        return ','.join(items)
 
     def configure(self, prompt, default, parent, section_name):
         each_prompt = '?'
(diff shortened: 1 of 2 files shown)


HumorBaby:
:cry: `test_config.py` fails on `test_listattribute_with_value_containing_comma` because now commas are supported, and the failure case is no longer valid. Should I submit a PR to update that test? Or would something like that come after the acceptance of this PR?

dgw:
Mmm, regex lookaround assertions :)

You can squash the existing test update at your leisure; it's preferable to change code and test in the same commit, if the test result is expected to change. (At the same time, you can fix "manaully" in the commit message for posterity :stuck_out_tongue:)

I briefly thought about how ZNC encodes special characters in its modules' `.registry` files, but that's annoying(ly verbose)—our `,` would become `%2c;`. Fine for an internal data format, but not great for a user-facing (and user-editable) file like Sopel's `.cfg`. Using `\,` is much simpler.

My only concern at present is potential edge cases. Take, for example, a list value that ends in `\`:

````
Original list: ['test', r'test\', 'test,foo', r'foo\,test']
Serialized:    r'test,test\,test\,foo,foo\\,test'
Parsed:        ['test', 'test,test,foo', r'foo\,test']
````

Granted, I've done that manually, and being sick this weekend can't be helping my ability to pretend I'm a Python interpreter, but I _think_ this is how it would work. It's not ideal. Such a case should also have a test (or three) written against it to make sure items can't be wrongly merged in this way.

The easiest solution is probably always (also) escaping `\` to `\\`. That will probably make the serialization and parsing steps "more fun" though.

**Comment by HumorBaby** [comment]
Mmm, *nested* negative lookbehinds :open_mouth: 

Tests are now squashed, and commit message typos fixed. @dgw, turns out your *in vivo* Python interpreter functions quite well even when sick :smile: The test case you mentioned failed spectacularly as expected for my first implementation. That specific scenario is now handled... but as I was designing more test cases, I thought up this one: 

https://github.com/sopel-irc/sopel/blob/01da8c6f2cebff9d2dff686201c794075bf3687b/test/test_config.py#L74-L76

That's when the serialization/parsing *really* starts to get **fun** :weary:... This is currently the failing test.

**Comment by dgw** [comment]
Mm, I was afraid this would become a rabbit hole once I brought up that first edge case… :laughing:

What do you think of a different parsing approach, that doesn't use regex? The problem, as I see it, is that you absolutely **cannot** process the same character more than once when serializing or parsing.

For serialization, that's easy: build the output string character-by-character, appending `\,` when the serializer sees `,`, and `\\` when it sees `\`. For parsing, it seems like a job for a state machine, e.g. https://rosettacode.org/wiki/Tokenize_a_string_with_escaping#Python

What do you think? Fighting with the regex could be fun, I guess, but there might be better uses to be made of your free time after this is finished. :man_shrugging:

**Comment by HumorBaby** [comment]
No-regex sounds good to me! Since I had already started using regex for the simpler cases when I started working on this, I figured I would keep digging a deeper hole, and maybe there would be some light on the other side :laughing: but _alas_, it doesn't seem worth it. The state machine approach should be pretty straight forward, and I will climb out of the crater I made with regex and go that route for parsing instead. Then, I'll be all out of ways to keep putting off the `.reload` issue and can work on that :sweat_smile:

**Comment by HumorBaby** [comment]
Yay Python 2.7! :tada: 
Boo Python 3.4 :-1: 

Gotta figure out the string differences between to ensure compatability with Python 3... _coming soon_

File: sopel/config/types.py [hunk 1]
```diff
@@ -236,17 +266,22 @@ def configure(self, prompt, default, parent, section_name):
...
-            prompt = '{} [{}]'.format(prompt, default)
+            default_prompt = ','.join(map(lambda item: '"{}"'.format(item), default))
+            prompt = '{} [{}]'.format(prompt, default_prompt)
         else:
-            default = ''
+            default = []
         print(prompt)
         values = []
         value = get_input(each_prompt + ' ') or default
+        if (value == default) and len(default) == 0:
```
Exirel: By convention, we should use `not default` instead of `len(default) == 0`.

sopel/config/types.py
```diff
@@ -236,17 +266,22 @@ def configure(self, prompt, default, parent, section_name):
             prompt = prompt[0]
 
         if default is not NO_DEFAULT:
-            default = ','.join(default)
-            prompt = '{} [{}]'.format(prompt, default)
+            default_prompt = ','.join(map(lambda item: '"{}"'.format(item), default))
```
Exirel: I would rather see a list-comprehension than a `map` + `lambda`:

    ','.join(['"{}"'.format(item) for item in default])

sopel/config/types.py
```diff
@@ -236,17 +266,22 @@ def configure(self, prompt, default, parent, section_name):
...
-            default = ''
+            default = []
         print(prompt)
         values = []
         value = get_input(each_prompt + ' ') or default
+        if (value == default) and len(default) == 0:
+            value = ''
         while value:
-            values.append(value)
+            if (value == default):
```
Exirel: You don't need the extra parenthesis here.

sopel/config/types.py
HumorBaby: :+1:

sopel/config/types.py
HumorBaby: :+1:

sopel/config/types.py
HumorBaby: :+1:

sopel/config/types.py
```diff
@@ -236,17 +266,22 @@ def configure(self, prompt, default, parent, section_name):
             prompt = prompt[0]
 
         if default is not NO_DEFAULT:
-            default = ','.join(default)
-            prompt = '{} [{}]'.format(prompt, default)
+            default_prompt = ','.join(['"{}"'.format(item) for item in  default])
```
dgw: This is what's failing Travis builds rn.
```suggestion
            default_prompt = ','.join(['"{}"'.format(item) for item in default])
```

HumorBaby: Aww snap. I fixed it after my first push, but I `commit`'ed again without `add`'ing :cry:

dgw: Lol, we've all done that. 😂

Also, TIL OctoDroid collapses whitespace in comments (it doesn't show the diff in my suggested change properly; both lines look identical).

dgw:
I guess it's time I tested this again, huh? :grin:

dgw:
This feature is reverted by current state of #1628. I'll remove it from the draft changelog and milestone after merging that… I hope.

PR #1462, opened 2019-01-29
version: include python version

HumorBaby:
Last unsolicited PR until I dig into `.reload` more, _I PROMISE_ :grin:

Sometimes I forget which version of python I am using :sweat_smile:.  Really though, this stemmed from testing issues/code across multiple versions of Python. I could _quickly_ check the Sopel version from the bot, but not the python version it was running on. This commit simply appends the Python version to the `.version` response.

I don't think it shouldn't really be a security issue to show the Python version of the bot... _right??_

### DIFF
--- a/sopel/modules/version.py
+++ b/sopel/modules/version.py
@@ -14,6 +14,8 @@ from sopel import __version__ as release
 from sopel.module import commands, intent, rate
 import re
 from os import path
+import platform
+
 
 log_line = re.compile(r'\S+ (\S+) (.*? <.*?>) (\d+) (\S+)\tcommit[^:]*: (.+)')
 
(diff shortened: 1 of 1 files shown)


HumorBaby:
What'r the chances of _that_ test failing for _that_ reason..?! :scream: ([Build #1470](https://travis-ci.org/sopel-irc/sopel/jobs/485898772))

dgw:
Travis getting a connection reset by peer… that's a new one. (I kicked build 1470.1, though, so it doesn't show the error any more. Could have easily made the test pass without your force-push. :smile_cat:)

I'm torn between including this in 6.6.2 (already planned to fix a few small bugs that weren't fixed in 6.6.x yet) and waiting until 7. I don't want to start a 6.7.x branch, and I kinda think this small added feature is slightly past the intent of a patch release.

Of course it's mostly useful for testing the current version, so putting this on `master` and not releasing it until Sopel 7 is probably fine. That'll also give plenty of time for feedback about the alleged security risk of including the Python version. (People who think the Python version is a risk probably also think the Sopel version is a risk, so I bet they just disable this module anyway. :rofl:)

PR #1464, opened 2019-01-29
cli: fix configpath issue for wizard

### LINKED ISSUE #1463: Config filename is mis-parsed when running configuration wizard from command line
~~`-w` / `--configure-all` and~~ `--configure-modules` does not drop the file extension `.cfg` from the specified file when searching for a config file to use (unlike just launching the bot, where as the extension does seem to be dropped, or at least accounted for when finding the config file).

A user stopped into #sopel (2019-01-29, ~14:15 UTC-5) who had a `default.cfg` and `custom_config.cfg`. They primarily used `custom_config.cfg` and when attempting to configure modules for said `custom_config.cfg`, the used the command `sopel -c custom_config.cfg --configure-modules`. Sopel could not find `custom_config.cfg.cfg`, and without error/warning, edited `default.cfg`. Launching the bot with `sopel -c custom_config.cfg` still used the correct file though, so any changes made with the configuration wizard were never propogated to the config file that was actually being loaded.

**EDIT**: the `--configure-all` flag will just make a new config file like `custom_config.cfg.cfg`

HumorBaby:
Now checks if the config file passed to `-c` already has a `.cfg` extensions before appending one.

Fixes #1463

### DIFF
--- a/sopel/config/__init__.py
+++ b/sopel/config/__init__.py
@@ -212,7 +212,7 @@ class Config(object):
 
 def _wizard(section, config=None):
     dotdir = os.path.expanduser('~/.sopel')
-    configpath = os.path.join(dotdir, (config or 'default') + '.cfg')
+    configpath = os.path.join(dotdir, ((config or 'default.cfg') + ('.cfg' if config and not config.endswith('.cfg') else '')))
     if section == 'all':
         _create_config(configpath)
     elif section == 'mod':


dgw:
So I guess you lied a little in #1462? :laughing:

> Last unsolicited PR until I dig into `.reload` more, _I PROMISE_ :grin:

I'm going to see if GitHub will let me change the base branch on this to `6.6.x` cleanly, because a bugfix like this is perfect for the planned maintenance release there.

HumorBaby:
Hey now... I _was_ solicited... by an issue! Just because _I_ submitted the issue doesn't mean anything :smiley:

Exirel:
> Ooh, it's _not_ touched in #1429… Well then. :D

Yeah, I tried to be very conservative with my changes. :)

PR #1465, opened 2019-01-30
cli: sopel-config

Exirel:
Yet another PR related to #1434 (and to #1445).

The goal, this time, is to be able to do this:

```
$ sopel-config init -c config_name
... wizard ...
$ sopel-config get -c config_name core nick
Sopel
$ sopel-config set -c config_name core nick GoodBot
$ sopel-config get -c config_name core nick
GoodBot
```

So far, both `init` and `get <section> <option>` work fine, but I'm not sure about the syntax yet: maybe `<section>.<option>` would be better, as in `core.nick` instead of `core nick`.

Anyway, this is just me doing experimental things and trying to make quality of life improvement for Sopel's sysadmin.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -76,6 +76,7 @@ setup(
         'console_scripts': [
             'sopel = sopel.run_script:main',
             'sopel-module = sopel.cli.modules:main',
+            'sopel-config = sopel.cli.config:main',
         ],
     },
 )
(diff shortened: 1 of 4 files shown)


Exirel:
This PR needs an heavy rework, for the same reason #1434 has been closed for now.

PR #1469, opened 2019-02-01
NEWS: Create and enforce format spec

dgw:
With the website using NEWS to generate a whole section of pages, a more formal definition of the NEWS file format seemed prudent.

Now we have a whole set of rules that should (mostly) cover all the basics and prevent anyone (who actually READS the rules) from breaking the website generator with a bad changelog entry.

Oh boy, did this require a LOT of back-and-forth. A second editor pane to keep the spec I just wrote in view at all times, and three or four full passes through the file, and I probably STILL missed something. It's massively cleaned up (and MUCH more consistent) than before though.

This PR is opened against the **6.6.x** branch because it should be merged immediately if another release in that series becomes necessary. However, it should cleanly merge into master also, if the actual next version turns out to be Sopel 7 (which I doubt). Going to pin it also, if GitHub will let me. (**Edit:** Apparently you can't pin PRs. Lame.)


dgw:
Let's merge this (and update the website's submodule pointer) so the live site can have better-formatted changelog entries :shipit:

PR #1470, opened 2019-02-04
irc/bot: implement echo-message

### LINKED ISSUE #526: Add a way for modules to listen on Willie's output
It would be neat if modules could act upon Willie's own output. We discussed this before and I think the general conclusion was "too hard to do". The thing is, it won't be to hard if we decide that callables are uni-directional (one callable won't be able to listen to both input and output).

If we can't build `Trigger()` properly for this, we can simply provide the raw line sent. We should encourage modules to not send out more data in response to parsing sent data, cause that might lead to an infinite loop. We especially don't want `url.py` or anything similar to use that.

Output triggers should use events and regexes and priorities as well.

This will be useful for chanlogs as it will allow as to log what the bot is saying to the channel.

calling the callables should be done after the sending lock is released, because you never want to hold that lock for too long.

As for defining such callables, I suggest `@willie.module.output_handler`

dgw:
Rebased & tweaked version of #1072, which stalled. Resolves #526. Original PR description by @maxpowa follows; see #1072 for the earlier evolution of this patch.

----

Allows modules to listen in on what's being sent by Sopel. Compatible with all the other decorators, as it goes through all the same logic. If `echo-message` is supported on the server and is toggled on by a module, only functions with the `@sopel.module.echo(True)` decorator will receive those echo messages. If `echo-message` isn't enabled, the same behavior still exists but the hostmask will not be as accurate.

### DIFF
--- a/sopel/bot.py
+++ b/sopel/bot.py
@@ -530,8 +530,8 @@ class Sopel(irc.Bot):
                                 match = True
                         if not match:
                             continue
-                    if (not (hasattr(func, 'echo') and func.echo is True) and
-                            trigger.nick.lower() == self.nick.lower()):
+                    if (trigger.nick.lower() == self.nick.lower() and
+                            not func.echo):
                         continue
                     if func.thread:
                         targs = (func, wrapper, trigger)
--- a/sopel/bot.py
+++ b/sopel/bot.py
@@ -530,6 +530,9 @@ class Sopel(irc.Bot):
                                 match = True
                         if not match:
                             continue
+                    if (not (hasattr(func, 'echo') and func.echo is True) and
+                            trigger.nick.lower() == self.nick.lower()):
+                        continue
                     if func.thread:
                         targs = (func, wrapper, trigger)
                         t = threading.Thread(target=self.call, args=targs)
--- a/sopel/irc.py
+++ b/sopel/irc.py
@@ -149,14 +149,21 @@ class Bot(asynchat.async_chat):
             self.send(temp.encode('utf-8'))
         finally:
             self.writing_lock.release()
+
         # Simulate echo-message
         if ('echo-message' not in self.enabled_capabilities and
                 args[0].upper() in ['PRIVMSG', 'NOTICE']):
-            # Since we have no way of knowing the hostmask the IRC server uses
-            # for us, we'll just use something reasonable
+            # Use the hostmask we think the IRC server is using for us,
+            # or something reasonable if that's not available
             host = 'localhost'
             if self.config.core.bind_host:
                 host = self.config.core.bind_host
+            else:
+                try:
+                    host = self.hostmask
+                except KeyError:
+                    pass  # we tried, and that's good enough
+
             pretrigger = PreTrigger(self.nick, ':{0}!{1}@{2} {3}'
                 .format(self.nick, self.user, host, temp))
             self.dispatch(pretrigger)
--- a/sopel/irc.py
+++ b/sopel/irc.py
@@ -150,7 +150,8 @@ class Bot(asynchat.async_chat):
         finally:
             self.writing_lock.release()
         # Simulate echo-message
-        if 'echo-message' not in self.enabled_capabilities:
+        if ('echo-message' not in self.enabled_capabilities and
+                args[0].upper() in ['PRIVMSG', 'NOTICE']):
             # Since we have no way of knowing the hostmask the IRC server uses
             # for us, we'll just use something reasonable
             host = 'localhost'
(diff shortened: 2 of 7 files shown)


sopel/module.py
```diff
@@ -116,6 +116,23 @@ def add_attribute(function):
     return add_attribute
 
 
+def echo(value):
```
dgw: I agree [with @embolalia](https://github.com/sopel-irc/sopel/pull/1072#issuecomment-258612306):

> I don't think the `@echo` decorator should take a value. The default should be definitely not to get echoes, meaning the decorator only needs to mean "yup this thing can get echo messages".

This decorator should just set the `echo` function attribute to `True` if called. That's all it needs to do.

sopel/irc.py
```diff
@@ -149,6 +149,16 @@ def write(self, args, text=None):
...
             self.writing_lock.release()
+        # Simulate echo-message
+        if 'echo-message' not in self.enabled_capabilities:
+            # Since we have no way of knowing the hostmask the IRC server uses
+            # for us, we'll just use something reasonable
+            host = 'localhost'
+            if self.config.core.bind_host:
+                host = self.config.core.bind_host
+            pretrigger = PreTrigger(self.nick, ':{0}!{1}@{2} {3}'
+                .format(self.nick, self.user, host, temp))
```
dgw: This line made me notice the overuse of a variable named `temp` in this function. Renaming that to something sensible should happen after this is merged.

sopel/bot.py
```diff
@@ -530,6 +530,9 @@ def dispatch(self, pretrigger):
                                 match = True
                         if not match:
                             continue
+                    if (not (hasattr(func, 'echo') and func.echo is True) and
+                            trigger.nick.lower() == self.nick.lower()):
+                        continue
```
dgw: I'd put this check first in the enclosing `for` loop, but the order isn't really an issue tbh.

dgw:
Also want to remind myself, when I come back to this, that at least the `find` module's `collectlines()` callable would benefit from `@echo`. Users could then do `Sopel: s/foo/bar/` and have it work as expected. (Yes, my users try to do this sometimes. I do, too, and am always disappointed when it does nothing.)

dgw:
Implemented the pending feedback from before:

  - Remove argument from module.echo() decorator and just always set the attribute to `True` on decorated functions
  - Only simulate `echo-message` for PRIVMSG and NOTICE lines

My tweaks should get a review from one or more @sopel-irc/rockstars before merging 🚀

Also updated the name of `temp` while I was in here, but that caused merge conflicts that I don't feel like dealing with today, so I removed that commit.

dgw:
Cool, I'mma squash that fixup now that you've seen it.

sopel/bot.py
```diff
@@ -530,6 +530,9 @@ def dispatch(self, pretrigger):
                                 match = True
                         if not match:
                             continue
+                    if (not (hasattr(func, 'echo') and func.echo is True) and
```
Exirel: You can replace this by `not getattr(func, 'echo', False)`.

sopel/irc.py
```diff
@@ -149,6 +149,17 @@ def write(self, args, text=None):
             self.send(temp.encode('utf-8'))
         finally:
             self.writing_lock.release()
+        # Simulate echo-message
+        if ('echo-message' not in self.enabled_capabilities and
+                args[0].upper() in ['PRIVMSG', 'NOTICE']):
+            # Since we have no way of knowing the hostmask the IRC server uses
```
Exirel: That might need some tweak after #1537 is released, I think.

dgw: s/might/will/, I know. Depends which goes first (should be this PR, as it's much older).

sopel/bot.py
dgw: I changed the loader's `clean_callable` to always set the `echo` attribute, so I need to revisit this line anyway.

dgw: Done in latest force-push. Now this just checks `not func.echo`, because the loader fills in that attribute (as `False`) for undecorated callables.

Exirel:
As far as I'm concerned, LGTM. I didn't test it live, but the code-style and the overall logic check out.

dgw:
Well, I confirmed via raw logs that the `echo-message` capability is requested, but all of the servers I tried to test on `NAK` it.

It's **hard** to find a network that actually supports this. The [client capability statistics](https://stats.ircdocs.horse/client-caps/) for "curiosity's sake" at `stats.ircdocs.horse` show a measly 5 networks (out of 270 surveyed at last update) advertising support for the `echo-message` capability.

The logic added in this PR should "just work" whether or not the `echo-message`s are real, but I would like to test at least briefly on a network that actually does echo messages back. Stay tuned, I guess.

dgw:
Once I realized I could just look for a network running [Oragono](https://oragono.io/), it was easy to test this. I even found another unrelated bug along the way (#1544).

Updated `find` module to collect Sopel's own lines, just for fun.

PR #1432, opened 2019-01-02
etymology: Back to basics

kwaaak:
Fixes #1248

PR #1474, opened 2019-02-08
etymology: fix py3.3 compatibility

dgw:
Travis wasn't testing on 3.3, so this slipped in with the rewrite for Sopel 6.6.0 (#1432). Hopefully Travis will just work without any additional tweaking, and we'll have tests for Python 3.3 until we eventually do drop support for it (since it's already EOL).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 language: python
 python:
   - "2.7"
+  - "3.3"
   - "3.4"
 git:
   submodules: false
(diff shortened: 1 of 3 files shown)


dgw:
Only took about fifty tries… not bad.

This is kludgy as hell, though. We shouldn't need to specify `setuptools` versions, but pip (at least the version on Travis's Python 3.3 image) isn't smart enough to upgrade what it has to a version that _works_. It sees IPython's requirement and upgrades too far.

PR #1476, opened 2019-02-09
Add docstrings for privilege levels

half-duplex:
At present, the privilege level constants (VOICE, OP, HALFOP, ...) aren't included in the [documentation](https://sopel.chat/docs/plugin.html?highlight=unblockable#module-sopel.module) because they lack docstrings. Docs `make html` appears correct with this patch, though ordered alphabetically with the other constant (NOLIMIT).

### DIFF
--- a/sopel/module.py
+++ b/sopel/module.py
@@ -23,19 +23,34 @@ be used, for example, to allow a user to retry a failed command immediately.
 """
 
 VOICE = 1
-"""Privilege level for the +v channel permission"""
+"""Privilege level for the +v channel permission
+
+.. versionadded:: 4.1
+"""
 
 HALFOP = 2
-"""Privilege level for the +h channel permission"""
+"""Privilege level for the +h channel permission
+
+.. versionadded:: 4.1
+"""
 
 OP = 4
-"""Privilege level for the +o channel permission"""
+"""Privilege level for the +o channel permission
+
+.. versionadded:: 4.1
+"""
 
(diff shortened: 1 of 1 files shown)


dgw:
Would you mind marking them as added in 4.1.0? The new constants (or even the whole feature of bitwise privilege tracking) aren't mentioned in the NEWS file, apparently, but that's the first tag that contains the relevant commit (943279e).

dgw:
Exploring `autodoc_member_order` options will come later, when I have more free time. That should make it possible to sort things in a sensible order.

dgw:
(Last comment, I promise!)

Would you also rebase this on the 6.6.x branch? I can do it if you aren't set up for that; it'll just be a few days until I get a chance.

half-duplex:
Should be done.
I don't see an obvious and easy way to reorder or group only some things, only putting the level constants into a "levels" object or reordering the source so `autodoc_member_order='bysource'` would be sane. Possibly putting all members into `__all__` manually in the desired order?
Thanks!

dgw:
Looks good. Thanks!

Sorting things in the docs is well outside the scope of this PR, and I want to play with it when I do have time. That `__all__` idea is interesting, aside from the risk of forgetting to update it when adding/removing things in the file. But 'bysource' does seem to be the only sane option, doesn't it? 🤔

Exirel:
Lovely! Thanks for the docstrings! ❤️

PR #1477, opened 2019-02-12
use platform to check OS name instead of try, except

deathbybandaid:
Removed the try/except method for a method to detect operating system.

os.name works fairly similar as well.

Testing on win10, and linux works great,,, don't have Mac to test "darwin"

### DIFF
--- a/sopel/run_script.py
+++ b/sopel/run_script.py
@@ -164,7 +164,7 @@ def check_not_root():
         if os.environ.get("USERNAME") == "Administrator":
             raise RuntimeError('Error: Do not run Sopel as Administrator.')
     else:
-        raise RuntimeError('Error: Could not detect Operating Sytem Type, or it is unsupported.')
+        stderr('Warning: Operating System is uncommon. Contact Sopel Devs.')
 
 
 def print_version():
--- a/sopel/run_script.py
+++ b/sopel/run_script.py
@@ -23,6 +23,7 @@ if sys.version_info.major == 3 and sys.version_info.minor < 3:
 import os
 import argparse
 import signal
+import platform
 
 from sopel import run, tools, __version__
 from sopel.config import (
@@ -148,19 +149,22 @@ def build_parser():
 
 
 def check_not_root():
+
     """Check if root is running the bot.
 
     It raises a ``RuntimeError`` if the user has root privileges on Linux or
     if it is the ``Administrator`` account on Windows.
     """
-    try:
+    if platform.system() in ["Linux", "Darwin"]:
         # Linux/Mac
         if os.getuid() == 0 or os.geteuid() == 0:
             raise RuntimeError('Error: Do not run Sopel with root privileges.')
-    except AttributeError:
+    elif platform.system() in ["Windows"]:
         # Windows
         if os.environ.get("USERNAME") == "Administrator":
             raise RuntimeError('Error: Do not run Sopel as Administrator.')
+    else:
+        raise RuntimeError('Error: Could not detect Operating Sytem Type, or it is unsupported.')
 
 
 def print_version():
--- a/sopel/run_script.py
+++ b/sopel/run_script.py
@@ -165,7 +165,7 @@ def check_not_root():
         if os.environ.get("USERNAME") == "Administrator":
             raise RuntimeError('Error: Do not run Sopel as Administrator.')
     else:
-        warnings.warn('Warning: Operating System is uncommon. Contact Sopel Devs regarding %s' % name, warnings.RuntimeWarning)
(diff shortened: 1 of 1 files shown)


sopel/run_script.py
```diff
@@ -148,19 +149,22 @@ def build_parser():
 
 
 def check_not_root():
+
```
Exirel: That extra space isn't necessary.

HumorBaby: I think you should remove this blank line. It makes the docstring for this function into a random multiline string/comment.

sopel/run_script.py
```diff
@@ -148,19 +149,22 @@ def build_parser():
...
         # Linux/Mac
         if os.getuid() == 0 or os.geteuid() == 0:
             raise RuntimeError('Error: Do not run Sopel with root privileges.')
-    except AttributeError:
+    elif platform.system() in ["Windows"]:
         # Windows
         if os.environ.get("USERNAME") == "Administrator":
             raise RuntimeError('Error: Do not run Sopel as Administrator.')
+    else:
+        raise RuntimeError('Error: Could not detect Operating Sytem Type, or it is unsupported.')
```
Exirel: I wouldn't raise an exception, but trigger a warning instead: an uncommon OS might run fine, and user may need to get a notification - that might raise questions, for sure, but also provide Sopel's dev a feedback on which OS is supported and how to support it (if it ever happens).

I think [RuntimeWarning](https://docs.python.org/3/library/exceptions.html#RuntimeWarning) would be a good fit.

See: https://docs.python.org/3.7/library/warnings.html?highlight=warning#available-functions

sopel/run_script.py
```diff
@@ -148,19 +149,22 @@ def build_parser():
...
 
 def check_not_root():
+
     """Check if root is running the bot.
 
     It raises a ``RuntimeError`` if the user has root privileges on Linux or
     if it is the ``Administrator`` account on Windows.
     """
-    try:
+    if platform.system() in ["Linux", "Darwin"]:
```
Exirel: You could use a temporary variable in order to reuse its value in exception (in the last `else` section):

```
name = platform.system()
if name in [...]:
    # doing fine
else:
    raise RuntimeError('Error: Could not detect Operating Sytem Type, or it is unsupported: %s' % name)
```

sopel/run_script.py
deathbybandaid: removed the blank line

deathbybandaid: already removed

sopel/run_script.py
HumorBaby: :+1:, for example Solaris based systems should be able to run Sopel, but this error would preclude that.

sopel/run_script.py
deathbybandaid: got this one fixed

HumorBaby: re: https://github.com/sopel-irc/sopel/pull/1477#discussion_r255975860, you should be able to  add "Solaris" to this list

sopel/run_script.py
deathbybandaid: used stderr from sopel.tools

dgw:
Is this supposed to output the message twice like this? @Exirel you suggested using `warnings`, so is it supposed to do this? Can it not?

```
/Users/dgw/github/sopel/sopel/run_script.py:168: RuntimeWarning: Warning: Operating System is uncommon. Contact Sopel Devs regarding Darwin
  warnings.warn('Warning: Operating System is uncommon. Contact Sopel Devs regarding %s' % opersystem, RuntimeWarning)
```

deathbybandaid:
@dgw that's why I was leaning towards the tried system of sopel.tools stderr

ultimately, the root check can't happen if the OS is unknown anyway

Exirel:
It is supposed to looks like this;

```
Warning (from warnings module):
  File "/home/sopelunking/sopel/run_script.py", line 168
    warnings.warn('Warning: Operating System is uncommon. Contact Sopel Devs regarding %s' % opersystem, RuntimeWarning)
RuntimeWarning: Warning: Operating System is uncommon. Contact Sopel Devs regarding Darwin
```

Which tells the user where the warning was emited first, like an Exception would do.
However, it doesn't stop the program.

For the average user, this can be, indeed, worrying but that's kind of the point. For an expert user, who is working on an uncommon OS, this is the perfect amount of information required to investigate themselve.

I would understand if you prefer using `sopel.tools.stderr`, yet I want to point out that it is a very "uncommon" case to manage.

Exirel:
@deathbybandaid meanwhile, I think you can squash your commits into one! :)

PR #1480, opened 2019-02-13
etymology: quote URL parameters

Exirel:
**Note**: quick fix for 6.6.x

The `ety` command now properly encodes the search term used in the URL
to request www.etymonline.com.

This way, the command `.ety coup d'etat` returns something like:

> "1640s, from French coup d'étate, literally 'stroke of the state'
> (see coup). […]" - https://www.etymonline.com/word/coup%20d%27etat

Instead of this:

> Can't find the etymology for "coup d'etat".
> Try https://www.etymonline.com/search?q=coup d'etat

Which was very unfortunate, because a) it didn't find the answer
and b) it gave an invalid URL to search for an answer.

It also encodes the search URL when it does not found a result:

    <me> .ety not found
    <Sopel> Can't find the etymology for "not found".
            Try https://www.etymonline.com/search?q=not%20found

So that the user can copy and paste the URL.

**Edit**: fix grammar

PR #1481, opened 2019-02-13
search: quote URL parameters

Exirel:
**Note**: quick fix for 6.6.x

Following #1480 my eyes turned to the search module, and they found the same kind of problems.

So I applied the same solution, and it works fine:

    <@Exirel> .suggest 2+2
    <Sopel> 4
    <@Exirel> .suggest coup d'ét
    <Sopel> coup d'état

_(also yes, I'm so French that I can't resi**s**t to use "coup d'état" in my example)_

### DIFF
--- a/sopel/modules/search.py
+++ b/sopel/modules/search.py
@@ -30,9 +30,13 @@ r_bing = re.compile(r'<h2(?: class=" b_topTitle")?><a href="([^"]+)"')
 
 
 def bing_search(query, lang='en-US'):
-    base = 'https://www.bing.com/search?mkt=%s&q=' % lang
-    bytes = requests.get(base + query).text
-    m = r_bing.search(bytes)
+    base = 'https://www.bing.com/search'
+    parameters = {
+        'mkt': lang,
+        'q': query,
+    }
+    response = requests.get(base, parameters)
+    m = r_bing.search(response.text)
     if m:
         return m.group(1)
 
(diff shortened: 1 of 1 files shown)


Exirel:
Ah! Gotta fix the checkstyle first.

dgw:
> _I can't resit_

resist?

_(spoiler: I'm going to say the same thing I did in #1480 about your commit message(s) in PR review :stuck_out_tongue:)_

Exirel:
> _(spoiler)_

I deserve that 100% my fault.

Exirel:
> I also am a little curious why you went with `requests.get`'s parameter handling for this, but `sopel.web.quote()` in #1480.

Because in #1480 the "word" is part of the URL's path, and not the URL's query string: `requests.get` encode the query string parameters, but doesn't touch the URL itself.

As for the 's'... I've no excuse.

dgw:
Fair enough on query params vs. URL path components.

Also, I forgot to ask: Did you intentionally remove the comment about #885? We probably don't need it, honestly, but I was surprised to see it removed in the diff.

Exirel:
I did, because query parameters **must** be encoded. If it was a weird quirk, the comment would be relevant, but it isn't: there was a bug because the query parameter wasn't encoded properly.

Exirel:
@dgw should be good now.

PR #1483, opened 2019-02-14
instagram: remove `u` prefix (and other cleanups)

Exirel:
**Note**: based on 6.6.x

It all started with a small hitch I wanted to scratch off my head, then I saw minor problems here and there, so I fixed them all:

* unicode prefix for string while `unicode_literals` is in the import from `__future__`
* bad import order
* usage of specific import (`from sopel.module import rule` instead of `from sopel import module`)

**Edit:** no more replacement of `module.rule` for now.

### DIFF
--- a/sopel/modules/instagram.py
+++ b/sopel/modules/instagram.py
@@ -72,7 +72,7 @@ def parse_insta_json(json):
         # Strip newlines
         icap = icap.replace('\n', ' ')
         # Truncate caption
-        icap = (icap[:256] + u'…') if len(icap) > 256 else icap
+        icap = (icap[:256] + '…') if len(icap) > 256 else icap
     except Exception:  # TODO: be specific
         icap = False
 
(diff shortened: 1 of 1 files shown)


sopel/modules/instagram.py
```diff
@@ -8,34 +8,25 @@
...
 
 # TODO: Parse Instagram profile page
 
 
-@rule(instagram_regex)
-def instaparse(bot, trigger):
+@module.url(instagram_regex)
+def instaparse(bot, trigger, match=None):
     # Get the embedded JSON
     json = get_insta_json(trigger.group(1))
```
dgw: Shouldn't this use `match` instead of `trigger`, now that this function is using the `url` decorator?

Exirel: Complicated, since it would require a change in the `instagram_regex` I'm not comfortable with at the moment, in order to catch the whole URL in one group. But that's doable, I guess. Not sure I want to risk it for now.

dgw:
So, changing this to use `@url` instead of `@rule` **will** prevent processing Instagram links on Sopel instances where the generic `url.py` module is disabled.

I'm not sure how to deal with that. It's very much less than ideal that disabling `url.py` (because one doesn't want just any random URL getting title'd) would also disable functionality of modules that were explicitly left enabled. Going to make a separate issue for that bit.

Exirel:
> Going to make a separate issue for that bit. (Edit: #1489)

I think that's the good call here!

dgw:
@Exirel In the interest of not "breaking" the Instagram module (on instances with `url.py` disabled) until #1489 is fixed, though, would you absolutely hate removing the migration to `@url` from this PR and reserving it for a later one? The rest of the cleanups are :+1:.

Exirel:
Will do that.

sopel/modules/instagram.py
dgw: Marking resolved; the change to `@url` has been postponed.

PR #1484, opened 2019-02-14
unicode: remove `u` prefix (and other cleanups)

Exirel:
**Note**: based on 6.6.x (and yes, this looks like #1483 a lot - but it's a different plugin!)

The `hex` function returns a `str` object, and you can not use a bytestring object when you need a unicode one. Which is a problem, because py2's str is py3's bytestring, and the py3's str is
py2's unicode.

The previous implementation used to convert everything into `str` and to call it a day. It "works", but requires some hack, for instance:

    ord(u'' + arg)
    point.rjust(4, str('0')

Which can be quite confusing. To make things easier to read, the new implementation defines `unicode` for py3 as an alias for `str`, and uses it in the code once. From that line:

    unicode(hex(ord(char)))[2:]

Everything is working with unicode strings (which is more consistent) instead of working with bytestrings in py2, and unicode in py3.

That being said, I'm amazed by the hack that was in place before, because it shows how complex the situation trully is. **Unicode is hard!**

### DIFF
--- a/sopel/modules/unicode_info.py
+++ b/sopel/modules/unicode_info.py
@@ -5,9 +5,10 @@
 # Licensed under the Eiffel Forum License 2.
 from __future__ import unicode_literals, absolute_import, print_function, division
 
-import unicodedata
 import sys
-from sopel.module import commands, example, NOLIMIT
+import unicodedata
+
+from sopel import module
 
 if sys.version_info.major >= 3:
     # Note on unicode and str (required for py2 compatibility)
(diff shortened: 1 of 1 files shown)


sopel/modules/unicode_info.py
```diff
@@ -5,22 +5,48 @@
...
-import unicodedata
 import sys
-from sopel.module import commands, example, NOLIMIT
+import unicodedata
+
+from sopel import module
 
 if sys.version_info.major >= 3:
+    # Note on unicode and str (required for py2 compatibility)
+    # the `hex` function returns an `str`, both in py2 and py3
```
dgw:
```suggestion
    # the `hex` function returns a `str`, both in py2 and py3
```

sopel/modules/unicode_info.py
```diff
@@ -5,22 +5,48 @@
...
+    # however, an `str` is an unicode string in py3, but a bytestrings in py2
```
dgw:
```suggestion
    # however, a `str` is a unicode string in py3, but a bytestring in py2
```

sopel/modules/unicode_info.py
Exirel: I swear I tried really hard to pay attention to these details. :stuck_out_tongue:

Exirel:
Grammar fixed.

Issue #1482, opened 2019-02-14
UnrealIRCd NAMES vs WHO mode symbols cause AssertionError


half-duplex:
```
>>1550121698.9370964    JOIN #test
<<1550121705.275868     :sopel!~sopel@bothost JOIN :#test
<<1550121705.2780008    :alpha.example.com 353 sopel = #test :sopel ~@me
<<1550121705.2786467    :alpha.example.com 366 sopel #test :End of /NAMES list.
<<1550121714.5684164    :alpha.example.com 352 sopel #test ~me myhost alpha.example.com me Hr~ :0 me
<<1550121714.5691297    :alpha.example.com 315 sopel #test :End of /WHO list.
<<1550121730.9563427    :me!~me@myhost PRIVMSG sopel :.exec bot.privileges["#test"]["me"]
>>1550121730.958703     PRIVMSG me :20
<<1550121741.7689545    :me!~me@myhost PRIVMSG sopel :.exec bot.channels["#test"].privileges["me"]
>>1550121741.7704542    PRIVMSG me :16
```

`Sopel 6.6.2 (running on python 3.7.2)`
`Unreal3.2.10.6`

It seems UnrealIRCd's RPL_NAMREPLY (353) reports all channel mode symbols (`~@me`, OP and OWNER) but its RPL_WHOREPLY (352) only gives the highest (`Hr~`, OWNER).
[handle_names](https://github.com/sopel-irc/sopel/blob/90a091614f7428fc6f419f5a50810850446b36f6/sopel/coretasks.py#L170) then sets `bot.privileges[][]` to 20 (OP|OWNER) and [_record_who](https://github.com/sopel-irc/sopel/blob/90a091614f7428fc6f419f5a50810850446b36f6/sopel/coretasks.py#L707) sets `bot.channels[].privileges[]` to 16 (OWNER).

It will explode the next time that user's modes are changed, with `AssertionError (file ".../sopel/coretasks.py", line 246, in track_modes)` since the two don't match.

dgw:
Exactly the kind of thing that assertion was meant to catch. I'm glad of leaving it in the released code, now.

On the one hand, this reveals that `handle_names()` only updates `bot.privileges` and `_record_who()` only updates `bot.channels`.

On the other hand, fixing that inconsistency wouldn't fix _this_ bug. Sopel would still have incorrect mode masks stored, depending on which reply type it received last (gotta love race conditions).

[...]

PR #1485, opened 2019-02-15
Replace deprecated bot.privileges accesses

half-duplex:
`bot.privileges[channel][user]` -> `bot.channels[channel].privileges[user]`. Smoke tested.

### DIFF
--- a/sopel/module.py
+++ b/sopel/module.py
@@ -330,7 +330,7 @@ def require_privilege(level, message=None):
             # If this is a privmsg, ignore privilege requirements
             if trigger.is_privmsg:
                 return function(bot, trigger, *args, **kwargs)
-            channel_privs = bot.privileges[trigger.sender]
+            channel_privs = bot.channels[trigger.sender].privileges
             allowed = channel_privs.get(trigger.nick, 0) >= level
             if not trigger.is_privmsg and not allowed:
                 if message and not callable(message):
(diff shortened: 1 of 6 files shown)


dgw:
Squash please!

Also, your PR title is much clearer than the commit message. We're not replacing `bot.privileges`, but only uses of it in core code.

dgw:
Now that I think about it, if this had been done way back when `bot.privileges` was first deprecated, I bet `bot.channels` never would have been released broken the way it was.

We should definitely be using the "new and improved" things internally, not the old deprecated stuff. This should have happened ages ago.

sopel/test_tools.py
```diff
@@ -47,7 +47,9 @@ def __init__(self, nick, admin=False, owner=False):
         self.nick = nick
         self.user = "sopel"
 
-        self.channels = ["#channel"]
+        channel = sopel.tools.Identifier("#Sopel")
+        self.channels = {
+            channel: sopel.tools.target.Channel(channel)}
```
dgw: What's up with this line break? The line wouldn't be that long without it.

half-duplex: I think I was using a longer variable name than `channel` when I made it.

dgw:
@half-duplex This is approved, but waiting on a hotfix for #1482 (consistency between `bot.privileges` and `bot.channels` data). Without it, this could cause modules to start behaving differently with certain ircds, due to the mismatched mode info, and we don't want to introduce such changes in a patch release.

PR #1486, opened 2019-02-15
Replace url_callbacks with @url decorators

half-duplex:


dgw:
(Retitled because it touches more than just `url.py`)

dgw:
Blocked on #1489

half-duplex:
The bugzilla module has a few options to remove the url_callbacks use:
- Needs a function to add a url handler, to add one for each configured domain from `setup()`
- Needs a module.CONTINUE or equivalent, to add a url handler for all `/show_bug.cgi?id=` but allow fallback to title grabber if it's not bugzilla
- Use only a `rule()`?

sopel/modules/instagram.py
```diff
@@ -8,39 +8,30 @@
...
-    del bot.memory['url_callbacks'][instagram_pattern]
-
 # TODO: Parse Instagram profile page
 
 
-@module.rule(instagram_regex)
-def instaparse(bot, trigger):
+@module.url(instagram_regex)
+def instaparse(bot, trigger, match=None):
+    match = match or trigger
```
Exirel: You don't need the trigger anymore.

sopel/modules/instagram.py
```diff
@@ -8,39 +8,30 @@
...
-def shutdown(bot):
-    del bot.memory['url_callbacks'][instagram_pattern]
-
 # TODO: Parse Instagram profile page
 
 
-@module.rule(instagram_regex)
-def instaparse(bot, trigger):
+@module.url(instagram_regex)
+def instaparse(bot, trigger, match=None):
```
Exirel: The `match` parameter should be always provided when `@url` is used instead of `@rule`.

```suggestion
def instaparse(bot, trigger, match):
```

sopel/modules/reddit.py
```diff
@@ -25,27 +27,13 @@
...
-
-
-def shutdown(bot):
-    del bot.memory['url_callbacks'][post_regex]
-    del bot.memory['url_callbacks'][user_regex]
-
-
-@rule('.*%s.*' % post_url)
+@url(post_url)
 def rpost_info(bot, trigger, match=None):
```
Exirel: The `match` parameter should be always provided when `@url` is used instead of `@rule`.

```suggestion
def rpost_info(bot, trigger, match):
```

sopel/modules/reddit.py
```diff
@@ -162,9 +150,9 @@ def redditor_info(bot, trigger, match=None):
 
 
 # If you change the groups here, you'll have to change some things above.
-@rule('.*%s.*' % user_url)
-def auto_redditor_info(bot, trigger):
-    redditor_info(bot, trigger)
+@url(user_url)
+def auto_redditor_info(bot, trigger, match=None):
```
Exirel: The `match` parameter should be always provided when `@url` is used instead of `@rule`.

```suggestion
def auto_redditor_info(bot, trigger, match):
```

sopel/modules/wikipedia.py
```diff
@@ -105,7 +101,8 @@ def mw_snippet(server, query):
     return snippet['extract']
 
 
-@rule(r'.*\/([a-z]+\.wikipedia\.org)\/wiki\/((?!File\:)[^ ]+).*')
+page_url = r'/([a-z]+\.wikipedia\.org)/wiki/((?!File\:)[^ ]+)'
+@url(page_url)
 def mw_info(bot, trigger, found_match=None):
```
Exirel: The `match` parameter should be always provided when `@url` is used instead of `@rule`.

```suggestion
def mw_info(bot, trigger, found_match):
```

half-duplex:
bugzilla still needs to be done, and the options above are still available, but now that #1508 is merged I can use that and kick the "would this be better" can to someone more ambitious 💤

sopel/modules/instagram.py
```diff
@@ -9,8 +9,6 @@
 from __future__ import unicode_literals, absolute_import, print_function, division
 
 from datetime import datetime
-import re
-
```
dgw: Leave this blank line in; when we start enforcing more import style rules, those will insist that stdlib and third-party imports be separated by a blank line.

sopel/modules/wikipedia.py
```diff
@@ -10,11 +10,10 @@
 
 import re
 import sys
-
```
dgw: Nope. See above.

sopel/modules/instagram.py
half-duplex: Forgot `requests` isn't stdlib, thanks >.>

Exirel:
@half-duplex very well played! Conflict fixed, code is simpler than before, and the goal is achieve.

:rocket: :ship:

PR #1487, opened 2019-02-15
use platform to check OS name instead of try, except

deathbybandaid:


deathbybandaid:
killed PR #1477 to PR from a different branch as well as squash the committs

sopel/run_script.py
```diff
@@ -153,14 +155,17 @@ def check_not_root():
...
         # Linux/Mac
         if os.getuid() == 0 or os.geteuid() == 0:
             raise RuntimeError('Error: Do not run Sopel with root privileges.')
-    except AttributeError:
+    elif opersystem in ["Windows"]:
         # Windows
         if os.environ.get("USERNAME") == "Administrator":
             raise RuntimeError('Error: Do not run Sopel as Administrator.')
+    else:
+        warnings.warn('Warning: Operating System is uncommon. Contact Sopel Devs regarding %s' % opersystem, RuntimeWarning)
```
dgw: I have some ideas for making this better.

```suggestion
        stderr("Warning: %s is an uncommon operating system platform. Sopel should still work, "
               "but please contact Sopel's developers if you experience issues." % opersystem)
```

Using `stderr` instead of `warnings.warn` makes the output cleaner. There's no advantage to the more verbose output of a `warnings` warning in this case, as this is really just an advisory message.

I also reworded the text pretty significantly, to emphasize that Sopel will probably run just fine anyway.

PR #1488, opened 2019-02-15
Py2 warning

deathbybandaid:
This will simply display a warning about py2.x EOL

It won't exit, but gives people most of a year to straighten out their python versions.

Not saying Sopel should/will drop py2.7 support Jan 1st 2020, but we shouldn't tie our hands with this either.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -37,6 +37,8 @@ Sopel requires ``backports.ssl_match_hostname`` to be installed. Use
 ``yum install python-backports.ssl_match_hostname`` to install it, or download
 and install it manually `from PyPI <https://pypi.org/project/backports.ssl_match_hostname>`_.
 
+Note: Python 2.x is near end of life. Sopel support at that point is TBD.
+
 In the source directory (whether cloned or from the tarball) run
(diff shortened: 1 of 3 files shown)


dgw:
I think this could be one commit, titled something like "Add warnings about py2 EOL".

For the record, Sopel 7 might not even come out much before Jan 1, 2020. I'll probably put this in anyway, but it could be worth putting the cutoff date in the warning text instead of just saying "is near end of life".

Actually, we could probably just say outright that Sopel 8 is unlikely to still support Python 2. There's no _way_ 8.0.0 will be done before 2020 based on the current typical version lifecycle. What say you?

deathbybandaid:
At my count, there are 26 files that use `sys.version_info` to make stuff work between py versions

dgw:
Finding the ones that use `try`/`except` to import modules that moved will be useful, too, but harder (because you can't just `grep` for `version_info`) :laughing:

dgw:
@deathbybandaid I was thinking you were going to squash this down. Were you?

Issue #1401, opened 2018-10-28
.reload for modules with subclasses that call super leads to TypeError


HumorBaby:
When a module that defines subclasses that call `super`, a reload will lead to an exception because `TypeError: super(type, obj): obj must be an instance or subtype of type`. Apparently in Python 2 (not sure about 3), when reloading modules, `isinstance` checks for running super will fail due to internal memory changes. 

See:
http://thomas-cokelaer.info/blog/2011/09/382/
https://thingspython.wordpress.com/2010/09/27/another-super-wrinkle-raising-typeerror/

HumorBaby:
Just don't use super. 
Do `Exception.__init__(self, msg)` instead of `super(MyException, self).__init__(msg)`.

Exirel:
I've been reading about this, and there is no good solution here in Python 2, and there is an excellent one with Python 3. Here is a Python shell session:

```
>>> import sopelunker
>>> import importlib
>>> 
>>> 
>>> c = sopelunker.Child()
>>> c.sopelunking()
making PR #3400
>>> c.py2compat()
making PR with py2 compat: #9007
>>> 
>>> 
>>> importlib.reload(sopelunker)
<module 'sopelunker' from '/home/sopel/workspace/unking/sopelunker.py'>
>>> 
>>> 
>>> c.sopelunking()
making PR #9162
>>> c.py2compat()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/sopel/workspace/unking/sopelunker.py", line 16, in py2compat
    super(Child, self).py2compat(random.randint(1000, 9999))
TypeError: super(type, obj): obj must be an instance or subtype of type
```

And here is the Python code itself:

```
import random

class Mother(object):
    def sopelunking(self, number):
        print('making PR #%d' % number)

    def py2compat(self, number):
        print('making PR with py2 compat: #%d' % number)


class Child(Mother):
    def sopelunking(self):
        super().sopelunking(random.randint(1000, 9999))

    def py2compat(self):
        super(Child, self).py2compat(random.randint(1000, 9999))
```

With Python 3, you can use `super()` directly, unlike Python 2, where you'll get an error: `TypeError: super() takes at least 1 argument (0 given)`.

So that's not something that can be fixed in Python 2, because that's how reload works, and there is nothing you can do about it - at least, nothing you can do while keeping a clean code, ready for Python 3.

HumorBaby:
[...]

Issue #1395, opened 2018-10-10
.reload doesn't run shutdown routines


Hanicef:
When `.reload` is invoked, none of the shutdown routines are ran before the modules are being loaded and initalized again.

The following text is from the output logs when I was testing this (Lines starting with `#` are added as additional info, and is not in the actual output):
```
Welcome to Sopel. Loading modules...




Registered 52 modules,
0 modules failed to load

Connecting to localhost:6667...
Connected.
# .reload is invoked here

Welcome to Sopel. Loading modules...




Registered 52 modules,
0 modules failed to load

# bot is signalled to shut down here
Got quit signal, shutting down.
ERROR:sopel.modules.sopel.irc:ERROR received from server: Closing Link: Sopel[localhost] (Quit: Closing)
Calling shutdown for 6 modules.
calling reddit.shutdown
calling bugzilla.shutdown
calling media.shutdown
calling reddit.shutdown
Error calling shutdown method for module reddit:re.compile('https?://(?:www\\.|np\\.)?reddit\\.com/r/(.*?)/comments/([\\w-]+)')
calling bugzilla.shutdown
calling media.shutdown
Closed!
```

Note that at the second module loading, there are no shutdown routines being called on modules.

dgw:
Merged #1412; this should be fixed now.

Issue #871, opened 2015-08-29
Watch modules for file changes and reload


hftf:
I suppose it should be enabled by a config flag, since this is mostly useful for developing and testing new modules, but not very useful in production when modules are assumed to be stable.

elad661:
I don't think it's a good idea, it would cause too much reloading when I'm working on a module with auto-saving enabled...

dgw:
@elad661 That's why it would be enabled by a config flag.

embolalia:
I'd prefer an option to the command, but yeah. It would be optional.

Exirel:
I wonder if a `inotify` script would do the trick, without having to add anything to sopel's code.

Like, `sopel --quit && sopel` shouldn't be too complicated to launch from `inotify`!

dgw:
Or just `sopel --restart` when #1333 is merged in?

I'm not against adding a new command-line flag, either.

Maybe Sopel could support a `--watch` mode that watches modules for changes during development, along the lines of `bundle exec jekyll server --incremental` or `webpack`'s `watch: true` config setting.

Alternatively (or additionally), `sopel --reload` could become a thing at some point. That could be tricky UX-wise because such an option is often used in e.g. HTTP servers to reload a config file, not plugins, but it's a thought. Might happen, might not.

Exirel:
This conversation reminds me that I wish #1429  and #1434 to be merge before working again on the `run_script` file, or anything that is related to the command line. Right now, as a user and a third-party dev, I'm not very happy with the toolings around command line, and there are a lot of obscure behavior here and there in this file.

That's why, even though I'm not against a `--restart/reload` command, I would suggest a low priority on that.

Exirel:
@dgw after further consideration, I don't think it's reasonable to plan this for Sopel 7.0, and probably not for the Sopel 7.x branch anyway. I'd suggest to remove the milestone for now.

dgw:
Mm, implementing this was predicated on fixing the issues with reloading. And we've had this discussion several times now—we can't properly fix reloading until we _at minimum_ no longer support old versions of Python (2.7, 3.3) that don't support the internal machinery needed for better reloading support.

There might be other things we need to do, too. It's interesting that you revived _this_ particular issue today, because I was just working on a new feature for our Twitter plugin and had reloading issues. That is, telling Sopel to reload it didn't work. Obviously #1056 is still a problem. I should just bump its milestone along with this issue's, since they both ultimately depend on better reload machinery.

[...]

PR #1490, opened 2019-02-16
plugin: reload using new interface

Exirel:
**Based on** #1479 - this PR starts at the `plugin: add AbstractPluginHandler.reload` commit.

Using the new plugin interface defined in `sopel.plugins`, I replaced how plugins are reloaded in `sopel.modules.reload`:

* `sopel.bot.Sopel` now exposes methods to add, remove, and reload plugins
* `sopel.modules.reload` uses the bot's method to reload plugins (all of them, or only one)
* `sopel.plugin.handlers.AbstractPluginHandler` defines a new `reload` method
* `sopel.plugin.handlers.AbstractPluginHandler`'s implementations use the bot's new methods to add or remove the plugin

## It Is Not Perfect And Here Is Why

There are problems unsolved and questions unanswered here:

1. Problem: module tree is not reloaded, only the main file (can and must be fixed),
2. Problem: Python's `imp.reload` and `importlib.reload` do not remove names undefined in a new version of a reloaded module (fix possible only with Py3.4+),
3. Problem: Job's management is still a mess (same as before),
4. Question: should we use (updated) configuration?

### Problem 1: Reload Module Tree

This one is a big problem to solve, because there is more than meets the eye:

* reloading the module tree must not reload built-in or externals modules,
* but because of that a reload may fail if an external module has been changed in a way that it requires a reload...

This leads me to conclude that `.reload` should not be used in production in these cases. And they are hard as hell to fix, or even to warn the user about.

So it's half a problem, half a question. And I don't have the answer either way.

### Problem 2: Python's `reload` Is Weird

**Note**: this behavior is the one currently in place in Sopel. **This is not new**.

Both `imp.reload` and `importlib.reload` have this weird behavior, explained by this line from the documentation ([python2 reload](https://docs.python.org/2/library/functions.html#reload) / [python3 importlib.reload](https://docs.python.org/3.5/library/importlib.html#importlib.reload)):

> If the new version of a module does not define a name that was defined by the old version, the old definition remains.

Meaning that:

1. a module has a function `something`, is loaded at Sopel's startup
2. this module is modified, and `something` does not exist anymore
3. `.reload <this-module>` command is used
4. `something` still exists and is seen by Sopel, no matter what we do, 

However, this can be fixed... with Python 3.4's public internals (namely, `importlib.util` and `importlib.machinery`) (see a [snippet of code](https://www.irccloud.com/pastebin/tE6j0SSM/load_file_py3)). By the way, using Py3.4's public machinery would fix Problem 1 too...

### Problem 3: Job Are Weird Too

**Note**: this behavior is the one currently in place in Sopel. **This is not new**.

I didn't change any of the "unregister a job" behavior, so this is something that requires more work. The good part is: this can be done in a future PR, without any consequences so far, because it's already like that right now.

### Question: Does `.reload` Should Use Updated Configuration?

In Sopel's current version, `.reload` uses the current configuration (potentially modified since startup), to gather a list of plugins:

* if a plugin has been disabled, it might not be unloaded properly,
* if a plugin has been enabled, it will load it, somehow

And that **is a problem for me**, because it means inconsistency and unexpected behavior. I would rather have something stable, that always work, in a predicable way, than the current behavior.

That's why this new version reloads the list of plugins as it exists at the moment in the bot's memory (ie. in `bot._plugins`). We could, if we want, change this behavior to unload a plugin (enabled or not), and to load only the plugin if it's still enabled, and enable all new enabled plugins.

And that's the question: what should we do?

### DIFF
--- a/docs/source/plugin.rst
+++ b/docs/source/plugin.rst
@@ -5,7 +5,12 @@ A Sopel plugin consists of a Python module containing one or more
 ``callable``\s. It may optionally also contain ``configure``, ``setup``, and
 ``shutdown`` hooks.
 
-.. py:method:: callable(bot, trigger)
+.. py:function:: callable(bot, trigger)
+
+    :param bot: the bot's instance
+    :type bot: :class:`sopel.bot.Sopel`
+    :param trigger: the object that triggered the call
+    :type trigger: :class:`sopel.trigger.Trigger`
 
     A callable is any function which takes as its arguments a
     :class:`sopel.bot.Sopel` object and a :class:`sopel.trigger.Trigger`
@@ -21,7 +26,10 @@ A Sopel plugin consists of a Python module containing one or more
     Note that the name can, and should, be anything - it doesn't need to be
     called "callable".
 
-.. py:method:: setup(bot)
+.. py:function:: setup(bot)
+
+    :param bot: the bot's instance
+    :type bot: :class:`sopel.bot.Sopel`
 
     This is an optional function of a plugin, which will be called while the
     module is being loaded. The purpose of this function is to perform whatever
@@ -41,7 +49,10 @@ A Sopel plugin consists of a Python module containing one or more
     execution of this function. As such, an infinite loop (such as an
     unthreaded polling loop) will cause the bot to hang.
 
-.. py:method:: shutdown(bot)
+.. py:function:: shutdown(bot)
+
+    :param bot: the bot's instance
+    :type bot: :class:`sopel.bot.Sopel`
 
     This is an optional function of a module, which will be called while the
     bot is quitting. Note that this normally occurs after closing connection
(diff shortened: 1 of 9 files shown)


Exirel:
Looking at issues:

* #1395 fixed: calling shutdown on reload
* #1056 half fixed: the job/interval command is not fixed (#831 can be considered a duplicate)
* #1401 not fixed: can not be done with Python 2
* #871 not fixed: it would make it easier, though

Overall, "reload" in Python is very hard, and I'm not sure we can really fix it. There are too many use cases and corner cases where we don't have the right tools for that.

Exirel:
So, after further consideration, I've some issues to fix first.

Exirel:
So, I decided to merge this branch into #1479 because it makes my job way easier, and I think it does not make sense anymore to separate the new loading from reloading using new loading.

So, I keep that around for history (I keep the branch as-is, too), but everything will happen in #1479 now!

dgw:
> So, I decided to merge this branch into #1479 because it makes my job way easier, and I think it does not make sense anymore to separate the new loading from reloading using new loading.

Makes sense to me. :)

PR #1492, opened 2019-02-28
admin: save configuration on .join/.part command

### LINKED ISSUE #563: Save joined channels in the configuration
Right now when you add a channel with !join it does not save the channel to the config and the channel is not rejoined after restarting the bot.

Personally I would like to see this added, however also with the option for a delay joining channels so that bots who join a lot of channels don't flood the server.

Chances are, that I'll write a module to do this as it's something I need personally.
However in my case it would not make sense to write this into the default.cfg, as more than 10 channels would already be too difficult to maintain in the config file.

Exirel:
For good measure, I added a configuration save on the `part` command too.

Fixes #563

### DIFF
--- a/sopel/modules/admin.py
+++ b/sopel/modules/admin.py
@@ -86,6 +86,18 @@ def temporary_join(bot, trigger):
     _join(bot, channel, key, save=False)
 
 
+def _part(bot, channel, msg=None, save=True):
+    bot.part(channel, msg or None)
+
+    if save and channel in bot.config.core.channels:
+        bot.config.core.channels = [
+            chan
+            for chan in bot.config.core.channels
+            if chan != channel
+        ]
+        bot.config.save()
+
+
 @sopel.module.require_privmsg
 @sopel.module.require_admin
 @sopel.module.commands('part')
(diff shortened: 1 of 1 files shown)


sopel/modules/admin.py
```diff
@@ -56,7 +56,11 @@ def join(bot, trigger):
         return
     elif not key:
         bot.join(channel)
+        if channel not in bot.config.core.channels:
+            bot.config.core.channels = bot.config.core.channels + [channel]
+            bot.config.save()
     else:
+        # TODO: a way to store a channel's key in configuration
```
dgw: Does this help?

https://github.com/sopel-irc/sopel/blob/af308621198f4e825063efbe472f8ec42d53e5d4/sopel/bot.py#L259-L266

dgw: Basically, you should be able to just store (in `bot.config.core.channels`) `' '.join(channel, key)` and have it work after restart unless the channel key changes.

Exirel: Ooooooooh... that is super weird, but yes, it helps!

dgw: It's a little weird, but that's also how IRC handles channel keys. `JOIN #channel key` is how a client joins a single channel with a key. (Sopel doesn't deal with more than one channel at a time, so the proper IRC syntax for multiple channels—`JOIN #chan1,#chan2 key1,key2`—isn't relevant… yet.)

Exirel:
@dgw the "save password" thing is fixed. One test failed because of some random error, unrelated.

Exirel:
Rebased upon master. @dgw merge when you went :)

Issue #1471, opened 2019-02-06
Rework internal APIs for Command Line Tools


Exirel:
After working on `sopel.run_script` and some internal features regarding configuration and modules loadings, it appears the internal APIs require a bit of love and care. There is a lot going on with that, so I'll try to list the things I think about:

* move CLI related code to `sopel.cli.*` sub-package,
* replace `sopel.run_script` by `sopel.cli.run` and adapt the setup.py's entry point accordingly,
* centralize how configuration files are found and loaded,
* centralize common CLI options (like `-c config`),

and I'm sure I'm forgetting some part.

For the record, I personaly worked on these PR before: #1429 #1434 #1445 #1465 - and others have worked toward the same goal (PR #1464 is a good example of that). Some PR have already been merged and some are closed until this issue is fixed.

**Note**: this is not an issue to discuss where the config files must be, or how to load modules. This is about making sure all the good ideas around how to structure internals APIs are shared.

HumorBaby:
Feature request: allow for configuration of a specific module (as opposed to the current `--configure-modules` flag that asks `y/n` for each possible module).

Let me know if you make an "open call for patches" regarding this, and I will gladly contribute what I can :)

Exirel:
> Feature request: allow for configuration of a specific module

Hm. In #1465 I played with the idea of a `sopel-config init`, and something like `sopel-module configure <module>` or `sopel-config module <mod-name>` or something similar might work.

For that, I think we need to make sure all Configuration Wizards comply with the configuration loading, and that we can find the module as Sopel does when running it.

HumorBaby:
Ah! I see now that's what `... init` was meant to do.

Exirel:
After a review of @dgw about `add_config_arguments` renamed `add_common_arguments`, I was thinking that there are 2 common options:

* `-v/--version`: show the program's version
* `--verbose` and `--quiet`: more/less output on stdout

dgw:
@Exirel I don't have time this morning to dig into code, but are we already using `-v` for version? Because it might be more useful to have only long-form for that and reserve the short args `-v`/`-q` as a pair for verbose/quiet.

If we are already using `-v`, we should (IMO) reassign that and give `--version` the short form of `-V` instead. After adequate warning, of course.

Exirel:
@dgw I fully agree @ --version/-V

Exirel:
[...]

PR #1493, opened 2019-02-28
cli: move sopel.run_script to sopel.cli.run

Exirel:
**Related** to #1471

This move the `sopel/run_script.py` file to `sopel/cli/run.py`, and updates the `sopel` command line tool's interface:

* `sopel start` replaces the default behavior (run the bot)
* `sopel stop` replaces the ``--quit/--kill`` options
* `sopel restart` replaces the ``--restart`` option
* `sopel configure` replaces the `-w/--configure-all/--configure-modules` options

The `-v` option for "version" is deprecated, `-V/--version` should be used instead.

By default, `sopel` works as before (100% backward compatible), but adds some warning here and there when needed (when using --restart/--quit/--kill for example).

This is all thanks to #1472 being merged.

### DIFF
--- a/sopel/cli/run.py
+++ b/sopel/cli/run.py
@@ -54,15 +54,15 @@ def run(config, pid_file, daemon=False):
     delay = 20
     # Inject ca_certs from config to web for SSL validation of web requests
     if not config.core.ca_certs:
-        stderr('Could not open CA certificates file. SSL will not '
-               'work properly!')
+        tools.stderr(
+            'Could not open CA certificates file. SSL will not work properly!')
 
     def signal_handler(sig, frame):
         if sig == signal.SIGUSR1 or sig == signal.SIGTERM or sig == signal.SIGINT:
-            stderr('Got quit signal, shutting down.')
+            tools.stderr('Got quit signal, shutting down.')
             p.quit('Closing')
         elif sig == signal.SIGUSR2 or sig == signal.SIGILL:
-            stderr('Got restart signal.')
+            tools.stderr('Got restart signal.')
             p.restart('Restarting')
 
     while True:
@@ -85,7 +85,7 @@ def run(config, pid_file, daemon=False):
         except Exception:  # TODO: Be specific
             trace = traceback.format_exc()
             try:
-                stderr(trace)
+                tools.stderr(trace)
             except Exception:  # TODO: Be specific
                 pass
             logfile = open(os.path.join(config.core.logdir, 'exceptions.log'), 'a')
@@ -107,7 +107,8 @@ def run(config, pid_file, daemon=False):
             return -1
         if p.hasquit:
             break
-        stderr('Warning: Disconnected. Reconnecting in %s seconds...' % delay)
+        tools.stderr(
+            'Warning: Disconnected. Reconnecting in %s seconds...' % delay)
         time.sleep(delay)
     # TODO: This should be handled by command_start
     # All we should need here is a return value, but making this
@@ -228,8 +229,11 @@ def check_not_root():
         if os.environ.get("USERNAME") == "Administrator":
             raise RuntimeError('Error: Do not run Sopel as Administrator.')
     else:
-        stderr("Warning: %s is an uncommon operating system platform. Sopel should still work, "
-               "but please contact Sopel's developers if you experience issues." % opersystem)
+        tools.stderr(
+            "Warning: %s is an uncommon operating system platform. "
+            "Sopel should still work, but please contact Sopel's developers "
+            "if you experience issues."
+            % opersystem)
 
 
 def print_version():
@@ -328,11 +332,11 @@ def command_start(opts):
     try:
         config_module = get_configuration(opts)
     except ConfigurationError as e:
-        stderr(e)
+        tools.stderr(e)
         return ERR_CODE_NO_RESTART
 
     if config_module.core.not_configured:
-        stderr('Bot is not configured, can\'t start')
+        tools.stderr('Bot is not configured, can\'t start')
         return ERR_CODE_NO_RESTART
 
     # Step Two: Manage logfile, stdout and stderr
@@ -344,8 +348,10 @@ def command_start(opts):
     pid = get_running_pid(pid_file_path)
 
     if pid is not None and tools.check_pid(pid):
-        stderr('There\'s already a Sopel instance running with this config file')
-        stderr('Try using either the `sopel stop` or the `sopel restart` command')
+        tools.stderr('There\'s already a Sopel instance running '
+                     'with this config file.')
+        tools.stderr('Try using either the `sopel stop` '
+                     'or the `sopel restart` command.')
         return ERR_CODE
 
     if opts.daemonize:
@@ -388,7 +394,7 @@ def command_stop(opts):
         return ERR_CODE
 
     if settings.core.not_configured:
-        stderr('Sopel is not configured, can\'t stop')
+        tools.stderr('Sopel is not configured, can\'t stop')
         return ERR_CODE
 
     # Redirect Outputs
@@ -399,16 +405,16 @@ def command_stop(opts):
     pid = get_running_pid(filename)
 
     if pid is None or not tools.check_pid(pid):
-        stderr('Sopel is not running!')
+        tools.stderr('Sopel is not running!')
         return ERR_CODE
 
     # Stop Sopel
     if opts.kill:
-        stderr('Killing the Sopel')
+        tools.stderr('Killing the Sopel')
         os.kill(pid, signal.SIGKILL)
         return
 
-    stderr('Signaling Sopel to stop gracefully')
+    tools.stderr('Signaling Sopel to stop gracefully')
     if hasattr(signal, 'SIGUSR1'):
         os.kill(pid, signal.SIGUSR1)
     else:
@@ -427,7 +433,7 @@ def command_restart(opts):
         return ERR_CODE
 
     if settings.core.not_configured:
-        stderr('Sopel is not configured, can\'t stop')
+        tools.stderr('Sopel is not configured, can\'t stop')
         return ERR_CODE
 
     # Redirect Outputs
@@ -438,10 +444,10 @@ def command_restart(opts):
     pid = get_running_pid(filename)
 
     if pid is None or not tools.check_pid(pid):
-        stderr('Sopel is not running!')
+        tools.stderr('Sopel is not running!')
         return ERR_CODE
 
-    stderr('Asking Sopel to restart')
+    tools.stderr('Asking Sopel to restart')
     if hasattr(signal, 'SIGUSR2'):
         os.kill(pid, signal.SIGUSR2)
     else:
@@ -506,11 +512,11 @@ def command_legacy(opts):
     try:
         config_module = get_configuration(opts)
     except ConfigurationError as e:
-        stderr(e)
+        tools.stderr(e)
         return ERR_CODE_NO_RESTART
 
     if config_module.core.not_configured:
-        stderr('Bot is not configured, can\'t start')
+        tools.stderr('Bot is not configured, can\'t start')
         return ERR_CODE_NO_RESTART
 
     # Step Three: Manage logfile, stdout and stderr
@@ -523,21 +529,23 @@ def command_legacy(opts):
 
     if old_pid is not None and tools.check_pid(old_pid):
         if not opts.quit and not opts.kill and not opts.restart:
-            stderr('There\'s already a Sopel instance running with this config file')
-            stderr('Try using either the `sopel stop` command or the `--restart` option')
+            tools.stderr(
+                'There\'s already a Sopel instance running with this config file')
+            tools.stderr(
+                'Try using either the `sopel stop` command or the `sopel restart` command')
             return ERR_CODE
         elif opts.kill:
             tools.stderr(
                 'option -k/--kill is deprecated, '
                 'use `sopel stop --kill` instead')
-            stderr('Killing the Sopel')
+            tools.stderr('Killing the Sopel')
             os.kill(old_pid, signal.SIGKILL)
             return
         elif opts.quit:
             tools.stderr(
                 'options -q/--quit is deprecated, '
                 'use `sopel stop` instead')
-            stderr('Signaling Sopel to stop gracefully')
+            tools.stderr('Signaling Sopel to stop gracefully')
             if hasattr(signal, 'SIGUSR1'):
                 os.kill(old_pid, signal.SIGUSR1)
             else:
@@ -549,7 +557,7 @@ def command_legacy(opts):
             tools.stderr(
                 'options --restart is deprecated, '
                 'use `sopel restart` instead')
-            stderr('Asking Sopel to restart')
+            tools.stderr('Asking Sopel to restart')
             if hasattr(signal, 'SIGUSR2'):
                 os.kill(old_pid, signal.SIGUSR2)
             else:
@@ -558,7 +566,7 @@ def command_legacy(opts):
                 os.kill(old_pid, signal.SIGILL)
             return
     elif opts.kill or opts.quit or opts.restart:
-        stderr('Sopel is not running!')
+        tools.stderr('Sopel is not running!')
         return ERR_CODE
 
     if opts.daemonize:
@@ -596,7 +604,7 @@ def main(argv=None):
         try:
             check_not_root()
         except RuntimeError as err:
-            stderr('%s' % err)
+            tools.stderr('%s' % err)
             return ERR_CODE
 
         # Step Three: Handle command
--- a/sopel/cli/run.py
+++ b/sopel/cli/run.py
@@ -94,6 +94,24 @@ def build_parser():
     add_legacy_options(parser_legacy)
     utils.add_common_arguments(parser_legacy)
 
+    # manage `start` sub-command
+    parser_start = subparsers.add_parser(
+        'start',
+        description='Start a Sopel instance',
+        help='Start a Sopel instance')
+    parser_start.add_argument(
+        '-d', '--fork',
+        dest='daemonize',
+        action='store_true',
+        default=False,
+        help='Run Sopel as a daemon (fork)')
+    parser_start.add_argument(
+        '--quiet',
+        action="store_true",
+        dest="quiet",
+        help="Suppress all output")
+    utils.add_common_arguments(parser_start)
+
     # manage `configure` sub-command
     parser_configure = subparsers.add_parser(
         'configure', help='Sopel\'s Wizard tool')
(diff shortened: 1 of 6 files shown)


sopel/cli/run.py
```diff
@@ -0,0 +1,619 @@
...
+        description='List of Sopel\'s sub-commands',
+        dest='action')
+
+    # manage `legacy` sub-command
+    parser_legacy = subparsers.add_parser(
+        'legacy', help='Launch Sopel\'s legacy command line')
+    add_legacy_options(parser_legacy)
+    utils.add_common_arguments(parser_legacy)
+
+    # manager `start` sub-command
```
dgw:
```suggestion
    # manage `start` sub-command
```

sopel/cli/run.py
```diff
@@ -0,0 +1,619 @@
...
+    for config in configs:
+        print('\t%s' % config)
+    if not config:
+        print('\tNone found')
+
+    print('-------------------------')
+
+
+def get_configuration(options):
+    """Get an instance of configuration from options.
```
dgw: "an instance of configuration" is kind of nonsense.

Should this read, "an instance of `Config`"? Something else?

sopel/cli/run.py
```diff
@@ -0,0 +1,619 @@
...
+
+    This may raise a ``sopel.config.ConfigurationError`` if the file is an
+    invalid configuration file.
```
dgw: "if the file is an invalid configuration file" is a bit redundant.

How about: "if the configuration file is invalid"?

No suggested change, because GitHub won't allow selecting multiple lines to edit. :disappointed:

sopel/cli/run.py
```diff
@@ -0,0 +1,619 @@
...
+        os.kill(pid, signal.SIGILL)
+
+
+def command_legacy(opts):
+    """Legacy Sopel run script
+
+    The ``legacy`` command manages the old-style ``sopel`` command line tool.
+    Most of its features are replaced by the following commands:
+
+    * ``sopel start`` replaces the default behavior (run the bot)
```
dgw: We probably don't want to _require_ `start`. Running just `sopel` with no arguments at all should always do something useful like running with a default configuration, just as it does now.

The default action should be `start` once the `legacy` parser is removed in (probably) Sopel 8. Any reason not to document it this way immediately? One less thing to forget about later (and goodness knows Sopel's docs/docstrings are already full of obsolete stuff that needs updating…).

sopel/cli/run.py
```diff
@@ -0,0 +1,619 @@
...
+    * ``sopel configure`` replaces the
+      ``-w/--configure-all/--configure-modules`` options
+
+    The ``-v`` option for "version" is deprecated, ``-V/--version`` should be
+    used instead.
+
+    .. seealso::
+
+       The github issue `#1471`__ tracks various changes requested for future
+       version of Sopel, some of them are related to this legacy command line
```
dgw:
```suggestion
       version of Sopel, some them related to this legacy command line tool.
```

And remove the orphaned word below (but GitHub suggestions won't let me do that).

I chose this over a semicolon because nobody uses semicolons in 2019, right?

sopel/cli/run.py
```diff
@@ -0,0 +1,619 @@
...
+        stderr("Warning: %s is an uncommon operating system platform. Sopel should still work, "
+               "but please contact Sopel's developers if you experience issues." % opersystem)
+
+
+def print_version():
+    """Print Python version and Sopel version on stdout."""
+    py_ver = '%s.%s.%s' % (sys.version_info.major,
+                           sys.version_info.minor,
+                           sys.version_info.micro)
+    print('Sopel %s (running on python %s)' % (__version__, py_ver))
```
dgw: It suddenly annoys me that this `python` is lowercase, but it's not your fault. I'll just probably forget about it before I have a chance to fix it post-merge… :/

sopel/cli/run.py
```diff
@@ -0,0 +1,619 @@
...
+
+    """
+    # Step One: Handle "No config needed" options
+    if opts.version:
+        print_version()
+        return
+    elif opts.version_legacy:
+        tools.stderr(
+            'option -v is deprecated, '
+            'use `sopel -V/--version` instead')
```
dgw: All of these deprecation warnings are pretty unobtrusive in actual output. The "error" from the server on quitting is much more conspicuous (because its line actually starts with `ERROR:`… even though it's _not_ an error).

I won't leave line notes on all of them (that would nearly double my line-note count, and be very intimidating when you get this review in your email), but perhaps you could change them all to this format:

```
'WARNING: Option {option_string} is deprecated; '
' use `sopel {replacement_string}` instead!'
```

(Yes, here I'm using a semicolon anyway. :stuck_out_tongue_closed_eyes:)

sopel/cli/run.py
```diff
@@ -0,0 +1,619 @@
...
+
+    # Step Four: Handle process-lifecycle options and manage the PID file
+    pid_dir = config_module.core.pid_dir
+    pid_file_path = get_pid_filename(opts, pid_dir)
+    old_pid = get_running_pid(pid_file_path)
+
+    if old_pid is not None and tools.check_pid(old_pid):
+        if not opts.quit and not opts.kill and not opts.restart:
+            stderr('There\'s already a Sopel instance running with this config file')
+            stderr('Try using either the `sopel stop` command or the `--restart` option')
```
dgw: There's a `sopel restart` command now. Suggest that instead of `--restart`?

```suggestion
            stderr('Try using the `sopel stop` or `sopel restart` command')
```

sopel/cli/run.py
```diff
@@ -0,0 +1,619 @@
...
+        return ret
+
+
+def main(argv=None):
+    """Sopel run script entry point"""
+    try:
+        # Step One: Parse The Command Line
+        parser = build_parser()
+
+        # make sure to have an action first (`legacy` by default)
```
dgw: Maybe tag this for updating? Not sure if necessary, but this ties in with my comment above about always doing _something useful_ when run with just `sopel` (no args), and `start` being a good default action.

sopel/cli/run.py
```diff
@@ -112,6 +112,13 @@ def build_parser():
         help='Kill Sopel without a graceful quit')
     utils.add_common_arguments(parser_stop)
 
+    # manage `restart` sub-command
+    parser_restart = subparsers.add_parser(
+        'restart',
+        description='Restart a running Sopel instance',
+        help='Restart a running Sopel instance')
+    utils.add_common_arguments(parser_restart)
```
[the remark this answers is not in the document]
Exirel: All that `restart` does is so send a signal, and there is no signal that allows us to say "force restart".

sopel/cli/run.py
```diff
@@ -270,6 +277,38 @@ def command_stop(opts):
...
+def command_restart(opts):
+    # Get Configuration
+    try:
+        settings = utils.load_settings(opts)
+    except ConfigurationNotFound as error:
+        tools.stderr('Configuration "%s" not found' % error.filename)
+        return ERR_CODE
+
+    if settings.core.not_configured:
+        stderr('Sopel is not configured, can\'t stop')
```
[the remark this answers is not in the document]
Exirel: I'll clean that, yes.

sopel/cli/run.py
Exirel: True.

sopel/cli/run.py
Exirel: This is more or less documented here in the `main` function:

        # make sure to have an action first (`legacy` by default)
        # TODO: `start` should be the default in Sopel 8

And this part of the docstring makes it clear enough IMHO that there is more to see:

    .. seealso::

       The github issue `#1471`__ tracks various changes requested for future
       version of Sopel, some of them are related to this legacy command line
       tool.

I would argue that, eventually, the function `command_legacy` will be removed for Sopel 8. This is more visible now than it was at the time of your review, thanks to all the little comments I added after you suggested them.

sopel/cli/run.py
Exirel: What is a semicolon anyway?

Exirel:
> I do love that this adds the new CLI structure in a transitional state while still keeping the old syntax around. The release notes for Sopel 7 can warn about the upcoming CLI changes, and users can have time to update their service units/scripts/etc.

Thank you! I'm quite happy with how you manage the project, and how you review and comment the PR & issues we send you. It's a happy collaboration so far!

sopel/cli/run.py
```diff
@@ -0,0 +1,658 @@
...
+                        help=(
+                            "Restart Sopel "
+                            "(deprecated, and will be removed in Sopel 8; "
+                            "use `sopel restart` instead)"))
+    parser.add_argument("-l", '--list', action="store_true",
+                        dest="list_configs",
+                        help="List all config files found")
+    parser.add_argument("-m", '--migrate', action="store_true",
+                        dest="migrate_configs",
+                        help="Migrate config files to the new format")
```
dgw: I was going to ask where the code to handle this option went, but… you had nothing to do with it.

Added in f85bf58 for (I guess) Willie 3.1, the `--migrate` option was neutered in 7e1dc1e (part of 4.0.0, I think). The code to handle the option was removed without mention in the commit log or changelog (that I can find).

Honestly, you can probably just drop this legacy option. It's likely nobody is using it any more after almost six years of Willie/Sopel doing nothing when passed the arg.

Or, keep it with a deprecation notice and remove it for real in Sopel 8, like its buddies. But it's probably fine to just remove (with a nice commit message about how I went Sopelunking and found the deleted code 😁).

sopel/cli/run.py
```diff
@@ -0,0 +1,658 @@
...
+    # TODO: This should be handled by command_start
+    # All we should need here is a return value, but making this
+    # a return makes Sopel hang on ^C after it says "Closed!"
+    os.unlink(pid_file)
+    os._exit(0)
+
+
+def add_legacy_options(parser):
+    parser.add_argument("-d", '--fork', action="store_true",
+                        dest="daemonize", help="Daemonize Sopel")
```
dgw: I'm guessing this has no deprecation note because e.g. `sopel -d` will work identically in Sopel 8, when `start` becomes the default action?

sopel/cli/run.py
```diff
@@ -0,0 +1,658 @@
...
+                            "(deprecated, and will be removed in Sopel 8; "
+                            "use `sopel stop --kill` instead)"))
+    parser.add_argument("-r", '--restart', action="store_true", dest="restart",
+                        help=(
+                            "Restart Sopel "
+                            "(deprecated, and will be removed in Sopel 8; "
+                            "use `sopel restart` instead)"))
+    parser.add_argument("-l", '--list', action="store_true",
+                        dest="list_configs",
+                        help="List all config files found")
```
dgw: Ditto here as for `--fork`/`-d`, but I can't find any reference to this arg outside `command_legacy`.

Will this become part of `start` for Sopel 8, when `legacy` goes away? (`# TODO`?)

sopel/cli/run.py
```diff
@@ -0,0 +1,658 @@
...
+    parser.add_argument("-m", '--migrate', action="store_true",
+                        dest="migrate_configs",
+                        help="Migrate config files to the new format")
+    parser.add_argument('--quiet', action="store_true", dest="quiet",
+                        help="Suppress all output")
```
dgw: Ditto as for `--fork`/`-d`, though this could (and probably should) apply to other actions besides `start`. `# TODO`?

sopel/cli/run.py
```diff
@@ -0,0 +1,658 @@
...
+                            "use `sopel configure --modules` instead)"))
+    parser.add_argument('-v', action="store_true",
+                        dest='version_legacy',
+                        help=(
+                            "Show version number and exit "
+                            "(deprecated, and will be removed in Sopel 8; "
+                            "use -V/--version instead)"))
+    parser.add_argument('-V', '--version', action='store_true',
+                        dest='version',
+                        help='Show version number and exit')
```
dgw: No deprecation note because this will be moved into `start` for Sopel 8? (Could be a good place for `# TODO`.)

sopel/cli/run.py
```diff
@@ -0,0 +1,658 @@
...
+    parser = argparse.ArgumentParser(description='Sopel IRC Bot',
+                                     usage='%(prog)s [options]')
+    subparsers = parser.add_subparsers(
+        title='sub-commands',
+        description='List of Sopel\'s sub-commands',
+        dest='action')
+
+    # manage `legacy` sub-command
+    parser_legacy = subparsers.add_parser(
+        'legacy', help='Launch Sopel\'s legacy command line')
```
dgw: I had a thought back when I had a chance to play with this a bit in an actual terminal:

Can the `legacy` sub-command be hidden from the list of choices given by `--help`? I wasn't confused by its presence because I'd read some of the PR already, but to a user it makes no sense to have this as a choice. It's an internal implementation detail, really.

dgw:
Leaving a note for @Exirel upon request: Amend `stderr` -> `tools.stderr` in import header, too. As we discussed on IRC, the version check up there might just get removed, but separately.

sopel/cli/run.py
Exirel: Nah, it's dead code, let's remove it.

sopel/cli/run.py
Exirel: Look at `get_configuration`, it's in here:

    bot_config._is_daemonized = options.daemonize
    return bot_config

Exirel: (nevermind the previous comment, wrong one)

For now,  -l/--list exists from `sopel` only. However, I would like to move this option to a separated command, ie. `sopel-config`, with this interface:

    $ sopel-config init
    $ sopel-config list
    $ sopel-config rename

And other features.

Cherry on top: `sopel` is like `git`, where `git commit` is, in fact, `git-commit`. It means that `sopel config` and `sopel-config` will be the same thing, code-wise.

I'm confident enough that this will be taken care of soon enough for us not to add extra TODO.

sopel/cli/run.py
Exirel: Exactly!

sopel/cli/run.py
Exirel: I'm not sure this option is useful outside of the start command. I would rather not have it for the other sub-command instead.

sopel/cli/run.py
Exirel: This option is one of the few that I don't know what to do for now. It's, like, `sopel start` don't need it, it makes little to no sense.

But `sopel --version` is useful. But maybe it should be replaced by `sopel version` instead?

So many questions, and I don't have an answer for now. Since the "legacy" mode is a big TODO by itself, I would argue that we should take care of it later, without adding more to it for now.

sopel/cli/run.py
Exirel: Fixed, nice find.

Exirel:
> Some of the legacy options have no counterpart in the revamped CLI, but aren't marked as deprecated, which makes me wonder.

Indeed! I would like to mark them as deprecated once there is a proper replacement. For example the list-config option should be replaced when I'll redo PR #1465 (cli: sopel-config), using the work done in this PR. So it's work for later.

> Loved the new, detailed docstrings for stuff like `get_configuration`!

Thanks!

> My next questions will be about test coverage of the new CLI [...] Both old and new should probably be tested, to be fair.

Yeah. These tests were added when I reworked the `run_script.py` file, and I should probably write more.

sopel/cli/run.py
dgw: Fair enough.

And even if it doesn't get done (long shot, because you're on top of your shit) it will be quite obvious that this needs replacing when it's time to remove the legacy parser, because this will be one of very few options without a deprecation notice.

sopel/cli/run.py
dgw: Probably any command that could be run from a script (init/service control or otherwise) should support `--quiet`. Currently that would be:

  * `start`
  * `stop`
  * `restart`

The other actions are decidedly user-facing and not useful to run from scripts, IMO. (Though I guess the Docker/Ansible guys might hackily use `sopel configure`, knowing them! :laughing:)

sopel/cli/run.py
dgw: Can the argument parser have a set of default flags with no associated action that falls through to a default subparser/action (`legacy` in 7, `start` in 8) if none of those flags is given?

Just spitballing, unless it's easy to move this argument out of the `legacy` parser right now. I agree that `sopel start` doesn't need this argument, and it should go somewhere else ideally.

sopel/cli/run.py
```diff
@@ -168,14 +168,17 @@ def build_parser():
...
                                      usage='%(prog)s [options]')
+    add_legacy_options(parser)
+    utils.add_common_arguments(parser)
+
     subparsers = parser.add_subparsers(
         title='sub-commands',
         description='List of Sopel\'s sub-commands',
-        dest='action')
+        dest='action',
+        metavar='{start,configure,stop,restart}')
```
dgw: Will anyone necessarily remember to undo this `metavar` when `legacy` is removed, or remember to update it when adding new CLI actions?

I'm guessing it's the only not-awful way to hide "legacy" from the action listing in help, though, so this isn't a change request per se.

Exirel: Yeah, it's annoying, but there is a bug listed for that: https://bugs.python.org/issue22848

sopel/cli/run.py
Exirel: There is no easy way to do that, because if you define a set of arguments on the main parser object, you will be able to do that:

    $ sopel --global-options start --start-options
    $ sopel start --start-options
    $ sopel-start --global-options --start-options

But not that:

    $ sopel start --global-options --start-options

Which is quite confusing when you think about it. Add to the mix the fact that, in Py2.7, the sub-parser argument **is always required**, and you have a recipe for nightmare and headaches.

sopel/cli/run.py
Exirel: Yeah, better keeping the same behavior as Sopel 6 on that one, I'll add the Output redirect in `stop` and `restart` sub-commands.

If I don't want it, I'll do another PR, or think about it more and provide better arguments later.

Exirel: Done!

sopel/cli/run.py
dgw: That _would_ be confusing… :frowning:

I guess `sopel start` is going to have it after `legacy` goes away, then. Ah, well.

sopel/cli/run.py
dgw: Not encouraging that that bug has been open for over four years… I guess we're stuck with this way.

dgw:
OK, I kicked Travis to get the PR status reported correctly (for some reason it finished all builds and reported the commit status, but not the PR status).

Future work on remaining legacy options like `--list` is left for the future. I assume more tests will be a future thing too?

If that's correct, all that's left is for me to sit down and actually run this code on both py2 and py3, when I have time to do that. Assuming that all works, this will get **:heavy_check_mark: Approved** and merged as-is. :tada:

Exirel:
> Future work on remaining legacy options like `--list` is left for the future. I assume more tests will be a future thing too?
> 
> If that's correct, all that's left is for me to sit down and actually run this code on both py2 and py3, when I have time to do that. Assuming that all works, this will get **✔️ Approved** and merged as-is. 🎉

Yes it is! Now glhf with your manual tests for these changes, it took me some times on my side!

Exirel:
Rebased upon master. Nothing broke, AFAICT, but it might be dangerous to keep this PR open for too long, given its scope.

dgw:
> it might be dangerous to keep this PR open for too long, given its scope

You're right about that. I merged it without performing excessive amounts of testing, but we will have plenty of time to fix bugs between now and when Sopel 7 actually ships. :sunglasses:

Exirel:
**BOLD MERGE** ❤️

dgw:
_\*sees :eyes: what you did there\*_

PR #1157, opened 2016-12-17
add /invite option

converge:
This module adds the /invite option to be used as a command (.invite nick #channel). The bot.py core file was touched to understand the /invite command.

PR #1497, opened 2019-03-10
Add .invite command

### LINKED ISSUE #1145: How do I let Sopel invite a user?
^

dgw:
This is based on #1157, which stalled after its original author deleted their fork.

Commits added in this version of the patch address feedback left in that original pull request, and make the end product a bit cleaner & more user-friendly (error messages, argument defaults, etc.).

Resolves #1145.

### DIFF
--- a/sopel/modules/invite.py
+++ b/sopel/modules/invite.py
@@ -11,10 +11,23 @@ from __future__ import unicode_literals, absolute_import, print_function, divisi
 from sopel.module import commands, example, OP
 
 @commands('invite')
+@example('.invite jenny')
 @example('.invite converge #sopel')
 def invite(bot, trigger):
-    if bot.privileges[trigger.sender][bot.nick] < OP:
-        return bot.reply("I'm not a channel operator!")
-    if not trigger.group(3) or not trigger.group(4):
-        return bot.reply("Usage: .invite user #channel (the bot must be operator and in the invited channel)")
-    bot.write(['INVITE', trigger.group(3), trigger.group(4)])
+    """
+    Invite the given user to the current channel, or (with optional
+    second argument) another channel that Sopel is in.
+    """
+    if not trigger.group(3):
+        return bot.reply("Who should I invite?")
+    nick = trigger.group(3)
+    if trigger.group(4):
+        channel = trigger.group(4)
+    else:
+        channel = trigger.sender
+    try:
+        if bot.privileges[channel][bot.nick] < OP:
+            return bot.reply("I'm not a channel operator!")
+    except KeyError:
+        return bot.reply("I'm not in {}!".format(channel))
+    bot.write(['INVITE', nick, channel])
--- a/sopel/modules/invite.py
+++ b/sopel/modules/invite.py
@@ -24,6 +24,8 @@ def invite(bot, trigger):
     if trigger.group(4):
         channel = trigger.group(4)
     else:
+        if trigger.is_privmsg:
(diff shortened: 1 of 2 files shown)


sopel/modules/invite.py
```diff
@@ -0,0 +1,35 @@
...
+    second argument) another channel that Sopel is in.
+    """
+    if not trigger.group(3):
+        return bot.reply("Who should I invite?")
+    nick = trigger.group(3)
+    if trigger.group(4):
+        channel = trigger.group(4)
+    else:
+        if trigger.is_privmsg:
+            return bot.say("Channel is required (.invite user #channel) when inviting from queries.")
```
dgw: Already noticed that this should use `help_prefix` instead of a hard-coded `.`…

half-duplex:
IMO this needs to either check for +i (and maybe others), have a per-channel knob for "anyone can invite", or be restricted to >halfop.
Allowing any user in a channel with +i and a +o sopel bot to invite other users would be surprise access control circumvention with no warning to channel staff.

dgw:
@half-duplex Looks like I opened this PR at some ungodly hour of the night/morning, so I probably somehow thought the command was already privilege-restricted. Good call on that, thanks! See dc1f98d.

half-duplex:
`@require_privilege` is a no-op in PMs, and in channels checks privilege in the current channel - so currently an op in any channel the bot is in, or any user in a PM with the bot, could successfully `.invite anyone #somechannel`.
In inspircd and probably others, /invite is enabled by +h or higher. Is there a reason it's +o for the bot?
Lastly, if it's restricted to only (half)ops, is there a reason sopel should be able to do it, vs just having the user /invite ?

dgw:
> `@require_privilege` is a no-op in PMs, and in channels checks privilege in the current channel - so currently an op in any channel the bot is in, or any user in a PM with the bot, could successfully `.invite anyone #somechannel`.

My brain's too fried tonight to do anything with proper privilege checking, but I do have enough presence of mind to back out the useless `@require_privilege` changes, so I did that.

> In inspircd and probably others, /invite is enabled by +h or higher. Is there a reason it's +o for the bot?

It's +o because that's what @converge wrote in the original PR. I didn't touch that except for updating the `if` check to use `bot.channels`.

> Lastly, if it's restricted to only (half)ops, is there a reason sopel should be able to do it, vs just having the user /invite ?

Networks without ChanServ (or similar), I guess, where Sopel can act as pseudo-services **if** the channel ops want it to (by giving it privileges). Adding a command to `.invite` is approximately as useful (or useless) as having commands for `.op`, `.voice`, etc. (#1498)

dgw:
@Exirel You think having to give the bot +o is enough of a gate, without checking channel privileges for the user who calls the command? The points @half-duplex made about +i channels make sense to me. It's just a question of complexity/security value trade-off. After all, if access control really matters, the `invite` module can always be excluded, and/or the bot just not given +o.

Exirel:
> You think having to give the bot +o is enough of a gate, without checking channel privileges for the user who calls the command?

Oh. That's what happens when I just review the code's style and not the deeper meaning. You're right, I didn't see that.

That's annoying. Why do we do that again?

dgw:
We do it because the original PR did, and (as you can see from the branch name) this is just a rebase/refresh of that PR to conform to current code standards.

At some point, I plan to revisit the whole privilege-checking thing and do something sensible.

PR #1500, opened 2019-03-11
module: optional `reply` arg for all `require_*` decorators

dgw:
Since @Exirel added `reply` as an optional argument to `module.require_admin` (in #1456), it seemed prudent to implement the same argument on all of its relatives for consistency's sake.

Also added the parameters to docstrings for the functions I touched, since I was in there anyway.

Now this is a full-blown API addition instead of a one-off oddity! :grin:

### DIFF
--- a/sopel/module.py
+++ b/sopel/module.py
@@ -371,19 +371,29 @@ def require_admin(message=None, reply=False):
     return actual_decorator
 
 
-def require_owner(message=None):
+def require_owner(message=None, reply=False):
     """Decorate a function to require the triggering user to be the bot owner.
 
-    If they are not, `message` will be said if given."""
+    :param str message: optional message said to non-owner user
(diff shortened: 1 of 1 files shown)


dgw:
@Exirel I couldn't figure out how to avoid using <tt>:meth:\`display.method() <.path.to.method>\`</tt>, but if you know of a way I'm happy to use it! I doubt it's possible to avoid, though. `~` always truncates everything but the last element (the method name here), and even if it is configurable to leave e.g. two elements (`module.method` or `Class.method`) it's not useful because we want to show `bot.method()` like a plugin might use with the standard `def triggerable(bot, trigger)` signature.

PS: Including Sphinx references in GitHub comments is a **pain**, wow.

dgw:
Merged so as to clear the way for an overhaul of docstrings in `module` for #1565, since while fixing up this I noticed a bunch of other stuff that *really* needs it.

PR #1507, opened 2019-03-19
cli: add `sopel-config {list,init,get}`

Exirel:
**Replace** #1465

The `sopel-config` is a CLI tool to:

* list existing configuration file from Sopel's homedir
* init a new configuration file in Sopel's homedir
* get a value from a section.option of Sopel's configuration

It could do much more, but I'll leave it here for now, since @dgw is quite busy.

Note: `sopel-config list` could replace `sopel --list`, and `sopel-config init` could replace `sopel --wizard` in the end. That requires more thought.

### DIFF
--- a/sopel/cli/config.py
+++ b/sopel/cli/config.py
@@ -1,0 +1,163 @@
+# coding=utf-8
+"""Sopel Config Command Line Interface (CLI): ``sopel-config``"""
+from __future__ import unicode_literals, absolute_import, print_function, division
+
+import argparse
+import os
+
+from sopel import tools, config
+from . import utils
+
+
+def build_parser():
+    """Configure an argument parser for ``sopel-config``"""
+    parser = argparse.ArgumentParser(
+        description='Sopel configuration tool')
+
+    # Subparser: sopel-config <sub-parser> <sub-options>
+    subparsers = parser.add_subparsers(
+        help='Actions to perform',
+        dest='action')
+
+    # sopel-config list
+    list_parser = subparsers.add_parser(
+        'list',
+        help="List available configurations from Sopel's homedir",
+        description="""
+            List available configurations from Sopel's homedir ({homedir})
+            with the extension "{ext}".
+        """.format(homedir=config.DEFAULT_HOMEDIR, ext='.cfg'))
+    list_parser.add_argument(
+        '-e', '--ext', '--extension',
+        dest='extension',
+        default='.cfg',
+        help='Filter by extension (default to "%(default)s)"')
+    list_parser.add_argument(
+        '-p', '--path',
+        action='store_true',
+        dest='display_path',
+        default=False,
+        help='Display a list of absolute filenames instead of their names')
+
+    # sopel-config init
+    init_parser = subparsers.add_parser(
+        'init',
+        help='Initialize Sopel configuration file',
+        description='Initialize Sopel configuration file')
+    utils.add_common_arguments(init_parser)
+
+    # sopel-config get <section> <key>
+    get_parser = subparsers.add_parser(
+        'get',
+        help="Get a configuration option's value",
+        description="Get a configuration option's value",
+    )
+    get_parser.add_argument('section')
+    get_parser.add_argument('option')
+    utils.add_common_arguments(get_parser)
+
+    return parser
+
+
+def handle_list(options):
+    """Display a list of configuration available from Sopel's homedir
+
+    :param options: parsed arguments
+    :type options: ``argparse.Namespace``
+
+    This command displays an unordered list of config names from Sopel's
+    default homedir, without their extensions::
+
+        $ sopel-config list
+        default
+        custom
+
+    It is possible to filter by extension using the ``-e/--ext/--extension``
+    option; default is ``.cfg`` (the ``.`` prefix is not required).
+    """
+    display_path = getattr(options, 'display_path', False)
+    extension = getattr(options, 'extension', '.cfg')
+    if not extension.startswith('.'):
+        extension = '.' + extension
+    configs = utils.enumerate_configs(config.DEFAULT_HOMEDIR, extension)
+
+    for config_filename in configs:
+        if display_path:
+            print(os.path.join(config.DEFAULT_HOMEDIR, config_filename))
+        else:
+            name, _ = os.path.splitext(config_filename)
+            print(name)
+
+
+def handle_init(options):
+    """Use config wizard to initialize a new configuration file for the bot
+
+    :param options: parsed arguments
+    :type options: ``argparse.Namespace``
+
+    .. note::
+
+       Due to how the config wizard works, the configuration filename's
+       extension **must be** ``.cfg``.
+
+    """
+    config_filename = utils.find_config(
+        config.DEFAULT_HOMEDIR,
+        getattr(options, 'config', None) or 'default')
+    config_name, ext = os.path.splitext(config_filename)
+
+    if ext and ext != '.cfg':
+        tools.stderr('Configuration wizard accepts .cfg files only')
+        return 1
+    elif not ext:
+        config_filename = config_name + '.cfg'
+
+    if os.path.isfile(config_filename):
+        tools.stderr('Configuration file %s already exists' % config_filename)
+        return 1
+
+    print('Starting Sopel config wizard for: %s' % config_filename)
+    config._wizard('all', config_name)
+
+
+def handle_get(options):
+    """Read the settings to display the value of <section> <key>"""
+    try:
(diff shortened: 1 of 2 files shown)


dgw:
I haven't gone very far into this, but I did find it odd that there's no argument to list configs from a non-default homedir. Have you just not built that yet or did you specifically leave it out for some reason?

Exirel:
I left it out because IIRC `sopel start` does not allow to use a different homedir: all you can do is to provide a configuration file. At the moment, Sopel is pretty strict about where its configuration file should be, and it lets you use another configuration file if you provide an absolute path or if you provide a relative path to the current directory. This can be seen with using the configuration wizard.

I don't really like it, but the bot isn't very good at managing that at the moment. Even though I think it's a very valid concern to have.

dgw:
I smell a future enhancement, then. For this PR I'll ignore that omission, since it's out of scope to do anything about it here.

sopel/cli/config.py
```diff
@@ -0,0 +1,161 @@
+# coding=utf-8
+"""Sopel Config Command Line Interfaces (CLI): ``sopel-config``"""
```
dgw: "Interfaces" plural?

sopel/cli/config.py
```diff
@@ -0,0 +1,161 @@
...
+        help='Actions to perform',
+        dest='action')
+
+    # sopel-config list
+    list_parser = subparsers.add_parser(
+        'list',
+        help='List available configurations from Sopel\'s homedir',
+        description="""
+        List available configurations from Sopel's homedir ({homedir}) with the
+        extension "{ext}".
```
dgw: Would it break things to indent this description one more step, so it's visually separate from the list of method args?

sopel/cli/config.py
```diff
@@ -0,0 +1,161 @@
...
+        action='store_true',
+        dest='display_path',
+        default=False,
+        help='Display a list of absolute filenames instead of their names')
+
+    # sopel-config init
+    init_parser = subparsers.add_parser(
+        'init',
+        help='Initialize sopel configuration file',
+        description='Initialize sopel configuration file')
```
dgw: Both of these should capitalize Sopel, no?

sopel/cli/config.py
```diff
@@ -0,0 +1,161 @@
...
+
+    # Subparser: sopel-config <sub-parser> <sub-options>
+    subparsers = parser.add_subparsers(
+        help='Actions to perform',
+        dest='action')
+
+    # sopel-config list
+    list_parser = subparsers.add_parser(
+        'list',
+        help='List available configurations from Sopel\'s homedir',
```
dgw: Not just here, but the only place I'll leave a note:

Are we so opposed to using double-quotes for strings that we won't even do so to avoid escaping `'`?

sopel/cli/config.py
```diff
@@ -0,0 +1,161 @@
...
+    get_parser.add_argument('option')
+    utils.add_common_arguments(get_parser)
+
+    return parser
+
+
+def handle_list(options):
+    """Display a list of configuration available from Sopel's homedir
+
+    :param options: argument parser's parsed option
```
dgw: More than one option. But maybe it should be "parsed arguments"?

sopel/cli/config.py
```diff
@@ -0,0 +1,161 @@
...
+
+    This command display an unordered list of config's name from the default
```
dgw:
```suggestion
    This command displays an unordered list of config names from Sopel's
```

(meant to be applied together with the next line's suggestion)

sopel/cli/config.py
```diff
@@ -0,0 +1,161 @@
...
+    Sopel's homedir, without their extension::
```
dgw:
```suggestion
    default homedir, without their extensions::
```

(meant to be applied together with the previous line's suggestion)

sopel/cli/config.py
```diff
@@ -0,0 +1,161 @@
...
+
+        $ sopel-config list
+        default
+        custom
+
+    It is possible to filter by extension using the ``-e/--ext/--extension``
+    option, default to ``.cfg`` (the ``.`` prefix is not required).
```
dgw:
```suggestion
    option; default is ``.cfg`` (the ``.`` prefix is not required).
```

sopel/cli/config.py
```diff
@@ -0,0 +1,161 @@
...
+    for config_filename in configs:
+        if display_path:
+            print(os.path.join(config.DEFAULT_HOMEDIR, config_filename))
+        else:
+            name, _ = os.path.splitext(config_filename)
+            print(name)
+
+
+def handle_init(options):
+    """Use config's wizard to initialize a new configuration file for the bot
```
dgw:
```suggestion
    """Use config wizard to initialize a new configuration file for the bot
```

sopel/cli/config.py
```diff
@@ -0,0 +1,161 @@
...
+
+    :param options: argument parser's parsed options
```
dgw: If you decide to change the similar param docstring I commented on above, also change this to match. :)

sopel/cli/config.py
```diff
@@ -0,0 +1,161 @@
...
+
+    .. note::
+
+       Due to how the config's wizard works, the configuration filename's
```
dgw:
```suggestion
       Due to how the config wizard works, the configuration filename's
```

sopel/cli/config.py
```diff
@@ -0,0 +1,161 @@
...
+       extension must be ``.cfg``.
```
dgw: Emphasis formatting (italic/bold) for "must"? Can we do that inside a `note` directive?

sopel/cli/config.py
```diff
@@ -0,0 +1,161 @@
...
+
+    """
+    config_filename = utils.find_config(
+        config.DEFAULT_HOMEDIR,
+        getattr(options, 'config', None) or 'default')
+    config_name, ext = os.path.splitext(config_filename)
+
+    if ext and ext != '.cfg':
+        tools.stderr('Configuration wizard accepts .cfg file only')
```
dgw:
```suggestion
        tools.stderr('Configuration wizard accepts .cfg files only')
```

sopel/cli/config.py
Exirel: Yes we can.

sopel/cli/config.py
Exirel: Ah ah, no, we can totally do double quotes here.

sopel/cli/config.py
Exirel: Silly me.

sopel/cli/config.py
Exirel: I can do that.

sopel/cli/config.py
Exirel: 🙈

Exirel:
Push changes without rebase, since everything will be squashed into one commit in the end.

dgw:
That makes it easy to see what you did, so :D

I was playing with the documentation and I tried making this change:

```patch
diff --git a/sopel/cli/config.py b/sopel/cli/config.py
index c8a2d29e..13462539 100644
--- a/sopel/cli/config.py
+++ b/sopel/cli/config.py
@@ -62,8 +62,7 @@ def build_parser():
 def handle_list(options):
     """Display a list of configuration available from Sopel's homedir
 
-    :param options: parsed arguments
-    :type options: ``argparse.Namespace``
+    :param argparse.Namespace options: parsed arguments
 
     This command displays an unordered list of config names from Sopel's
     default homedir, without their extensions::
@@ -92,8 +91,7 @@ def handle_list(options):
 def handle_init(options):
     """Use config wizard to initialize a new configuration file for the bot
 
-    :param options: parsed arguments
-    :type options: ``argparse.Namespace``
+    :param argparse.Namespace options: parsed arguments
 
     .. note::
 
```

Doing it that way makes Sphinx auto-link to the Python documentation for `argparse.Namespace`, which is nice. It also makes use of the fact that `:param:` takes a type inline, and eliminates excess lines.

It's not useful right now (because Sphinx isn't documenting anything in `sopel.cli`—that needs fixing too, in a separate PR), but in playing with the output I like this better.

(`sopel.cli.run` also outputs a not-so-nice docstring header because it still has that copyright bit in it, and Sphinx autodoc pulls the whole thing in… Again, separate PR for actually putting the whole `cli` submodule into the docs.)

Exirel:
@dgw rebased, squashed, Travis's happy, everything's good now!

dgw:
Today's the day, @Exirel: You can start working on the next phase of this CLI. :grin:

Issue #1489, opened 2019-02-15
Disabling url.py module also disables @url decorator


dgw:
The point of adding `@module.url()` to the module API was to simplify the song-and-dance of adding a module's URL patterns to the global list of "callbacks" so that `url.py` wouldn't process links already handled by a more specific plugin.

However, if `url.py` is not loaded, **any** callable using `@url` will simply never run, because these functions are dispatched **in `url.py`** instead of by Sopel's core:

https://github.com/sopel-irc/sopel/blob/0543f2f59ad0ff08b8f22491be6edffcc8694cd4/sopel/modules/url.py#L239-L245

I consider this to be a pretty major bug in the API. The point of having `url_callbacks` is for `url.py` to avoid double-processing links that another module already handles. Module API functionality **must not** depend on loading any module besides `coretasks` (which cannot be disabled, for obvious reasons).

_(On a personal note, this also explains why I couldn't get `@url`-decorated callables to trigger on my own Sopel instance when I tried a few years ago. I've had `url.py` disabled approximately forever.)_

----

I think the solution here is to move `url.py`'s URL-detection and callback-dispatch logic into `coretasks`, and the `exclude` & `exclusion_char` settings from the `[url]` config section into `[core]`. This `@rule`:

https://github.com/sopel-irc/sopel/blob/0543f2f59ad0ff08b8f22491be6edffcc8694cd4/sopel/modules/url.py#L157-L158

would become just another `@url` decorator.

That said, making all modules equal by moving dispatch of `@url` into `coretasks` presents a dilemma: How to determine which callback/pattern is "most appropriate", and avoid having the same URL processed by both `url.py` and (an)other module(s)? I'm almost positive that this conundrum is the reason why the current implementation was implemented as-is, but I'm also sure it's a solvable problem.

Given the lack of immediate ideas (at least on my end), and the fact that this is a fairly major overhaul, I'm assigning this issue to Sopel 7. That's the ideal, but if implementing a solution requires an API change, this will be pushed to Sopel 8 so as to provide advance warning via release notes. (It will also, of course, be pushed off if it simply isn't ready in a reasonable amount of time.)

Exirel:
Thanks for the thorough description of the issue! I saw your message on my PR, and I'm thinking about it since then. I'll see what I can come up with, and I'm sure we'll find a solution.

Exirel:
While we are at it, the `url` plugin has some weird behavior:

```
[...]

PR #1508, opened 2019-03-19
core: URL Callbacks new interface

Exirel:
This new interface adds:

* `bot.register_url_callback(regex, callback)`, to register a `callback` to call when a URL matches `regex`
* `bot.unregister_url_callback(regex)` to remove this regex from triggering callbacks
* `bot.search_url_callbacks(url)` to get a list of (callback, match) for each pattern matched by the URL

This provides a unified interface to work with URL Callbacks, but does not change how the built-in url plugin works (yet).

So this:

    if not bot.memory.contains('url_callbacks'):
        bot.memory['url_callbacks'] = tools.SopelMemory()
    bot.memory['url_callbacks'][instagram_pattern] = instaparse

Become this:

    bot.register_url_callback(instagram_pattern, instaparse)

It does not change the fact that the `url` built-in plugin is required at the moment in order to work with URL Callback (see #1489 for that).

@dgw now my question is, should I keep moving forward, and replace the `url` built-in plugin? Or should we merge this PR as-is, and work from there in another PR?

### DIFF
--- a/sopel/bot.py
+++ b/sopel/bot.py
@@ -247,10 +247,8 @@ class Sopel(irc.Bot):
                 job = sopel.tools.jobs.Job(interval, func)
                 self.scheduler.add_job(job)
 
-        if not self.memory.contains('url_callbacks'):
-            self.memory['url_callbacks'] = tools.SopelMemory()
         for func in urls:
-            self.memory['url_callbacks'][func.url_regex] = func
+            self.register_url_callback(func.url_regex, func)
 
     def part(self, channel, msg=None):
         """Part a channel."""
@@ -648,3 +646,97 @@ class Sopel(irc.Bot):
             entry.append(_CapReq(prefix, module_name, failure_callback, arg,
                                  success_callback))
             self._cap_reqs[cap] = entry
+
+    def register_url_callback(self, pattern, callback):
+        """Register a ``callback`` for URLs matching the regex ``pattern``
+
+        :param pattern: compiled regex pattern to register
+        :param callback: callable object to handle matching URLs
+
+        .. versionadded:: 7.0
+
+            This method replaces manual management of ``url_callbacks`` in
+            Sopel's plugins, so instead of doing this in ``setup()``::
+
+                if not bot.memory.contains('url_callbacks'):
+                    bot.memory['url_callbacks'] = tools.SopelMemory()
+
+                regex = re.compile(r'http://example.com/path/.*')
+                bot.memory['url_callbacks'][regex] = callback
+
+            use this much more concise pattern::
+
+                regex = re.compile(r'http://example.com/path/.*')
+                bot.register_url_callback(regex, callback)
+
+        """
+        if not self.memory.contains('url_callbacks'):
+            self.memory['url_callbacks'] = tools.SopelMemory()
+
+        if isinstance(pattern, basestring):
+            pattern = re.compile(pattern)
+
+        self.memory['url_callbacks'][pattern] = callback
+
+    def unregister_url_callback(self, pattern):
+        """Unregister the callback for URLs matching the regex ``pattern``
+
+        :param pattern: compiled regex pattern to unregister callback
+
+        .. versionadded:: 7.0
+
+            This method replaces manual management of ``url_callbacks`` in
+            Sopel's plugins, so instead of doing this in ``shutdown()``::
+
+                regex = re.compile(r'http://example.com/path/.*')
+                try:
+                    del bot.memory['url_callbacks'][regex]
+                except KeyError:
+                    pass
+
+            use this much more concise pattern::
+
+                regex = re.compile(r'http://example.com/path/.*')
+                bot.unregister_url_callback(regex)
+
+        """
+        if not self.memory.contains('url_callbacks'):
+            # nothing to unregister
+            return
+
+        if isinstance(pattern, basestring):
+            pattern = re.compile(pattern)
+
+        try:
+            del self.memory['url_callbacks'][pattern]
+        except KeyError:
+            pass
+
+    def search_url_callbacks(self, url):
+        """Yield callbacks found for ``url`` matching their regex pattern
+
+        :param str url: URL found in a trigger
+        :return: yield 2-value tuples of ``(callback, match)``
+
+        For each pattern that matches the ``url`` parameter, it yields a
+        2-value tuple of ``(callable, match)`` for that pattern.
+
+        The ``callable`` is the one registered with
+        :meth:`register_url_callback`, and the ``match`` is the result of
+        the regex pattern's ``search`` method.
+
+        .. versionadded:: 7.0
+
+        .. seealso::
+
+            The Python documentation for the `re.search`__ function and
+            the `match object`__.
+
+        .. __: https://docs.python.org/3.6/library/re.html#re.search
+        .. __: https://docs.python.org/3.6/library/re.html#match-objects
+
+        """
+        for regex, function in tools.iteritems(self.memory['url_callbacks']):
+            match = regex.search(url)
+            if match:
+                yield function, match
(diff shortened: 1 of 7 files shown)


dgw:
You can forge on ahead with a new PR branch based on this one, like you've done before, unless you need to discuss direction (in #1489) first. That way, no matter how much bikeshedding happens with replacing the url module, we'll have a clean new interface in the bag.

I like that this really is just adding convenience methods to do what plugins already do, so nothing will break (yet) if module code isn't updated. 😁

Do you think it's worth checking in the register/unregister functions added here whether the `pattern` is actually compiled? I can picture users who merely skimmed the documentation (or coders drunk on lack of sleep XD) trying to pass in an uncompiled regex string…

sopel/bot.py
```diff
@@ -648,3 +646,71 @@ def cap_req(self, module_name, capability, arg=None, failure_callback=None,
...
                                  success_callback))
             self._cap_reqs[cap] = entry
+
+    def register_url_callback(self, pattern, callback):
+        """Register a ``callback`` for URLs matching the regex ``pattern``
+
+        :param pattern: compiled regex pattern to register
+        :param callback: callable object to handle matching URLs
+
+        .. versionadded: 7.0.0
```
dgw: This directive won't appear in Sphinx's output as written; it needs two colons. (Also applies to the other two occurrences.)

I would also just put `7.0`, since we're not _ever_ (if I have anything to say about it) going to add/change/remove API methods in a patch release.

```suggestion
        .. versionadded:: 7.0
```

sopel/bot.py
```diff
@@ -648,3 +646,71 @@ def cap_req(self, module_name, capability, arg=None, failure_callback=None,
...
+
+            This method has been added in Sopel 7.0.0 in order to offer
+            a unified interface for URL callbacks.
```
dgw: Redundant to say "This method has been added in Sopel 7.0.0"—the `versionadded` directive already says that for us. (Applies to all three added docstrings.)

This explanatory text is OK, but I think it would be more useful to show module authors what code it replaces in their `setup` functions (the old `bot.memory` check and subsequent manual assignment into the dict).

sopel/bot.py
```diff
@@ -648,3 +646,71 @@ def cap_req(self, module_name, capability, arg=None, failure_callback=None,
...
+
+    def unregister_url_callback(self, pattern):
+        """Unregister the callback for URLs matching the regex ``pattern``
+
+        :param pattern: compiled regex pattern to unregister callback
+
+        .. versionadded: 7.0.0
+
+            This method has been added in Sopel 7.0.0 in order to offer
+            a unified interface to URL callbacks.
```
dgw: Similar to the `register_` function docstring, it would be useful here to show module authors what to replace in their `shutdown` methods.

sopel/bot.py
```diff
@@ -648,3 +646,71 @@ def cap_req(self, module_name, capability, arg=None, failure_callback=None,
...
+        2-value tuple of ``(callable, match)`` for that pattern.
+
+        The ``callable`` is the one registered with
+        :meth:`register_url_callback`, and the ``match`` is the result of
+        the regex pattern's ``search`` method.
+
+        .. versionadded: 7.0.0
+
+            This method has been added in Sopel 7.0.0 in order to offer
+            a unified interface for URL callbacks.
```
dgw: Not sure what would be more useful to put here, but the comment I made about redundancy on `register_` above applies here. Maybe this one doesn't need an explanation for the `versionadded` directive at all?

Exirel:
> I like that this really is just adding convenience methods to do what plugins already do, so nothing will break (yet) if module code isn't updated. 😁

I knew you would understand. 👍 

> Do you think it's worth checking in the register/unregister functions added here whether the `pattern` is actually compiled?

Yeah, I was wondering too. It's not that complicated, I'll do that.

Exirel:
Also: of course I failed at docstring again. x)

sopel/bot.py
Exirel: Done.

sopel/bot.py
Exirel: Done.

sopel/bot.py
Exirel: Removed explanation - done.

sopel/bot.py
Exirel: Fixed.

Exirel:
Hm. I'm not sure "re-request a review" was the right idea. Meh, I guess I need to learn what the new github has to offer with all these features.

Anyway @dgw I fixed the docstrings and added the safety measures to check if pattern is a compiled regex.

Exirel:
Ahem. Checkstyle?

sopel/bot.py
```diff
@@ -648,3 +646,97 @@ def cap_req(self, module_name, capability, arg=None, failure_callback=None,
...
+            This method replaces manual management of ``url_callbacks`` in
+            Sopel's plugins, so instead of doing this in ``setup()``::
+
+                if not bot.memory.contains('url_callbacks'):
+                    bot.memory['url_callbacks'] = tools.SopelMemory()
+
+                regex = re.compile(r'http://example.com/path/.*')
+                bot.memory['url_callbacks'][regex] = callback
+
+            Plugin's maintainers must do this::
```
dgw: "must" is a bit strong?

```suggestion
            use this much more concise pattern::
```

(starts with lowercase because it's really continuing the previous sentence)

Also applies to `unregister_`.

sopel/test_tools.py
```diff
@@ -74,6 +75,12 @@ def _init_config(self):
             os.mkdir(home_dir)
         cfg.parser.set('core', 'homedir', home_dir)
 
+    def search_url_callbacks(self, url):
```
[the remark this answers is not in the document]
Exirel: I honestly don't know why. :-/

sopel/bot.py
Exirel: I like your version. I never know how to properly use "must". :(

sopel/test_tools.py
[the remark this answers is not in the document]
Exirel: Ok, I totally need to add the register/unregister method too, because it might fail in the future.

It's complicated, but TL;DR: it works now because none of the plugin that use register/unregister have test example, so their setup function is never called, so these methods are never called. However, the URL module is tested, and it uses the search method, so it needs to be here.

So, right now, it's pure luck that it works. I'll fix that and make sure it works *because that's the right thing to do*.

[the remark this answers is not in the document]
dgw: OK, now we're adding three of `bot`'s methods to the `MockSopel` but not any of the others that modules might call. Something feels very wrong about this.

`say`, `reply`, and `action` are defined in `MockSopelWrapper` to store the command's output:

https://github.com/sopel-irc/sopel/blob/d2e7a163ff06e7697e857e77b14681cac251c845/sopel/test_tools.py#L87

We're not already looking at test problems because, for example, none of the core modules that use `bot.db` have testable examples. If they did have example results, errors would happen.

Same for `notice`—no core module command both has a testable example and calls `bot.notice()` while doing its work. If that happened, error.

I'm not going to "Unresolve" this conversation, because doing anything about any of this would be way out of scope for this PR, but… Basically, next time you feel ambitious about making the _tests_ better, it would be worth finding an alternative solution, like using the real Sopel but mocking out only its connection to IRC instead. Keeping the "real" and "mock" implementations/APIs in sync is only going to get sillier as we add features to the API and try to improve test coverage of modules.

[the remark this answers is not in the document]
Exirel: I fully agree with you. I'm not happy with these mocks, and they don't cover everything we need.

As said on IRC, we need time to figure something out that will be better and fit our needs. I'm sure we just need some time for this to sink in and come back to that topic with fresh ideas.

dgw:
Merging so @Exirel can properly get started on a PR to fix the mess that is `url.py`. :vulcan_salute:

PR #1413, opened 2018-11-14
url: add punctuation cleaning

dgw:
Resolves #726
Resolves #1390

I'm still prodding the code for a good way to handle parentheses/brackets, but it'll be done soon™.

PR #1510, opened 2019-03-21
core: handle URL callbacks in coretasks

### LINKED ISSUE #1489: Disabling url.py module also disables @url decorator

Exirel:
This fix #1489 - at last!

PR #1508 added a new interface for plugin maintainers, that would help us work with URL callbacks. In this PR, I:

* never trigger URL Callbacks from the `url` built-in plugin
* share the "find URL" feature into `sopel.web` (with unit tests for good measure)
* trigger URL callbacks from a `@url` defined in `coretasks`

This way, one may disable the `url` plugin, and the bot won't auto-title URLs. However, other plugins will still be able to define and use URL Callbacks.

Plugins that use `@rule` and define their own callback will still work, and that won't trigger callbacks.

By the way, this should fix https://github.com/sopel-irc/sopel-github/issues/13 too as soon as the github plugin uses `@url` instead of `@rule`.

### DIFF
--- a/sopel/config/core_section.py
+++ b/sopel/config/core_section.py
@@ -91,7 +91,13 @@ class CoreSection(StaticSection):
         'auto_url_schemes',
         strip=True,
         default=['http', 'https', 'ftp'])
-    """List of URL schemes that will trigger URL callbacks."""
+    """List of URL schemes that will trigger URL callbacks.
+
+    Used by the URL callbacks feature; see :func:`sopel.module.url` decorator
+    for plugins.
+
+    The default value allows ``http``, ``https``, and ``ftp``.
+    """
 
     bind_host = ValidatedAttribute('bind_host')
     """Bind the connection to a specific IP"""
--- a/sopel/config/core_section.py
+++ b/sopel/config/core_section.py
@@ -87,6 +87,12 @@ class CoreSection(StaticSection):
 
     May not apply, depending on ``auth_method``."""
 
+    auto_url_schemes = ListAttribute(
+        'auto_url_schemes',
+        strip=True,
+        default=['http', 'https', 'ftp'])
+    """List of URL schemes that will trigger URL callbacks."""
+
     bind_host = ValidatedAttribute('bind_host')
     """Bind the connection to a specific IP"""
 
--- a/sopel/coretasks.py
+++ b/sopel/coretasks.py
@@ -18,6 +18,7 @@ import sys
 import time
 import sopel
 import sopel.module
+import sopel.web
 from sopel.bot import _CapReq
 from sopel.tools import Identifier, iteritems, events
 from sopel.tools.target import User, Channel
@@ -797,3 +798,20 @@ def track_topic(bot, trigger):
     if channel not in bot.channels:
         return
     bot.channels[channel].topic = trigger.args[-1]
+
+
+@sopel.module.rule(r'(?u).*(https?://\S+).*')
+@sopel.module.unblockable
+def handle_url_callbacks(bot, trigger):
+    """Dispatch callbacks on URLs
+
+    For each URL found in the trigger, trigger the URL callback registered by
+    the ``@url`` decorator.
+    """
+    # find URLs in the trigger
+    for url in sopel.web.search_urls(trigger):
+        # find callbacks for said URL
+        for function, match in bot.search_url_callbacks(url):
+            # trigger callback defined by the `@url` decorator
+            if hasattr(function, 'url_regex'):
+                function(bot, trigger, match=match)
--- a/sopel/coretasks.py
+++ b/sopel/coretasks.py
@@ -800,7 +800,7 @@ def track_topic(bot, trigger):
     bot.channels[channel].topic = trigger.args[-1]
 
 
-@sopel.module.rule(r'(?u).*(https?://\S+).*')
+@sopel.module.rule(r'(?u).*(.+://\S+).*')
 @sopel.module.unblockable
 def handle_url_callbacks(bot, trigger):
     """Dispatch callbacks on URLs
@@ -808,8 +808,9 @@ def handle_url_callbacks(bot, trigger):
     For each URL found in the trigger, trigger the URL callback registered by
     the ``@url`` decorator.
     """
+    schemes = bot.config.core.auto_url_schemes
     # find URLs in the trigger
-    for url in sopel.web.search_urls(trigger):
+    for url in sopel.web.search_urls(trigger, schemes=schemes):
         # find callbacks for said URL
         for function, match in bot.search_url_callbacks(url):
             # trigger callback defined by the `@url` decorator
--- a/sopel/modules/url.py
+++ b/sopel/modules/url.py
@@ -95,6 +95,10 @@ def setup(bot):
     if not bot.memory.contains('last_seen_url'):
         bot.memory['last_seen_url'] = tools.SopelMemory()
 
+    # Initialize shortened_urls as a dict if it doesn't exist.
+    if not bot.memory.contains('shortened_urls'):
+        bot.memory['shortened_urls'] = tools.SopelMemory()
+
 
 @commands('title')
 @example('.title http://google.com', '[ Google ] - google.com')
@@ -192,9 +196,6 @@ def process_urls(bot, trigger, urls):
             if (shorten_url_length > 0) and (len(url) > shorten_url_length):
                 # Check bot memory to see if the shortened URL is already in
                 # memory
-                if not bot.memory.contains('shortened_urls'):
-                    # Initialize shortened_urls as a dict if it doesn't exist.
-                    bot.memory['shortened_urls'] = tools.SopelMemory()
                 if bot.memory['shortened_urls'].contains(url):
                     tinyurl = bot.memory['shortened_urls'][url]
                 else:
--- a/sopel/modules/url.py
+++ b/sopel/modules/url.py
@@ -115,26 +115,15 @@ def title_command(bot, trigger):
         else:
             urls = [bot.memory['last_seen_url'][trigger.sender]]
     else:
-        urls = list(
-            web.search_urls(
-                trigger,
-                exclusion_char=bot.config.url.exclusion_char
-            )
-        )
+        urls = web.search_urls(
+            trigger,
+            exclusion_char=bot.config.url.exclusion_char)
 
-    results = list(process_urls(bot, trigger, urls))
-    for _url, title, domain, tinyurl in results[:4]:
+    for _url, title, domain, tinyurl in process_urls(bot, trigger, urls):
         message = '[ %s ] - %s' % (title, domain)
         if tinyurl:
             message += ' ( %s )' % tinyurl
         bot.reply(message)
-
-    # Nice to have different failure messages for one-and-only requested URL
-    # failed vs. one-of-many failed.
-    if len(urls) == 1 and not results:
-        bot.reply('Sorry, fetching that title failed. Make sure the site is working.')
-    elif len(urls) > len(results):
-        bot.reply('I couldn\'t get all of the titles, but I fetched what I could!')
 
 
 @module.rule(r'(?u).*(https?://\S+).*')
@@ -155,8 +144,7 @@ def title_auto(bot, trigger):
     urls = web.search_urls(
         trigger, exclusion_char=bot.config.url.exclusion_char, clean=True)
 
-    for i, info in enumerate(process_urls(bot, trigger, urls)):
-        url, title, domain, tinyurl = info
+    for url, title, domain, tinyurl in process_urls(bot, trigger, urls):
         message = '[ %s ] - %s' % (title, domain)
         if tinyurl:
             message += ' ( %s )' % tinyurl
@@ -164,10 +152,6 @@ def title_auto(bot, trigger):
         if message != trigger:
             bot.say(message)
             bot.memory['last_seen_url'][trigger.sender] = url
-
-        # stop at 4th iteration
-        if i == 3:
-            break
 
 
 def process_urls(bot, trigger, urls):
--- a/sopel/modules/url.py
+++ b/sopel/modules/url.py
@@ -15,10 +15,8 @@ import re
 
 import requests
 
-from sopel import web, tools, __version__
-from sopel.config.types import ValidatedAttribute, ListAttribute, StaticSection
-from sopel.module import commands, rule, example
-
+from sopel import __version__, module, tools, web
+from sopel.config.types import ListAttribute, StaticSection, ValidatedAttribute
 
 USER_AGENT = 'Sopel/{} (https://sopel.chat)'.format(__version__)
 default_headers = {'User-Agent': USER_AGENT}
@@ -100,8 +98,8 @@ def setup(bot):
         bot.memory['shortened_urls'] = tools.SopelMemory()
 
 
-@commands('title')
-@example('.title http://google.com', '[ Google ] - google.com')
+@module.commands('title')
+@module.example('.title https://www.google.com', '[ Google ] - www.google.com')
 def title_command(bot, trigger):
     """
     Show the title or URL information for the given URL, or the last URL seen
@@ -139,7 +137,7 @@ def title_command(bot, trigger):
         bot.reply('I couldn\'t get all of the titles, but I fetched what I could!')
 
 
-@rule(r'(?u).*(https?://\S+).*')
+@module.rule(r'(?u).*(https?://\S+).*')
 def title_auto(bot, trigger):
     """
     Automatically show titles for URLs. For shortened URLs/redirects, find
--- a/sopel/modules/url.py
+++ b/sopel/modules/url.py
@@ -22,7 +22,6 @@ from sopel.module import commands, rule, example
 
 USER_AGENT = 'Sopel/{} (https://sopel.chat)'.format(__version__)
 default_headers = {'User-Agent': USER_AGENT}
-find_urls = None
 # These are used to clean up the title tag before actually parsing it. Not the
 # world's best way to do this, but it'll do for now.
 title_tag_data = re.compile('<(/?)title( [^>]+)?>', re.IGNORECASE)
@@ -73,8 +72,6 @@ def configure(config):
 
 
 def setup(bot):
-    global find_urls
-
     bot.config.define_section('url', UrlSection)
 
     if bot.config.url.exclude:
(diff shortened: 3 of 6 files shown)


sopel/coretasks.py
```diff
@@ -18,6 +18,7 @@
 import time
 import sopel
 import sopel.module
+import sopel.web
```
dgw: It's unnecessary to import `sopel.submodule` if `sopel` is already imported, unless doing `from sopel.submodule import SpecificThing`.

sopel/modules/url.py
```diff
@@ -226,21 +198,15 @@ def process_urls(bot, trigger, urls):
     return results
 
 
-def check_callbacks(bot, trigger, url, run=True):
+def check_callbacks(bot, trigger, url):
```
dgw: Update docstring to remove the `run` param too.

test/test_web.py
```diff
@@ -0,0 +1,69 @@
...
+    urls = list(search_urls('!http://example.com', exclusion_char='!'))
+    assert not urls, 'Must not find URL, found %d' % len(urls)
+
+    urls = list(
+        search_urls(
+            '!http://example.com http://example.com',
+            exclusion_char='!'
+        )
+    )
+    assert len(urls) == 1, 'Must find 1 URL, found %d' % len(urls)
```
dgw: This isn't a useful test of the exclusion character, since even if the `exclusion_char` doesn't work properly the `search_urls` function will only return one item anyway (because it only returns unique URLs).

Better to double down on the following test, and assert that `http://a.com` is not in the returned list.

sopel/web.py
```diff
@@ -237,3 +237,39 @@ def iri_to_uri(iri):
     urlencode = urllib.urlencode
 else:
     urlencode = urllib.parse.urlencode
+
+
+def search_urls(text, exclusion_char=None, clean=False):
+    def trim_url(url):
```
dgw: I think this should either become its own utility function (in `sopel.web` or `sopel.tools`), or become inlined code in the only place where it's used. Defining a function inside another is one of those oddball things that we should probably avoid doing unless there's a good reason.

sopel/coretasks.py
Exirel: I was keeping the style of import used in `coretasks.py`. I don't like it. I'd rather have `from sopel import module, web, [...]`... but I would also keep that for another PR, if you agree.

dgw: > another PR

Yeah, OK, I'll lay off for now and you can fix a bunch of those all at once. :+1:

sopel/web.py
Exirel: I moved this function mostly "as-is", adding the `exclusion_char` parameter.

Fun fact: since `find_url` was never call with `clean=True`, `trim_url` was (and still is) never called, ever. And I've no idea why.

test/test_web.py
Exirel: I agree that it's a very edge-case, but here is my thought process.

First, this is what we do: ignore **this particular instance** of a URL:

    >>> search_urls('!http://a.com http://b.com http://a.com', '!')
    ['http://b.com', 'http://a.com']

Which is fine by me. But it could mean something else: blacklist **all instance** of a URL:

    >>> search_urls('!http://a.com http://b.com http://a.com', '!')
    ['http://b.com']

I think I should change my test to add this extra `http://b.com` URL, maybe that would speak to you?

dgw: No, it should _not_ blacklist **all instances** of that URL in the same line, haha.

I think I get it—this test is checking that the `exclusion_char` only applies to the URL it abuts, and not all identical URLs in the line. The previous and next tests check that the `exclusion_char` actually excludes the single URL it abuts.

You needn't change the tests. Just a couple comments about what each is testing (since they're grouped under the same function) would be enough.

sopel/web.py
dgw: > Fun fact: since `find_url` was never call with `clean=True`, `trim_url` was (and still is) never called, ever. And I've no idea why.

I know why. It's because, before the flood of collaborators Sopel gained in the last few months, I left a small but very important change out of #1413 and there was nobody else looking at the code to call me out on my mistake. >.> So that'll be a good (bugfix) bugfix to add to the 6.6.x pile.

For this PR, you'll want to add a similar bugfix in the same place. I haven't written the patch just yet, but it's probably going to involve just passing `clean=True` from `title_auto` in the `url` module, so you can tentatively do the equivalent here.

test/test_web.py
Exirel: Will do! :)

sopel/web.py
dgw: Update: The relevant bugfix is in #1515.

Exirel: >  so you can tentatively do the equivalent here

I'll add some tests and do this.

Exirel: I did extract it, I let you see if you agree with that.

Exirel:
@dgw I think I fixed most of the issue, you're welcome for the second round, and once everything's fine I'll rebase/squash accordingly. :)

Exirel:
> multiple spaces before operator

Damn, I used `flake8 sopel` but forgot the test-suite. I was quite in a hurry this morning, my bad!

test/test_web.py
```diff
@@ -93,3 +94,50 @@ def test_search_urls_exclusion_char_only_once():
...
+
+@pytest.mark.parametrize('trailing_char', TRAILING_CHARS)
+@pytest.mark.parametrize('left, right', ENCLOSING_PAIRS)
+def test_trim_url_trailing_char_and_enclosing(trailing_char, left, right):
+    test_url = 'http://exemple.com/'
+    assert test_url == trim_url(test_url + right + trailing_char)
+
+    # assert the trailing char is kept if there is something else
+    test_url = 'http://exemple.com/' + trailing_char
+    assert test_url  == trim_url(test_url + right)
```
Exirel: The double-space offender is here!

test/test_web.py
Exirel: Done, have a look at the new tests! :)

Exirel:
What the hell is wrong with Python 2.7 this time???

dgw:
Is that really an issue in py2? If so it probably needs to be applied on 6.6.x too :/

Exirel:
> Is that really an issue in py2? If so it probably needs to be applied on 6.6.x too :/

Yes. :sob:

dgw:
> > Is that really an issue in py2? If so it probably needs to be applied on 6.6.x too :/
> 
> Yes. 😭

Well. You'd think I would have caught that myself, given that my primary Sopel instance still runs in py2 and I'm therefore invested in maintaining compatibility…

Added this bugfix to my todo list for 6.6.x. :/

Exirel:
@dgw From my point of view, everything's fine. Yet it appears that Travis isn't happy with an unrelated test (it has something to do with Google stuff).

So if you want to merge, it's ready for the last review.

Exirel:
Py2.7 & Py3.4 tests pass with the exception of the `ip` test, so I think we are good. :)

dgw:
I'll get Travis to re-run the tests after #1523 makes it into `master`. Did you want to do any squashing on this branch at all, @Exirel, or just leave all the commits in? (I'm fine with having the full evolution, just curious if you planned to squash down the little fixups like for flake8 failures and py2.7 compat etc.)

sopel/coretasks.py
```diff
@@ -791,3 +792,20 @@ def track_topic(bot, trigger):
     if channel not in bot.channels:
         return
     bot.channels[channel].topic = trigger.args[-1]
+
+
+@sopel.module.rule(r'(?u).*(https?://\S+).*')
```
dgw: I wonder about protocol support. Like, there's no reason that plugins shouldn't be able to register handlers for any kind of URL, not just HTTP(S) links, especially since your added `sopel.web.search_urls()` function includes `ftp` in the list of protocols. (Side note: `ftps` exists, too. 😸)

But like, a gaming channel might want a bot to handle `steam://` links, for example.

Could be a separate PR, sure! But I think it's worth putting in a bit of extra work on supporting arbitrary protocols if enabled by the bot owner and/or a plugin—and that means not hard-coding `https?` here.

sopel/modules/url.py
```diff
@@ -92,33 +89,13 @@ def setup(bot):
...
-            urls = [trim_url(url) for url in urls]
-        return urls
-
-    find_urls = find_func
-
-
-@commands('title')
-@example('.title http://google.com', '[ Google ] - google.com')
+@module.commands('title')
+@module.example('.title http://google.com', '[ Google ] - google.com')
```
dgw: Want to bring this link into the modern era while you're in here?

```suggestion
@module.example('.title https://www.google.com', '[ Google ] - www.google.com')
```

sopel/modules/url.py
```diff
@@ -167,20 +148,22 @@ def title_auto(bot, trigger):
...
         message = '[ %s ] - %s' % (title, domain)
         if tinyurl:
             message += ' ( %s )' % tinyurl
         # Guard against responding to other instances of this bot.
         if message != trigger:
             bot.say(message)
+            bot.memory['last_seen_url'][trigger.sender] = url
+
+        # stop at 4th iteration
+        if i == 3:
```
dgw: Why is this one (auto-title) `enumerate(process_urls([blah]))` and the other (manual command) is `results[:4]`?

I feel like we shouldn't need an `i` counter variable in Python. :/ But I also think both places should use the same pattern, not have one store all URLs in a `results` list and the other stop iteration before processing all URLs.

sopel/web.py
```diff
@@ -237,3 +237,40 @@ def iri_to_uri(iri):
...
+
+    return url
+
+
+def search_urls(text, exclusion_char=None, clean=False):
+    re_url = r'((?:http|https|ftp)(?::\/\/\S+))'
+    if exclusion_char is not None:
+        re_url = r'((?<!%s)(?:http|https|ftp)(?::\/\/\S+))' % (exclusion_char)
+
+    r = re.compile(re_url, re.IGNORECASE | re.UNICODE)
```
dgw: It shouldn't be necessary to recompile the regex every time this function is called. Runtime changes to the `exclusion_char` (via `.set`) are the only things that could invalidate it, and at the moment I don't think supporting that is worth the efficiency loss.

Even if it's only running every time a link is found (with the arbitrary protocol support I'd like to add eventually, at least every line containing `://`), it's still not good to recompile regexes all the time for no reason. It's a bad habit. :)

sopel/web.py
```diff
@@ -237,3 +237,40 @@ def iri_to_uri(iri):
...
+
+    urls = re.findall(r, text)
+    if clean:
+        urls = (trim_url(url) for url in urls)
+
+    # yield unique URL in their order of appearance
```
dgw: ♪ Nitpick time~ ♪

```suggestion
    # yield unique URLs in their order of appearance
```

test/test_web.py
```diff
@@ -0,0 +1,143 @@
...
+    assert 'http://b.com' in urls
+
+
+TRAILING_CHARS = list('.,?!\'":;')
+ENCLOSING_PAIRS = [('(', ')'), ('[', ']'), ('{', '}'), ('<', '>')]
+
+
+@pytest.mark.parametrize('trailing_char', TRAILING_CHARS)
+def test_trim_url_remove_trailing_char(trailing_char):
+    test_url = 'http://exemple.com/'
```
dgw: It doesn't actually _matter_ in the slightest, but above here we have `example.com` and from here to the end of the file it's `exemple.com` instead. 😕

sopel/coretasks.py
Exirel: Hey, I wonder the same thing, because when I decided to kept the old behavior, I found it odd. At this point, I changed so many little things that I think it's safe to change a bit more.

Say I change the rule to `.*`, then perform a search based on a list of protocol provided by configuration. Say `core.url_allowed_schemes`?

sopel/modules/url.py
Exirel: Yeah, but there is a problem:

* we fetch all URLs, no matter how many we display
* we display only 4 of them at most (like it's a common thing)
* in the other function, we try to be smart with an error message if the number of URL does not match the number of result
* one of the function check for malicious links, the other doesn't,
* etc.

I think the differences between these functions are irrelevant and should be restructured into this:

* always display *all* URL, it's not like people will flood with tinyurl on 512 bytes (minus their hostmask)
* display an error message per URL that we fail to get (and not just URL that are handled by a callback)
* apply the same protection on all URLs
* register the "last seen URL" on both .title and auto-url

This way I can fix that `enumerate` problem, and never rely on `list(generator-function)` again. :D

sopel/web.py
Exirel: Don't worry, internaly, `re.compile` use caching: it won't re-process the same pattern twice.

test/test_web.py
Exirel: lol, French being bad at English :D

dgw:
Shall we get back on this horse now that you have a laptop again, @Exirel? :grin:

Exirel:
Oh yes, right. I'll resume working on that very soon.

sopel/config/core_section.py
```diff
@@ -216,3 +216,9 @@ def homedir(self):
 
     verify_ssl = ValidatedAttribute('verify_ssl', bool, default=True)
     """Whether to require a trusted SSL certificate for SSL connections."""
+
+    auto_url_schemes = ListAttribute(
+        'auto_url_schemes',
+        strip=True,
+        default=('http', 'https', 'ftp'))
```
dgw: `ListAttribute` should default to a `list`, hmm?
```suggestion
        default=['http', 'https', 'ftp'])
```

sopel/config/core_section.py
```diff
@@ -216,3 +216,9 @@ def homedir(self):
 
     verify_ssl = ValidatedAttribute('verify_ssl', bool, default=True)
     """Whether to require a trusted SSL certificate for SSL connections."""
+
+    auto_url_schemes = ListAttribute(
```
dgw: This belongs below the `auth_*` values. The properties defined here are alphabetized.

sopel/config/core_section.py
```diff
@@ -216,3 +216,9 @@ def homedir(self):
...
+        'auto_url_schemes',
+        strip=True,
+        default=('http', 'https', 'ftp'))
+    """List of URL schemes the URL callbacks are triggered."""
```
dgw:
```suggestion
    """List of URL schemes that will trigger URL callbacks."""
```

Need to think about whether it's possible, but this should link to documentation for that API feature. Maybe `bot.register_url_callback()`?

sopel/web.py
```diff
@@ -237,3 +237,43 @@ def iri_to_uri(iri):
...
+
+    # clean unmatched parentheses/braces/brackets
+    for (opener, closer) in [('(', ')'), ('[', ']'), ('{', '}'), ('<', '>')]:
+        if url[-1] == closer and url.count(opener) < url.count(closer):
+            url = url[:-1]
+
+    return url
+
+
+def search_urls(text, exclusion_char=None, clean=False, schemes=None):
```
dgw: I've this feeling that the function name should be `parse_urls` or `find_urls` instead… Thoughts?

Exirel: I'd argue that `parse_urls` should parse URLs, ie. get a list of URLs and extract something from them. `find_urls` on the other hand would do the same thing as `search_urls`, but I can't see the difference. Would you tell me what your gut feeling are saying?

Exirel:
@dgw I manually tested that everything worked fine. Now I'll try to write some documentation. This can be done in this PR, or in a separated one. As you wish. :)

Exirel:
I added more information in the `sopel.module.url` decorator's docstring, with a link to `auto_url_schemes` and to `bot.register_url_callback`.

Note that I've tested the example added in this docstring to make sure it worked as intented.

sopel/config/core_section.py
```diff
@@ -87,6 +87,18 @@ class CoreSection(StaticSection):
...
+    auto_url_schemes = ListAttribute(
+        'auto_url_schemes',
+        strip=True,
+        default=['http', 'https', 'ftp'])
+    """List of URL schemes that will trigger URL callbacks.
+
+    Used by the URL callbacks feature, see :func:`sopel.module.url` decorator
+    for plugins.
+
+    The default value will allows ``http``, ``https``, and ``ftp``.
```
dgw:
```suggestion
    The default value will allow ``http``, ``https``, and ``ftp``.
```

or

```suggestion
    The default value allows ``http``, ``https``, and ``ftp``.
```

sopel/config/core_section.py
```diff
@@ -87,6 +87,18 @@ class CoreSection(StaticSection):
 
     May not apply, depending on ``auth_method``."""
 
+    auto_url_schemes = ListAttribute(
+        'auto_url_schemes',
+        strip=True,
+        default=['http', 'https', 'ftp'])
+    """List of URL schemes that will trigger URL callbacks.
+
+    Used by the URL callbacks feature, see :func:`sopel.module.url` decorator
```
dgw:
```suggestion
    Used by the URL callbacks feature; see :func:`sopel.module.url` decorator
```

:smirk:

Exirel:
@dgw ready for a last review to :ship:

Exirel:
Fixup commits squashed, ready to :ship:

PR #1511, opened 2019-03-22
systemd service files: use network-online.target

ddstreet:
The service files also specify their unit to start After=network.target
but that target only defines the start of networking, and network
configuration may not be complete yet; instead it should use
network-online.target to ensure it starts after the network is
ready to use.

[EDIT: I removed the PIDFile= removal that was originally in this pull request]

### DIFF
--- a/contrib/sopel.service
+++ b/contrib/sopel.service
@@ -1,7 +1,7 @@
 [Unit]
 Description=Sopel IRC bot
 Documentation=https://sopel.chat/
-After=network.target
+After=network-online.target
 
 [Service]
 Type=simple
--- a/contrib/sopel@.service
+++ b/contrib/sopel@.service
@@ -1,7 +1,7 @@
 [Unit]
 Description=Sopel IRC bot
 Documentation=https://sopel.chat/
-After=network.target
+After=network-online.target
 DefaultInstance=sopel
 
 [Service]


dgw:
Switching to `network-online.target` is a good catch, and thanks!

About the `PIDFile` options, though:

Sopel writes its own PID file to the location given in each example systemd configuration. Keeping those lines means systemd will clean up after Sopel if it crashes in such a way that it leaves the PID file behind. I don't see any reason _not_ to keep them in [the systemd docs](https://www.freedesktop.org/software/systemd/man/systemd.service.html). All they have to say is that setting the option is "recommended" for `forking` services.

Since @Exirel recently did a major overhaul of the `run` code (including the functions that manage Sopel's PID file), I'd like to get his opinion too.

ddstreet:
> Sopel writes its own PID file to the location given in each example systemd configuration

ah, ok.  So, I could update the pull req to change the service to 'forking' (and add -d).  That may make more sense also, as it would delay systemd from considering the service 'started' until after the actual fork() - currently, systemd considers the sopel service 'started' immediately.

or, i could just remove the PIDFile change from the pull req and leave as-is, with only the After= change.

ddstreet:
ok, i removed the PIDFile= removal and updated the commit log - this is only changing the After=network.target to use network-online.target now.

Exirel:
Very good catch on `network-online.target`!

About the PID, there are 2 options:

* old-style service using a fork, that manage their own PID file,
* new-style service, using systemd to manage both the PID and the process,

Sopel was written to be an old-style service that manage its own PID file, but if we really want to do that, we should use the `--fork` option, and to provide the appropriate filename. By default, it's `~/.sopel/sopel.pid` (ie. Sopel's homedir + `sopel.pid`). It would mean to set `Type=forking` in the systemd service unit too, to be totally accurate.

However, it's totally possible to use Sopel as a new-style service, and ignore the PID file entirely: this is more or less how it works by default, with the exception of writing its own PID file. If `systemd` see `PIDFile=...` but can't read the PID, it will ignore it and do its job as it used to do - so it "works".

Further reading: [Type](https://www.freedesktop.org/software/systemd/man/systemd.service.html#Type=) and [PIDFile](https://www.freedesktop.org/software/systemd/man/systemd.service.html#PIDFile=).

dgw:
:trophy: Congratulations on your first merged PR in the Sopel project, @ddstreet!

Issue #1097, opened 2016-07-01
date/time in logs


dasu:
Would it be possible to add date/time to exceptions.log and stdio.log ?

Exirel:
In Sopel 7.x it'll be possible to configure the log format, to include the date. This was done in #1512 and is documented in #1646.

PR #1512, opened 2019-03-22
logging: add format and datefmt to allow configuring the log output from cfg file

ddstreet:
logging currently uses the default msg format and date format; add cfg file parameters to allow users to specify what format they would like for both msgs and timestamps.

### DIFF
--- a/sopel/logger.py
+++ b/sopel/logger.py
@@ -21,10 +21,8 @@ class IrcLoggingHandler(logging.Handler):
 
 
 class ChannelOutputFormatter(logging.Formatter):
-    def __init__(self):
-        super(ChannelOutputFormatter, self).__init__(
-            fmt='[%(filename)s] %(message)s'
-        )
+    def __init__(self, fmt='[%(filename)s] %(message)s', datefmt=None):
+        super(ChannelOutputFormatter, self).__init__(fmt=fmt, datefmt=datefmt)
 
     def formatException(self, exc_info):
         # logging will through a newline between the message and this, but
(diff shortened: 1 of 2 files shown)


sopel/logger.py
```diff
@@ -33,10 +33,15 @@ def formatException(self, exc_info):
 
 
 def setup_logging(bot):
-    level = bot.config.core.logging_level or 'WARNING'
-    logging.basicConfig(level=level)
+    cfg = bot.config.core
```
dgw: I'd rather not shortcut-reference the core config section like this. It makes the code harder to understand when coming back to it months/years later.

ddstreet:
> level is no longer defined after the changes above. This failed the linter.

oops!  will fix that

> I'd rather not shortcut-reference the core config section like this

will change this also.

sopel/logger.py
```diff
@@ -33,12 +31,22 @@ def formatException(self, exc_info):
...
+    params = {'level': bot.config.core.logging_level or 'WARNING'}
+    if bot.config.core.logging_format:
+        params['format'] = bot.config.core.logging_format
+    if bot.config.core.logging_datefmt:
+        params['datefmt'] = bot.config.core.logging_datefmt
+    logging.basicConfig(**params)
     logger = logging.getLogger('sopel')
     if bot.config.core.logging_channel:
+        level = (bot.config.core.logging_channel_level or
+                 bot.config.core.logging_level or 'WARNING')
```
Exirel: I think you should define a variable `base_level = bot.config.core.logging_level or 'WARNING'`, and the use it everywhere you need it:

```
base_level = bot.config.core.logging_level or 'WARNING'
params = {'level': base_level}

# ...
level = (bot.config.core.logging_channel_level or base_level)
```

Exirel: And same for `bot.config.core.logging_format` with `bot.config.core.logging_channel_format` btw.

sopel/logger.py
Exirel: Agree on that!

sopel/logger.py
```diff
@@ -33,10 +33,15 @@ def formatException(self, exc_info):
...
+    cfgparams = {'level': cfg.logging_level or 'WARNING'}
+    if cfg.logging_format:
+        cfgparams['format'] = cfg.logging_format
+    if cfg.logging_datefmt:
+        cfgparams['datefmt'] = cfg.logging_datefmt
+    logging.basicConfig(**cfgparams)
     logger = logging.getLogger('sopel')
-    if bot.config.core.logging_channel:
+    if cfg.logging_channel:
         handler = IrcLoggingHandler(bot, level)
```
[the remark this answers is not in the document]
ddstreet: corrected with latest version in git

sopel/logger.py
ddstreet: corrected in latest version in git

sopel/logger.py
ddstreet: just updated to use base_level et. al. as suggested; I also added channel_* vars and a channel_params dict, so only those params that are defined are included in the formatter contstructor.

ddstreet:
updated as suggested; please let me know if you have any add'l requests/suggestions.  thnx!

ddstreet:
(pushed minor indentation correction)

Exirel:
**LGTM** :+1: :shipit: :unicorn:

Exirel:
This should fix #1097 or at least it should allow users to configure logging in a way that fixes it for them.

Exirel:
@dgw I was looking at logging & the output redirect we use for Sopel, and I want to do some work around that. This PR would bring some very interesting feature I'd really like to reuse as soon as possible. If you didn't pick yet your next PR to merge, here is a very good candidate!

PR #1514, opened 2019-03-22
config: add [core] 'interpolation' and 'interpolation_extended' params

ddstreet:
RawConfigParser is the legacy python2 parser and does not allow/perform
interpolation.  More complex config files can benefit greatly from
interpolation and so we should use ConfigParser and allow interpolation.

The new params allow backwards compatibility if neither param is set
(or they are set to False).  If either basic or extended interpolation
is desired, either param can be added to the [core] section and set to
True to cause sopel to change its parser to the newer ConfigParser
and re-read the file, with either basic (the default) or extended
interpolation enabled.

Note that as extended interpolation is not available in Python2, if
the config file has 'interpolation_extended' enabled when running
Sopel under Python2, a RuntimeError is raised.

### DIFF
--- a/sopel/config/__init__.py
+++ b/sopel/config/__init__.py
@@ -26,10 +26,10 @@ import sopel.loader
 import os
 import sys
 if sys.version_info.major < 3:
-    import ConfigParser
+    import ConfigParser as configparser
 else:
     basestring = str
-    import configparser as ConfigParser
+    import configparser
 import sopel.config.core_section
 from sopel.config.types import StaticSection
 
@@ -70,8 +70,34 @@ class Config(object):
         """
         self.filename = filename
         """The config object's associated file, as noted above."""
-        self.parser = ConfigParser.RawConfigParser(allow_no_value=True)
+        parserparams = {'allow_no_value': True}
+        self.parser = configparser.RawConfigParser(**parserparams)
         self.parser.read(self.filename)
+        # Check if interpolation is desired.
+        try:
+            basic = self.parser.getboolean('core', 'interpolation')
+        except (configparser.NoSectionError, configparser.NoOptionError):
+            basic = False
+        except ValueError:
+            stderr("interpolation set to invalid value, use True/False")
+            basic = False
+        try:
+            extended = self.parser.getboolean('core', 'interpolation_extended')
+        except (configparser.NoSectionError, configparser.NoOptionError):
+            extended = False
+        except ValueError:
+            stderr("interpolation_extended set to invalid value, use True/False")
+            extended = False
+        if basic or extended:
+            if extended:
+                try:
+                    parserparams['interpolation'] = configparser.ExtendedInterpolation()
+                except AttributeError:
+                    msg = "Cannot use extended interpolation format config with Python2"
+                    stderr(msg)
+                    raise RuntimeError(msg)
+            self.parser = configparser.ConfigParser(**parserparams)
+            self.parser.read(self.filename)
         self.define_section('core', sopel.config.core_section.CoreSection,
                             validate=validate)
         self.get = self.parser.get
@@ -103,7 +129,7 @@ class Config(object):
         """
         try:
             return self.parser.add_section(name)
-        except ConfigParser.DuplicateSectionError:
+        except configparser.DuplicateSectionError:
             return False
 
     def define_section(self, name, cls_, validate=True):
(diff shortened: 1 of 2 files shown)


dgw:
Since @Exirel has done considerable work recently around config loading (and plans to do quite a bit more, if memory serves), I'll request that he review this first.

As with the PR of yours I just commented on, @ddstreet, you will need to update the `versionadded` directives in this one as well once we decide which release this will go into, if accepted. It might want to wait until Sopel 8; I'm not sure.

ddstreet:
> you will need to update the versionadded directives

will do - should i just remove them, or change them to 7.0 or 8.0?

ddstreet:
ok, removed versionadded - let me know if they should be added in later and what version to use for them.

Exirel:
I'm not a huge fan of the "read the configuration until it works" bit of logic. I'm also not very confortable replacing `RawConfigParser` by `ConfigParser` right off the bat. Since Sopel's Config object does a lot of things in the background, I'd rather see a thorough review of that first.

And then, I'd allow basic interpolation in Sopel 7.x, asking for a command-line option in Legacy mode to be activated, and activated by default with `sopel start|stop|restart` with the option to prevent it (like `--raw-config`:

* `sopel [legacy] --new-config` => RawConfig by default, --new-config to activate
* `sopel [start|stop|restart] --raw-config` => Config by default, --raw-config to deactivate

ddstreet:
> I'm not a huge fan of the "read the configuration until it works" bit of logic. 

Well, it has to be done, because parsing is done differently for no interpolation (default with RawConfigParser), basic interpolation (default with ConfigParser) and extended interpolation (python3 only with interpolation param set to extended).

I think it might be possible to do a single cfg file read, then set the parser-internal _interpolation parameter to the appropriate interpolator, but there's no public api to do that, so it would rely on python lib implementation details - I much prefer to just re-read the cfg file after we know what interpolation method to use.

> I'm also not very confortable replacing `RawConfigParser` by `ConfigParser` right off the bat.

It's not, unless the cfg file contains the special (new) 'interpolation' parameter in the [core] section.  Without that - which no existing cfg file should have - it remains parsed by RawConfigParser with no interpolation done.

> And then, I'd allow basic interpolation in Sopel 7.x, asking for a command-line option in Legacy mode to be activated, and activated by default with `sopel start|stop|restart` with the option to prevent it (like `--raw-config`:

well...cmd line param is possible, but will make the end user do a LOT more work to make it happen.  End users can very easily edit the default /etc/sopel.cfg file, but there's no provision made in the systemd service for adding cmdline params.  So forcing interpolation to use a cmdline param is pretty heavy imposition on the end user, IMHO.  We could create a /etc/[default|sysconfig]/sopel default params config file that the systemd service uses as EnvironmentFile and then passes cmdline params on to ExecStart line, but all this seems like a huge amount of work instead of just parsing a single param inside the existing cfg file.

I think I don't really understand your objection - are you objecting to re-reading the config file?  Or objecting to using a parameter inside the cfg file to control how it's parsed (i.e. with interpolation or without)?  Or something else I'm missing?

ddstreet:
I'll add one more point of consideration - the interpolation method depends on the cfg file itself, and cfg file formats are incompatible between (no, basic, extended) interpolation - for example, a basic interpolation format cfg file, when read with interpolation disabled, will have '%%' instead of '%' in its param values, and values using substitution will have e.g. '%(myotherparam)s' instead of actual values.

With the custom sopel@.service, if you force the cfg file interpolation method to be on the cmdline, then ALL users of sopel with custom cfg files (i.e. anyone using the sopel@ service instead of main /etc/sopel.cfg and main sopel.service) will have to all use the same interpolation method.  And depending on how the new cmdline param gets placed into the service files, maybe everyone would have to use the same interpolation method (sopel.service and sopel@.service).

Would you prefer if the interpolation cfg mechanism is moved out of the [core] section, to instead have something like

[interpolation]
method = none|basic|extended

ddstreet:
Ok, I've thought about this more, and while interpolation is still a very important use-case for me, being able to load multiple config files (instead of just a single cfg file) is more important - and since the ConfigParser interpolation setting affects the parser itself, not just an individual file the parser reads (i.e. if a parser reads multiple cfg files, all need to be in the same interpolation format), I think I agree with your assessment this should be a cmdline param.  I'm not totally clear on how that will get included in the systemd service script, though.

So, I think this pull req should be rejected - I'll work on a new one to use a cmdline param to control the parser's interpolation setting, and a separate one to allow multiple cfg files - unless you are already working on that and/or have ideas about how it should work?

ddstreet:
Closing this pull request - I'll work on different one(s) as mentioned above.

PR #1515, opened 2019-03-22
url: Actually clean URLs when auto-fetching titles

dgw:
A change that should have been made in #1413, but was overlooked. This proves why having more than one developer review changes is useful.

@Exirel indirectly found this omission during an overhaul of Sopel's URL-handling for version 7.

See https://github.com/sopel-irc/sopel/pull/1510#discussion_r268139814


Exirel:
Deal. 😉

PR #1517, opened 2019-03-24
url: Fix py2 compatibility in trim_url()

dgw:
Python 2 doesn't like `thing is unicode`; it wants `thing == unicode`. Of course, it just works with regular str in both py2 and py3… Oh well.

This is another find from @Exirel's overhaul of URL handling for 7.x.

As long as I was touching the line anyway, I added parentheses to make the operator precedence explicit. Likely unnecessary, but probably easier to read by some small measure. (@Exirel, if you don't think the added parens are useful, I can drop it from this PR so you don't have to mirror it in #1510. :smile_cat:)

### DIFF
--- a/sopel/modules/url.py
+++ b/sopel/modules/url.py
@@ -99,7 +99,7 @@ def setup(bot):
 
             # clean unmatched parentheses/braces/brackets
             for (opener, closer) in [('(', ')'), ('[', ']'), ('{', '}'), ('<', '>')]:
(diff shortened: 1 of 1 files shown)


dgw:
Fortunately, `find_urls` was never actually told to `clean` what it found (#1515), so this bug was only dormant in Sopel <=6.6.4 and shouldn't have affected anything.

PR #1520, opened 2019-03-25
admin: Catch set commands without arguments

kwaaak:
Before:
<Owner> .set
<Sopel> AttributeError: 'NoneType' object has no attribute 'split'

After:
<Owner> .set
<Sopel> Owner: Usage: .set section.option value

### DIFF
--- a/sopel/modules/admin.py
+++ b/sopel/modules/admin.py
@@ -182,7 +182,11 @@ def set_config(bot, trigger):
     If value is None, the option will be deleted.
     """
     # Get section and option from first argument.
-    arg1 = trigger.group(3).split('.')
+    match = trigger.group(3)
+    if match is None:
+        bot.reply("Usage: .set section.option value")
+        return
+    arg1 = match.split('.')
     if len(arg1) == 1:
         section_name, option = "core", arg1[0]
     elif len(arg1) == 2:


sopel/modules/admin.py
```diff
@@ -182,7 +182,10 @@ def set_config(bot, trigger):
     If value is None, the option will be deleted.
     """
     # Get section and option from first argument.
-    arg1 = trigger.group(3).split('.')
+    if trigger.group(3) is not None:
+        arg1 = trigger.group(3).split('.')
+    else:
+        return bot.reply("Usage: .set section.option value")
```
Exirel: The `return` here is to stop the function, and not to return `bot.reply`'s result, and I'd rather see this instead:

```python
match = trigger.group(3)
if match is None:
    bot.reply("Usage: .set section.option value")
    return

args = match.split('.')
```

dgw:
@Exirel I would like to point out that `arg1` was the preexisting variable name, so let's not change it in the whole command in this PR.

Exirel:
>  `arg1` was the preexisting variable name

Fair enough. Let's keep it that way for now then!

kwaaak:
Philosophical question whether
```python
return bot.say()
```
or
```python
bot.say()
return
```
is more readable :boom: 

Changed it to the latter, to keep the style consistent with the rest of the file.

Exirel:
> is more readable boom

As I said on IRC, I think it's not about readability, but about intent and being explicit. My reasons is that `return bot.say(...)` asks these questions:

* does bot.say return something?
* does it matter?
* why a command return something, albeit being run in a separated thread?
* what's the meaning of returning something?

which is quite confusing. As pointed out by Pilate on IRC:

> Explicit is better than implicit

and in that regard, an empty `return` statement is, by convention, the explicit way to tell "stop and quit". There is extra question, no confusion possible.

That's my opinion, at least.

PR #1521, opened 2019-03-26
dev: add a Makefile to the project and use setup.cfg to configure flake8

Exirel:
**TL;DR**: I'm super lazy and it shows.

There are 2 things I can't remember:

* the full line to checkstyle + coverage + tests
* the exact parameters for flake8

And I really like `make`, it's a great tool for any projects, not just for C/C++ stuff. So here is a short one, with few commands that I find useful. This is the kind of commands I use in my own project, be they personal or professional.

For `flake8`, I wish I could just run it without having to resort to `checkstyle.sh`, and the `setup.cfg` file is perfect for that. I tested it worked, by adding/removing ignored errors/warnings. Isn't it great?

Bonus point: I hate having to cd into docs to build the doc, then to cd back in the project.  So I added that, too.

### DIFF
--- a/checkstyle.sh
+++ b/checkstyle.sh
@@ -12,22 +12,6 @@ if ! flake8; then
     exit_code=1
 fi
 
-fail_coding=false
-for file in $(find_source_files); do
-    line=$(head -n 1 $file)
-    if echo $line | grep -q '#!/usr/bin/env python'; then
-        line=$(head -n 2 $file | tail -n 1)
-    fi
-    if ! echo $line | grep -q '# coding=utf-8'; then
-        echo $file
-        fail_coding=true
-    fi
-done
-if $fail_coding; then
-    echo "ERROR: Above files do not have utf-8 coding declared."
-    exit_code=1
-fi
-
 # Find files which use the unicode type but (heuristically) don't make it py3
 # safe
 fail_py3_unicode=false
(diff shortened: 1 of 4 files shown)


Exirel:
Checking encoding can be done with [flake8-coding](https://pypi.org/project/flake8-coding/). I will check this out for you, that might be a good addition!

For `fail_py3_unicode`, it'll be unnecessary once we drop support for Py2.7, and I think it's safe to assume we won't change the excluded files/folders that much in a year (if ever!).

Exirel:
So I tested `flake8-coding` and it looks like it works like a charm:

```
(sopel) $ pip install flake8-coding
... skip ...
(sopel) $ flake8 --version
3.6.0 (flake8_coding: 1.3.1, mccabe: 0.6.1, pycodestyle: 2.4.0, pyflakes: 2.0.0) CPython 3.5.2 on Linux
(sopel) $ flake8
./sopel/__init__.py:0:1: C101 Coding magic comment not found
$ ./checkstyle.sh 
./sopel/__init__.py:0:1: C101 Coding magic comment not found
ERROR: flake8 does not pass.
```

checkstyle.sh
```diff
@@ -4,44 +4,14 @@ find_source_files() {
...
-    fi
-    if ! echo $line | grep -q '# coding=utf-8'; then
-        echo $file
-        fail_coding=true
-    fi
-done
-if $fail_coding; then
-    echo "ERROR: Above files do not have utf-8 coding declared."
-    exit_code=1
-fi
```
dgw: One blank line (from above the removed section) is enough. This is just Bash, after all. :laughing:

```suggestion

```

Exirel:
Travis is fast again! @dgw no excuse to merge now. :shipit:

Issue #1311, opened 2018-04-05
Overhaul setup.py metadata


dgw:
Sopel's `setup.py` could do with some love. Most importantly, it would be beneficial to add project classifiers, and to specify which Python versions the package is compatible with directly in the `setup` function.

This is kind of a tracking issue, so I'll keep the task list here:

- [x] Classify project using appropriate [trove classifiers](https://pypi.python.org/pypi?%3Aaction=list_classifiers) [#1458]
- [x] Add `python_requires` entry to `setup()` call in `setup.py` [#1522]
- [x] Sort out author/maintainer/contributor/whatever people [#1787]
  - Embolalia is still listed as author, despite stepping down as maintainer, so it would be useful to update that field and add other major contributors for the Sopel 7 release series

dgw:
"Overhaul", to me, implies "for next major release".

Exirel:
We should also have a look at a possible `setup.cfg` file, while we are at it. I leave a note here as a reminder that I should suggest something - if not a PR!

Exirel:
I know right... :(

Exirel:
I added `python_requires` to #1522 and the trove classifiers were added by #1458

Exirel:
So what's left? Authors?

dgw:
Yeah, the authors/maintainers/contributors should be updated at the end of 7.0 dev cycle, right before release. I'll take care of it, don't worry. :)

dgw:
I'm going to call this done. Nitpicks of the final steps will be handled in #1787.

Too bad Python packaging metadata doesn't seem to support the concept of multiple project authors… but that's why we have a CREDITS file.

PR #1522, opened 2019-03-26
install: use setup.cfg to configure Sopel's package metadata

Exirel:
Did you know it's not safe to `import sopel` in `setup.py`? It's because when you do  `pip install sopel`, what it does is:

* fetch `sopel-{version}.tgz` from PyPI
* un-tar and uncompress it
* run `python setup.py install` (more or less)

It means that, anything that is imported in `setup.py` must be available *before* the project is installed, meaning none of Sopel's requirements are available at this very moment. Which is not ideal!

On the other side, how do you get Sopel's version once it's installed? Or even before? This is what we want (Python 3.5.2):

```console
$ python setup.py --version
6.6.5
$ python sopel.py --version
Sopel 6.6.5 (running on Python 3.5.2)
https://sopel.chat/
```

and with Python 2.7:

```console
$ python setup.py --version
Warning: Python 2.x is near end of life. Sopel support at that point is TBD.
6.6.5
$ python sopel.py --version
Warning: Python 2.x is near end of life. Sopel support at that point is TBD.
Sopel 6.6.5 (running on Python 2.7.12)
https://sopel.chat/
```

For that to work, we need:

* setuptools 30.3 or above
* an updated `setup.cfg` file

The trick is here, in `sopel/__init__.py`:

```
import pkg_resources

__version__ = pkg_resources.get_distribution('sopel').version
```

```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -65,4 +65,5 @@ setup(
             'sopel-config = sopel.cli.config:main',
         ],
     },
+    python_requires='>=2.7, !=3.0.*, !=3.1.*, !=3.2.*, <4',
 )
--- a/setup.py
+++ b/setup.py
@@ -2,21 +2,35 @@
 # coding=utf-8
 from __future__ import unicode_literals, absolute_import, print_function, division
 
-from sopel import __version__
 import sys
 
 try:
-    from setuptools import setup
+    from setuptools import setup, __version__ as setuptools_version
 except ImportError:
     print(
         'You do not have setuptools, and can not install Sopel. The easiest '
         'way to fix this is to install pip by following the instructions at '
-        'http://pip.readthedocs.org/en/latest/installing.html\n'
-        'Alternately, you can run sopel without installing it by running '
+        'https://pip.readthedocs.io/en/latest/installing/\n'
+        'Alternately, you can run Sopel without installing it by running '
         '"python sopel.py"',
         file=sys.stderr,
     )
     sys.exit(1)
+else:
+    version_info = setuptools_version.split('.')
+    major = int(version_info[0])
+    minor = int(version_info[1])
+
+    if major < 30 or (major == 30 and minor < 3):
+        print(
+            'Your version of setuptools is outdated: version 30.3 or above '
+            'is required to install Sopel. You can do that with '
+            '"pip install -U setuptools"\n'
+            'Alternately, you can run Sopel without installing it by running '
+            '"python sopel.py"',
+            file=sys.stderr,
+        )
+        sys.exit(1)
 
 if sys.version_info < (2, 7) or (
         sys.version_info[0] > 3 and sys.version_info < (3, 3)):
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Comment by dgw** [comment]
I'm not going to review yet, but uh… Something is broken here. I don't think it's ready?

**Comment by Exirel** [comment]
> I'm not going to review yet, but uh… Something is broken here. I don't think it's ready?

Looks good to me now!

It was just Travis, setuptools, and Python 3.3 :rofl: at me, making fun of me thinking that py.test would work, that setuptools would be up to date, and stuffs.

**Comment by dgw** [comment]
I'm generally in favor of this, but I might leave it sit for a while. Merging this would make pulling changes in from the 6.6.x branch much more tedious because of moving the version number. :)

**Comment by Exirel** [comment]
Fair enough.

File: setup.py [hunk 1]
```diff
@@ -2,21 +2,35 @@
...
+    version_info = setuptools_version.split('.')
+    major = int(version_info[0])
+    minor = int(version_info[1])
+
+    if major < 30 or (major == 30 and minor < 3):
+        print(
+            'Your version of setuptools is outdated: version 30.3 or above '
+            'is required to install Sopel. You can do that with '
+            '"pip install -U setuptools"\n'
+            'Alternately, you can run sopel without installing it by running '
```
dgw: Always capitalization that bites you, isn't it? 😉

```suggestion
            'Alternately, you can run Sopel without installing it by running '
```

Exirel:
Okay, just one last problem with pkg_resource and Python 3.3. Because obviously.

Exirel:
@dgw rebase and fix commit got squashed. Ready for the last review.

Sorry for the `pip install -U "setuptools<=39.2.0"` that is rather ugly, but setuptools supports Python version that are not past their EOL only-or at least there is no effort of any kind to support obsolete Python version.

dgw:
Somehow, looking at the commit messages, I feel like `pip` is the wrong "module name" to use… Shouldn't that be `setup` or `install`?

Exirel:
@dgw `install` then!

Exirel:
Rebased after version bumped to 6.6.7.

Exirel:
This should fix part of #1311.

Exirel:
Merge conflict fixed!

Exirel:
> why `MANIFEST.in` is including itself

Because by default, distutils/setuptools don't include it in the package, as it's a non-python file. You have to include it yourself, so that the package is complete.

dgw:
You don't need to include it:

```
$ tar --wildcards -ztvf ~/Desktop/sopel-6.6.8.tar.gz '*/MANIFEST*'
-rw-rw-r-- travis/travis    26 2019-05-09 00:11 sopel-6.6.8/MANIFEST.in
```

I downloaded the latest package (v6.6.8) from PyPI and despite [`MANIFEST.in`](https://github.com/sopel-irc/sopel/blob/v6.6.8/MANIFEST.in) not including itself, it's still included in the final package. That was a practical check to make after I looked at several different pieces of Python packaging documentation and found no mention of manually including the manifest anywhere.

dgw:
Indeed, I removed both `MANIFEST.in` and `setup.cfg` from the manifest and ran a test `python3 ./setup.py sdist`. Both files were still included in the finished archive.

Listing them in the manifest is redundant, IMO.

Exirel:
> Listing them in the manifest is redundant, IMO.

Well, it should be redundant as of today, but it wasn't with older version of `setuptools`. To be honest, they are so many bugs with `setuptools` that I keep a set of habits that are no-op at worst, and fix the problem at best.

So yeah, probably redundant today.

dgw:
> but it wasn't with older version of `setuptools`

How _much_ older? We're requiring 30.3+ to install, and Travis appears (as of last build from v6.6.8 tag) to come with 35.x and upgrade to 41.x before running deploys.

> So yeah, probably redundant today.

"Probably" enough to leave them out? :smile_cat: I'm only worried about Travis deploys at the moment (which should be fine, due to aforementioned upgrade behavior), not people building packages locally using outdated `setuptools`—and we're dropping support for older Pythons soon enough.

Can you tell I really don't want to add pointless stuff in our manifest? :sweat_smile:

Exirel:
> Can you tell I really don't want to add pointless stuff in our manifest? sweat_smile

I can tell! I think we can remove it. :+1:

dgw:
Welp, this isn't actually ready. It's merge-conflicted, and it still needs the extraneous manifest entries removed as talked about above.

Exirel:
@dgw merge conflict fixed, and `MANIFEST.in` updated.

dgw:
👏 Let's 👏 ship 👏 it! 👏

PR #1523, opened 2019-03-27
ip: fix tests (again)

dgw:
`dns.google` is quite a catchy name, but our tests were failing because of this sudden (not that I know where they would announce it) change.

On the plus side, we now have robust tests that should correctly handle both the old and new hostnames, and probably most potential future changes as well. And they're *actually* using regex syntax, unlike before with the unescaped brackets and pipes.

### DIFF
--- a/sopel/modules/ip.py
+++ b/sopel/modules/ip.py
@@ -99,7 +99,7 @@ def _find_geoip_db(bot):
 
 @commands('iplookup', 'ip')
 @example('.ip 8.8.8.8',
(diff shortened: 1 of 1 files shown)


dgw:
It could be a while. Travis is super slow today due to some issues they're having ([incident page](https://www.traviscistatus.com/incidents/my5wm56npf7q)), but we'll get this in when we can. Then I'll do a mid-stream merge to `master` (without waiting for 6.6.6 to be finished) so PRs' tests against that stop failing too.

PR #1524, opened 2019-03-28
Allow bot to utilize multiple destinations at one time

SpiceBot:
This adds the ability to send lists of recipients to bot. say/msg/reply/notice to send the same message to multiple destinations/recipients simultaneously.


dgw:
See #1525

Issue #1082, opened 2016-05-24
coretasks: parse RPL_ISUPPORT


maxpowa:
We should parse `RPL_ISUPPORT` to ensure we're doing the right thing with regards to Identifier parsing (`CHANTYPES`), account tracking (`WHOX`/`CASEMAPPING`), channel permission levels (`PREFIX`) and so on.

Documentation on individual params http://www.irc.org/tech_docs/005.html
Rough usage statistics http://stats.ircdocs.horse/isupport/
- [x] network name (`NETWORK`)
- [x] privilege prefixes (`PREFIX`)
- [x] channel prefixes (`CHANTYPES`)
- [x] account tracking (`WHOX`/`CASEMAPPING`)
- [x] multiple NAMES prefixes (`NAMESX`)
- [x] maximum message destinations (`MAXTARGETS`)
  - [x] also target maximums in general (`TARGMAX`)

----

All of this and more is parsed and exposed through #1758 and #1769. – @dgw edit

dgw:
Hijacked the OP with an edit, sorry @maxpowa. :)

dgw:
Done by @Exirel in #1758 (with a little bonus MYINFO stuff in #1769).

PR #1525, opened 2019-03-28
Allow bot to utilize multiple destinations at one time

deathbybandaid:
This adds the ability to send lists of recipients to bot. say/msg/reply/notice to send the same message to multiple destinations/recipients simultaneously.

### DIFF
--- a/sopel/bot.py
+++ b/sopel/bot.py
@@ -267,14 +267,14 @@ class Sopel(irc.Bot):
         else:
             self.write(['JOIN', channel, password])
 
-    def msg(self, recipient, text, max_messages=1):
+    def msg(self, recipients, text, max_messages=1):
         # Deprecated, but way too much of a pain to remove.
-        self.say(text, recipient, max_messages)
+        self.say(text, recipients, max_messages)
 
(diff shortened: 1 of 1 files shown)


dgw:
Leaving this be for now, as @deathbybandaid recently learned about the fact that IRC already supports [sending to multiple destinations at once](https://tools.ietf.org/html/rfc1459#section-4.4.1).

deathbybandaid:
I think this is a pretty good step forward.
#1322 should probably be handled for this to be truly effective.

deathbybandaid:
I have a good idea of what I want to do to tackle issue #1322 . I think it might be better to build that off of this, and not in this commit.

deathbybandaid:
as per

`<+dgw> you know it's probably safest to assume a default target limit of maybe 4, or even 2, if MAXTARGETS (or TARGMAX) isn't defined`

I'm now going to build in a maxtargets value, and maybe by time 
issue #1082 is handled, we can try to draw that value from the server capabilities

deathbybandaid:
I will fix the issues shortly

PR #1526, opened 2019-03-28
db: add methods to delete a channel or nick's key from database

deathbybandaid:
These are functions I've had for my bot for a long time,

* reset - sets a value to None
* adjust - allows math adjustments to existing values

Reset can likely use a better method like `DELETE FROM` instead of replacing values with None.

Adjust is something I've been using for a long time now to retrieve a current value and add a positive or negative number to it. Handy for modules that need to update stat counters.

### DIFF
--- a/sopel/db.py
+++ b/sopel/db.py
@@ -247,6 +247,26 @@ class SopelDB(object):
         finally:
             session.close()
 
+    def delete_nick_value(self, nick, key):
+        """Deletes the value for a given key associated with a nick."""
+        nick = Identifier(nick)
+        nick_id = self.get_nick_id(nick)
+        session = self.ssession()
+        try:
+            result = session.query(NickValues) \
+                .filter(NickValues.nick_id == nick_id) \
+                .filter(NickValues.key == key) \
+                .one_or_none()
+            # NickValue exists, delete
+            if result:
+                session.delete(result)
+                session.commit()
+        except SQLAlchemyError:
+            session.rollback()
+            raise
+        finally:
+            session.close()
+
     def get_nick_value(self, nick, key):
         """Retrieves the value for a given key associated with a nick."""
         nick = Identifier(nick)
@@ -360,6 +380,25 @@ class SopelDB(object):
             else:
                 new_channelvalue = ChannelValues(channel=channel, key=key, value=value)
                 session.add(new_channelvalue)
+                session.commit()
+        except SQLAlchemyError:
+            session.rollback()
+            raise
+        finally:
+            session.close()
+
+    def delete_channel_value(self, channel, key):
+        """Deletes the value for a given key associated with a channel."""
+        channel = Identifier(channel).lower()
+        session = self.ssession()
+        try:
+            result = session.query(ChannelValues) \
+                .filter(ChannelValues.channel == channel)\
+                .filter(ChannelValues.key == key) \
+                .one_or_none()
+            # ChannelValue exists, delete
+            if result:
+                session.delete(result)
                 session.commit()
         except SQLAlchemyError:
             session.rollback()
--- a/test/test_db.py
+++ b/test/test_db.py
@@ -151,6 +151,14 @@ def test_get_nick_value(db):
         assert found_value == value
 
 
+def test_delete_nick_value(db):
+    nick = 'Embolalia'
+    db.set_nick_value(nick, 'wasd', 'uldr')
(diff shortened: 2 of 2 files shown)


dgw:
It's easy to 👍 a function to delete values from the database, though I would definitely name the functions `delete_{nick|channel}_value()` rather than `reset` (and use a `DELETE` statement to match the name instead of setting the value to `None`).

Getting on board with an adjustment function is much harder. Sopel's database doesn't care what you put in it—everything becomes strings under the hood. Because of that, I don't think the `db` layer is a good place to put manipulator functions. Changing a value in the DB is more of a "program logic" operation, and I say the code to do it should live in the program (read: plugin) that wants to do it.

Exirel:
I agree with @dgw.

I would like to point out that Sopel's DB interface will be modified for 7.x, by using SQLAlchemy. So if you want to add or change DB-related features, I invite you to either wait for #1446 to be merged, or to base your work on this branch (but that's risky, considering it's a very living branch with rebase every now and then).

dgw:
Well it's not changing much at this point, but I agree that rebasing onto a branch that has been rebased is annoying.

Shouldn't be long, though!

deathbybandaid:
I removed the `adjust_` and figured out how to properly `DELETE` via sqlite

sopel/db.py
```diff
@@ -131,6 +131,12 @@ def alias_nick(self, nick, alias):
         except sqlite3.IntegrityError:
             raise ValueError('Alias already exists.')
 
+    def reset_nick_value(self, nick, key):
```
Exirel: As pointed out by @dgw the `delete` keyword is more appropriate when it comes to database operation.

And it's not the value that is reset, it's the key that is deleted.

```suggestion
    def delete_nick_key(self, nick, key):
```

sopel/db.py
```diff
@@ -195,6 +201,11 @@ def merge_nick_groups(self, first_nick, second_nick):
 
     # CHANNEL FUNCTIONS
 
+    def reset_channel_value(self, channel, key):
```
Exirel: ditto

```suggestion
    def delete_channel_key(self, channel, key):
```

dgw:
Reminder for myself (and/or the ambitious PR owner) to give the function names and docstrings a good once-over after this is reworked on top of SQLAlchemy/#1446. (I still prefer `_value` to `_key` in the names, and "delete" instead of "reset" in the docs.)

dgw:
Ready to rework, @deathbybandaid. #1446 just merged in. ☺

deathbybandaid:
Assuming #1603 gets fixed, my research shows that this is how to delete a value via sqlalchemy. This will need some testing pending that fix.

sopel/db.py
```diff
@@ -367,6 +387,25 @@ def set_channel_value(self, channel, key, value):
         finally:
             session.close()
 
+    def delete_channel_value(self, channel, key):
+        """Sets the value for a given key to be associated with the channel."""
```
dgw: `delete_channel_value` "Sets" something?

deathbybandaid: I will fix that later tonight, it wouldnt be a dbb PR without a docstring issue

RustyBower:
Yeah it's definitely `session.delete(result)`, otherwise you're gonna run into errors.

deathbybandaid:
Applied suggestions

dgw:
Closed/reopened to kick Scrutinizer in the pants (it errored during the last inspection and I didn't notice until this weekend). I guess we'll also get a new Travis build out of it, so yay?

deathbybandaid:
updated, and squashed

deathbybandaid:
I went an added step and added tests

sopel/db.py
```diff
@@ -247,6 +247,26 @@ def set_nick_value(self, nick, key, value):
         finally:
             session.close()
 
+    def delete_nick_value(self, nick, key):
+        """Deletes the value for a given key to be associated with the nick."""
```
dgw:
```suggestion
        """Deletes the value for a given key associated with a nick."""
```

sopel/db.py
```diff
@@ -367,6 +387,25 @@ def set_channel_value(self, channel, key, value):
         finally:
             session.close()
 
+    def delete_channel_value(self, channel, key):
+        """Deletes the value for a given key to be associated with the channel."""
```
dgw:
```suggestion
        """Deletes the value for a given key associated with a channel."""
```

Issue #1478, opened 2019-02-12
Broken interrupt handling when connection is inactive


dgw:
For the record, I'm testing on macOS today because that's what I have available. However, the same behavior occurs with identical or similar tracebacks on my Ubuntu machine at home, so I don't think it's related to the fact that I'm using Homebrew Python or anything. It's probably related to Sopel's sometimes-wonky exception handling.

Here's what I did:

1. Run `sopel`
2. Press Ctrl-C after "Connecting to &lt;server&gt;..."
3. Run `sopel` again
4. Press Ctrl-C while Sopel is waiting to reconnect after an expected SSL failure

Interrupting the connection phase resulted in an `AttributeError`, probably as expected (since quitting tries to send something to the socket, and the socket object doesn't exist before connecting finishes):

```
Connecting to irc.network.net:6667...
^CGot quit signal, shutting down.
Traceback (most recent call last):
  File "/Users/dgw/github/sopel/sopel/__init__.py", line 91, in run
    p.run(config.core.host, int(config.core.port))
  File "/Users/dgw/github/sopel/sopel/irc.py", line 167, in run
    self.initiate_connect(host, port)
  File "/Users/dgw/github/sopel/sopel/irc.py", line 177, in initiate_connect
    source_address=source_address))
  File "/usr/local/Cellar/python/3.7.2/Frameworks/Python.framework/Versions/3.7/lib/python3.7/socket.py", line 707, in create_connection
    for res in getaddrinfo(host, port, 0, SOCK_STREAM):
  File "/usr/local/Cellar/python/3.7.2/Frameworks/Python.framework/Versions/3.7/lib/python3.7/socket.py", line 748, in getaddrinfo
    for res in _socket.getaddrinfo(host, port, family, type, proto, flags):
  File "/Users/dgw/github/sopel/sopel/__init__.py", line 73, in signal_handler
    p.quit('Closing')
  File "/Users/dgw/github/sopel/sopel/irc.py", line 199, in quit
    self.write(['QUIT'], message)
  File "/Users/dgw/github/sopel/sopel/bot.py", line 166, in write
    irc.Bot.write(self, args, text=text)
  File "/Users/dgw/github/sopel/sopel/irc.py", line 161, in write
    self.send(temp.encode('utf-8'))
  File "/usr/local/Cellar/python/3.7.2/Frameworks/Python.framework/Versions/3.7/lib/python3.7/asyncore.py", line 360, in send
    result = self.socket.send(data)
[...]

PR #1527, opened 2019-03-30
core: fix broken interrupt handling

HumorBaby:
@dgw did a good job describing the issue in #1478, so I refer the reviewer there for the details.

The fixes provided here use simple checks to ensure 1) that a socket is present before attempting to send a `QUIT` message, and 2) a `hasquit` flag was not set on a bot instance during a disconnected phase.

Generally, in both cases, the issue was the signal setting the `hasquit` flag too late, and allowing a new `while` loop to begin. The bot would still have the `hasquit` flag set, but this would not be caught until a new `Exception`/`signal`.

### DIFF
--- a/sopel/cli/run.py
+++ b/sopel/cli/run.py
@@ -67,7 +67,11 @@ def run(config, pid_file, daemon=False):
             tools.stderr('Got restart signal.')
             p.restart('Restarting')
 
+    # Define empty variable `p` for bot
+    p = None
     while True:
+        if p and p.hasquit:  # Check if `hasquit` was set for bot during disconnected phase
+            break
         try:
             p = bot.Sopel(config, daemon=daemon)
             if hasattr(signal, 'SIGUSR1'):


sopel/irc.py
```diff
@@ -196,7 +196,8 @@ def restart(self, message):
 
     def quit(self, message):
         """Disconnect from IRC and close the bot."""
-        self.write(['QUIT'], message)
+        if self.connected:  # Only send QUIT message if socket is open
```
HumorBaby: `connected` is only set if the socket has been created for the `async_chat` object. This does not require a successful IRC connection, just an open connection to the server.

HumorBaby:
grammar peeve :cry:... will change _connection_ --> _connecting_ `fix: interrupt during connection phase` to `fix: interrupt during connecting phase` in subsequent pushes

HumorBaby:
woohoo! Seriously though, I only say this b/c of https://github.com/sopel-irc/sopel/issues/1478#issuecomment-478268578

dgw:
Time to make a new issue label here called "Affects: Docker" /s

Exirel:
I'm honestly quite impressed by how small the changes are. And it looks good so far!

HumorBaby:
Sad news :cry: I found another case where exit can be messy during testing for #1543: during the connection phase before the connection is accepted.

Test this by trying to connect to a server on a wrong port (where it is most likely not to be accepted, but not explicitly rejected).

dgw:
Worth making a quick PR for 6.6.x so this can be fixed in the 6.6.6 release on Monday?

HumorBaby:
Yep.

dgw:
Looks like only 1171e4e (required because of the `sopel/run_script.py` -&gt; `sopel/cli/run.py` refactor from #1493) will be needed after `6.6.x` is merged back to `master`. The other commits in this PR only touch `sopel/irc.py`, which isn't much changed, and will make it into `master` by way of #1534 and #1558.

HumorBaby:
>Looks like only 1171e4e (required because of the sopel/run_script.py -> sopel/cli/run.py refactor from #1493) will be needed after 6.6.x is merged back to master

That’s a Texas-sized 10-4, good buddy. 

Confirmed with local test merges. Preemptive squash coming right up...

dgw:
@HumorBaby Would you be put out if I asked to reference 9dd772d in this commit message? Like:

```
reapply fix: interrupt during disconnected phase (9dd772d)
```

Also, in "Because the `time.sleep`", there's a missing "of". :)

HumorBaby:
> Would you be put out if I asked to reference 9dd772d in this commit message?

Too much. :upside_down_face:

dgw:
Perfect!

PR #1528, opened 2019-03-30
core: perform commands on connect

HumorBaby:
This PR is meant to address #1455. It adds a `core` setting that is a `ListAttribute` of commands to perform upon successful connection with the IRC server. These commands are executed at the end of the `startup` sequence (set `MODE`s, `JOIN` channels, and now, _perform_ commands).

It depends on #1460.

**Example setting**
```
perform_commands = NICKSERV identify testbot ThisIsMy\,complexP@$sW0RD!!,JOIN #testing,PRIVMSG #testing :I am here!!
```
`raw.log`, as expected:
```
...
>>1553928396.52 NICKSERV identify testbot ThisIsMy,complexP@$sW0RD!!
>>1553928396.53 JOIN #testing
>>1553928396.53 PRIVMSG #testing :I am here!!
<<1553928396.57 :testbot!sopel@pumba.local JOIN #testing * :Sopel: https://sopel.chat
>>1553928396.57 TOPIC #testing
>>1553928396.57 WHO #testing a%nuachtf,387
<<1553928396.57 :e9b9f479e8eb.example.net 353 testbot = #testing :@testbot
<<1553928396.57 :e9b9f479e8eb.example.net 366 testbot #testing :End of /NAMES list.
<<1553928396.57 :e9b9f479e8eb.example.net 331 testbot #testing :No topic is set.
<<1553928396.57 :e9b9f479e8eb.example.net 352 testbot #testing sopel pumba.local e9b9f479e8eb.example.net testbot H@ :0 Sopel:
https://sopel.chat
<<1553928396.57 :e9b9f479e8eb.example.org 315 testbot #testing :End of /WHO list.
<<1553928396.61 :NickServ!NickServ@services.int NOTICE testbot :You are now identified for testbot.
...
```
_Note_: 
- There is no space around commas delimiting commands 
- Only the comma of the password is escaped; all other characters do not need to be escaped.

**Why `[core]`?**
I had considered two alternatives to adding `perform_commands` directly into core:
1) Make `perform` its own module
  - _Pros_
    - Get its own config section
    - Easily add the `list`, `swap`, `add`, `del` commands from ZNC's perform as commands that the bot's admin can use
    - No need to modify the core bot
  - _Cons_
    - Would have to do something like: set `@interval` on a function that checks if `bot.connection_registered`; once connected, execute the commands; remove the `@interval` function from the job scheduler (from within the same function itself)
    - The above approach would go crazy on `.reload` without additional measures to track the execution status of commands
2) Introduce a signal/slot system to connect the _successful startup_ event (basically represented by the `startup` function to perform actions)
  - _Pros_
    - Same as (1), except modifying the core bot part
 - _Cons_
    - Need to modify the core bot to add signals
    - Don't even get me started on `.reload` :boom:... it's bad enough register/unregistering commands
    - I imagine something like this will be in the works soon anyway :smile: :crossed_fingers:, so have a dedicated PR for that, instead of this being introduced as a by-product of this PR.

**Plans**
I hope a solution like (2) can be achieved at some point. With all the changes to the module loading interface coming along, I think it would be reasonable (and relatively simple) to migrate this to its own module.

### DIFF
--- a/docs/source/configuration.rst
+++ b/docs/source/configuration.rst
@@ -223,7 +223,47 @@ policy.
    options have been added: ``flood_burst_lines``, ``flood_empty_wait``, and
    ``flood_refill_rate``.
 
+Perform commands on connect
+---------------------------
 
+The bot can be configured to send custom commands upon successful connection to
+the IRC server. This can be used in situations where the bot's built-in
+capabilities are not sufficient, or further automation is desired. The list of
+commands to send is set with :attr:`~CoreSection.commands_on_connect`.
+
+For example, the following configuration:
+
+.. code-block:: ini
+
+    [core]
+    commands_on_connect = PRIVMSG X@Channels.undernet.org :LOGIN MyUserName A$_Strong\,*pasSWord,PRIVMSG IDLEBOT :login IdleUsername idLEPasswoRD
+
+
+will, upon connection:
+    
+    1) identify to Undernet services
+    2) login with ``IDLEBOT``
+
+.. important::
+
+    Commas are used to delimit separate commands, so any comma found within a
+    command must be escaped with ``\``. In the example above, the password
+    ``A$_Strong,*pasSWord`` is escaped as ``A$_Strong\,pasSWord`` (note the
+    escaped comma in the middle of the password, but not immediately following,
+    which is delimiting the next command).
+
+    No other text needs to be escaped.
+
+.. 
+    TODO: update this note (and the example config) once #1628 is merged in,
+    changing the delimiter to newlines (from commas).
+
+.. seealso::
+    
+    This functionality is analogous to ZNC's ``perform`` module:
+    https://wiki.znc.in/Perform
+    
+    
 Authentication
 ==============
 
--- a/docs/source/configuration.rst
+++ b/docs/source/configuration.rst
@@ -228,21 +228,25 @@ Perform commands on connect
 
 The bot can be configured to send custom commands upon successful connection to
 the IRC server. This can be used in situations where the bot's built-in
-capabilities are not sufficient, or further automation is desired. The list of
-commands to send is set with :attr:`~CoreSection.commands_on_connect`.
+capabilities are not sufficient, or further automation is desired.
+``$nickname`` can be used in a command as a placeholder, and it will be
+replaced with the bot's nickname, as specified in the configuration
+(:attr:`~CoreSection.nick`). 
 
-For example, the following configuration:
+The list of commands to send is set with
+:attr:`~CoreSection.commands_on_connect`. For example, the following
+configuration:
 
 .. code-block:: ini
 
     [core]
-    commands_on_connect = PRIVMSG X@Channels.undernet.org :LOGIN MyUserName A$_Strong\,*pasSWord,PRIVMSG IDLEBOT :login IdleUsername idLEPasswoRD
+    commands_on_connect = PRIVMSG X@Channels.undernet.org :LOGIN MyUserName A$_Strong\,*pasSWord,PRIVMSG IDLEBOT :login $nickname idLEPasswoRD
 
 
 will, upon connection:
     
-    1) identify to Undernet services
-    2) login with ``IDLEBOT``
+    1) identify to Undernet services (``PRIVMSG X@Channels...``)
+    2) login with ``IDLEBOT`` using the bot's nickname (``PRIVMSG IDLEBOT ...``)
 
 .. important::
 
--- a/docs/source/configuration.rst
+++ b/docs/source/configuration.rst
@@ -231,43 +231,28 @@ the IRC server. This can be used in situations where the bot's built-in
 capabilities are not sufficient, or further automation is desired.
 ``$nickname`` can be used in a command as a placeholder, and it will be
 replaced with the bot's nickname, as specified in the configuration
-(:attr:`~CoreSection.nick`). 
+(:attr:`~CoreSection.nick`).
 
 The list of commands to send is set with
 :attr:`~CoreSection.commands_on_connect`. For example, the following
-configuration:
-
-.. code-block:: ini
+configuration::
 
     [core]
-    commands_on_connect = PRIVMSG X@Channels.undernet.org :LOGIN MyUserName A$_Strong\,*pasSWord,PRIVMSG IDLEBOT :login $nickname idLEPasswoRD
-
+    commands_on_connect =
+        PRIVMSG X@Channels.undernet.org :LOGIN MyUserName A$_Strong,*pasSWord
+        PRIVMSG IDLEBOT :login $nickname idLEPasswoRD
 
 will, upon connection:
-    
-    1) identify to Undernet services (``PRIVMSG X@Channels...``)
-    2) login with ``IDLEBOT`` using the bot's nickname (``PRIVMSG IDLEBOT ...``)
 
-.. important::
-
-    Commas are used to delimit separate commands, so any comma found within a
-    command must be escaped with ``\``. In the example above, the password
-    ``A$_Strong,*pasSWord`` is escaped as ``A$_Strong\,pasSWord`` (note the
-    escaped comma in the middle of the password, but not immediately following,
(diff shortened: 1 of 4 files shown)


sopel/config/core_section.py
```diff
@@ -181,6 +185,15 @@ def homedir(self):
...
     """
 
+    perform_commands = ListAttribute('perform_commands')
+    """A list of commands to perform upon successful connection to IRC server.
+
+    When entered using the config wizard, commas will be escaped automatically.
+    Otherwise, commas must be escaped, e.g.: `PRIVMSG Q@CServe.quakenet.org
+    :AUTH my_username MyPassword\,HasAComma@#$%!` Nothing else needs to be
+    escaped.
+    """  # noqa: W605; ignore escaped comma error
```
dgw: This [isn't very future-proof](https://docs.python.org/3.6/reference/lexical_analysis.html#index-23).

> *Changed in version 3.6:* Unrecognized escape sequences produce a DeprecationWarning. In some future version of Python they will be a SyntaxError.

Instead of ignoring the warning, please fix it.

sopel/config/core_section.py
```diff
@@ -181,6 +185,15 @@ def homedir(self):
     capabilities.
     """
 
+    perform_commands = ListAttribute('perform_commands')
```
dgw: While I'm not crazy about this attribute name, my own ideas aren't much better:

  - `startup_commands`
  - `connect_commands`
  - `do_on_connect`
  - … more lame ideas

But I definitely think the choice to use "perform" came from my ZNC analogy, which it probably shouldn't. The name of this config setting should make sense to someone who's ever even _heard_ of a ZNC.

sopel/config/core_section.py
HumorBaby: Done.

Of the two options to resolve this:
1) convert to "raw" string literal
2) escape the `\`

I went with option (1), to avoid confusion. Assuming someone is reading the source to determine how to correctly escape commands, I imagine it would be easier to realize that a comma is escaped with a single `\`, instead of having to brain-parse the docstring to realize that `\\` actually refers to `\`, but is double-`\` only because the `\` needs to be escaped... 🤯

sopel/config/core_section.py
HumorBaby: I went with `commands_on_connect`...
- `startup_commands` and `connect_commands` (despite being documented) connote that these commands will be executed _during_ the startup/connect, and not _after_
- `do_on_connect`... well... not bad, but I hope that `commands_on_connect` is slightly clearer.

HumorBaby:
>(And you'll probably rebase this anyway after #1460 goes in—up to you if you address feedback now or wait until I merge that… Soon™.)

@Exirel pointed out that this would be best re-implemented after the new plugin interface goes in anyway :smile: Addressed the requested changes anyway, because, why not?

Exirel:
@HumorBaby this need to be rebased since the PR it requires has been merged.

About the feature itself... I'm not super happy about it. I feel like we are trying to provide a half-baked hook into Sopel to help admin users implement a feature Sopel does not handle yet.

So... meh? Maybe this comment should be in #1455 though...

dgw:
> I feel like we are trying to provide a half-baked hook into Sopel to help admin users implement a feature Sopel does not handle yet.

My perspective on this is, we aren't going to add support for _every_ little feature on _every_ obscure IRCd or network. Sending arbitrary commands at connect is frequently requested, and is (anecdotally) supported by most other popular IRC clients.

sopel/config/core_section.py
dgw: Didn't notice until today, but… The name of the setting should be changed, too.

Also, in other code, the old name is still used. This PR would fail tests if it had them. :laughing:

dgw:
This is one of only three "review required" PRs not authored by me. One is deferred until Sopel 8, one is just so big I haven't had time to look at it, and this one has a merge conflict now after waiting so long for #1628 to tweak how `ListAttribute`s get stored.

@HumorBaby do you need to make any updates to this, besides the merge conflict? Or is it drop-in compatible with the `ListAttribute` changes?

HumorBaby:
> @HumorBaby do you need to make any updates to this, besides the merge conflict? Or is it drop-in compatible with the `ListAttribute` changes?

Should be a drop in. Will confirm, of course.

dgw:
Docs too, I hope. Any examples should use the new style, for example.

docs/source/configuration.rst
```diff
@@ -215,7 +215,51 @@ policy.
...
+    command must be escaped with ``\``. In the example above, the password
+    ``A$_Strong,*pasSWord`` is escaped as ``A$_Strong\,pasSWord`` (note the
+    escaped comma in the middle of the password, but not immediately following,
+    which is delimiting the next command).
+
+    No other text needs to be escaped.
+
+.. 
+    TODO: update this note (and the example config) once #1628 is merged in,
+    changing the delimiter to newlines (from commas).
```
dgw: Yep, escaped commas aren't a thing any more. This all needs updating as per the TODO.

sopel/config/core_section.py
```diff
@@ -336,6 +340,22 @@ def homedir(self):
     capabilities.
     """
 
+    commands_on_connect = ListAttribute('commands_on_connect')
+    r"""A list of commands to perform upon successful connection to IRC server.
+
+    When entered using the config wizard, commas will be escaped automatically.
+    Otherwise, commas must be escaped, e.g.: ``PRIVMSG Q@CServe.quakenet.org
+    :AUTH my_username MyPassword\,HasAComma@#$%!`` Nothing else needs to be
```
dgw: This, too, needs updating now that escaped commas aren't a thing any more.

Eliminating that false escape sequence means the whole docstring can quit being raw (`r"""`) also.

sopel/coretasks.py
```diff
@@ -61,6 +61,27 @@ def auth_after_register(bot):
...
+    for i, command in enumerate(bot.config.core.commands_on_connect):
+        command = command.replace('$nickname', bot.config.core.nick)
+        LOGGER.debug(command)
+        bot.write((command,))
+
+
+@sopel.module.require_privmsg("This command only works as a private message.")
+@sopel.module.require_admin("This command requires admin privileges.")
+@sopel.module.commands('execute')
+def _execute_perform(bot, trigger):
```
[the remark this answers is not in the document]
dgw: Still of the opinion that the command callable should have an unprefixed name, and the helper version should get the `_` instead.

Exirel:
So, I've no idea how I can push my commits to @HumorBaby 's branch. So I've my own branch, available here: https://github.com/sopel-irc/sopel/compare/master...Exirel:feat-perform-commands-on-connect

It contains an up-to-date version of this branch, rebased upon the latest sopel/master branch, with the tests modified to fit with the new tests tools.

dgw:
@Exirel short answer: `git push -f HumorBaby feat-perform-commands-on-connect`, assuming you added HB's remote with that name.

Exirel:
> **EDIT:** So I don't forget later, this should be able to drop the `bot.write` patch commit, since the testing mocks have been redone.

And also fix the PR so it can be merged. I'll work on that later (probably tomorrow morning).

Exirel:
@dgw rebased upon master, docstrings fixed, and `_execute_perform`  is now properly private.

docs/source/configuration.rst
```diff
@@ -223,6 +223,35 @@ policy.
...
+will, upon connection:
+
+1) identify to Undernet services (``PRIVMSG X@Channels...``)
+2) login with ``IDLEBOT`` using the bot's nickname (``PRIVMSG IDLEBOT ...``)
+
+.. seealso::
+
+   This functionality is analogous to ZNC's ``perform`` module:
+   https://wiki.znc.in/Perform
+
```
dgw: At some point we'll standardize the indent for blocks like this (I tend to use 4 because I just press <kbd>Tab</kbd> in my editor), but I'm not going to quibble over it until we actually have a documentation style guide. Writing one is on my to-do list for 7.1-ish.

PR #1530, opened 2019-03-30
Running instance check

jbrunink:
Sopel keeps track of the current process ID in a file to be used to check if Sopel is already running an instance with the same configuration file. 

When running Sopel in Docker and when the container restarts it retains the filesystem and thus has still the file with the process ID. Due to Dockers nature it's likely that Sopel is going to be running with the same process ID as before and thus will fail the running instance check on itself. 

To mitigate this I've added a check to this system which checks first if the current process ID does not equal with the one in the file and then proceeds with the check. If it does equal it should just go ahead with running.

### DIFF
--- a/sopel/run_script.py
+++ b/sopel/run_script.py
@@ -168,21 +168,22 @@ def main(argv=None):
                 except ValueError:
                     old_pid = None
             if old_pid is not None and tools.check_pid(old_pid):
-                if not opts.quit and not opts.kill:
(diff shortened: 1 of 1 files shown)


dgw:
Issues specific to running inside Docker should be handled on the Docker side.

If there's a situation that would require this in normal environments, I'll consider it, but Sopel does not "officially" support running inside Docker and I don't want to put code in to handle Docker corner cases.

PR #1531, opened 2019-03-30
Running instance check @master

jbrunink:
Please see #1530 but for master branch.


dgw:
See close reason for #1530

PR #1535, opened 2019-04-01
fix _store() so it accepts any named arguments

jbrunink:
When running a testsuite where the bot.say, reply or action function has any other than the defined arguments set, for example max_arguments, it will fail. With this change it should accept any named arguments. 

Would be nice if this could also be done for the 6.6.x branch.

```
TypeError: _store() got an unexpected keyword argument 'max_messages'    
```

### DIFF
--- a/sopel/modules/ipython.py
+++ b/sopel/modules/ipython.py
@@ -9,6 +9,7 @@ Sopel: https://sopel.chat/
 from __future__ import unicode_literals, absolute_import, print_function, division
 
 import sopel
+import sopel.module
 import sys
 if sys.version_info.major >= 3:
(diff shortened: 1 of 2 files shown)


dgw:
> Would be nice if this could also be done for the 6.6.x branch.

You can just rebase this PR on the `6.6.x` branch and change the base on GitHub after force-pushing. I don't think there are any conflicting changes on `master` that would block merging it back after 6.6.6 drops.

jbrunink:
Meh, I'll leave that to folks that actually know how to do that then.

dgw:
Replaced by #1538

PR #1537, opened 2019-04-02
core: add `bot.hostmask` getter

### LINKED ISSUE #1533: bot.hostmask
While bot.users.get(bot.nick).hostmask is nice to use, I think it would be slightly better to have a more direct approach like bot.hostmask .

HumorBaby:
Adds shortcut for `bot.users.get(bot.nick).hostmask`.

Closes #1533.

### DIFF
--- a/sopel/bot.py
+++ b/sopel/bot.py
@@ -141,6 +141,17 @@ class Sopel(irc.Bot):
             self.config.core.host_blocks = []
         self.setup()
 
+    @property
+    def hostmask(self):
+        """str: the current hostmask for the bot :class:`sopel.tools.target.User`
+
+        Bot must be connected and in at least one channel.
+        """
+        if not self.users or not self.users.contains(self.nick):
+            raise KeyError("'hostmask' not available: bot must be connected and in at least one channel.")
+
+        return self.users.get(self.nick).hostmask
+
     # Backwards-compatibility aliases to attributes made private in 6.2. Remove
     # these in 7.0
     times = property(lambda self: getattr(self, '_times'))


dgw:
Since this adds to the API, I bumped it to `master` (for 7.x) instead of the next patch release for `6.6.x`.

Rebased for you, since it was a really easy one.

HumorBaby:
Sounds good.

HumorBaby:
`bot.hostmask` getter has been fortified according to discussions on IRC.

sopel/bot.py
```diff
@@ -141,6 +141,20 @@ def __init__(self, config, daemon=False):
...
         self.setup()
 
+    @property
+    def hostmask(self):
+        if self.users.contains(self.nick):
+            # If the user exists, the `User` object will provide a hostmask
+            return self.users.get(self.nick).hostmask
+
+        if not self.users:
+            raise KeyError("'hostmask' not available yet: bot must be connected and in at least one channel.")
```
Exirel: The `KeyError` makes no sense to me here. A `KeyError` means that a key is not in an index, a list, or a dict - any kind of data-structure that is act as a container. Here, `self.users` act as a container, and should be treated as such. Can you look at this so it's the other other way arround:

```python
if not self.users:
    raise RuntimeErrror('not available')

if not self.users.contains(self.nick):
    raise KeyError('Not in its user-list')

return self.users.get(self.nick).hostsmask
```

* No user at all: something is wrong (probably)
* Not the bot: KeyError
* Otherwise: the bot's hostsmask (if it's known)

I'd even dare to say that we could squash the first two cases in one: no user at all or no bot in it should throw the exact same `KeyError` - no `RuntimeError` needed, in my opinion.

HumorBaby: Sounds good. Message for the `KeyError` okay in the changes?

Exirel: Looking good!

sopel/bot.py
```diff
@@ -141,6 +141,13 @@ def __init__(self, config, daemon=False):
             self.config.core.host_blocks = []
         self.setup()
 
+    @property
+    def hostmask(self):
```
Exirel: Can we have a docstring now? 😜

HumorBaby: Done + squashed.

PR #1539, opened 2019-04-02
Provide a simpler way to kick users from channels.

deathbybandaid:
this adds a `bot.kick(text, channel, nick)` ability to the bot, instead of using `bot.write(['KICK', channel, nick], text)`.

If the bot isn't OP, nothing happens (expected).

### DIFF
--- a/sopel/bot.py
+++ b/sopel/bot.py
@@ -485,6 +485,16 @@ class Sopel(irc.Bot):
         else:
             self.say(text, dest)
 
+    def kick(self, nick, channel, text=None):
+        """Send an IRC KICK command.
+        Within the context of a triggered callable, ``channel`` will default to the
+        channel in which the call was triggered. If triggered from a private message,
+        ``channel`` is required (or the call to ``kick()`` will be ignored).
+        The bot must be a channel operator in specified channel for this to work.
+        .. versionadded:: 7.0
+        """
+        self.write(['KICK', channel, nick], text)
+
     def call(self, func, sopel, trigger):
         nick = trigger.nick
         current_time = time.time()
@@ -887,3 +897,13 @@ class SopelWrapper(object):
         if reply_to is None:
             reply_to = self._trigger.nick
         self._bot.reply(message, destination, reply_to, notice)
+
+    def kick(self, nick, channel=None, message=None):
+        if channel is None:
+            if self._trigger.is_privmsg:
+                raise RuntimeError('Error: KICK requires a channel.')
+            else:
+                channel = self._trigger.sender
+        if nick is None:
+            raise RuntimeError('Error: KICK requires a nick.')
+        self._bot.kick(nick, channel, message)
--- a/sopel/modules/reddit.py
+++ b/sopel/modules/reddit.py
@@ -74,8 +74,8 @@ def rpost_info(bot, trigger, match):
             sfw = bot.db.get_channel_value(trigger.sender, 'sfw')
             if sfw:
                 link = '(link hidden)'
-                bot.write(['KICK', trigger.sender, trigger.nick,
-                           'Linking to NSFW content in a SFW channel.'])
+                bot.kick(trigger.nick, trigger.sender,
+                     'Linking to NSFW content in a SFW channel.')
         else:
             nsfw = ''
 
--- a/sopel/modules/adminchannel.py
+++ b/sopel/modules/adminchannel.py
@@ -119,7 +119,7 @@ def kick(bot, trigger):
         reasonidx = 3
     reason = ' '.join(text[reasonidx:])
     if nick != bot.config.core.nick:
-        bot.write(['KICK', channel, nick], reason)
+        bot.kick(nick, channel, reason)
 
 
 def configureHostMask(mask):
@@ -290,7 +290,7 @@ def kickban(bot, trigger):
     if mask == '':
         return
     bot.write(['MODE', channel, '+b', mask])
-    bot.write(['KICK', channel, nick], reason)
+    bot.kick(nick, channel, reason)
 
 
 @require_chanmsg
--- a/sopel/modules/safety.py
+++ b/sopel/modules/safety.py
@@ -183,8 +183,7 @@ def url_handler(bot, trigger):
         msg += '(confidence %s - %s/%s)' % (confidence, positives, total)
         bot.say('[' + bold(color('WARNING', 'red')) + '] ' + msg)
         if strict:
-            bot.write(['KICK', trigger.sender, trigger.nick,
-                       'Posted a malicious link'])
+            bot.kick(trigger.nick, trigger.sender, 'Posted a malicious link')
 
 
 @sopel.module.commands('safety')


sopel/bot.py
```diff
@@ -376,6 +376,11 @@ def reply(self, text, dest, reply_to, notice=False):
         else:
             self.say(text, dest)
 
+    def kick(self, text, channel, nick):
+        """Send an IRC KICK to a user in a channel.
+        """
```
dgw: This closing `"""` should go at the end of the previous line (single-line docstrings don't need line-breaks).

sopel/bot.py
```diff
@@ -376,6 +376,11 @@ def reply(self, text, dest, reply_to, notice=False):
         else:
             self.say(text, dest)
 
+    def kick(self, text, channel, nick):
```
dgw: `text` should be the last argument, and default to `''` or `None`, so it can be omitted. The `KICK` IRC command itself only requires `channel` and `nick`.

sopel/bot.py
```diff
@@ -417,6 +422,13 @@ def reply(self, message, destination=None, reply_to=None, notice=False):
                 reply_to = self._trigger.nick
             self._bot.reply(message, destination, reply_to, notice)
 
+        def kick(self, message, channel, nick):
```
dgw: `message` should be the last argument, and default to `''` or `None`, so it can be omitted. The `KICK` IRC command itself only requires `channel` and `nick`.

sopel/bot.py
```diff
@@ -376,6 +376,10 @@ def reply(self, text, dest, reply_to, notice=False):
         else:
             self.say(text, dest)
 
+    def kick(self, nick, channel, text=''):
+        """Send an IRC KICK to a user in a channel."""
```
dgw: Going to propose a bit of a doozy here, along the lines of other docstrings in this class. Wrapped it roughly in GitHub's editor, but you can wrap to a proper length in your editor before committing. 😸

```suggestion
        """Send an IRC KICK command.

        Within the context of a triggered callable, ``channel`` will default to the
        channel in which the call was triggered. If triggered from a private message,
        ``channel`` is required (or the call to ``kick()`` will be ignored).
        """
```

sopel/bot.py
```diff
@@ -417,6 +421,13 @@ def reply(self, message, destination=None, reply_to=None, notice=False):
                 reply_to = self._trigger.nick
             self._bot.reply(message, destination, reply_to, notice)
 
+        def kick(self, nick, channel=None, message=''):
+            if channel is None:
+                channel = self._trigger.sender
```
dgw: As I alluded to in my suggested docstring above, if the trigger is in privmsg, not a channel, this should still fall through to a `return` (or throw a nice exception, if we follow @Exirel's way of thinking).

sopel/bot.py
```diff
@@ -417,6 +421,13 @@ def reply(self, message, destination=None, reply_to=None, notice=False):
...
+            if nick is None:
+                return
```
dgw: `nick` is a REQUIRED argument, and calling `kick()` without it should fail loudly.

sopel/bot.py
```diff
@@ -376,6 +376,16 @@ def reply(self, text, dest, reply_to, notice=False):
         else:
             self.say(text, dest)
 
+    def kick(self, nick, channel, text=''):
```
dgw: This (and below) should default the text to `None` instead of `''` to clean up the raw line generated by `irc.Bot.write()` later down the call stack.
```suggestion
    def kick(self, nick, channel, text=None):
```

sopel/bot.py
```diff
@@ -417,6 +427,13 @@ def reply(self, message, destination=None, reply_to=None, notice=False):
                 reply_to = self._trigger.nick
             self._bot.reply(message, destination, reply_to, notice)
 
+        def kick(self, nick, channel=None, message=''):
```
dgw: Same as above.
```suggestion
        def kick(self, nick, channel=None, message=None):
```

sopel/bot.py
```diff
@@ -376,6 +376,16 @@ def reply(self, text, dest, reply_to, notice=False):
         else:
             self.say(text, dest)
 
+    def kick(self, nick, channel, text=''):
+        """Send an IRC KICK command.
+
+        Within the context of a triggered callable, ``channel`` will default to the
+        channel in which the call was triggered. If triggered from a private message,
+        ``channel`` is required (or the call to ``kick()`` will be ignored).
+
```
dgw: Unnecessary blank line.
```suggestion
```

sopel/bot.py
```diff
@@ -376,6 +376,15 @@ def reply(self, text, dest, reply_to, notice=False):
         else:
             self.say(text, dest)
 
+    def kick(self, nick, channel, text=None):
+        """Send an IRC KICK command.
+
+        Within the context of a triggered callable, ``channel`` will default to the
+        channel in which the call was triggered. If triggered from a private message,
+        ``channel`` is required (or the call to ``kick()`` will be ignored).
```
dgw: Accepting this part of the suggested docstring means that if triggered in privmsg, outside a channel, `SopelWrapper.kick()` should return without doing anything if not given a channel name.

sopel/bot.py
```diff
@@ -417,6 +426,13 @@ def reply(self, message, destination=None, reply_to=None, notice=False):
                 reply_to = self._trigger.nick
             self._bot.reply(message, destination, reply_to, notice)
 
+        def kick(self, nick, channel=None, message=None):
+            if channel is None:
+                channel = self._trigger.sender
```
dgw: Easiest way to match behavior to the desired outcome described in the docstring:

```suggestion
                if self._trigger.is_privmsg:
                    return
                else:
                    channel = self._trigger.sender
```

sopel/bot.py
HumorBaby: I may have missed something elsewhere, but this doesn't appear resolved :confused: 

`return`ing seems more like quitely hiding the problem than "should fail loudly".

dgw: User would have to explicitly pass in `None`, since the default parameter value was removed. If passed no `nick` argument at all, this will fail quite loudly.

Probably should remove this `if` block entirely.

sopel/bot.py
```diff
@@ -376,6 +376,15 @@ def reply(self, text, dest, reply_to, notice=False):
         else:
             self.say(text, dest)
 
+    def kick(self, nick, channel, text=None):
+        """Send an IRC KICK command.
+
+        Within the context of a triggered callable, ``channel`` will default to the
+        channel in which the call was triggered. If triggered from a private message,
+        ``channel`` is required (or the call to ``kick()`` will be ignored).
+        """
```
dgw: In agreement with @half-duplex on IRC:

```suggestion

        .. versionadded:: 7.0
        """
```

sopel/bot.py
dgw: NAK on removing the `if`. Failing silently is fine if explicitly passed `None`, which is in line with how other `bot` API methods work if something required for the command to be valid is None-y.

dgw:
Now needs a squash.

deathbybandaid:
Some other thoughts for consideration in either this, or a seperate PR.

1) Some servers allow KICK to have multiple recipients based on the TARGMAX.KICK value

2) I believe that KICK has a message size limit of 200 bytes, we may want to truncate that just in case.

3) We may want to check that the bot has the channel permissions of the specified channel before `bot.write`

dgw:
@deathbybandaid You could rebase the current commit and then add (2) and (3) here. We should keep (1) on the radar, but not try to implement it until #1536 is finalized and merged.

deathbybandaid:
After some deeper reading, it looks like the message length of KICK is determined my the KICKLEN value in 005_ISUPPORT so that will also be after #1536 is done as well.

I rebased just now, and will look at adding (3)

deathbybandaid:
I went an extra step with this and updated `adminchannel.py` `reddit.py` and `safety.py` with `bot.kick`

sopel/bot.py
```diff
@@ -777,3 +788,13 @@ def reply(self, message, destination=None, reply_to=None, notice=False):
         if reply_to is None:
             reply_to = self._trigger.nick
         self._bot.reply(message, destination, reply_to, notice)
+
+    def kick(self, nick, channel=None, message=None):
+        if channel is None:
+            if self._trigger.is_privmsg:
+                return
```
Exirel: This should raise an error.

sopel/bot.py
```diff
@@ -777,3 +788,13 @@ def reply(self, message, destination=None, reply_to=None, notice=False):
...
+            else:
+                channel = self._trigger.sender
+        if nick is None:
+            return
```
Exirel: This should raise an error.

deathbybandaid: What error would you like it to raise?

sopel/bot.py
deathbybandaid: What error would you like it to raise?

deathbybandaid:
> Is that me or the `kick` method never checks that it has the right to do so?

https://github.com/sopel-irc/sopel/blob/e4b4bed27e7fddca07a45fcc024038917cbdcf17/sopel/bot.py#L479

dgw:
I don't think the `bot` object should do the permissions check. If told to send a command, it should send that command. Whether the IRC server will execute that command or return an error is inconsequential.

So, the check should be made in `adminchannel.py` ([it already is](https://github.com/sopel-irc/sopel/blob/e4b4bed27e7fddca07a45fcc024038917cbdcf17/sopel/modules/adminchannel.py#L104-L105)) and other plugins that want to kick, as appropriate. That's the only way appropriate error messages will get sent back to the user if the bot has insufficient permissions to kick.

_Some feature to listen for certain error numerics (such as 481 `ERR_NOPRIVILEGES`, 482 `ERR_CHANOPRIVSNEEDED`, and/or 499 `ERR_CHANOWNPRIVNEEDED`) and report them to the log file, log channel, and/or bot owner is out of scope for this PR, but could be valuable._

sopel/bot.py
Exirel: A `RuntimeError` is a good start, since this error happens when the developer mess up with the code.

sopel/bot.py
Exirel: Same as above, `RuntimeError`.

deathbybandaid:
Updated to Raise RuntimeErrors, and removes the check for bot privileges in a channel.

sopel/bot.py
```diff
@@ -468,6 +468,16 @@ def reply(self, text, dest, reply_to, notice=False):
         else:
             self.say(text, dest)
 
+    def kick(self, nick, channel, text=None):
+        """Send an IRC KICK command.
```
```suggestion
        """Send an IRC KICK command.

```

sopel/bot.py
```diff
@@ -468,6 +468,16 @@ def reply(self, text, dest, reply_to, notice=False):
...
+        Within the context of a triggered callable, ``channel`` will default to the
+        channel in which the call was triggered. If triggered from a private message,
+        ``channel`` is required (or the call to ``kick()`` will be ignored).
```
```suggestion
        ``channel`` is required (or the call to ``kick()`` will be ignored).

```

sopel/bot.py
```diff
@@ -468,6 +468,16 @@ def reply(self, text, dest, reply_to, notice=False):
...
+        The bot must be a channel operator in specified channel for this to work.
```
```suggestion
        The bot must be a channel operator in specified channel for this to work.

```

sopel/modules/reddit.py
```diff
@@ -91,8 +91,8 @@ def rpost_info(bot, trigger, match=None):
         sfw = bot.db.get_channel_value(trigger.sender, 'sfw')
         if sfw:
             link = '(link hidden)'
-            bot.write(['KICK', trigger.sender, trigger.nick,
-                       'Linking to NSFW content in a SFW channel.'])
+            bot.kick(trigger.nick, trigger.sender,
+                    'Linking to NSFW content in a SFW channel.')
```
dgw:
```suggestion
                     'Linking to NSFW content in a SFW channel.')
```

deathbybandaid:
suggestions applied, squashed, but scrutinizer is throwing a hissy-fit

dgw:
Ugh, merge conflict.

@deathbybandaid There's one `bot.write` -> `bot.kick` migration to reapply in `reddit.py`.

deathbybandaid:
@dgw fixed the conflict

dgw:
Thanks, @deathbybandaid!

PR #1540, opened 2019-04-03
coretasks: add tests

HumorBaby:
Test coverage for `coretasks.py` is a little lacking. Here is my attempt to remedy that... :hankey: 

Requires #1542 to be merged first.

PR #1542, opened 2019-04-04
tests: make MockSopelWrapper subclass actual SopelWrapper class

HumorBaby:
Despite notes like this: https://github.com/sopel-irc/sopel/blob/229f3bfacaf393f331bf79569783f23d88d05a03/sopel/bot.py#L270-L272
each outgoing method (`bot.msg`, `bot.say`, `bot.notice`, `bot.action`,
`bot.reply`) should be present in the mock as long as they are present
in the codebase such that they can be used in testing.

These "mocks" need to exist so that more tests can be written (e.g., #1540) that use _actual_ mockers (`e.g., pytest-mock`).

### DIFF
--- a/sopel/bot.py
+++ b/sopel/bot.py
@@ -376,47 +376,6 @@ class Sopel(irc.Bot):
         else:
             self.say(text, dest)
 
-    class SopelWrapper(object):
-        def __init__(self, sopel, trigger):
-            # The custom __setattr__ for this class sets the attribute on the
-            # original bot object. We don't want that for these, so we set them
-            # with the normal __setattr__.
-            object.__setattr__(self, '_bot', sopel)
-            object.__setattr__(self, '_trigger', trigger)
-
-        def __dir__(self):
-            classattrs = [attr for attr in self.__class__.__dict__
-                          if not attr.startswith('__')]
-            return list(self.__dict__) + classattrs + dir(self._bot)
-
-        def __getattr__(self, attr):
-            return getattr(self._bot, attr)
-
-        def __setattr__(self, attr, value):
-            return setattr(self._bot, attr, value)
-
-        def say(self, message, destination=None, max_messages=1):
-            if destination is None:
-                destination = self._trigger.sender
-            self._bot.say(message, destination, max_messages)
-
-        def action(self, message, destination=None):
-            if destination is None:
-                destination = self._trigger.sender
-            self._bot.action(message, destination)
-
-        def notice(self, message, destination=None):
-            if destination is None:
-                destination = self._trigger.sender
-            self._bot.notice(message, destination)
-
-        def reply(self, message, destination=None, reply_to=None, notice=False):
-            if destination is None:
-                destination = self._trigger.sender
-            if reply_to is None:
-                reply_to = self._trigger.nick
-            self._bot.reply(message, destination, reply_to, notice)
-
     def call(self, func, sopel, trigger):
         nick = trigger.nick
         current_time = time.time()
@@ -493,7 +452,7 @@ class Sopel(irc.Bot):
                 user_obj = self.users.get(pretrigger.nick)
                 account = user_obj.account if user_obj else None
                 trigger = Trigger(self.config, pretrigger, match, account)
-                wrapper = self.SopelWrapper(self, trigger)
+                wrapper = SopelWrapper(self, trigger)
 
                 for func in funcs:
                     if (not trigger.admin and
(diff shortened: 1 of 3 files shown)


HumorBaby:
Sorry for the review request ping Exirel, but I think this is simple enough that anyone can review :smile:

Exirel:
If you are up to the task, you can extract `SopelWrapper` from `Sopel`, and then extend and override it in `MockSopelWrapper`.

That would make our life so much easier...

HumorBaby:
> If you are up to the task, you can extract `SopelWrapper` from `Sopel`, and then extend and override it in `MockSopelWrapper`.

I'll give it a go :alembic:... it certainly would make test-writing much cleaner/simpler!

Exirel:
I think that's the right PR to do that. :)

dgw:
I'm 💯% on board with having `SopelWrapper` as a first-class citizen. Just let me know when you have something for me to look at :grin:

HumorBaby:
- [x] Extract `SopelWrapper` from `Sopel` (into its own module because of circular-import issues<sup>[1]</sup>)

- [x] Update `MockSopelWrapper` with a subclass of `SopelWrapper`, and override `MockSopel` methods (`MockSopelWrapper` is not overriden or extended in any way<sup>[2]</sup>)

- [x] Update `get_example_test` in `sopel/test_tools.py` to count `output` stack correctly<sup>[3]</sup>

**Notes**
1. `from sopel.bot import SopelWrapper` leads to import errors because `sopel.bot` wants `sopel.module` which wants `sopel.test_tools`, which will want `SopelWrapper` from `sopel.bot`...

2. `SopelWrapper` does some processing of triggers which affects how bot methods are called, and so I did not want to override this behavior. I feel that only the methods that do something that cannot be tested should be overridden (essentially, methods that directly interact with the `socket`, i.e., actual sending to server). These methods are the last time the bot does anything, and so overriding _these_ methods minimizes the amount of the actual call stack that gets truncated during testing. 

    The implication of this is that `MockSopel` will extend/override `Sopel`, and `MockSopelWrapper` will cease to exist because (or do anything meaningful) because the base `SopelWrapper` class will work fine. This _should_ allow more robust tests of responses to triggers.

3. Previously, each replicate of a test in [`get_example_test(..., repeat, ...)`](https://github.com/sopel-irc/sopel/blob/95812b540e082ff9d24fbd77c19662349ab119e3/sopel/test_tools.py#L132-L133) would get a fresh `output` stack because `MockSopelWrapper` was tracking it, and a new one was created for each replicate. Now,  `MockSopel` handles it, and since it is only instantiated once, the `output` stack grows with each replicate.

    Rather than manually setting `bot.output = []`, the logic of counting expected replies was updated.

**Next**
Now, I plan to change `MockSopel` to subclass `Sopel` and start digging into which methods will need to be overridden. Let me know if I'm going down the wrong track here @Exirel / @dgw.

HumorBaby:
> ~~Now, I plan to change MockSopel to subclass Sopel and start digging into which methods will need to be overridden. Let me know if I'm going down the wrong track here @Exirel / @dgw.~~

This would be better left for another PR, as it is no small task :cry: 

In the mean time, the current state of this PR satisfies the goal set out in the title, and is enough for those with grander plans for writing tests to work with, for now.

sopel/test_tools.py
```diff
@@ -65,11 +66,18 @@ def __init__(self, nick, admin=False, owner=False):
...
+
         if admin:
             self.config.core.admins = [self.nick]
         if owner:
             self.config.core.owner = self.nick
 
+    def _store(self, string, *args, **kwargs):
+        self.output.append(string.strip())
+
+    write = msg = say = notice = action = reply = _store
```
Exirel: I wonder which of these functions are actually high-level or low-leve methods. That means: which one use the `send` method, and which one are just fancy shortcut over a low-level method (which uses `send`)?

Basically, I'm wondering which of these methods really requires the `send` method.

sopel/bot.py
```diff
@@ -376,47 +377,6 @@ def reply(self, text, dest, reply_to, notice=False):
         else:
             self.say(text, dest)
 
-    class SopelWrapper(object):
```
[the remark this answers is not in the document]
dgw: I'm on board with this. Adding yet another sub-namespace in the form of `sopel.wrapper` seems like overkill.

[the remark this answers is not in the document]
HumorBaby: Done. Resolved the resulting circular dependency: 
> from sopel.bot import SopelWrapper leads to import errors because sopel.bot wants sopel.module which wants sopel.test_tools, which will want SopelWrapper from sopel.bot...

in a less-than-ideal way (violates "convention" of imports at the top; this file already defers one import: `import sys`, so...) by moving `import sopel.test_tools` into `module.example.__call__` where it was needed. 

Let me know if instead this issue should be resolved with a better refactoring of the current test functions between `test_tools.py` and `module.example`.

sopel/test_tools.py
HumorBaby:
```
+----------------+    +--------------+    +-------------+
|  Sopel.action  +--->+  Sopel.say   +--->+             |
|  Sopel.msg     |    +--------------+    |             |    +---------------+    +--------------------------+
|  Sopel.reply   |                        | Sopel.write +--->+ irc.Bot.write +--->+ asynchat.async_chat.send |
+----------------+    +--------------+    |             |    +---------------+    +--------------------------+
                      | Sopel.notice +--->+             |
                      +--------------+    +-------------+
```
So ultimately, the bot's `write` method calls `send`. Since `MockSopel` doesn't subclass `Sopel`, all the other methods need to be defined (and point to `_store` in the mean time). Ideally, `MockSopel` will inherit the actual shortcut methods from `Sopel` so that any additional processing that happens
(e.g., in `reply` or even `say`) is actually executed.

This is what I imagined when I said: https://github.com/sopel-irc/sopel/pull/1542#issuecomment-481223596

Do you have a preference for how `write = msg = say = notice = action = reply = _store` should be organized differently? For example, more explicitly, like:
```python
def _store(self, string, *args, **kwargs):
    ...

write = _store
say = write
notice = write
msg = reply = action = say
```
to more directly represent how the methods are routed in `Sopel`?

Exirel: This is fine as-is. Big up for the lovely graph you made, that helped me a lot understanding the issue. Thank you!

sopel/module.py
```diff
@@ -472,6 +471,8 @@ def __call__(self, func):
 
         import sys
 
+        import sopel.test_tools
```
Exirel: Could you add a `# TODO: fix circular import with sopel.bot and sopel.test_tools`?

sopel/bot.py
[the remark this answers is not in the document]
Exirel: Yeah, that makes sense for now. It's not ideal, but I consider this PR to be a first step toward a cleaner interface for both production and testing purposes.

So go for it, just add a TODO comment. :)

sopel/module.py
HumorBaby: :+1: Done.

Exirel:
:ship:

dgw:
Travis needs a real kick in the pants, and I'm told closing/reopening the PR will do that.

Exirel:
@HumorBaby well done!

dgw:
Let's get this in so @Exirel can continue working on #1479. :shipit: :rocket:

PR #1547, opened 2019-04-05
logging: use config's basename in log filenames

deathbybandaid:
This is a fairly simple tweak to the way log files are named.

The use case is for people that run more than one Sopel bot, distinguishing exceptions and stdio log files would be a bit easier this way.

### DIFF
--- a/sopel/irc.py
+++ b/sopel/irc.py
@@ -102,7 +102,7 @@ class Bot(asynchat.async_chat):
                 stderr('%s %s' % (str(e.__class__), str(e)))
                 stderr('Please fix this and then run Sopel again.')
                 os._exit(1)
-        f = codecs.open(os.path.join(self.config.core.logdir, 'raw.log'),
+        f = codecs.open(os.path.join(self.config.core.logdir, self.config.basename + '.raw.log'),
                         'a', encoding='utf-8')
         f.write(prefix + unicode(time.time()) + "\t")
         temp = line.replace('\n', '')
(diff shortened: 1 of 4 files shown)


dgw:
Better to use config name, IMO. Either way, this patch is broken as-is because it tries to use `self.config` in `cli/run.py`. Use just `config`.

HumorBaby:
I would like to make the case that this should use the config file name instead of the bot nick. The config file is tied more closely to a "bot" than the bot's nick.

This is not a valuation of whether this change is actually needed.

sopel/irc.py
```diff
@@ -102,7 +102,7 @@ def log_raw(self, line, prefix):
                 stderr('%s %s' % (str(e.__class__), str(e)))
                 stderr('Please fix this and then run Sopel again.')
                 os._exit(1)
-        f = codecs.open(os.path.join(self.config.core.logdir, 'raw.log'),
+        f = codecs.open(os.path.join(self.config.core.logdir, self.config.basename + 'raw.log'),
```
dgw: For consistency with the others:

```suggestion
        f = codecs.open(os.path.join(self.config.core.logdir, self.config.basename + '.raw.log'),
```

dgw:
Needs a squishin' and a squashin'.

deathbybandaid:
ammended

Exirel:
Hmmmm I totally forgot that one. I need to update #1564 again. 😬

PR #1518, opened 2019-03-25
 config, bot: Configurable message rate

kwaaak:
Code originally by @larsks in #908 

I just made it compatible with the current state.

PR #1552, opened 2019-04-11
core: limit maximum flood penalty

### LINKED ISSUE #1550: .dice denial of service
If I spam sopel with `.dice 0xff^0xff^0xff`, sopel's replies to arbitrary commands (e.g., .`version`)  become indefinitely delayed. 

.version: `Sopel v. 6.6.5 (Python 3.4.8) at commit: 824ae7484b9e04310079a27d4d6ec40edeafdfd1`

environment: gentoo, ran from source, no special configuration

logs: nothing relevant in the logs

### LINKED ISSUE #1551: .py denial of service
Similar to #1550; spam `.py [0]*100000`.

HumorBaby:
The penalty for a message that trips flood protection scales directly with the message length and has no upper bound. As seen in #1550 and #1551, a 64bk message (which can apparently be returned during normal bot function) would incur a penalty of ~37 minutes between message sends.

I think the best option is to just cap the maximum wait time between messages. I followed the example of ZNC, which has a default of `1.00` seconds between messages that trip its flood protection.

This option is configurable under `[core]` with the `flood_trickle_rate` setting.

Fixes #1550 and fixes #1551

### DIFF
--- a/sopel/bot.py
+++ b/sopel/bot.py
@@ -313,7 +313,7 @@ class Sopel(irc.Bot):
                 elapsed = time.time() - self.stack[recipient_id][-1][0]
                 if elapsed < 3:
                     penalty = float(max(0, len(text) - 40)) / 70
-                    wait = 0.8 + penalty
+                    wait = min(0.8 + penalty, 2)  # Never wait more than 2 seconds
                     if elapsed < wait:
                         time.sleep(wait - elapsed)
 


HumorBaby:
> I followed the example of ZNC, which has a default of 1.00 seconds between messages that trip its flood protection.

While the docs show a default of `1.00` seconds, this shows `2.00` seconds.
https://github.com/znc/znc/pull/1416, so the `fixup: ...` updates this. Let me know what the final value should be.

That discussion also refers to a spec that outlines a potential flood penalty logic, but it seems that a hard limit on the penalty will suffice in most cases (I'm following ZNC's lead here).

HumorBaby:
@kwaaak pointed out #1518 on IRC. It deals directly with the message rate, but it still would have the underlying issue, because of the penalty being directly proportional to the message length: `penalty = float(max(0, len(text) - 50)) / 70`...

If #1550 and #1551 are not urgent, then I guess this PR can be closed in favor of #1518, and the necessary changes can built from there.

dgw:
> If #1550 and #1551 are not urgent, then I guess this PR can be closed in favor of #1518, and the necessary changes can built from there.

Would be nice to at least reduce the severity of the issue in 6.x. Try to get some insight into a sensible cap for the flood penalty, and/or work with @kwaaak to coordinate moving the penalty calculation to the right place (i.e. after `get_sendable_message` runs and truncates the text).

HumorBaby:
>i.e. after get_sendable_message runs and truncates the text).

I had considered this first, but the problem is that `get_sendable_message` splits text at word boundaries when possible, and so a response from `.dice ...` with a number that is a 64kb string would never actually get split/truncated with just `get_sendable_message` if `max_messages=1` (the default), and instead only the `You rolled a: ` would be sent, without an actual result.

The only way `get_sendable_message` would work here would be if  `max_messages` was automatically adjusted to a higher value, which defeats its purpose.... I guess `max_messages` could have a higher default instead, but the results of such a change would need to be considered...

HumorBaby:
>Try to get some insight into a sensible cap for the flood penalty

The peeps over at #ircv3 had this to say:
```
<Z...>  most delays do something like a [2-8] message burst with [1-2] second delay afterwards. there is no need to make it backoff exponentially
<Z...>  there may be configurations where it *does* make sense, but imo the server is misconfigured at that point, not the client
```

---

>and/or work with @kwaaak to coordinate moving the penalty calculation to the right place

Looks like dd3c407 in #1518 was since changed to match this one :smile: 

---

>(i.e. after get_sendable_message runs and truncates the text).

I think the current location is the quickest/safest fix for now (until the entire message rate logic is moved to the correct location, somewhere in `irc.Bot.write` or `Sopel.write`). See previous comment.

dgw:
This is a weird thing to ask for, but would you mind hard-coding the trickle rate for 6.6.x so we don't add a new core config value in a patch release?

We'll figure out what to do with #1518 on or after April 15 (when 6.6.x gets merged back to master). The new config knob fits in perfectly; I'll just leave it up to you and @kwaaak to work out commit authorship.

HumorBaby:
Done. I was fretting about the setting name anyway. It's more of an _interval_ anyway, as opposed to a _rate_. Glad to not have to worry about that anymore :smile:

PR #1555, opened 2019-04-11
test: add an --offline option to pytest

Exirel:
Usage:

    $ py.test --offline

This option skip all tests that can not be run offline, as they are marked as "online". Either from manual tag:

    @pytest.mark.online
    def test_call_google():
        request.get('https://google.com')

Or from a command example:

    @example('.title http://google.com', '[ Google ] - google.com', online=True)
    def title_command(bot, trigger):
        # ...

This is especially useful for people like me who are always traveling and can't have access to a reliable Internet connection.

Existing online examples have been marked as such.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -1,2 +1,19 @@
+import pytest
+
 # This file lists files which should be ignored by pytest
 collect_ignore = ["setup.py", "sopel.py", "sopel/modules/ipython.py"]
+
+
+def pytest_addoption(parser):
+    parser.addoption('--offline', action='store_true', default=False)
+
+
+def pytest_collection_modifyitems(config, items):
+    if not config.getoption('--offline'):
+        # nothing to skip
+        return
+
+    skip_online = pytest.mark.skip(reason='deactivated when --offline is used')
+    for item in items:
+        if 'online' in item.keywords:
+            item.add_marker(skip_online)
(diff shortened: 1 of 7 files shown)


dgw:
Before I dive into this—I like the idea—we might get some utility from a set of options:

  - `--offline`, alias `--offline-only`: run only offline-safe tests
  - `--online`, alias `--online-only`: run only online tests (which call APIs or whatever)
  - `--no-offline`: skip offline tests
  - `--no-online`: skip online tests

This might be overkill, true. If it makes implementation easier we can do just the `-only` args and skip the `no-` options. My thinking here is we could later configure Travis to run only the offline tests during required-pass jobs, and have separate test runs for the online tests that are allowed to fail. Then we wouldn't have random transient changes to DuckDuckGo search results (or whatever) failing related PRs.

Exirel:
I got reminded of this PR when I had trouble testing locally because some random website was inaccessible and/or very slow.

@dgw I think this PR is becoming more and more important, since Travis can literally fails for the wrong reasons because of online-only tests.

dgw:
Hmm… @Exirel I think you did something weird in your last rebase. The commit is showing _me_ as author, which can't be right. This code is all you. :confused:

Exirel:
@dgw I've honestly no idea how or why. :|

Exirel:
Yay `git reset HEAD~1` worked as intended! Should be good now.

PR #1556, opened 2019-04-11
 admin: add unset command

### LINKED ISSUE #1554: admin: .set cannot clear values
The `.set` command of `admin.py` does not work as described.

https://github.com/sopel-irc/sopel/blob/824ae7484b9e04310079a27d4d6ec40edeafdfd1/sopel/modules/admin.py#L252-L263

says that `if value is None, the option will be deleted`. Currently, a `None` value does not get parsed properly, and the setting just ends up getting converted to a literal `None` string, as opposed to an actual `NoneType`.

I considered that the sentence meant that no value should be supplied, but that would preclude the *See* part of the `See and modify values of Sopel's config object`, where no argument should be provided for setting (which is working as expected). Thus, :bug:

---

Steps to replicate:
```python
# admin_set_test.py

from __future__ import unicode_literals, absolute_import, division, print_function

from sopel.config.types import StaticSection, ValidatedAttribute
import sopel.module


class AdminSetTestSection(StaticSection):
    test_setting = ValidatedAttribute('test_setting', default=None)


def setup(bot):
    bot.config.define_section('admin_set_test', AdminSetTestSection)


@sopel.module.require_privmsg
@sopel.module.commands('ast-show')
def show(bot, trigger):
    bot.say('repr: "{}", type: "{}"'.format(
        repr(bot.config.admin_set_test.test_setting),
        type(bot.config.admin_set_test.test_setting)
    ))
    
```

```
# my-config.cfg
## Note: there is no `test_setting` set, so it defaults to `None` (`NoneType`)

[admin_set_test]

```

```
# _See_ current setting value
<HumorBaby>  .set admin_set_test.test_setting
<testbot>  HumorBaby: admin_set_test.test_setting = None

# See repr and type of setting
<HumorBaby>  .ast-show
<testbot>  repr: "None", type: "<type 'NoneType'>"

# Set a random value to setting
<HumorBaby>  .set admin_set_test.test_setting abc123

# Confirm change
<HumorBaby>  .set admin_set_test.test_setting
<testbot>  HumorBaby: admin_set_test.test_setting = abc123
<HumorBaby>  .ast-show
<testbot>  repr: "u'abc123'", type: "<type 'unicode'>"

# Try to "delete"/unset setting
<HumorBaby>  .set admin_set_test.test_setting None

# Confirm change
<HumorBaby>  .set admin_set_test.test_setting
<testbot>  HumorBaby: admin_set_test.test_setting = None
# everything looks okay...

# But wait... wat?
<HumorBaby>  .ast-show
<testbot>  repr: "u'None'", type: "<type 'unicode'>"
```

So, I wondered if `.save`'ing the config would somehow solve the issue...
```
<HumorBaby>  .save

# Restart bot here...

<HumorBaby>  .set admin_set_test.test_setting
<testbot>  HumorBaby: admin_set_test.test_setting = None
<HumorBaby>  .ast-show
<testbot>  repr: "'None'", type: "<type 'str'>"
```
Nope...

HumorBaby:
If the value to be set is ever literally 'None' (string), then the setting is set to actual `NoneType` `None`. When a true `None` is set to a config attribute, the setting will essentially be reset to the default value. I feel this should be the expected behavior; i.e., a user should not be able to set a null value for something with a default, because presumably, that default is there for a reason.

---

Testing: _(see #1554 for steps)_

```python
# admin_set_test.py
...
class AdminSetTestSection(StaticSection):
    test_setting = ValidatedAttribute('test_setting', default=None)
    test_list = ListAttribute('test_list', default=['a','b','c'])
...
@sopel.module._privmsg
@sopel.module.commands('ast-show')
def show(bot, trigger):
    bot.say('repr: "{}", type: "{}"'.format(
        repr(bot.config.admin_set_test.test_setting),
        type(bot.config.admin_set_test.test_setting)
    ))
    bot.say('repr: "{}", type: "{}"'.format(
        repr(bot.config.admin_set_test.test_list),
        type(bot.config.admin_set_test.test_list)
    ))
```

```
# _See_ current setting value
<HumorBaby>  .set admin_set_test.test_setting
<testbot>  HumorBaby: admin_set_test.test_setting = None

# See repr and type of setting
<HumorBaby>  .ast-show
<testbot>  repr: "None", type: "<type 'NoneType'>"

# Set a random value to setting
<HumorBaby>  .set admin_set_test.test_setting abc123

# Confirm change
<HumorBaby>  .set admin_set_test.test_setting
<testbot>  HumorBaby: admin_set_test.test_setting = abc123
<HumorBaby>  .ast-show
<testbot>  repr: "u'abc123'", type: "<type 'unicode'>"

# Try to "delete"/unset setting
<HumorBaby>  .set admin_set_test.test_setting None

# Confirm change
<HumorBaby>  .set admin_set_test.test_setting
<testbot>  HumorBaby: admin_set_test.test_setting = None
# everything looks okay...

# It is!
<HumorBaby>  .ast-show
<testbot>  repr: "None", type: "<type 'NoneType'>"
```

```
# Test the `ListAttribute` type
<HumorBaby>  .set admin_set_test.test_list
<testbot>  HumorBaby: admin_set_test.test_list = [u'a', u'b', u'c']
<HumorBaby>  .ast-show
<testbot>  repr: "[u'a', u'b', u'c']", type: "<type 'list'>"

# Change the list
<HumorBaby>  .set admin_set_test.test_list 1,2,3
<HumorBaby>  .set admin_set_test.test_list
<testbot>  HumorBaby: admin_set_test.test_list = [u'1', u'2', u'3']
<HumorBaby>  .ast-show
<testbot>  repr: "[u'1', u'2', u'3']", type: "<type 'list'>"

# Unset the list
<HumorBaby>  .set admin_set_test.test_list None
<HumorBaby>  .set admin_set_test.test_list
<testbot>  HumorBaby: admin_set_test.test_list = [u'a', u'b', u'c']
<HumorBaby>  .ast-show
<testbot>  repr: "[u'a', u'b', u'c']", type: "<type 'list'>"
```

Fixes #1554

### DIFF
--- a/sopel/modules/admin.py
+++ b/sopel/modules/admin.py
@@ -386,7 +386,10 @@ def unset_config(bot, trigger):
         bot.reply('Invalid command; no value should be provided to unset.')
         return
 
-    setattr(section, option, None)
+    try:
+        setattr(section, option, None)
+    except ValueError:
+        bot.reply('Cannot unset {}.{}; it is a required option.'.format(section_name, option))
 
 
 @sopel.module.require_privmsg
--- a/sopel/modules/admin.py
+++ b/sopel/modules/admin.py
@@ -44,6 +44,19 @@ def setup(bot):
     bot.config.define_section('admin', AdminSection)
 
 
+class InvalidSection(Exception):
+    def __init__(self, section):
+        super(InvalidSection, self).__init__(self, 'Section [{}] does not exist.'.format(section))
+        self.section = section
+
+
+class InvalidSectionOption(Exception):
+    def __init__(self, section, option):
+        super(InvalidSectionOption, self).__init__(self, 'Section [{}] does not have option \'{}\'.'.format(section, option))
+        self.section = section
+        self.option = option
+
+
 def _get_config_channels(channels):
     """List"""
     for channel_info in channels:
@@ -245,6 +258,53 @@ def mode(bot, trigger):
     bot.write(('MODE', bot.nick + ' ' + mode))
 
 
+def parse_section_option_value(config, trigger):
+    """Parse trigger for set/unset to get relevant config elements.
+
+    :param config: Sopel's config
+    :param trigger: IRC line trigger
+    :return: A tuple with ``(section, section_name, static_sec, option, value)``
+    :raises InvalidSection: section does not exist
+    :raises InvalidSectionOption: option does not exist for section
+
+    The ``value`` is optional and can be returned as ``None`` if omitted from command.
+    """
+    match = trigger.group(3)
+    if match is None:
+        raise ValueError  # Invalid command
+
+    # Get section and option from first argument.
+    arg1 = match.split('.')
+    if len(arg1) == 1:
+        section_name, option = "core", arg1[0]
+    elif len(arg1) == 2:
+        section_name, option = arg1
+    else:
+        raise ValueError  # invalid command format
+
+    section = getattr(config, section_name, False)
+    if not section:
+        raise InvalidSection(section_name)
+    static_sec = isinstance(section, StaticSection)
+
+    if static_sec and not hasattr(section, option):
+        raise InvalidSectionOption(section_name, option)  # Option not found in section
+
+    if not static_sec and not config.parser.has_option(section_name, option):
+        raise InvalidSectionOption(section_name, option)  # Option not found in section
+
+    delim = trigger.group(2).find(' ')
+    # Skip preceding whitespaces, if any.
+    while delim > 0 and delim < len(trigger.group(2)) and trigger.group(2)[delim] == ' ':
+        delim = delim + 1
+
+    value = trigger.group(2)[delim:]
+    if delim == -1 or delim == len(trigger.group(2)):
+        value = None
+
+    return (section, section_name, static_sec, option, value)
+
+
 @sopel.module.require_privmsg("This command only works as a private message.")
 @sopel.module.require_admin("This command requires admin privileges.")
 @sopel.module.commands('set')
@@ -257,40 +317,19 @@ def set_config(bot, trigger):
         arg2 - value
 
     If there is no section, section will default to "core".
-    If value is None, the option will be deleted.
+    If value is not provided, the current value will be displayed.
     """
-    # Get section and option from first argument.
-    match = trigger.group(3)
-    if match is None:
-        bot.reply("Usage: .set section.option value")
+    try:
+        section, section_name, static_sec, option, value = parse_section_option_value(bot.config, trigger)
+    except ValueError:
+        bot.reply('Usage: .set section.option [value]')
         return
-    arg1 = match.split('.')
-    if len(arg1) == 1:
-        section_name, option = "core", arg1[0]
-    elif len(arg1) == 2:
-        section_name, option = arg1
-    else:
-        bot.reply("Usage: .set section.option value")
-        return
-    section = getattr(bot.config, section_name)
-    static_sec = isinstance(section, StaticSection)
-
-    if static_sec and not hasattr(section, option):
-        bot.say('[{}] section has no option {}.'.format(section_name, option))
+    except (InvalidSection, InvalidSectionOption) as exc:
+        bot.say(exc.args[1])
         return
 
-    delim = trigger.group(2).find(' ')
-    # Skip preceding whitespaces, if any.
-    while delim > 0 and delim < len(trigger.group(2)) and trigger.group(2)[delim] == ' ':
-        delim = delim + 1
-
-    # Display current value if no value is given.
-    if delim == -1 or delim == len(trigger.group(2)):
-        if not static_sec and bot.config.parser.has_option(section, option):
-            bot.reply("Option %s.%s does not exist." % (section_name, option))
-            return
-        # Except if the option looks like a password. Censor those to stop them
-        # from being put on log files.
+    # Display current value if no value is given
+    if not value:
         if option.endswith("password") or option.endswith("pass"):
             value = "(password censored)"
         else:
@@ -305,8 +344,7 @@ def set_config(bot, trigger):
                 .format(section_name, option))
         return
 
-    # Otherwise, set the value to one given as argument 2.
-    value = trigger.group(2)[delim:]
+    # Otherwise, set the value to one given
     if static_sec:
         descriptor = getattr(section.__class__, option)
         try:
--- a/sopel/modules/admin.py
+++ b/sopel/modules/admin.py
@@ -322,7 +322,7 @@ def set_config(bot, trigger):
     try:
         section, section_name, static_sec, option, value = parse_section_option_value(bot.config, trigger)
     except ValueError:
-        bot.reply('Usage: .set section.option [value]')
+        bot.reply('Usage: {}set section.option [value]'.format(bot.config.core.help_prefix))
         return
     except (InvalidSection, InvalidSectionOption) as exc:
         bot.say(exc.args[1])
@@ -356,6 +356,37 @@ def set_config(bot, trigger):
             bot.say("Can't set attribute: " + str(exc))
             return
     setattr(section, option, value)
+
+
+@sopel.module.require_privmsg("This command only works as a private message.")
+@sopel.module.require_admin("This command requires admin privileges.")
+@sopel.module.commands('unset')
+@sopel.module.example('.unset core.owner')
+def unset_config(bot, trigger):
+    """Unset value of Sopel's config object.
+
+    Unsetting a value will reset it to the default specified in the config
+    definition.
+
+    Trigger args:
+        arg1 - section and option, in the form "section.option"
+
+    If there is no section, section will default to "core".
+    """
+    try:
+        section, section_name, static_sec, option, value = parse_section_option_value(bot.config, trigger)
+    except ValueError:
+        bot.reply('Usage: {}unset section.option [value]'.format(bot.config.core.help_prefix))
+        return
+    except (InvalidSection, InvalidSectionOption) as exc:
+        bot.say(exc.args[1])
+        return
+
+    if value:
+        bot.reply('Invalid command; no value should be provided to unset.')
+        return
+
+    setattr(section, option, None)
 
 
 @sopel.module.require_privmsg
--- a/sopel/modules/admin.py
+++ b/sopel/modules/admin.py
@@ -334,7 +334,7 @@ def set_config(bot, trigger):
             value = "(password censored)"
         else:
             value = getattr(section, option)
-        bot.reply("%s.%s = %s" % (section_name, option, value))
+        bot.reply("%s.%s = %s (%s)" % (section_name, option, value, type(value).__name__))
         return
 
     # Owner-related settings cannot be modified interactively. Any changes to these
@@ -356,6 +356,7 @@ def set_config(bot, trigger):
             bot.say("Can't set attribute: " + str(exc))
             return
     setattr(section, option, value)
+    bot.say("OK. Set '{}.{}' successfully.".format(section_name, option))
 
 
 @sopel.module.require_privmsg("This command only works as a private message.")
@@ -388,6 +389,7 @@ def unset_config(bot, trigger):
 
     try:
         setattr(section, option, None)
+        bot.say("OK. Unset '{}.{}' successfully.".format(section_name, option))
     except ValueError:
         bot.reply('Cannot unset {}.{}; it is a required option.'.format(section_name, option))
 
--- a/sopel/config/types.py
+++ b/sopel/config/types.py
@@ -156,6 +156,8 @@ class BaseValidated(object):
 
     def __set__(self, instance, value):
         if value is None:
+            if self.default == NO_DEFAULT:
+                raise ValueError('Cannot unset an option with a required value.')
             instance._parser.remove_option(instance._section_name, self.name)
             return
         value = self.serialize(value)


Exirel:
Issue #1556 raises valid concern about how `.set` work, and the lack of "unset a config value". However, what you do here is a hack that relies on the interpretation of a string that may or may not be what the user actually wants. So I'm not on board with this solution.

IMHO, I'd rather have `.unset section.option` to remove the option and `.reset section.option` to set its value to its default value, and no way to set it to `None` (as-in the `NoneType` object `None` which make no sense for a configuration file that is all about strings).

HumorBaby:
> However, what you do here is a hack that relies on the interpretation of a string that may or may not be what the user actually wants. So I'm not on board with this solution.

I'm with you completely. I felt a little uneasy when I was doing it, but somehow I ignore my instincts and convinced myself it was okay :sweat_smile: 

>no way to set it to None (as-in the NoneType object None which make no sense for a configuration file that is all about strings).

Yes, I agree :smile: _If_ a string None is present in the config, then it will be interpreted as a string, as expected. Thus, it makes no sense to try to set a `NoneType` `None` to a configuration setting and expecting it to be written/saved that way. 

Also, given how setting values are pulled from a config object, just means that a `None` value will return the default for that setting. If a truly `None` value is desired, then it must be the default for that item. If there is a non-`None` default, and a user wants to set a `None` value to the setting, it will not be possible to save that value to config. Another point: I cannot think of a setting that would need to be nullified when the default value is non-`None`. Overall, this relates to your point: 
>.unset section.option to remove the option and .reset section.option to set its value to its default value

There is no real meaning to having removed an option. It will just return the default. And because setting a value to `None` will also return the default, I don't see a reason to have `.unset` and `.reset` as they will have the same result. I think `.unset` is the better option to keep here, just because it is more parallel to the `.set` terminology.

HumorBaby:
Since this is a relatively large change... changing to `master`, but not rebasing to hopefully avoid conflicts once `6.6.x` is merged in.

dgw:
I definitely think adding an `.unset` command is the way to go, but handling both commands in the same function (and defining all the logic to manipulate the config inside that function) is… kinda ew.

Not that I have any modules that look at `trigger.group(1)` to see how they're being called… but they're much simpler. This gets pretty complex by trying to do everything in one place. :sweat_smile:

HumorBaby:
>but handling both commands in the same function (and defining all the logic to manipulate the config inside that function) is… kinda ew.

Yea... was trying to minimize duplication of the section and option parsing. Refactor coming soon, then.

Exirel:
> Refactor coming soon, then.

That's the way to go! 👍

HumorBaby:
> Refactor coming soon.

'tis here.

dgw:
I guess I'll come back to this in a few days, after 6.6.x is merged into master and you can rebase. :grin:

dgw:
Ready to rebase; I merged 6.6.x into master just now.

HumorBaby:
:tada:

sopel/modules/admin.py
```diff
@@ -247,61 +247,85 @@ def mode(bot, trigger):
...
-        arg1 - section and option, in the form "section.option"
-        arg2 - value
-
-    If there is no section, section will default to "core".
-    If value is None, the option will be deleted.
+def parse_section_option_value(bot, trigger):
+    """Parse trigger for set/unset to get relevant config elements.
+
+    Returns:
+        - False, if invalid command format; let command handle sending usage.
```
Exirel: I see where you are coming from: language like C or Go uses `if err == -1` or alike as their control-flow mechanism—which is fine for these languages. The pythonic way for control flow is to use exceptions instead.

For example, instead of `return False` you can use `raise ValueError`.

Then, in both `set_config` and `unset_config`, you can do that:

```
try:
    section, section_name, static_sec, option, value = parse_section_option_value(
        bot.config, trigger)
except ValueError:
    bot.reply('Usage: .set section.option [value]')
    return
except InvalidSection as error:
    bot.reply("Section %s does not exist." % (error.section))
    return
except InvalidOption as error:
    bot.reply("Option %s.%s does not exist." % (error.section, error.option))
    return
```

Provided you have `InvalidSection` and `InvalidOption ` exception class at your disposal—if not, create them!

Also, notice how I replace the parameter `bot` by `bot.config` in `parse_section_option_value`.

sopel/modules/admin.py
```diff
@@ -247,61 +247,85 @@ def mode(bot, trigger):
...
     arg1 = match.split('.')
     if len(arg1) == 1:
         section_name, option = "core", arg1[0]
     elif len(arg1) == 2:
         section_name, option = arg1
     else:
-        bot.reply("Usage: .set section.option value")
-        return
+        return False  # Invalid command
     section = getattr(bot.config, section_name)
```
Exirel: This will raise `AttributeError` if `bot.config` does not have a section named `section_name`. You should provide a default value here, in order to handle your own control flow.

sopel/modules/admin.py
```diff
@@ -247,61 +247,85 @@ def mode(bot, trigger):
...
     static_sec = isinstance(section, StaticSection)
 
     if static_sec and not hasattr(section, option):
         bot.say('[{}] section has no option {}.'.format(section_name, option))
-        return
+        return ()  # Option not found.
+
+    if not static_sec and bot.config.parser.has_option(section, option):
```
Exirel: I do  not understand this line. I must be missing something.

If `section` is *not* a `StaticSection`, AND if `section.option` **does exist**, then it's an error case?

sopel/modules/admin.py
HumorBaby: Nice! I like that a lot better. I'm happy to be more idiomatic :grin: 

Suggested changes have been made.

sopel/modules/admin.py
HumorBaby: Ah yes, good catch.

Done.

sopel/modules/admin.py
HumorBaby: Another good catch! This exact line has been around for years, so I didn't give it a proper thought :hushed: 

After dedicating the required brain-power, this line has been corrected (in two ways):
- `... and`**`not`**`bot.config...`
- `...has_option(`**`section_name`**`, ...)`: the `parser.has_option` takes a string as its first argument, but `section` is a `sopel.config.ConfigSection` instance instead

and so this condition was in fact completely borked from its early days.

sopel/modules/admin.py
```diff
@@ -44,6 +44,27 @@ def setup(bot):
     bot.config.define_section('admin', AdminSection)
 
 
+class InvalidSection(Exception):
+    def __init__(self, section):
+        super(InvalidSection, self).__init__(self)
+        self.section = section
+
+    @property
```
HumorBaby: I was between this and overriding the `__str__` method to get a pretty message. I went with this to help differentiate this exception from most where a message is actually passed in as the first argument to the constructor (which is not done/desired here).

Exirel: Well, at least the message looks good. :wink:

sopel/modules/admin.py
```diff
@@ -44,6 +44,27 @@ def setup(bot):
     bot.config.define_section('admin', AdminSection)
 
 
+class InvalidSection(Exception):
+    def __init__(self, section):
+        super(InvalidSection, self).__init__(self)
```
Exirel: No need for a `@property` on `message` for that:

```suggestion
        super(InvalidSection, self).__init__(self, 'Section [{}] does not exist.'.format(section))
```

sopel/modules/admin.py
```diff
@@ -44,6 +44,27 @@ def setup(bot):
...
+        self.section = section
+
+    @property
+    def message(self):
+        return 'Section [{}] does not exist.'.format(self.section)
+
+
+class InvalidSectionOption(Exception):
+    def __init__(self, section, option):
+        super(InvalidSectionOption, self).__init__(self)
```
Exirel: Just make sure to stay inside the line length limit (which is complicated to do with Github suggestions):

```suggestion
        super(InvalidSectionOption, self).__init__(self, 'Section [{}] does not have option \'{}\'.'.format(section, option))
```

sopel/modules/admin.py
```diff
@@ -44,6 +44,27 @@ def setup(bot):
...
+        self.section = section
+        self.option = option
+
+    @property
```
Exirel: ditto, this is not needed if you use Exception's `__init__` properly.

sopel/modules/admin.py
```diff
@@ -247,61 +268,85 @@ def mode(bot, trigger):
...
-        arg1 - section and option, in the form "section.option"
-        arg2 - value
+def parse_section_option_value(config, trigger):
+    """Parse trigger for set/unset to get relevant config elements.
 
-    If there is no section, section will default to "core".
-    If value is None, the option will be deleted.
+    Returns:
+        - (section, section_name, static_sec, option, value) otherwise.
+            - value is `None` if omitted from command.
```
Exirel: It's me or there is an extra tabulation here?

sopel/modules/admin.py
```diff
@@ -247,61 +268,85 @@ def mode(bot, trigger):
...
-
-    Trigger args:
-        arg1 - section and option, in the form "section.option"
-        arg2 - value
+def parse_section_option_value(config, trigger):
+    """Parse trigger for set/unset to get relevant config elements.
 
-    If there is no section, section will default to "core".
-    If value is None, the option will be deleted.
+    Returns:
```
Exirel: You could use the Sphinx directive here:

```
:param config: Sopel's config
:param trigger: IRC line trigger
:return: A tuple with ``(section, section_name, static_sec, option, value)``
:raises InvalidSection: section does not exist
:raises InvalidSectionOption: option does not exist for section

The ``value`` is optional and can be returned as ``None`` if omitted from command.
```

sopel/modules/admin.py
```diff
@@ -247,61 +268,85 @@ def mode(bot, trigger):
...
+
+    If there is no section, section will default to "core".
+    If value is not provided, the current value will be displayed.
+    """
+    try:
+        section, section_name, static_sec, option, value = parse_section_option_value(bot.config, trigger)
+    except ValueError:
+        bot.reply('Usage: .set section.option [value]')
+        return
+    except InvalidSection as exc:
```
Exirel: You can catch both at the same time:

```
except (InvalidSection, InvalidSectionOption) as error:
    bot.say(error.message)
```

sopel/modules/admin.py
```diff
@@ -315,6 +360,43 @@ def set_config(bot, trigger):
...
+        arg1 - section and option, in the form "section.option"
+
+    If there is no section, section will default to "core".
+    """
+    try:
+        section, section_name, static_sec, option, value = parse_section_option_value(bot.config, trigger)
+    except ValueError:
+        bot.reply('Usage: .unset section.option [value]')
+        return
+    except InvalidSection as exc:
```
Exirel: Same here. :wink:

sopel/modules/admin.py
dgw: Good thing we don't enforce the line length limit (Yet™) then :smirk:

sopel/modules/admin.py
HumorBaby: Done.

sopel/modules/admin.py
HumorBaby: Much more elegant!

Done.

sopel/modules/admin.py
HumorBaby: It is not just you.

This is removed anyway with the addition of the sphinx directives.

sopel/modules/admin.py
HumorBaby: I am very new to sphinx, so I defer to your expertise here. It has been added as you suggested :smile:

sopel/modules/admin.py
HumorBaby: Done.

sopel/modules/admin.py
```diff
@@ -315,6 +353,40 @@ def set_config(bot, trigger):
...
 
+@sopel.module.require_privmsg("This command only works as a private message.")
+@sopel.module.require_admin("This command requires admin privileges.")
+@sopel.module.commands('unset')
+@sopel.module.example('.unset core.owner')
+def unset_config(bot, trigger):
+    """Unset value of Sopel's config object.
+
+    Unsetting a value will reset it to the default specified in the config
+    defintion.
```
dgw:
```suggestion
    definition.
```

sopel/modules/admin.py
```diff
@@ -315,6 +353,40 @@ def set_config(bot, trigger):
...
+
+    Trigger args:
+        arg1 - section and option, in the form "section.option"
+
+    If there is no section, section will default to "core".
+    """
+    try:
+        section, section_name, static_sec, option, value = parse_section_option_value(bot.config, trigger)
+    except ValueError:
+        bot.reply('Usage: .unset section.option [value]')
```
dgw: Use `help_prefix` here instead of hardcoded `.`.

sopel/modules/admin.py
```diff
@@ -247,61 +260,86 @@ def mode(bot, trigger):
...
+        arg1 - section and option, in the form "section.option"
+        arg2 - value
+
+    If there is no section, section will default to "core".
+    If value is not provided, the current value will be displayed.
+    """
+    try:
+        section, section_name, static_sec, option, value = parse_section_option_value(bot.config, trigger)
+    except ValueError:
+        bot.reply('Usage: .set section.option [value]')
```
dgw: Use `help_prefix` here instead of hardcoded `.`.

I know this and its counterpart are copypasta, but since you're touching this _anyway_ it makes sense to fix it.

dgw:
Nitpicks have been addressed! 🎉 Now there's just a squash (of the fixup commits) and a rebase (to resolve the conflicts introduced on `master`) before final review & approval. 👍

HumorBaby:
> Now there's just a squash (of the fixup commits) and a rebase (to resolve the conflicts introduced on master) before final review & approval. :+1:

Done.

sopel/modules/admin.py
```diff
@@ -318,6 +356,42 @@ def set_config(bot, trigger):
...
+
+    If there is no section, section will default to "core".
+    """
+    try:
+        section, section_name, static_sec, option, value = parse_section_option_value(bot.config, trigger)
+    except ValueError:
+        bot.reply('Usage: {}unset section.option [value]'.format(bot.config.core.help_prefix))
+        return
+    except (InvalidSection, InvalidSectionOption) as exc:
+        bot.say(exc.args[1])
```
Exirel: Ditto.

sopel/modules/admin.py
```diff
@@ -245,57 +258,83 @@ def mode(bot, trigger):
...
+    If there is no section, section will default to "core".
+    If value is not provided, the current value will be displayed.
+    """
+    try:
+        section, section_name, static_sec, option, value = parse_section_option_value(bot.config, trigger)
+    except ValueError:
+        bot.reply('Usage: {}set section.option [value]'.format(bot.config.core.help_prefix))
+        return
+    except (InvalidSection, InvalidSectionOption) as exc:
+        bot.say(exc.args[1])
```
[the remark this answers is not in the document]
HumorBaby: Thought the same thing and tried that, but not what I expected :cry: `str(exc)` would result in the following string: `(InvalidSection(...), _the message here_)`. While Py2 allowed for using the deprecated `exc.message` to get what I needed, Py3 doesn't anymore. So, without making the custom exceptions overly complicated (e.g., by adding a message parameter), I went this route; I think I will leave it as is :smile:

sopel/modules/admin.py
HumorBaby: See prev comment

PR #1557, opened 2019-04-12
jobs: make JobScheduler easier to stop

Exirel:
Two main ideas:

* instead of a `while True` loop to run the scheduler, a `threading.Event` object is used: the scheduler run until this event is set
* instead of a custom `PriorityQueue`, a built-in list object is used to store job: every now and then, the scheduler will look for jobs that are ready to run, run them, and never modify its list, keeping everything thread-safe and simpler

It is kept thread-safe by using properly a `lock`, and it has the advantage of not having to release it or rely to heavily on `time.sleep()` - or at least, only for a second each loop.

### DIFF
--- a/sopel/bot.py
+++ b/sopel/bot.py
@@ -553,10 +553,23 @@ class Sopel(irc.Bot):
         return False
 
     def _shutdown(self):
+        # Stop Job Scheduler
+        stderr('Stopping the Job Scheduler.')
+        self.scheduler.stop()
+
+        try:
+            self.scheduler.join(timeout=15)
+        except RuntimeError:
+            stderr('Unable to stop the Job Scheduler.')
+        else:
+            stderr('Job Scheduler stopped.')
+
+        self.scheduler.clear_jobs()
+
+        # Shutdown plugins
         stderr(
             'Calling shutdown for %d modules.' % (len(self.shutdown_methods),)
         )
-
         for shutdown_method in self.shutdown_methods:
             try:
                 stderr(
--- a/sopel/tools/jobs.py
+++ b/sopel/tools/jobs.py
@@ -1,44 +1,21 @@
 # coding=utf-8
+"""Sopel's Job Scheduler: internal tool for job management.
+
+.. note::
+
+    As of Sopel 5.3, :mod:`sopel.tools.jobs` is an internal tool. Therefore,
+    it is not shown in the public documentation.
+
+"""
 from __future__ import unicode_literals, absolute_import, print_function, division
 
-import copy
 import datetime
 import sys
 import threading
 import time
 
-if sys.version_info.major >= 3:
-    unicode = str
-    basestring = str
-    py3 = True
-else:
-    py3 = False
 
-try:
-    import Queue
-except ImportError:
-    import queue as Queue
-
-
-class PriorityQueue(Queue.PriorityQueue):
-    """A priority queue with a peek method.
-
-    .. deprecated:: 7.0
-
-        This will be removed in Sopel 8. Use Python's built-in
-        :class:`queue.PriorityQueue` instead.
-    """
-    def peek(self):
-        """Return a copy of the first element without removing it."""
-        self.not_empty.acquire()
-        try:
-            while not self._qsize():
-                self.not_empty.wait()
-            # Return a copy to avoid corrupting the heap. This is important
-            # for thread safety if the object is mutable.
-            return copy.deepcopy(self.queue[0])
-        finally:
-            self.not_empty.release()
+py3 = sys.version_info.major >= 3
 
 
 class JobScheduler(threading.Thread):
@@ -205,28 +182,13 @@ class Job(object):
 
         return self
 
-    def __cmp__(self, other):
-        """Compare Job objects according to attribute next_time."""
-        return self.next_time - other.next_time
-
-    if py3:
-        def __lt__(self, other):
-            return self.next_time < other.next_time
-
-        def __gt__(self, other):
-            return self.next_time > other.next_time
-
     def __str__(self):
         """Return a string representation of the Job object.
 
-        Example result:
+        Example result::
+
             <Job(2013-06-14 11:01:36.884000, 20s, <function upper at 0x02386BF0>)>
 
         """
-        iso_time = str(datetime.fromtimestamp(self.next_time))
-        return "<Job(%s, %ss, %s)>" % \
-            (iso_time, self.interval, self.func)
-
-    def __iter__(self):
-        """This is an iterator. Never stops though."""
-        return self
+        iso_time = str(datetime.datetime.fromtimestamp(self.next_time))
+        return "<Job(%s, %ss, %s)>" % (iso_time, self.interval, self.func)
--- a/sopel/tools/jobs.py
+++ b/sopel/tools/jobs.py
@@ -68,6 +68,7 @@ class JobScheduler(threading.Thread):
         """Requires bot as argument for logging."""
         threading.Thread.__init__(self)
         self.bot = bot
+        self.stopping = threading.Event()
         self._jobs = PriorityQueue()
         # While PriorityQueue it self is thread safe, this mutex is needed
         # to stop old jobs being put into new queue after clearing the
@@ -90,9 +91,18 @@ class JobScheduler(threading.Thread):
             self._cleared = True
             self._jobs = PriorityQueue()
 
+    def stop(self):
+        """Ask the job scheduler to stop.
+
+        The scheduler thread will stop its loop over jobs to process, but it
+        won't join the thread, or clear its queue—this has to be done
+        separately by the calling thread.
+        """
+        self.stopping.set()
+
     def run(self):
         """Run forever."""
-        while True:
+        while not self.stopping.is_set():
             try:
                 self._do_next_job()
             except Exception:  # TODO: Be specific
@@ -110,7 +120,7 @@ class JobScheduler(threading.Thread):
         with self._mutex:
             # Wait until the next job should be executed.
             # This has to be a loop, because signals stop time.sleep().
-            while True:
+            while not self.stopping.is_set():
                 job = self._jobs.peek()
                 difference = job.next_time - time.time()
                 duration = min(difference, self.min_reaction_time)
--- a/sopel/tools/jobs.py
+++ b/sopel/tools/jobs.py
@@ -20,20 +20,14 @@ except ImportError:
     import queue as Queue
 
 
-class released(object):
-    """A context manager that releases a lock temporarily."""
-    def __init__(self, lock):
-        self.lock = lock
-
-    def __enter__(self):
-        self.lock.release()
-
-    def __exit__(self, _type, _value, _traceback):
-        self.lock.acquire()
-
-
 class PriorityQueue(Queue.PriorityQueue):
-    """A priority queue with a peek method."""
+    """A priority queue with a peek method.
+
+    .. deprecated:: 7.0
+
+        This will be removed in Sopel 8. Use Python's built-in
+        :class:`queue.PriorityQueue` instead.
+    """
     def peek(self):
         """Return a copy of the first element without removing it."""
         self.not_empty.acquire()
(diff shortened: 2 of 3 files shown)


Exirel:
Okay... for some reason Py2 don't know about `TimeoutError`:

    ./sopel/bot.py:607:16: F821 undefined name 'TimeoutError'

I've to fix that.

Exirel:
Welp, on the other hand, it should be a `RuntimeError`...

sopel/tools/jobs.py
```diff
@@ -48,53 +55,61 @@ def peek(self):
...
+    X seconds, where X is a property of the Job.
 
-    """
-
-    min_reaction_time = 30.0  # seconds
-    """How often should scheduler checks for changes in the job list."""
+    Thread safety is ensured with an internal mutext.
 
+    It runs forever until the :attr:`stopping` event is set using the
+    :meth`stop` method.
```
dgw: Doesn't this need to be:
```suggestion
    :meth:`stop` method.
```
in order to render properly in Sphinx?

sopel/tools/jobs.py
```diff
@@ -48,53 +55,61 @@ def peek(self):
...
-            self._cleared = True
-            self._jobs = PriorityQueue()
+            self._jobs = []
+
+    def stop(self):
+        """Ask the job scheduler to stop.
+
+        The scheduler thread will stop its loop over jobs to process, but it
+        won't join the thread, or clear its queue - this has to be done
+        separatly by the calling thread.
```
dgw:
```suggestion
        separately by the calling thread.
```

sopel/tools/jobs.py
```diff
@@ -48,53 +55,61 @@ def peek(self):
...
         with self._mutex:
-            self._cleared = True
-            self._jobs = PriorityQueue()
+            self._jobs = []
+
+    def stop(self):
+        """Ask the job scheduler to stop.
+
+        The scheduler thread will stop its loop over jobs to process, but it
+        won't join the thread, or clear its queue - this has to be done
```
dgw: I'm a sucker for real dashes. :grin:
```suggestion
        won't join the thread, or clear its queue—this has to be done
```

sopel/tools/jobs.py
```diff
@@ -48,53 +55,61 @@ def peek(self):
...
+                for job in self._get_ready_jobs(now):
+                    self._run_job(job)
+
+                # Wait up to a second
+                time_spent = time.time() - now
+                wait_time = max(0, 1 - time_spent)
+                if wait_time:
+                    time.sleep(wait_time)
+            except KeyboardInterrupt:
+                # Do not lock on KeyboardInterrupt
```
dgw: This at least helps with the previously-hidden ignored ^C I found when testing #1534, mentioned in https://github.com/sopel-irc/sopel/issues/1478#issuecomment-480043097, right? :+1:

(Note made here instead of next line so I could suggest a change to the comment, as usual…)
```suggestion
                # Do not block on KeyboardInterrupt
```

sopel/tools/jobs.py
```diff
@@ -105,47 +120,34 @@ def run(self):
...
+            self._call(job.func)
+        job.next()
 
     def _call(self, func):
         """Wrapper for collecting errors from modules."""
-        # Sopel.bot.call is way too specialized to be used instead.
         try:
             func(self.bot)
+        except KeyboardInterrupt:
+            # Do not lock on KeyboardInterrupt
```
dgw: (Same as above one.)
```suggestion
            # Do not block on KeyboardInterrupt
```

sopel/tools/jobs.py
```diff
@@ -105,47 +120,34 @@ def run(self):
...
+        return jobs
+
+    def _run_job(self, job):
+        if job.func.thread:
+            t = threading.Thread(
+                target=self._call, args=(job.func,)
+            )
+            t.start()
+        else:
+            self._call(job.func)
```
dgw: Wouldn't it be better to fire off all the threaded jobs first, so they can run while the main thread processes the unthreaded ones? Or the reverse—run all unthreaded jobs, then fire off the threaded ones?

Or maybe this pseudo-random way is better, since firing off a bunch of threaded jobs at once could lead to problems with GIL contention…

No matter what, hopefully threaded jobs are the (vast) majority, since blocking Sopel's main thread for any significant amount of time could be noticeable by users.

sopel/tools/jobs.py
```diff
@@ -48,53 +55,61 @@ def peek(self):
...
-    Thread safety is maintained with a mutex that is released during long
-    operations, so methods add_job and clear_jobs can be safely called from
-    the main thread.
+    X seconds, where X is a property of the Job.
 
-    """
-
-    min_reaction_time = 30.0  # seconds
-    """How often should scheduler checks for changes in the job list."""
+    Thread safety is ensured with an internal mutext.
```
dgw:
```suggestion
    Thread safety is ensured with an internal mutex.
```

sopel/tools/jobs.py
```diff
@@ -179,6 +181,17 @@ def __init__(self, interval, func):
         self.interval = interval
         self.func = func
 
+    def is_ready_to_run(self, now):
+        """Tell if this job is ready to run at ``now`` time.
+
+        :param int now: Timestamp in seconds at which the job might be ready to
+                        run at
```
dgw: Might be clearer to describe it like this:

```
:param int now: Timestamp to check, in seconds
```

Eliminates the duplicate "at"s, and leaves describing the purpose of the function to its description line and return value.

I understand why the timestamp is passed in instead of being fetched dynamically, but _because_ it's passed in, maybe "`now`" isn't the best name for the parameter?

sopel/tools/jobs.py
```diff
@@ -179,6 +181,17 @@ def __init__(self, interval, func):
         self.interval = interval
         self.func = func
 
+    def is_ready_to_run(self, now):
+        """Tell if this job is ready to run at ``now`` time.
```
dgw:
```suggestion
        """Check if this job is ready to run at the given time.
```

sopel/tools/jobs.py
```diff
@@ -33,7 +33,14 @@ def __exit__(self, _type, _value, _traceback):
...
 
 class PriorityQueue(Queue.PriorityQueue):
-    """A priority queue with a peek method."""
+    """A priority queue with a peek method.
+
+    .. deprecated:: 7.0
+
+        This will be removed in Sopel 8. Use Python's built-in priority queue
+        instead.
+
```
dgw: Blank line left here when the others like it in the file were removed? :confused:
```suggestion
```

(Or does leaving out this particular blank line break something in Sphinx?)

sopel/bot.py
```diff
@@ -597,6 +597,20 @@ def _nick_blocked(self, nick):
         return False
 
     def _shutdown(self):
+        self.scheduler.stop()
+
+        stderr(
+            'Stopping the Job Scheduler.'
```
dgw: Would prefer to print the message, then run the action, instead of the current "backwards" ordering.

sopel/test_tools.py
```diff
@@ -22,6 +22,7 @@
 import sopel.config
 import sopel.config.core_section
 import sopel.tools
+import sopel.tools.target
```
dgw: This is done in the `6.6.x` branch already (from #1529), and will cause a merge conflict.

sopel/tools/jobs.py
```diff
@@ -33,7 +33,14 @@ def __exit__(self, _type, _value, _traceback):
 
 
 class PriorityQueue(Queue.PriorityQueue):
-    """A priority queue with a peek method."""
+    """A priority queue with a peek method.
+
+    .. deprecated:: 7.0
+
+        This will be removed in Sopel 8. Use Python's built-in priority queue
```
dgw: Can this "priority queue" link directly to the Python docs for that type (I guess you refer to [`queue.PriorityQueue`](https://docs.python.org/3/library/queue.html#queue.PriorityQueue))?

sopel/tools/jobs.py
Exirel: It should help, but I noticed while testing that it doesn't completly remove the problem. I think there are other parts of the bot that don't like `^C`. I guess our multiple efforts will pay in the end.

sopel/tools/jobs.py
Exirel: I don't have an opinion at the moment. I'd argue that we need more field testing at this point, and get user's feedbacks.

sopel/tools/jobs.py
Exirel: You can clearly see that I wrote that on the train... x)

Exirel:
Thanks for the review @dgw 👍 I'll fix that later this week.

sopel/tools/jobs.py
Exirel: Indeed...

sopel/test_tools.py
Exirel: Not the worst offender. :D

sopel/tools/jobs.py
Exirel: It can, and now it does.

Exirel:
Actually fixed that way earlier than expected. Oh well, I guess it's good for your github merge streak. :wink: :wink: :wink: :wink:

sopel/tools/jobs.py
```diff
@@ -179,6 +168,16 @@ def __init__(self, interval, func):
         self.interval = interval
         self.func = func
 
+    def is_ready_to_run(self, at_time):
+        """Check if this job is ready to run at the given time.
+
+        :param int at_time: Timestamp to check, in seconds
+        :return: ``True`` if the job is ready to run ``at_time`` time,
+                 ``False`` otherwise
```
Exirel: I'm not sure about the grammar of this sentence @dgw

dgw: Hmm, naming the parameter `at_time` makes the grammatically perfect version sound awkward:

```
:return: ``True`` if the job is ready to run at ``at_time``,
         or ``False`` otherwise
```

The double-"at", :roll_eyes:

Also, "is" could become "is (or will be)", but that's getting awfully complicated again. That kind of thing should be reserved for the first line (would become "Check if this job is (or will be) ready to run at the given time.") instead of the return value line.

Exirel:
@dgw while I'm working on #1510 you can review this for the very last time (I hope).

sopel/tools/jobs.py
```diff
@@ -20,20 +20,14 @@
     import queue as Queue
 
 
-class released(object):
```
dgw: I would make noise about marking this as deprecated, but the [changelog for 5.3.0](https://sopel.chat/changelog/5.3.0/) makes this easy:

> * Functions and classes used internally for scheduling jobs are moved. They are still available in their old location, but are no longer documented and will not be considered part of the public API in 6.0.

Since this is part of the job tools, it's not public, so we don't have to go through a deprecation cycle. :tada:

sopel/tools/jobs.py
```diff
@@ -20,20 +20,14 @@
...
-    """A context manager that releases a lock temporarily."""
-    def __init__(self, lock):
-        self.lock = lock
-
-    def __enter__(self):
-        self.lock.release()
-
-    def __exit__(self, _type, _value, _traceback):
-        self.lock.acquire()
+class PriorityQueue(Queue.PriorityQueue):
```
dgw: Per the same 5.3.0 changelog entry, we shouldn't consider this deprecated, as everything in this module is already considered internal since 6.0. Just removing it should be fine.

sopel/tools/jobs.py
Exirel: Oh, that's interesting! I'll add a comment about that in the code, so future devs won't be surprised it's not included in the public documentation.

sopel/tools/jobs.py
Exirel: Good point, thank you!

Exirel:
> Close! clap

Done. :shipit:

sopel/tools/jobs.py
```diff
@@ -1,100 +1,79 @@
 # coding=utf-8
+"""Sopel's Job Scheduler: internal tool for job management.
+
+.. note::
+
+    As of Sopel 5.3, :mod:`sopel.tools.jobs` is an internal tool. Therefore,
+    it is not shown is the public documentation.
```
dgw:
```suggestion
    it is not shown in the public documentation.
```

Also, is the blank line below this required for Sphinx to behave?

sopel/tools/jobs.py
```diff
@@ -1,100 +1,79 @@
...
         with self._mutex:
-            self._cleared = True
-            self._jobs = PriorityQueue()
+            self._jobs = []
+
+    def stop(self):
+        """Ask the job scheduler to stop.
+
+        The scheduler thread will stop its loop over jobs to process, but it
+        won't join the thread, or clear its queue—this has to be done
```
dgw: Apropos your comment about dashes in #1402, this is fine as-is without spaces because it's a docstring. :)

sopel/tools/jobs.py
Exirel: Fixed!

PR #1560, opened 2019-04-15
tools: Add formatter for timedeltas

kwaaak:
datetime.timedelta has a string representation, I don't like how it looks, though.

The most obvious uses for this would be the `seen` and `tell` module.


Exirel:
Since this function is pure (no side-effect), I'd like to see proper unit-testing.

dgw:
@kwaaak I'm ready to approve this as soon as it has proper commit messages. :+1:

You can keep the commits separate, or not. If you merge them, just use the first suggestion:

* tools.time: add human-readable timedelta formatter
* test: add tests for tools' timedelta formatter

kwaaak:
A push and a force, and here we are!

Issue #1318, opened 2018-04-26
Move sopel.web utilities into sopel.tools


dgw:
Since `sopel.web` has been officially deprecated for over two years (and unofficially deprecated for even longer), I think it's time to consider what should actually be removed and what should be kept. The module as it currently exists in `sopel.web` should probably go away in Sopel 7, as we've had plenty of warning that it's deprecated. But some of its functions aren't marked `@deprecated` and do not seem to have direct replacements in `requests` or Python's stdlib.

Specifically, the `quote`, `quote_query`, `iri_to_uri`, and `decode` utility functions seem to have use cases still. I believe it would be best to put those where they seem like they belong: `sopel.tools`. That would get rid of `sopel.web` entirely, but preserve the bits that are not replaced by dropping `requests` in instead. There would, of course, be a transitional period (between the 6.x series and Sopel 7) during which the existing functions would be available in both namespaces.

The names of the new functions in `sopel.tools` are undecided, as I've just thought up this plan tonight. I'm not against having a `sopel.tools.web` subpackage, for example. I welcome input from module developers, who will be affected if staple functions for API interactions, like `sopel.web.quote()`, disappear.

Milestone target and details very much subject to change, as this is still in the conceptual/planning phase.

freeboson:
Also user_agent/default_headers

freeboson:
`sopel.web.quote` is a good example of #1326. It's something that's useful for maintaining 2/3 compatibility, but it's already done by `six`, or could be simplified by dropping python2.

dgw:
I started working on this this weekend. Not quite ready to open a PR for it just yet, but I did write up a plan in sopel-irc/sopel.chat@56a665d68decd3396179ce1cedbb227d6f8e973b, copied here in case that commit eventually disappears (it's on a WIP branch that will probably get rebased several times before going live). The WIP code changes are also pushed here, for those curious enough to scan the branch list.

The tl;dr: Sopel 7 will have both `sopel.web` and `sopel.tools.web`. Stuff we're not keeping stays only in `sopel.web`; new stuff gets added only to `sopel.tools.web`. Renamed stuff has the old name in `sopel.web` and the new name in `sopel.tools.web`, but only one shared implementation/definition (in `sopel.tools.web`). Sopel 8 will do away with `sopel.web` entirely, keeping only `sopel.tools.web`.

----

While the whole `sopel.web` module was marked as deprecated in Sopel
[...]

PR #1563, opened 2019-04-21
tools: Overhaul docstrings

dgw:
It started from noticing some extra blank lines and a few typos, and snowballed into this reasonably large changeset. Sometimes I just can't stop fixing documentation once I start…

This might or might not get added onto. I only touched `tools/__init__.py` so far, and haven't looked at any of the subpackages. (`web` is safe until #1318 gets sorted—that branch is still WIP, but largely done.)

Officially marking the `SopelMemory.contains()` method (and its fraternal twin in `SopelMemoryWithDefault`) as deprecated felt really good. Can't wait to get rid of them for good…

### DIFF
--- a/sopel/tools/__init__.py
+++ b/sopel/tools/__init__.py
@@ -14,11 +14,12 @@
 
 from __future__ import unicode_literals, absolute_import, print_function, division
 
-import sys
+import codecs
+import functools
 import os
 import re
+import sys
 import threading
-import codecs
 import traceback
 from collections import defaultdict
 
@@ -232,14 +233,13 @@ def deprecated(old):
     Any time a decorated function is used, a deprecation warning will be printed
     to the console/log-file.
     """
+    @functools.wraps(old)
     def new(*args, **kwargs):
         print('Function %s is deprecated.' % old.__name__, file=sys.stderr)
         trace = traceback.extract_stack()
         for line in traceback.format_list(trace[:-1]):
             stderr(line[:-1])
         return old(*args, **kwargs)
-    new.__doc__ = old.__doc__
-    new.__name__ = old.__name__
     return new
 
 
--- a/sopel/tools/__init__.py
+++ b/sopel/tools/__init__.py
@@ -42,7 +42,11 @@ _regex_type = type(re.compile(''))
 
 
 def get_input(prompt):
-    """Get decoded input from the terminal (equivalent to python 3's ``input``).
+    """Get decoded input from the terminal (equivalent to Python 3's ``input``).
+
+    :param str prompt: what to display as a prompt on the terminal
+    :return: the user's input
+    :rtype: str
     """
     if sys.version_info.major >= 3:
         return input(prompt)
@@ -51,10 +55,11 @@ def get_input(prompt):
 
 
 def get_raising_file_and_line(tb=None):
-    """Return the file and line number of the statement that raised the tb.
+    """Get the file and line number where an exception happened.
 
-    Returns: (filename, lineno) tuple
-
+    :param tb: the traceback (uses the most recent exception if not given)
+    :return: a tuple of the filename and line number
+    :rtype: (str, int)
     """
     if not tb:
         tb = sys.exc_info()[2]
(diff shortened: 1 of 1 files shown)


sopel/tools/__init__.py
```diff
@@ -197,18 +235,14 @@ def new(*args, **kwargs):
     new.__name__ = old.__name__
     return new
 
-
 # from
 # http://parand.com/say/index.php/2007/07/13/simple-multi-dimensional-dictionaries-in-python/
-# A simple class to make mutli dimensional dict easy to use
+# A simple class to make multi-dimensional dict easy to use
```
Exirel: If it's an above-the-class in-line comment, it can be a docstring for the `Ddict` class below.

sopel/tools/__init__.py
```diff
@@ -379,22 +420,29 @@ def check_pid(pid):
...
-
     """A simple thread-safe dict implementation.
 
-    *Availability: 4.0; available as ``Sopel.SopelMemory`` in 3.1.0 - 3.2.0*
-
     In order to prevent exceptions when iterating over the values and changing
     them at the same time from different threads, we use a blocking lock on
     ``__setitem__`` and ``contains``.
 
+    .. versionadded:: 3.1
```
Exirel: I think it's safe, by now, to remove this information. Since  6.x branch is the only one supported at the moment, anything older than 6.x doesn't need to be mentioned. Keep the `versionchanged` for 6.0, and remove what's from 3.1 and 4.0.

sopel/tools/__init__.py
```diff
@@ -409,21 +457,31 @@ def __setitem__(self, key, value):
...
+            operator, you should be.
+        """
         return self.__contains__(key)
 
 
 class SopelMemoryWithDefault(defaultdict):
-    """Same as SopelMemory, but subclasses from collections.defaultdict."""
+    """Same as SopelMemory, but subclasses from collections.defaultdict.
+
+    .. versionadded:: 4.3
```
Exirel: ditto

sopel/tools/__init__.py
```diff
@@ -437,14 +495,18 @@ def __setitem__(self, key, value):
...
-        It locks it for writes when doing so.
-
+        The dict is locked for writes while doing so.
         """
         self.lock.acquire()
         result = defaultdict.__contains__(self, key)
         self.lock.release()
         return result
 
     def contains(self, key):
```
Exirel: You can add `@deprecated` here too!

dgw: Good catch!

sopel/tools/__init__.py
dgw: Because old modules exist, I disagree on this one point. I made a point of finding out which version added these classes (both `SopelMemory` and `SopelMemoryWithDefault`) and adding the notations here to aid anyone who is trying to get a module written for an older version working again.

Actually, this is the only suggestion you made that I didn't take. :grin:

dgw:
@Exirel Pushed all my changes based on your review without rebasing, so you can see them more easily. Also fixed a related conflict between the `@deprecated` decorator and how Sphinx builds docs for wrapped decorated functions, so their signatures will work now. Just have to figure out how I managed to break decorators I didn't even touch, now…

sopel/tools/__init__.py
Exirel: My last argument [is this video](https://www.youtube.com/watch?v=yXyPvhISkRQ). But I'm pretty sure that won't convince you. 😆 

Don't worry, I think it's reasonable to keep that history.

dgw:
Unrelated decorator breakage sorted, and it turns out we don't need that monkey-patch I added into Sphinx's config in a previous version of this branch. Good to re-review :+1:

sopel/tools/__init__.py
dgw: Nice clip!

I actually streamlined these in the second pass tonight, so they ought to look better in the docs and take up less space now.

Exirel:
🚢

dgw:
Squished. Waiting for Travis…

Issue #1565, opened 2019-04-22
Modernize docstrings & make everything consistent


dgw:
This whole initiative started with #1563, but I only dealt with one file in that PR.

The whole public API (at _least_) should get an overhaul before we release 7.0, to make sure the API docs are really nice and easy to use. We only use Sphinx magic like `:param:` descriptions in a few places so far, and it would be **awesome** to have those everywhere.

Here's the important (public) stuff:

  - [x] `sopel.bot` (#1772)
  - [x] `sopel.cli` (#1922)
    - Probably already modern given that @Exirel wrote it, but ~~I'll still go~~ @Exirel went through it again
    - Review/overhaul pushed to 7.1
  - [x] `sopel.config` (#1778)
    - [x] `sopel.config.core_section` (#1778)
    - [x] `sopel.config.types` (#1719)
  - [x] `sopel.db` (#1776)
  - [x] `sopel.formatting` (#1766)
  - [x] `sopel.irc` (#1772)
    - [x] `sopel.irc.abstract_backends` (#1772)
    - [x] `sopel.irc.backends` (#1772)
    - [x] `sopel.irc.utils` (#1772)
  - [x] `sopel.logger` (#1775)
  - [x] `sopel.module` (#1770)
  - [x] `sopel.tools` (#1563)
    - [x] `sopel.tools.calculation` (#1771)
    - [x] `sopel.tools.jobs` (#1771)
    - [x] `sopel.tools.target` (#1771)
    - [x] `sopel.tools.time` (#1771)
    - [x] `sopel.tools.web` (moved from `sopel.web` in #1616; doc updates in #1669)
  - [x] `sopel.trigger` (#1782)

This stuff is more internal, and it's not as important to get done in time for 7.0, but it would still be great to have nice, consistent docs for at some point:

  - [x] `sopel.coretasks` (#2001)
    - Proper documentation of everything in here would help core development immensely, even if it's not published in the public API docs
  - [x] `sopel.loader` (#2004)
  - [x] `sopel.plugins` (#2002)
    - [x] `sopel.plugins.handlers` (#2002)
    - [x] `sopel.plugins.exceptions` (already done, checked by #2002)
  - [x] `sopel.test_tools` (deprecated, see #2003)

[...]

PR #1566, opened 2019-04-24
First pass of docstring additions.

nsnw:
Hey! With reference to #1565, I've started tidying up some of the docstrings in `bot.py`. I'm happy to help do others as well, but this is just to check that this is the kind of thing you're looking for. Feedback is obviously requested and will be gladly received!

### DIFF
--- a/sopel/bot.py
+++ b/sopel/bot.py
@@ -65,20 +65,26 @@ class Sopel(irc.Bot):
         self._plugins = {}
         self.config = config
         """The :class:`sopel.config.Config` for the current Sopel instance."""
+
         self.doc = {}
+        """A dictionary of command names to their documentation.
+
+        Each command is mapped to its docstring and any available examples, if
+        declared in the module's code.
+
+        .. versionchanged:: 3.2
+            Use the first item in each callable's commands list as the key,
+            instead of the function name as declared in the source code.
         """
-        A dictionary of command names to their docstring and example, if
-        declared. The first item in a callable's commands list is used as the
-        key in version *3.2* onward. Prior to *3.2*, the name of the function
-        as declared in the source code was used.
-        """
+
         self._command_groups = collections.defaultdict(list)
         """A mapping of module names to a list of commands in it."""
+
         self.stats = {}  # deprecated, remove in 7.0
         self._times = {}
         """
-        A dictionary mapping lower-case'd nicks to dictionaries which map
-        funtion names to the time which they were last used by that nick.
+        A dictionary mapping lowercased nicks to dictionaries which map
+        function names to the time which they were last used by that nick.
         """
 
         self.server_capabilities = {}
@@ -88,14 +94,17 @@ class Sopel(irc.Bot):
         it will be here as ``{"sasl": "EXTERNAL"}``. Capabilities specified
         without any options will have ``None`` as the value.
 
-        For servers that do not support IRCv3, this will be an empty set."""
+        For servers that do not support IRCv3, this will be an empty set.
+        """
+
         self.enabled_capabilities = set()
         """A set containing the IRCv3 capabilities that the bot has enabled."""
+
         self._cap_reqs = dict()
-        """A dictionary of capability names to a list of requests"""
+        """A dictionary of capability names to a list of requests."""
 
         self.privileges = dict()
-        """A dictionary of channels to their users and privilege levels
+        """A dictionary of channels to their users and privilege levels.
 
         The value associated with each channel is a dictionary of
         :class:`sopel.tools.Identifier`\\s to
@@ -109,17 +118,18 @@ class Sopel(irc.Bot):
         self.channels = tools.SopelMemory()  # name to chan obj
         """A map of the channels that Sopel is in.
 
-        The keys are Identifiers of the channel names, and map to
-        :class:`sopel.tools.target.Channel` objects which contain the users in
-        the channel and their permissions.
+        The keys are :class:`sopel.tools.Identifier`\\s of the channel names,
+        and map to :class:`sopel.tools.target.Channel` objects which contain
+        the users in the channel and their permissions.
         """
+
         self.users = tools.SopelMemory()  # name to user obj
         """A map of the users that Sopel is aware of.
 
-        The keys are Identifiers of the nicknames, and map to
-        :class:`sopel.tools.target.User` instances. In order for Sopel to be
-        aware of a user, it must be in at least one channel which they are also
-        in.
+        The keys are :class:`sopel.tools.Identifier`\\s of the nicknames, and
+        map to :class:`sopel.tools.target.User` instances. In order for Sopel
+        to be aware of a user, it must be in at least one channel which they
+        are also in.
         """
 
         self.db = SopelDB(config)
@@ -128,11 +138,11 @@ class Sopel(irc.Bot):
         self.memory = tools.SopelMemory()
         """
         A thread-safe dict for storage of runtime data to be shared between
-        modules. See :class:`sopel.tools.Sopel.SopelMemory`
+        modules. See :class:`sopel.tools.SopelMemory`.
         """
 
         self.shutdown_methods = []
-        """List of methods to call on shutdown"""
+        """List of methods to call on shutdown."""
 
         self.scheduler = sopel.tools.jobs.JobScheduler(self)
         self.scheduler.start()
@@ -147,7 +157,10 @@ class Sopel(irc.Bot):
 
     @property
     def hostmask(self):
-        """str: the current hostmask for the bot :class:`sopel.tools.target.User`
+        """The current hostmask for the bot :class:`sopel.tools.target.User`.
+
+        :return: the bot's current hostmask
+        :rtype: str
 
         Bot must be connected and in at least one channel.
         """
@@ -164,6 +177,11 @@ class Sopel(irc.Bot):
     def write(self, args, text=None):  # Shim this in here for autodocs
         """Send a command to the server.
 
+        :param args: an iterable of strings, which will be joined by spaces
+        :type args: :term:`iterable`
+        :param str text: a string that will be prepended with a ``:`` and added
+                         to the end of the command
+
         ``args`` is an iterable of strings, which are joined by spaces.
         ``text`` is treated as though it were the final item in ``args``, but
         is preceeded by a ``:``. This is a special case which  means that
@@ -174,13 +192,14 @@ class Sopel(irc.Bot):
         and ``sopel.write(('PRIVMSG', ':Hello, world!'))`` will send
         ``PRIVMSG :Hello, world!`` to the server.
 
-        Newlines and carriage returns ('\\n' and '\\r') are removed before
-        sending. Additionally, if the message (after joining) is longer than
-        than 510 characters, any remaining characters will not be sent.
+        Newlines and carriage returns (``'\\n'`` and ``'\\r'``) are removed
+        before sending. Additionally, if the message (after joining) is longer
+        than than 510 characters, any remaining characters will not be sent.
         """
         irc.Bot.write(self, args, text=text)
 
     def setup(self):
+        """Set up the Sopel instance."""
         load_success = 0
         load_error = 0
         load_disabled = 0
@@ -305,6 +324,11 @@ class Sopel(irc.Bot):
         return name in self._plugins
 
     def unregister(self, obj):
+        """Unregister a callable.
+
+        :param obj: the callable to unregister
+        :type obj: :term:`object`
+        """
         if not callable(obj):
             return
         if hasattr(obj, 'rule'):  # commands and intents have it added
@@ -321,6 +345,17 @@ class Sopel(irc.Bot):
             self.shutdown_methods.remove(obj)
 
     def register(self, callables, jobs, shutdowns, urls):
+        """Register a callable.
+
+        :param callables: an iterable of callables to register
+        :type callables: :term:`iterable`
+        :param jobs: an iterable of functions to periodically invoke
+        :type jobs: :term:`iterable`
+        :param shutdowns: an iterable of functions to call on shutdown
+        :type shutdowns: :term:`iterable`
+        :param urls: an iterable of functions to call when matched against a URL
+        :type urls: :term:`iterable`
+        """
         # Append module's shutdown function to the bot's list of functions to
         # call on shutdown
         self.shutdown_methods += shutdowns
@@ -347,16 +382,23 @@ class Sopel(irc.Bot):
             self.register_url_callback(func.url_regex, func)
 
     def part(self, channel, msg=None):
-        """Part a channel."""
+        """Leave a channel.
+
+        :param str channel: the channel to leave
+        :param str msg: the message to display when leaving a channel
+        """
         self.write(['PART', channel], msg)
 
     def join(self, channel, password=None):
-        """Join a channel
+        """Join a channel.
 
-        If `channel` contains a space, and no `password` is given, the space is
-        assumed to split the argument into the channel to join and its
-        password.  `channel` should not contain a space if `password` is given.
+        :param str channel: the channel to join
+        :param str password: an optional channel password
 
+        If ``channel`` contains a space, and no ``password`` is given, the
+        space is assumed to split the argument into the channel to join and its
+        password. ``channel`` should not contain a space if ``password``
+        is given.
         """
         if password is None:
             self.write(('JOIN', channel))
@@ -372,7 +414,12 @@ class Sopel(irc.Bot):
         self.say(text, recipient, max_messages)
 
     def say(self, text, recipient, max_messages=1):
-        """Send ``text`` as a PRIVMSG to ``recipient``.
+        """Send a PRIVMSG to a user or channel.
+
+        :param str text: the text to send
+        :param str recipient: the message recipient
+        :param int max_messages: the maximum number of messages to break the
+                                 text into
 
         In the context of a triggered callable, the ``recipient`` defaults to
         the channel (or nickname, if a private message) from which the message
@@ -450,7 +497,10 @@ class Sopel(irc.Bot):
             self.say(excess, max_messages - 1, recipient)
 
     def notice(self, text, dest):
-        """Send an IRC NOTICE to a user or a channel.
+        """Send an IRC NOTICE to a user or channel.
+
+        :param str text: the text to send in the NOTICE
+        :param str dest: the destination of the NOTICE
 
         Within the context of a triggered callable, ``dest`` will default to
         the channel (or nickname, if a private message), in which the trigger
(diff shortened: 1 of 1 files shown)


sopel/bot.py
```diff
@@ -174,10 +183,16 @@ def write(self, args, text=None):  # Shim this in here for autodocs
         Newlines and carriage returns ('\\n' and '\\r') are removed before
         sending. Additionally, if the message (after joining) is longer than
         than 510 characters, any remaining characters will not be sent.
+
+        :param collections.Iterable args: an iterable of strings, which will be joined by spaces
```
dgw: You'd need to use `:type:` lines also, but it's possible to link things like this to the Python glossary:

```
:param args: command arguments, which will be joined by spaces
:type args: :term:`iterable`
```

That produces the type hint "([iterable](https://docs.python.org/3/glossary.html#term-iterable))" for `args`, linking to the term in Python's own docs. Feel free to use this trick all over the place! There are many references to such things in this PR.

dgw:
I had all but given up trying to figure out how to link `re.Pattern` as an example, so I left it out of my review.

But! This worked when I tweaked the current PR state locally and built the HTML docs:

```
    def register_url_callback(self, pattern, callback):
        """Register a ``callback`` for URLs matching the regex ``pattern``

        :param pattern: compiled regex pattern to register
        :type pattern: :ref:`re.Pattern <python:re-objects>`
        :param function callback: callable object to handle matching URLs
[... etc ...]
```

`pattern`'s type then links to https://docs.python.org/3/library/re.html#re-objects by doing this, which is precisely the section needed. Why it's not linkable as `#re.Pattern` despite describing what appear to be objects of that type, I don't know. But this is _a_ solution to the "More intersphinx links" point I requested.

nsnw:
Thanks for the feedback! I'll make a start on incorporating the feedback :-D

sopel/bot.py
```diff
@@ -25,7 +25,6 @@
 from sopel.logger import get_logger
 import sopel.loader
 
-
```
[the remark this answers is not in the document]
nsnw: Not actually sure how I ended up deleting that line - I've put it back in the latest revision!

sopel/bot.py
```diff
@@ -463,6 +519,11 @@ def call(self, func, sopel, trigger):
                 self._times[trigger.sender][func] = current_time
 
     def dispatch(self, pretrigger):
+        """Match an incoming, parsed message from the server and dispatch it to any registered
+        callables.
```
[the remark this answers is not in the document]
nsnw: Yeah, it was clumsy wording - I've updated it with your suggestion.

sopel/bot.py
```diff
@@ -263,7 +291,11 @@ def register(self, callables, jobs, shutdowns, urls):
             self.register_url_callback(func.url_regex, func)
 
     def part(self, channel, msg=None):
-        """Part a channel."""
+        """Part a channel.
+
+        :param str channel: the channel to leave
+        :param str msg: the message to display when leaving a channel
```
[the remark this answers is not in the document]
nsnw: I settled on **leave** (even though the method is `part`) as it read better - updated.

sopel/bot.py
```diff
@@ -237,6 +256,15 @@ def unregister(self, obj):
...
     def register(self, callables, jobs, shutdowns, urls):
+        """Register a callable.
+
+        :param collections.Iterable callables: an iterable of callables to register
+        :param collections.Iterable jobs: an iterable of functions to periodically invoke
+        :param collections.Iterable shutdowns: an iterable of functions to call on shutdown
+        :param collections.Iterable urls: an iterable of functions to call when matched against a
+            URL
+        """
+
```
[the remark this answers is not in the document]
nsnw: Noted!

sopel/bot.py
nsnw: Thanks for this hint - I'm reasonably new to Sphinx so this is a learning experience. I've used `:term:` in a couple of places in the latest revision!

nsnw:
I've incorporated the feedback, and:-

* docstrings are now limited to 80 characters wide
* moved parameter definitions to the top of the docstrings (under the initial one-line description)
* switching to hanging indentation for line-wrapped parameter definitions
* added glossary links for concepts (using `:term:` as suggested)
* removed blank newlines between docstring definitions and method bodies

sopel/bot.py
```diff
@@ -438,12 +447,15 @@ def reply(self, text, dest, reply_to, notice=False):
...
 
     def call(self, func, sopel, trigger):
-        """Call a function, after checking whether the function is restricted or subject to limits.
-
-        :param function func: the function to call
-        :param SopelWrapper sopel: a SopelWrapper instance
-        :param Trigger trigger: the Trigger object for the line from the server that triggered
-            this call
+        """Call a function (``func``), after checking if the function is
+        restricted or subject to limits.
```
dgw: As above, no need to point out the parameter name in the short description when we have param lists.

Would also really like to keep short descriptions to one line only. Can't make suggestions that apply to multiple lines, so apply this (and tweak it, if you have a better idea) however you may manually:

```
    """Call a function, applying any rate-limits or restrictions.
```

sopel/bot.py
```diff
@@ -519,8 +531,7 @@ def call(self, func, sopel, trigger):
                 self._times[trigger.sender][func] = current_time
 
     def dispatch(self, pretrigger):
-        """Match an incoming, parsed message from the server and dispatch it to any registered
-        callables.
+        """Dispatch a parsed message to any registered callables
```
dgw: I think we're trying to be consistent about having `.` at the end of these short ones? Maybe?

Even if we aren't yet, we probably should be.

sopel/bot.py
```diff
@@ -387,35 +392,44 @@ def say(self, text, recipient, max_messages=1):
...
 
     def notice(self, text, dest):
-        """Send an IRC NOTICE to a user or a channel.
+        """Send an IRC NOTICE (of ``text``) to a user or a channel (``dest``).
 
-        Within the context of a triggered callable, ``dest`` will default to
-        the channel (or nickname, if a private message), in which the trigger
-        happened.
+        :param str text: the text to send in the NOTICE
+        :param str dest: the destination of the NOTICE
```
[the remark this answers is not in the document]
nsnw: Whoops - missed that one when doing a local merge!

sopel/bot.py
```diff
@@ -387,35 +392,44 @@ def say(self, text, recipient, max_messages=1):
             self.msg(recipient, excess, max_messages - 1)
 
     def notice(self, text, dest):
-        """Send an IRC NOTICE to a user or a channel.
+        """Send an IRC NOTICE (of ``text``) to a user or a channel (``dest``).
```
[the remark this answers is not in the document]
nsnw: Updated.

sopel/bot.py
```diff
@@ -387,35 +392,44 @@ def say(self, text, recipient, max_messages=1):
...
         self.say('\001ACTION {}\001'.format(text), dest)
 
     def reply(self, text, dest, reply_to, notice=False):
         """Prepend ``reply_to`` to ``text``, and send as a PRIVMSG to ``dest``.
 
+        :param str text: the text of the reply
+        :param str dest: the destination of the reply
+        :param str reply_to: the nickname that the reply will be prepended with
+        :param bool notice: whether to send the reply as a NOTICE or not,
+                            defaults to False
```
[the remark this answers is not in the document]
nsnw: Done!

sopel/bot.py
nsnw: Done!

sopel/bot.py
nsnw: That one slipped through - added.

sopel/bot.py
```diff
@@ -387,35 +392,44 @@ def say(self, text, recipient, max_messages=1):
...
         the channel (or nickname, if a private message), in which the trigger
         happened.
-
-        :param str text: the text to send in the CTCP ACTION
-        :param str dest: the destination of the CTCP ACTION
         """
         self.say('\001ACTION {}\001'.format(text), dest)
 
     def reply(self, text, dest, reply_to, notice=False):
         """Prepend ``reply_to`` to ``text``, and send as a PRIVMSG to ``dest``.
```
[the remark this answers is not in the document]
nsnw: Assuming you're referring to 407-409, I've tried to think of an alternative way to write it and I can't think of anything that is that much shorter. Alternatively, maybe it's better to not mention it at all?

sopel/bot.py
```diff
@@ -372,6 +421,12 @@ def action(self, text, dest):
...
 
+        :param str text: the text of the reply
+        :param str dest: the destination of the reply
+        :param str reply_to: the nickname that the reply will be prepended with
+        :param bool notice: whether to send the reply as a NOTICE or not,
+                            defaults to `False`
```
dgw: We need the double backticks, actually. Single backticks produce the wrong output (usually italics).

```suggestion
                            defaults to ``False``
```

sopel/bot.py
```diff
@@ -351,14 +394,20 @@ def say(self, text, recipient, max_messages=1):
...
+
         Within the context of a triggered callable, ``dest`` will default to
         the channel (or nickname, if a private message), in which the trigger
         happened.
         """
         self.write(('NOTICE', dest), text)
 
     def action(self, text, dest):
-        """Send ``text`` as a CTCP ACTION PRIVMSG to ``dest``.
+        """Send a CTCP ACTION PRIVMSG to a user or a channel.
```
```suggestion
        """Send a CTCP ACTION PRIVMSG to a user or channel.
```

Nitpicking, only because I already have other comments to make in this review, and I noticed inconsistencies.

sopel/bot.py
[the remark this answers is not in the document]
dgw: I actually do mean this specific line. Let me take a stab at it, to give you at least a starting point maybe?

```suggestion
        """Send a PRIVMSG addressing someone.
```

Could end this with "to a user or channel" like the above `action` one, but probably not needed. In fact, I think I'll suggest removing most of those.

Might have shortened it a bit _too_ much, but it does eliminate naming the params… Hmm.

sopel/bot.py
```diff
@@ -351,14 +394,20 @@ def say(self, text, recipient, max_messages=1):
     def notice(self, text, dest):
         """Send an IRC NOTICE to a user or a channel.
```
dgw: Matching below.

```suggestion
        """Send an IRC NOTICE to a user or channel.
```

sopel/bot.py
```diff
@@ -263,16 +295,22 @@ def register(self, callables, jobs, shutdowns, urls):
...
-        """Part a channel."""
+        """Leave a channel.
+
+        :param str channel: the channel to leave
+        :param str msg: the message to display when leaving a channel
+        """
         self.write(['PART', channel], msg)
 
     def join(self, channel, password=None):
         """Join a channel
```
```suggestion
        """Join a channel.
```

Didn't notice the missing `.` before. :/

sopel/bot.py
```diff
@@ -144,7 +151,10 @@ def __init__(self, config, daemon=False):
 
     @property
     def hostmask(self):
-        """str: the current hostmask for the bot :class:`sopel.tools.target.User`
+        """The current hostmask for the bot :class:`sopel.tools.target.User`
```
```suggestion
        """The current hostmask for the bot :class:`sopel.tools.target.User`.
```

sopel/bot.py
```diff
@@ -286,6 +324,11 @@ def msg(self, recipient, text, max_messages=1):
     def say(self, text, recipient, max_messages=1):
         """Send ``text`` as a PRIVMSG to ``recipient``.
```
dgw: Overlooked this technical-soup one the other day. Sorry about that! Probably because you didn't touch it, and I was focused on the changed/added lines.

```suggestion
        """Send a PRIVMSG to a user or channel.
```

nsnw:
You're not being too demanding!

I've pushed the latest version. I've added the "to a user or channel" wording to all the ones that it makes for now. I personally can't decide which is better - with or without it - but I'll leave it to you to decide.

sopel/bot.py
```diff
@@ -419,13 +419,14 @@ def action(self, text, dest):
         self.say('\001ACTION {}\001'.format(text), dest)
 
     def reply(self, text, dest, reply_to, notice=False):
-        """Prepend ``reply_to`` to ``text``, and send as a PRIVMSG to ``dest``.
+        """Send a PRIVMSG to a user or channel, optionally prepended with a
```
dgw: Thing is, the text to prepend is _not_ optional. The argument has no default value, and is therefore required.

dgw:
> You're not being too demanding!

That's a relief! All the same, I thank you for putting up with the lengthy iteration process. :smile_cat:

> I've added the "to a user or channel" wording to all the ones that it makes for now. I personally can't decide which is better - with or without it - but I'll leave it to you to decide.

Obviously, I choose to keep "to a user or channel" in places where it makes sense to include, but without the second "a" that (to me) makes the phrase slightly redundant.

sopel/bot.py
```diff
@@ -60,17 +60,21 @@ def __init__(self, config, daemon=False):
...
         self.config = config
         """The :class:`sopel.config.Config` for the current Sopel instance."""
+
         self.doc = {}
         """
         A dictionary of command names to their docstring and example, if
         declared. The first item in a callable's commands list is used as the
         key in version *3.2* onward. Prior to *3.2*, the name of the function
         as declared in the source code was used.
         """
```
dgw: This docstring could be rewritten to use Sphinx notations like `.. versionchanged:: 3.2`, with the notes below that. For example:

```py
        """A dictionary of command names to their documentation.

        Each command is mapped to its docstring and any available examples, if
        declared in the module's code.

        .. versionchanged:: 3.2
            Use the first item in each callable's commands list as the key,
            instead of the function name as declared in the source code.
        """
```

sopel/bot.py
```diff
@@ -110,6 +116,7 @@ def __init__(self, config, daemon=False):
         :class:`sopel.tools.target.Channel` objects which contain the users in
         the channel and their permissions.
         """
+
```
dgw: Sigh, more out-of-reach lines.

Both above and below here, there are mentions of `Identifiers` that could be linked to <tt>:class:\`sopel.tools.Identifier\`</tt>

Below, there are also two docstrings (for `memory` and `shutdown_methods`) that should have trailing `.` added.

sopel/bot.py
```diff
@@ -677,6 +759,7 @@ def unregister_url_callback(self, pattern):
         """Unregister the callback for URLs matching the regex ``pattern``
```
```suggestion
        """Unregister the callback for URLs matching the regex ``pattern``.
```

sopel/bot.py
```diff
@@ -646,7 +726,9 @@ def register_url_callback(self, pattern, callback):
         """Register a ``callback`` for URLs matching the regex ``pattern``
```
```suggestion
        """Register a ``callback`` for URLs matching the regex ``pattern``.
```

sopel/bot.py
```diff
@@ -263,16 +295,22 @@ def register(self, callables, jobs, shutdowns, urls):
...
         self.write(['PART', channel], msg)
 
     def join(self, channel, password=None):
-        """Join a channel
+        """Join a channel.
+
+        :param str channel: the channel to join
+        :param str password: an optional channel password
 
         If `channel` contains a space, and no `password` is given, the space is
```
dgw: Figured it out. Single-backtick gives us `<cite>text</cite>`, which is grossly incorrect for these uses both semantically and visually (it renders in italicized prose font). Using double-backticks yields (I'm simplifying) `<code>text</code>`, which is what we want (renders as monospace).

```suggestion
        If ``channel`` contains a space, and no ``password`` is given, the space is
```

sopel/bot.py
```diff
@@ -263,16 +295,22 @@ def register(self, callables, jobs, shutdowns, urls):
...
         assumed to split the argument into the channel to join and its
         password.  `channel` should not contain a space if `password` is given.
```
```suggestion
        password. ``channel`` should not contain a space if ``password`` is given.
```

sopel/bot.py
```diff
@@ -578,8 +647,19 @@ def cap_req(self, module_name, capability, arg=None, failure_callback=None,
...
+                               with ``+`` or ``=``
+        :param str arg: arguments for the capability request
+        :param failure_callback: a function that will be called if the
+                                 capability request fails
+        :type failure_callback: :term:`function`
+        :param success_callback: a function that will be called if the
+                                 capability is successfully requested
+        :type success_callback: :term:`function`
+
         By prefixing the capability with `-`, it will be ensured that the
```
dgw: Some of them are out of line-note range (GitHub is really pissing me off tonight with that limitation), but every instance of <tt>\`thing\`</tt> in this docstring needs to be replaced with <tt>\`\`thing\`\`</tt>. It's the only place where single-backticks are used (based on searching the generated HTML for incorrect occurrences of `<cite>`) other than the `join` method docstring.

I would also appreciate a general cleanup of any remaining double-spaces after periods, as occurs on what is currently line 671 of the HEAD revision (but that no doubt will have changed by the time you address this note :sweat_smile:)!

nsnw:
I am still working on this! I've just been busy with work. Hoping to get this finished in the next few days.

Exirel:
Just a quick note to say: I love it! :+1: 
Also you'll have to rebase and resolve some conflict. And the more PR get merged, the more conflicts you'll have to fix.

nsnw:
Ok! Sorry for the delay - currently in the process of buying a house so things got in the way -_-

In the latest commit I've implemented the suggestions, including:-
* switching single backticks to double backticks where appropriate - a scan of the generated HTML after this doesn't show any instances of `<cite>`/`</cite>` 
* added periods at the end of sentences - I think I've now caught them all
* removed 2 instances of double spaces after a period

There were a couple of other bits I've found as well that I fixed.

nsnw:
I've merged in the latest changes from `master`, so this should be current against those.

sopel/bot.py
```diff
@@ -186,9 +191,10 @@ def write(self, args, text=None):  # Shim this in here for autodocs
         and ``sopel.write(('PRIVMSG', ':Hello, world!'))`` will send
         ``PRIVMSG :Hello, world!`` to the server.
 
-        Newlines and carriage returns ('\\n' and '\\r') are removed before
-        sending. Additionally, if the message (after joining) is longer than
-        than 510 characters, any remaining characters will not be sent.
+        Newlines and carriage returns (``'\\n'`` and ``'\\r'``) are
+        removed before sending. Additionally, if the message (after joining) is
+        longer than than 510 characters, any remaining characters will not be
+        sent.
```
dgw: Rewrapping this paragraph should be able to eliminate this widow.

sopel/bot.py
```diff
@@ -66,10 +66,14 @@ def __init__(self, config, daemon=False):
 
         self.doc = {}
         """
-        A dictionary of command names to their docstring and example, if
-        declared. The first item in a callable's commands list is used as the
-        key in version *3.2* onward. Prior to *3.2*, the name of the function
-        as declared in the source code was used.
+        A dictionary of command names to their documentation.
```
dgw: This should go on the same line as `"""`.

nsnw:
Excellent! Been a bit busy (getting married in a few weeks!) but I'll make the tweaks above and tidy up the PR. Once that's done I'll start looking at others :-)

Exirel:
@nsnw that's lovely! :heart_eyes:

nsnw:
Okay - I believe I've squashed everything down to a single commit now. I have to admit that squashing and rebasing is not my strong point so if I've butchered it I apoplogise!

PR #1567, opened 2019-04-24
style: replace deprecated `SopelMemory.contains()` with `in`

HumorBaby:
The title says it all!

The flood of deprecation warnings from `.contains()` was quite unpleasant.


dgw:
So much output for so few actual calls to the deprecated methods. Oy.

dgw:
Future note: #1479 reintroduced a single instance of `contains()`, fixed by #1685.

PR #1569, opened 2019-04-24
modules: always safely set up/clear memory

HumorBaby:
Modules will safely initialize (ensure key does not exist in memory already) in `setup()` and clean up (remove module-specific keys from memory) in `shutdown()`, where necessary.

### DIFF
--- a/sopel/modules/find.py
+++ b/sopel/modules/find.py
@@ -20,7 +20,15 @@ from sopel.formatting import bold
 
 
 def setup(bot):
-    bot.memory['find_lines'] = SopelMemory()
+    if 'find_lines' not in bot.memory:
+        bot.memory['find_lines'] = SopelMemory()
+
+
+def shutdown(bot):
+    try:
+        del bot.memory['find_lines']
+    except KeyError:
+        pass
 
 
 @echo
--- a/sopel/modules/safety.py
+++ b/sopel/modules/safety.py
@@ -81,7 +81,8 @@ def configure(config):
 def setup(bot):
     bot.config.define_section('safety', SafetySection)
 
-    bot.memory['safety_cache'] = sopel.tools.SopelMemory()
+    if 'safety_cache' not in bot.memory:
+        bot.memory['safety_cache'] = sopel.tools.SopelMemory()
     for item in bot.config.safety.known_good:
         known_good.append(re.compile(item, re.I))
 
@@ -97,6 +98,13 @@ def setup(bot):
             clean_line = unicode(line).strip().lower()
             if clean_line != '':
                 malware_domains.add(clean_line)
+
+
+def shutdown(bot):
+    try:
+        del bot.memory['safety_cache']
+    except KeyError:
+        pass
 
 
 def _download_malwaredomains_db(path):
--- a/sopel/modules/tell.py
+++ b/sopel/modules/tell.py
@@ -76,8 +76,18 @@ def setup(bot):
         else:
             f.write('')
             f.close()
-    bot.memory['tell_lock'] = threading.Lock()
-    bot.memory['reminders'] = loadReminders(bot.tell_filename, bot.memory['tell_lock'])
+    if 'tell_lock' not in bot.memory:
+        bot.memory['tell_lock'] = threading.Lock()
+    if 'reminders' not in bot.memory:
+        bot.memory['reminders'] = loadReminders(bot.tell_filename, bot.memory['tell_lock'])
+
+
+def shutdown(bot):
+    for key in ['tell_lock', 'reminders']:
+        try:
+            del bot.memory[key]
+        except KeyError:
+            pass
 
 
 @commands('tell', 'ask')
--- a/sopel/modules/tell.py
+++ b/sopel/modules/tell.py
@@ -65,19 +65,19 @@ def dumpReminders(fn, data, lock):
     return True
 
 
-def setup(self):
-    fn = self.nick + '-' + self.config.core.host + '.tell.db'
-    self.tell_filename = os.path.join(self.config.core.homedir, fn)
-    if not os.path.exists(self.tell_filename):
+def setup(bot):
+    fn = bot.nick + '-' + bot.config.core.host + '.tell.db'
+    bot.tell_filename = os.path.join(bot.config.core.homedir, fn)
+    if not os.path.exists(bot.tell_filename):
         try:
-            f = open(self.tell_filename, 'w')
-        except (OSError, IOError):  # Remove IOError when dropping py2 support
+            f = open(bot.tell_filename, 'w')
+        except (OSError, IOError):  # TODO: Remove IOError when dropping py2 support
             pass
         else:
             f.write('')
             f.close()
-    self.memory['tell_lock'] = threading.Lock()
-    self.memory['reminders'] = loadReminders(self.tell_filename, self.memory['tell_lock'])
+    bot.memory['tell_lock'] = threading.Lock()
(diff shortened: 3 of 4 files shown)


HumorBaby:
Thinking more about the word `safely`, I am brought to the question: 

Wouldn't it be _safer_ to `raise` something in `setup()` if a key is already present (where applicable)? Presumably `shutdown()` should remove keys, and a fresh load should not have the keys already; thus, the only time a key already exists would be an unwanted collision...

dgw:
It depends. Sometimes it's OK to leave stuff in `bot.memory` so it can persist across module reloads (since we actually run `shutdown()` on unload now, IIRC). Things like the short-URL cache from `url.py` come to mind, where we don't necessarily want to have the bot hit the API again if it sees the same URL before & after reload. Basically, I think that just blanket-adding all of these might not be the best approach.

Let's see if I can come up with pros/cons for each of these.

HumorBaby:
hmmm, I was worried about that. I will rack my brain also to see where it makes sense to persist memory.

sopel/modules/find.py
```diff
@@ -20,7 +20,15 @@
...
-    bot.memory['find_lines'] = SopelMemory()
+    if 'find_lines' not in bot.memory:
+        bot.memory['find_lines'] = SopelMemory()
+
+
+def shutdown(bot):
+    try:
+        del bot.memory['find_lines']
+    except KeyError:
+        pass
```
dgw: This one might make sense to delete, because if the plugin misses any lines while it's inactive it could behave unexpectedly the next time someone tries to make a substitution.

sopel/modules/help.py
```diff
@@ -160,6 +160,13 @@ def setup(bot):
     bot.config.define_section('help', HelpSection)
 
 
+def shutdown(bot):
+    try:
+        del bot.memory['command-list']
+    except KeyError:
+        pass
```
dgw: I'm not sure about deleting this one. The cache validation logic should take care of any settings changes or command additions/removals that would require regenerating the command list.

sopel/modules/ipython.py
```diff
@@ -39,6 +39,13 @@
 console = None
 
 
+def shutdown(bot):
+    try:
+        del bot.memory['iconsole_running']
+    except KeyError:
+        pass
```
dgw: Worth looking into whether the console actually stops if the module is unloaded (and then reloaded, or not) while the console is active. If it's not stopped, then this should either not remove the "running" flag from memory, or it should terminate the console at `shutdown()`.

sopel/modules/safety.py
```diff
@@ -99,6 +100,13 @@ def setup(bot):
                 malware_domains.add(clean_line)
 
 
+def shutdown(bot):
+    try:
+        del bot.memory['safety_cache']
+    except KeyError:
+        pass
```
dgw: Probably OK to clear this I guess. It shouldn't cause that much repeated work.

After tracing through how the cache works though, either I don't understand `_clean_cache()` at all or it's almost completely useless and doesn't do what it should. Feel free to look at that if you get biologist's block again. 😉

sopel/modules/tell.py
```diff
@@ -65,7 +65,7 @@ def dumpReminders(fn, data, lock):
     return True
 
 
-def setup(self):
+def setup(self):  # TODO: `self`?!? but why!? change to `bot` eventually...
```
dgw: Just change it now, tbh. Separate commit if you like.

Basically, either fix it (and don't leave a comment) or remove the comment (because it's unrelated to the PR).

sopel/modules/tell.py
```diff
@@ -76,8 +76,18 @@ def setup(self):
...
+    if 'reminders' not in self.memory:
+        self.memory['reminders'] = loadReminders(self.tell_filename, self.memory['tell_lock'])
+
+
+def shutdown(bot):
+    for key in ['tell_lock', 'reminders']:
+        try:
+            del bot.memory[key]
+        except KeyError:
+            pass
```
dgw: Yep. Reloading this module should purge ALL the things and reload them from the reminders file on disk. 👍

sopel/modules/uptime.py
```diff
@@ -18,6 +18,13 @@ def setup(bot):
         bot.memory["start_time"] = datetime.datetime.utcnow()
 
 
+def shutdown(bot):
+    try:
+        del bot.memory['start_time']
+    except KeyError:
+        pass
```
dgw: Don't clear this one. Doing so will make it so reloading the `uptime` module resets the bot's uptime to 0.

sopel/modules/url.py
```diff
@@ -98,6 +98,14 @@ def setup(bot):
         bot.memory['shortened_urls'] = tools.SopelMemory()
 
 
+def shutdown(bot):
+    for key in ['url_exclude', 'last_seen_url', 'shortened_urls']:
+        try:
+            del bot.memory[key]
+        except KeyError:
+            pass
```
dgw: Clearing `last_seen_url` makes sense, for the same reason that it makes sense to clear `find_lines` (the module could miss a URL while unloaded, and `.title` would then return possibly an unpredictable result).

Clearing `url_exclude` makes it possible to edit the exclude patterns and then reload the module for them to take effect, so that's good.

But clearing `shortened_urls` can only cause extra API calls, and I can't think of a way that leaving it behind on `shutdown()` could result in unexpected behavior.

sopel/modules/ipython.py
HumorBaby: The console doesn't stop, so keeping the "running" flag around instead (no `shutdown()`).

sopel/modules/tell.py
HumorBaby: Separate commit. Updated this as well:
https://github.com/sopel-irc/sopel/blob/39e3680db18c4bed9801d040f7486c655b95a9a0/sopel/modules/tell.py#L74

```diff
-    except (OSError, IOError):  # Remove IOError when dropping py2 support 
+    except (OSError, IOError):  # TODO: Remove IOError when dropping py2 support 
```

sopel/modules/uptime.py
HumorBaby: Ah yes, good call.

sopel/modules/url.py
HumorBaby: Cleared all but `shortened_urls`.

sopel/modules/safety.py
HumorBaby: > After tracing through how the cache works though, either I don't understand _clean_cache() at all or it's almost completely useless and doesn't do what it should.

Seems like the latter. If I understand it correctly, every 24 hours (or if there are > 1024 entries), _one_ (the oldest) is removed... Not really a `_clean`'ing if you ask me :stuck_out_tongue_closed_eyes:

sopel/modules/url.py
```diff
@@ -98,6 +98,14 @@ def setup(bot):
         bot.memory['shortened_urls'] = tools.SopelMemory()
 
 
+def shutdown(bot):
+    for key in ['url_exclude', 'last_seen_url']:
```
dgw: Should have mentioned yesterday: Leaving a comment here explaining (briefly) why `shortened_urls` isn't cleared would help ensure that it's not re-added later without good reason.

sopel/modules/safety.py
dgw: That's the conclusion I came to, also. Going to make a new issue for this one.

sopel/modules/url.py
HumorBaby: :+1: Done.

PR #1574, opened 2019-04-26
db: make database path relative to `config.core.homedir`

### LINKED ISSUE #1047: db_filename is relative to the directory where the config file resides
This should be relative to homedir. The way is currently handle makes no sense IMO.

HumorBaby:
Makes the default base directory for the database `core.homedir`. Non-absolute paths are relative to that directory, as opposed to previously, where they were relative to the config file path.

Additionally, now `SopelDB.__init__` will  `raise OSError` if the directory in which the database is to be created does not exist.

Closes #1047.


HumorBaby:
Any @sopel-irc/rockstars takers to review this? I'm hoping its pretty straightforward :grin: _we shall see..._

Exirel:
The conflicts look... interesting. Good luck. :-/

dgw:
On the plus side, there's only "one" conflict? 😕

HumorBaby:
> The conflicts look... interesting. Good luck. :-/

hehe, wasn't too bad actually! RustyB developed _around_ the part I modified :grin: Anyway, the conflict is handled now.

sopel/db.py
```diff
@@ -87,8 +88,8 @@ class SopelDB(object):
     database. It simplifies those common operations, and allows direct access
     to the database, wherever the user has configured it to be.
 
-    When configured with a relative filename, it is assumed to be in the same
-    directory as the config."""
+    When configured with a relative filename, it is assumed to be in the directory
+    set (or defaulted to) in the core setting 'homedir'."""
```
Exirel: While we are changing docstring, can we put these `"""` on their own line? They like to be alone on their own.

sopel/db.py
```diff
@@ -98,13 +99,21 @@ def __init__(self, config):
         # Handle SQLite explicitly as a default
         if db_type == 'sqlite':
             path = config.core.db_filename
-            config_dir, config_file = os.path.split(config.filename)
+            _, config_file = os.path.split(config.filename)
```
Exirel: You can actually replace all of that by using `config.basename`.

See https://github.com/sopel-irc/sopel/blob/master/sopel/config/__init__.py#L109

sopel/db.py
```diff
@@ -98,13 +99,21 @@ def __init__(self, config):
...
             config_name, _ = os.path.splitext(config_file)
             if path is None:
-                path = os.path.join(config_dir, config_name + '.db')
+                path = os.path.join(config.core.homedir, config_name + '.db')
```
Exirel:
```suggestion
                path = os.path.join(config.core.homedir, config.basename + '.db')
```

sopel/db.py
HumorBaby: :+1:

sopel/db.py
HumorBaby: :+1:

sopel/db.py
HumorBaby: :+1:

sopel/db.py
```diff
@@ -87,8 +88,9 @@ class SopelDB(object):
     database. It simplifies those common operations, and allows direct access
     to the database, wherever the user has configured it to be.
 
-    When configured with a relative filename, it is assumed to be in the same
-    directory as the config."""
+    When configured with a relative filename, it is assumed to be in the directory
+    set (or defaulted to) in the core setting 'homedir'.
```
Exirel:
```suggestion
    set (or defaulted to) in the core setting ``homedir``.
```

HumorBaby: Done :+1:

PR #1576, opened 2019-04-29
bot: prevent KeyError from search_url_callbacks

deathbybandaid:
Discovered after removing stock modules. If not in memory, the search fails with a KeyError.

### DIFF
--- a/sopel/bot.py
+++ b/sopel/bot.py
@@ -824,6 +824,10 @@ class Sopel(irc.Bot):
         .. __: https://docs.python.org/3.6/library/re.html#match-objects
 
         """
+        if 'url_callbacks' not in self.memory:
+            # nothing to search
+            return
+
         for regex, function in tools.iteritems(self.memory['url_callbacks']):
             match = regex.search(url)
             if match:


sopel/bot.py
```diff
@@ -731,6 +731,9 @@ def search_url_callbacks(self, url):
         .. __: https://docs.python.org/3.6/library/re.html#match-objects
 
         """
+        if 'url_callbacks' not in self.memory:
+            # nothing to search
+            return
```
dgw: This just makes me happier for some reason.

```suggestion
            return

```

Exirel:
> why he had to duplicate the code instead of inheriting/importing the class methods somehow.

Because none of MockSopel and MockSopelWrapper were subclasses of Sopel and SopelWrapper at the time of writing. Good news, that's something we should be able to change in a near future!

Exirel:
@deathbybandaid please squash your commits and rename it like to follow the project's naming conventions:

    bot: prevent KeyError from search_url_callbacks

deathbybandaid:
Squished, renamed, and all checks pass

PR #1577, opened 2019-04-30
dice: Add better comment functionality

rileywilddog:
Uses # to denote comments so users can say `.dice 1d20 # for initiative`. It just discards the comment so the output is cleaner.
Old output: `You roll 1d20 # for initiative: (12) # for initiative = 12`
New output: `You roll 1d20: (13) = 13`

### DIFF
--- a/sopel/modules/dice.py
+++ b/sopel/modules/dice.py
@@ -176,12 +176,13 @@ def _roll_dice(bot, dice_expression):
 @sopel.module.example(".roll 1001d1", 'I only have 1000 dice. =(')
 @sopel.module.example(".roll 1d1 + 1d1", 'You roll 1d1 + 1d1: (1) + (1) = 2')
 @sopel.module.example(".roll 1d1+1d1", 'You roll 1d1+1d1: (1)+(1) = 2')
+@sopel.module.example(".roll 1d6 # initiative", r'You roll 1d6: \(\d\) = \d', re=True)
 def roll(bot, trigger):
-    """.dice XdY[vZ][+N], rolls dice and reports the result.
+    """.dice XdY[vZ][+N][#COMMENT], rolls dice and reports the result.
 
     X is the number of dice. Y is the number of faces in the dice. Z is the
     number of lowest dice to be dropped from the result. N is the constant to
-    be applied to the end result.
+    be applied to the end result. Comment is for easily noting the purpose.
     """
     # This regexp is only allowed to have one capture group, because having
     # more would alter the output of re.findall.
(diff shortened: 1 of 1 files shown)


dgw:
I, uh… hmm. The ability to include "comments" like that in `.dice` expressions isn't exactly an intended behavior. One might even reasonably consider the fact that it works to be a bug of sorts.

Officially supporting this isn't really something _I_ want to do. @sopel-irc/rockstars, dissenting opinions?

rileywilddog:
I was similarly surprised. I intended to just clean up the output, but discovered the "eval preprocessor" accident instead of a feature to change. Either way - usage exists in the real world.
Previously we had a simpler dice roll command in a channel that only used the first argument, leading to unintended but eventually _ubiquitous_ usage of the form `!roll 1d20 to seduce x`. When my sopel-based bot replaced that one, the inability to add a remark was a frequent complaint until someone found that `.roll 1d6 #comment` worked.
Making `#comment` output better is the best way I could find to make the existing usage less gross without changing the command format. Would adding it without the documentation help?

dgw:
How about just silently outputting only the matched part of the input? Ignore anything that doesn't actually match the dice regex, and don't include it in the output at all.

Could be more complicated than I think (am mobile, with limited ability to check the code), but much closer to ideal than explicitly handling and documenting unintended behavior that we don't want to continue supporting if we rewrite the dice module later to move away from the scary use of eval'ing arbitrary input.

Exirel:
To be honest... I quite like the idea of allowing for people to type `.dice 1D100+1 # initiative`. Like, I can see how I could use that myself.

However, I'm chasing untested utility function, and not keen to accept or to review PR when they touch at something that can clearly be tested. If you are interested in doing so, I'll be happy to help! Otherwise, I think you raised a very valid issue, and I'd be happy to keep track of it, so someone (maybe me or someone else) can fix it while adding tests. I totally get if you don't have the time for that, as you probably didn't expect this amount of work given the current state of Sopel's test suite (that is to say: it is not good).

Yet, if you're still reading this... then here is what I would do, and what you can do if you like:

* extract the utility function from the command function, make it a public function in the `dice.py` module
* add a test suite in `test/modules/test_modules_dice.py`
* fix the regex match to accept an optional comment, starting with `#`
* handle both case, and test that the regex parsing works as intended

... and we should be good. What do you think?

Again, I totally understand if you don't have the time or the knowledge to do what I suggest, so let me know because I can see me doing the job for you, as the feature speaks to me!

dgw:
@Exirel You could refactor and add tests in a new commit within this PR, probably, since you've write access to the repo. (I'd suggest `test/modules/dice.py` for the test file though, or at most `test_dice.py`. Otherwise it's long and redundant!)

rileywilddog:
I haven't abandoned you, I'm just short on time. If someone else would like to take it over feel free.

I like the idea of ignoring anything not resembling dice or math, especially since that might enable hash-less comments (`.dice 1d20 for initiative`), but worry that might cause odd behavior for ambiguous parameters (`.dice 1d6 +20 sword steal`, `.dice 1d6 steal +20 sword`).

Greenfield, my preference would be that the dice definition cannot contain spaces: `.dice 1d20+1d10-14 +20 which isn't used`

I was under the impression the reason for the wide variety of arguments to `@example` was for simple test functionality, did I misunderstand?

Exirel:
Sound good to me.

PR #1578, opened 2019-04-30
meetbot: Line length, linting, tweaks, fixes

half-duplex:
This should satisfy the easy scrutinizer, flake8, and human complaints, and removes a couple TODOs (display durations in minutes instead of seconds).

PR #1579, opened 2019-04-30
Unignore even more style checks

half-duplex:
Unignores (or updates to add) and fixes the following:
E117 over-indented
E127 continuation line over-indented for visual indent
E128 continuation line under-indented for visual indent
F632 use ==/!= to compare str, bytes, and int literals

With the above fixed, flake8 3.7.7 finds no errors, so the 3.6 pin is replaced with 3.7

`checkstyle.sh`'s py3-unsafe `unicode()` detection seemed to have some inverted logic and was reporting files _containing_ `unicode = str` as potentially unsafe, instead of only files _not_ containing that.

### DIFF
--- a/sopel/coretasks.py
+++ b/sopel/coretasks.py
@@ -180,11 +180,13 @@ def handle_names(bot, trigger):
     # it'd be worth it.
     # If this ever needs to be updated, remember to change the mode handling in
     # the WHO-handler functions below, too.
-    mapping = {'+': sopel.module.VOICE,
-               '%': sopel.module.HALFOP,
-               '@': sopel.module.OP,
-               '&': sopel.module.ADMIN,
-               '~': sopel.module.OWNER}
+    mapping = {
+        "+": sopel.module.VOICE,
+        "%": sopel.module.HALFOP,
+        "@": sopel.module.OP,
+        "&": sopel.module.ADMIN,
+        "~": sopel.module.OWNER,
+    }
 
     for name in names:
         priv = 0
@@ -240,11 +242,13 @@ def track_modes(bot, trigger):
     modestring = trigger.args[1]
     nicks = [Identifier(nick) for nick in trigger.args[2:]]
 
-    mapping = {'v': sopel.module.VOICE,
-               'h': sopel.module.HALFOP,
-               'o': sopel.module.OP,
-               'a': sopel.module.ADMIN,
-               'q': sopel.module.OWNER}
+    mapping = {
+        "v": sopel.module.VOICE,
+        "h": sopel.module.HALFOP,
+        "o": sopel.module.OP,
+        "a": sopel.module.ADMIN,
+        "q": sopel.module.OWNER,
+    }
 
     # Parse modes before doing anything else
     modes = []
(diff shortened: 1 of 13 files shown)


sopel/modules/dice.py
```diff
@@ -224,7 +224,7 @@ def _get_pretty_str(dice):
         result = eval_equation(eval_str)
     except TypeError:
         bot.reply("The type of this equation is, apparently, not a string. " +
```
Exirel: While you are at it, you can remove the `+` sign at the end, since Python strings are automatically concatenated with each other.

```suggestion
         bot.reply("The type of this equation is, apparently, not a string. "
```

sopel/modules/dice.py
```diff
@@ -233,7 +233,7 @@ def _get_pretty_str(dice):
         return
     except (SyntaxError, eval_equation.Error):
         bot.reply("I don't know how to process that. " +
```
Exirel:
```suggestion
         bot.reply("I don't know how to process that. "
```

Exirel:
@half-duplex aw, now there are conflict to fix first. :(

half-duplex:
Had to make the same changes in some new code after the rebase

sopel/coretasks.py
```diff
@@ -764,11 +768,13 @@ def _record_who(bot, channel, user, host, nick, account=None, away=None, modes=N
     user.away = away
     priv = 0
     if modes:
-        mapping = {'+': sopel.module.VOICE,
-           '%': sopel.module.HALFOP,
-           '@': sopel.module.OP,
-           '&': sopel.module.ADMIN,
-           '~': sopel.module.OWNER}
+        mapping = {
+            "+": sopel.module.VOICE,
```
Exirel: Why do we change `'+'` to `"+"` already?

test/modules/test_modules_remind.py
```diff
@@ -28,23 +28,35 @@ def sopel():
     ('13:37 message', TimeReminder('13', '37', None, None, 'message')),
     ('13:37:00 message', TimeReminder('13', '37', '00', None, 'message')),
     # Make sure numbers are not confused with anything else
-    ('13:37:00 10 message',
-     TimeReminder('13', '37', '00', None, '10 message')),
+    (
```
Exirel: You'll hate me so hard if #1590 get merged first, and I won't blame you for that. I mean, I already hate myself a little bit for that...

sopel/coretasks.py
half-duplex: Requoting is an "I'm touching it, and where it wouldn't necessitate escaping, double quotes are the eventual direction for consistency."

test/modules/test_modules_remind.py
half-duplex: :stuck_out_tongue:

sopel/coretasks.py
Exirel: Meh. I don't know who said that, but I slightly disagree. Whatever, not important.

dgw:
Of course merging #1578 caused conflicts with this PR, because why wouldn't it? 😑

@half-duplex, would you fix this one up for me over the weekend? 😸

Exirel:
Oh, that's a very easy conflict to fix! :)

dgw:
Tested merging this to a local branch, and I'm gonna merge it before anything breaks the newly fixed checks. I want better code style just like everyone else!

dgw:
Me from the future would like to ask himself from the past why he didn't insist on reordering things to prevent a commit with failed tests from being merged into `master`… Oh well.

PR #1583, opened 2019-05-02
cli: `cli.utils.find_config` returns absolute path, as expected

HumorBaby:
According to this:
https://github.com/sopel-irc/sopel/blob/0952f4443f831ef7eeba0c04ab34ff3f6dc76197/sopel/cli/utils.py#L153-L154
`find_config` is expected to always return an absolute file path.

However, here:
https://github.com/sopel-irc/sopel/blob/0952f4443f831ef7eeba0c04ab34ff3f6dc76197/sopel/cli/utils.py#L185-L186
it is possible for a simple filename (or relative path, which is not as bad, but still not the expected result) to be returned.

This is usually not an issue, but in the wizard, an attempt to `os.makedirs` is made, which breaks if there is no directory to even try to be made :smile: 

Fixed now, though.

### DIFF
--- a/sopel/cli/utils.py
+++ b/sopel/cli/utils.py
@@ -183,7 +183,7 @@ def find_config(config_dir, name, extension='.cfg'):
 
     """
     if os.path.isfile(name):
-        return name
+        return os.path.abspath(name)
     name_ext = name + extension
     for filename in enumerate_configs(config_dir, extension):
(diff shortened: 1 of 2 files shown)


Exirel:
Heh, you found the last bit of code I left untouched because I wasn't sure what to do about that. Well done!

PR #1585, opened 2019-05-03
plugins: load plugins from setuptools entry points

Exirel:
By defining a `sopel.plugins` entry point in a `setup.cfg` (or `setup.py`) it is now possible to expose a Sopel plugin:

    setup(
        name='my_plugin',
        version='1.0'
        entry_points={
            'sopel.plugins': [
                'custom = my_plugin.path.to.plugin',
            ],
        }
    )

This definition will allow Sopel to load the `custom` plugin, from the Python sub-module `my_plugin.path.to.plugin` from the `my_plugin` package.

I'm not 100% happy with my abstractclass/subclasses, but I think that's for another day. Here, I wanted to showcase how quick and easy it is to implement new ways to discover Sopel plugins.

I'd like to mention that, in my opinion, entry points are probably a better option than `sopel_modules.*` subpackages.

### DIFF
--- a/sopel/plugins/__init__.py
+++ b/sopel/plugins/__init__.py
@@ -13,6 +13,7 @@ and only one plugin per unique name, using a specific order:
 
 * extra directories defined in the settings
 * homedir's ``modules`` directory
+* ``sopel.plugins`` setuptools entry points
 * ``sopel_modules``'s subpackages
 * ``sopel.modules``'s core plugins
 
@@ -29,6 +30,8 @@ import collections
 import imp
 import itertools
 import os
+
+import pkg_resources
 
 from . import exceptions, handlers  # noqa
 
@@ -80,6 +83,18 @@ def find_sopel_modules_plugins():
             yield handlers.PyModulePlugin(name, 'sopel_modules')
 
 
+def find_entry_point_plugins(group='sopel.plugins'):
+    """List plugins from a setuptools entry point group
+
+    :param str group: setuptools entry point group to look for
+                      (defaults to ``sopel.plugins``)
+    :return: Yield instance of :class:`~.handlers.EntryPointPlugin`
+             created from setuptools entry point given ``group``
+    """
+    for entry_point in pkg_resources.iter_entry_points(group):
+        yield handlers.EntryPointPlugin(entry_point)
+
+
 def find_directory_plugins(directory):
     """List plugins from a ``directory``
 
@@ -110,12 +125,15 @@ def enumerate_plugins(settings):
 
        * :func:`find_internal_plugins` for internal plugins
        * :func:`find_sopel_modules_plugins` for ``sopel_modules.*`` plugins
+       * :func:`find_entry_point_plugins` for plugins exposed by setuptools
+         entry points
        * :func:`find_directory_plugins` for modules in ``$homedir/modules``
          and in extra directories, as defined by ``settings.core.extra``
 
     """
     from_internals = find_internal_plugins()
     from_sopel_modules = find_sopel_modules_plugins()
+    from_entry_points = find_entry_point_plugins()
     # load from directories
     source_dirs = [os.path.join(settings.homedir, 'modules')]
     if settings.core.extra:
@@ -131,6 +149,7 @@ def enumerate_plugins(settings):
     all_plugins = itertools.chain(
         from_internals,
         from_sopel_modules,
+        from_entry_points,
         *from_directories)
 
     # Get module settings
@@ -161,6 +180,7 @@ def get_usable_plugins(settings):
 
     * extra directories defined in the settings
     * homedir's ``modules`` directory
+    * ``sopel.plugins`` setuptools entry points
     * ``sopel_modules``'s subpackages
     * ``sopel.modules``'s core plugins
 
--- a/sopel/plugins/handlers.py
+++ b/sopel/plugins/handlers.py
@@ -330,3 +330,62 @@ class PyFilePlugin(PyModulePlugin):
         module might not be available through ``sys.path``.
         """
         self._module = self._load()
+
+
+class EntryPointPlugin(PyModulePlugin):
+    """Sopel plugin loaded from a ``setuptools`` entry point
+
+    :param entry_point: a ``setuptools`` entry point object
+
+    This handler loads a Sopel plugin exposed by a ``setuptools`` entry point.
+    It expects to be able to load a module object from the entry point, and to
+    work as a :class:`~.PyModulePlugin` from that module.
+
+    By default, Sopel uses the entry point ``sopel.plugins``. To use that for
+    their plugin, developers must define an entry point either in their
+    ``setup.py`` file or their ``setup.cfg`` file::
+
+        # in setup.py file
+        setup(
+            name='my_plugin',
+            version='1.0'
+            entry_points={
+                'sopel.plugins': [
+                    'custom = my_plugin.path.to.plugin',
+                ],
+            }
+        )
+
+    And this plugin can be loaded with::
+
+        >>> from pkg_resources import iter_entry_points
+        >>> from sopel.plugins.handlers import EntryPointPlugin
+        >>> plugin = [
+        ...     EntryPointPlugin(ep)
+        ...     for ep in iter_entry_points('sopel.plugins', 'custom')
+        ... ][0]
+        >>> plugin.load()
+        >>> plugin.name
+        'custom'
+
+    In this example, the plugin ``custom`` is loaded from an entry point.
+    Unlike the :class:`~.PyModulePlugin`, the name is not derived from the
+    actual python module, but from its entry point's name.
+
+    .. seealso::
+
+        Sopel uses the :func:`~sopel.plugins.find_entry_point_plugins` function
+        internally to search entry points.
+
+        Entry point is a `standard feature of setuptools`__ for Python, used
+        by other applications (like ``pytest``) for their plugins.
+
+        .. __: https://setuptools.readthedocs.io/en/stable/setuptools.html#dynamic-discovery-of-services-and-plugins
+
+    """
+    def __init__(self, entry_point):
+        self.entry_point = entry_point
+        super(EntryPointPlugin, self).__init__(entry_point.name)
+
+    def load(self):
+        self._module = self.entry_point.load()
--- a/test/test_plugins.py
+++ b/test/test_plugins.py
@@ -2,6 +2,9 @@
 """Test for the ``sopel.plugins`` module."""
 from __future__ import unicode_literals, absolute_import, print_function, division
 
+import sys
+
+import pkg_resources
 import pytest
 
 from sopel import plugins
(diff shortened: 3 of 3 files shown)


sopel/plugins/handlers.py
```diff
@@ -330,3 +330,51 @@ def reload(self):
         module might not be available through ``sys.path``.
         """
         self._module = self._load()
+
+
+class EntrypointPluginHandler(PyModulePlugin):
```
dgw: Any particular reason this class is named with the `Handler` suffix when the others are just `TypeOfPlugin`, without mentioning `Handler` anywhere?

This goes along with another line comment I haven't added yet in another file: Why isn't it `EntryPoint` in CamelCase? All over the Python docs (and the ones you added here), they're referred to as "entry points"—two words—so why does the code itself refer to `entrypoint(s)`?

Based on those two things, I think this class should actually be called `EntryPointPlugin`.

sopel/plugins/handlers.py
```diff
@@ -330,3 +330,51 @@ def reload(self):
...
+    """Sopel plugin loaded from a ``setuptools`` entry points
```
dgw: Either "a" or "points", but not both. 😁

```suggestion
    """Sopel plugin loaded from a ``setuptools`` entry point
```

sopel/plugins/handlers.py
```diff
@@ -330,3 +330,51 @@ def reload(self):
...
+        >>> plugin.load()
+        >>> plugin.name
+        'custom'
+
+    In this example, the plugin ``custom`` is loaded from an entry point.
+    Unlike the :class:`~.PyModulePlugin`, the name is not derived from the
+    actual python module, but from its entry point's name.
+    """
+    def __init__(self, entrypoint):
+        self.entrypoint = entrypoint
```
dgw: Probably should make the parameter AND the instance variable both `entry_point`, for consistency with Python's documentation, the terminology used in `setup.py`/`setup.cfg`, and the documentation here that refers to "entry points" with that two-word phrase (instead of squashing it into one word).

test/test_plugins.py
```diff
@@ -123,3 +126,34 @@ def test_plugin_load_pypackage_bad_dir_no_init(tmpdir):
...
+def test_plugin_load_entrypoint(tmpdir):
+    root = tmpdir.mkdir('loader_mods')
+    mod_file = root.join('file_mod.py')
+    mod_file.write(MOCK_MODULE_CONTENT)
+
+    # generate setuptools Distribution object
+    distrib = pkg_resources.Distribution(root.strpath)
+    sys.path.append(root.strpath)
+
+    # load the entrypoint
```
dgw: The one place in your docstrings/comments I found that uses "entrypoint" as one word instead of two! 😁

```suggestion
    # load the entry point
```

test/test_plugins.py
```diff
@@ -123,3 +126,34 @@ def test_plugin_load_pypackage_bad_dir_no_init(tmpdir):
...
+    try:
+        entrypoint = pkg_resources.EntryPoint(
```
dgw: 🙈

```suggestion
        entry_point = pkg_resources.EntryPoint(
```

sopel/plugins/__init__.py
```diff
@@ -80,6 +83,18 @@ def find_sopel_modules_plugins():
             yield handlers.PyModulePlugin(name, 'sopel_modules')
 
 
+def find_entrypoints_plugins(group='sopel.plugins'):
```
dgw: Other comments go into why I'm suggesting this.

```suggestion
def find_entry_point_plugins(group='sopel.plugins'):
```

sopel/plugins/__init__.py
```diff
@@ -80,6 +83,18 @@ def find_sopel_modules_plugins():
...
+    """List plugins from a setuptools entry point group
+
+    :param str group: setuptools entry point group to look for
+                      (default to ``sopel.plugins``)
+    :return: Yield instance of :class:`~.handlers.EntrypointPluginHandler`
+             created from setuptools entry point given ``group``
+    """
+    for entrypoint in pkg_resources.iter_entry_points(group):
```
dgw:
```suggestion
    for entry_point in pkg_resources.iter_entry_points(group):
```

And the reference to it on the next line, of course.

sopel/plugins/handlers.py
Exirel: Sometimes, I don't make the right decision. That's why review exists and I'm glad they do!

sopel/plugins/__init__.py
```diff
@@ -80,6 +83,18 @@ def find_sopel_modules_plugins():
...
+        yield handlers.EntrypointPluginHandler(entrypoint)
```
Exirel:
```suggestion
        yield handlers.EntryPointPlugin(entry_point)
```

sopel/plugins/__init__.py
```diff
@@ -80,6 +83,18 @@ def find_sopel_modules_plugins():
             yield handlers.PyModulePlugin(name, 'sopel_modules')
 
 
+def find_entrypoints_plugins(group='sopel.plugins'):
+    """List plugins from a setuptools entry point group
+
+    :param str group: setuptools entry point group to look for
+                      (default to ``sopel.plugins``)
+    :return: Yield instance of :class:`~.handlers.EntrypointPluginHandler`
```
Exirel:
```suggestion
    :return: Yield instance of :class:`~.handlers.EntryPointPlugin`
```

sopel/plugins/__init__.py
```diff
@@ -110,12 +125,15 @@ def enumerate_plugins(settings):
 
        * :func:`find_internal_plugins` for internal plugins
        * :func:`find_sopel_modules_plugins` for ``sopel_modules.*`` plugins
+       * :func:`find_entrypoints_plugins` for plugins exposed by setuptools
```
Exirel:
```suggestion
       * :func:`find_entry_points_plugins` for plugins exposed by setuptools
```

sopel/plugins/__init__.py
```diff
@@ -110,12 +125,15 @@ def enumerate_plugins(settings):
...
+         entry points
        * :func:`find_directory_plugins` for modules in ``$homedir/modules``
          and in extra directories, as defined by ``settings.core.extra``
 
     """
     from_internals = find_internal_plugins()
     from_sopel_modules = find_sopel_modules_plugins()
+    from_entry_points = find_entrypoints_plugins()
```
Exirel:
```suggestion
    from_entry_points = find_entry_points_plugins()
```

sopel/plugins/handlers.py
Exirel:
```suggestion
class EntryPointPlugin(PyModulePlugin):
```

sopel/plugins/handlers.py
```diff
@@ -330,3 +330,51 @@ def reload(self):
         module might not be available through ``sys.path``.
         """
         self._module = self._load()
+
+
+class EntrypointPluginHandler(PyModulePlugin):
+    """Sopel plugin loaded from a ``setuptools`` entry points
+
+    :param entrypoint: a ``setuptools`` entry point object
```
Exirel:
```suggestion
    :param entry_point: a ``setuptools`` entry point object
```

sopel/plugins/handlers.py
```diff
@@ -330,3 +330,51 @@ def reload(self):
...
+                'sopel.plugins': [
+                    'custom = my_plugin.path.to.plugin',
+                ],
+            }
+        )
+
+    And this plugin can be loaded with::
+
+        >>> from pkg_resources import iter_entry_points
+        >>> from sopel.plugins.handlers import EntrypointPluginHandler
```
Exirel:
```suggestion
        >>> from sopel.plugins.handlers import EntryPointPlugin
```

sopel/plugins/handlers.py
```diff
@@ -330,3 +330,51 @@ def reload(self):
...
+        >>> plugin = [
+        ...     EntrypointPluginHandler(ep)
```
Exirel:
```suggestion
        ...     EntryPointPlugin(ep)
```

sopel/plugins/handlers.py
```diff
@@ -330,3 +330,51 @@ def reload(self):
...
+        ... ][0]
+        >>> plugin.load()
+        >>> plugin.name
+        'custom'
+
+    In this example, the plugin ``custom`` is loaded from an entry point.
+    Unlike the :class:`~.PyModulePlugin`, the name is not derived from the
+    actual python module, but from its entry point's name.
+    """
+    def __init__(self, entrypoint):
```
Exirel:
```suggestion
    def __init__(self, entry_point):
```

sopel/plugins/handlers.py
Exirel:
```suggestion
        self.entry_point = entry_point
```

sopel/plugins/handlers.py
```diff
@@ -330,3 +330,51 @@ def reload(self):
...
+        self.entrypoint = entrypoint
+        super(EntrypointPluginHandler, self).__init__(entrypoint.name)
```
Exirel:
```suggestion
        super(EntryPointPlugin, self).__init__(entry_point.name)
```

sopel/plugins/handlers.py
```diff
@@ -330,3 +330,51 @@ def reload(self):
...
+
+    def load(self):
+        self._module = self.entrypoint.load()
```
Exirel:
```suggestion
        self._module = self.entry_point.load()
```

test/test_plugins.py
```diff
@@ -123,3 +126,34 @@ def test_plugin_load_pypackage_bad_dir_no_init(tmpdir):
 
     with pytest.raises(Exception):
         plugins.handlers.PyFilePlugin(package_dir.strpath)
+
+
+def test_plugin_load_entrypoint(tmpdir):
```
Exirel:
```suggestion
def test_plugin_load_entry_point(tmpdir):
```

test/test_plugins.py
```diff
@@ -123,3 +126,34 @@ def test_plugin_load_pypackage_bad_dir_no_init(tmpdir):
...
+
+    # generate setuptools Distribution object
+    distrib = pkg_resources.Distribution(root.strpath)
+    sys.path.append(root.strpath)
+
+    # load the entrypoint
+    try:
+        entrypoint = pkg_resources.EntryPoint(
+            'test_plugin', 'file_mod', dist=distrib)
+        plugin = plugins.handlers.EntrypointPluginHandler(entrypoint)
```
Exirel:
```suggestion
        plugin = plugins.handlers.EntryPointPlugin(entry_point)
```

Exirel:
Just a quick note: fixed as asked. It looks good for a last review & merge. Can't wait to play with that feature for future plugins!

sopel/plugins/__init__.py
```diff
@@ -80,6 +83,18 @@ def find_sopel_modules_plugins():
             yield handlers.PyModulePlugin(name, 'sopel_modules')
 
 
+def find_entry_point_plugins(group='sopel.plugins'):
+    """List plugins from a setuptools entry point group
+
+    :param str group: setuptools entry point group to look for
+                      (default to ``sopel.plugins``)
```
dgw: I think we should do our best to stick to this verb tense in these "default value" notes. Yes, I'll probably miss some, and (unlike this instance) not catch them before merging PRs. :laughing:

```suggestion
                      (defaults to ``sopel.plugins``)
```

sopel/plugins/handlers.py
```diff
@@ -330,3 +330,51 @@ def reload(self):
...
         self._module = self._load()
+
+
+class EntryPointPlugin(PyModulePlugin):
+    """Sopel plugin loaded from a ``setuptools`` entry point
+
+    :param entry_point: a ``setuptools`` entry point object
+
+    This handler loads a Sopel plugin exposed by a ``setuptools`` entry point
+    from the ``sopel.plugins`` group. It expects to be able to load a module
```
dgw: > <tt>from the \`\`sopel.plugins\`\` group</tt>

The entry point plugin search function takes an optional argument to change the search group, so… should this class's docstring really talk about `sopel.plugins` as if it's the only possibility? (We won't bother changing it in this PR—I just want to start a discussion about future intentions.)

Exirel: Hm, you're right. I didn't think about that. I've an idea to fix that.

Exirel:
@dgw done! :wink:

PR #1564, opened 2019-04-21
doc: autoprogram & configuration

Exirel:
I stole the work from #1274 by @nfearnley to addapt it to the latest changes in the codebase (rework of Sopel's cli in particular).

I had almost nothing to do: check things here and there, and here we go.

Quick notes:

* I had to trick the autoprogram directive to prevent the `legacy` sub-command to appear
* `:maxdepth: 1` shows only the fisrt-level of the command, ie. none of its sub-commands
* public sub-commands are listed manually, but they still use autoprogram at its best

Again, thank you @nfearnley for your work, I make sure to co-author everything. I didn't know about `sphinxcontrib-autoprogram`, and it's a great addition to my toolkit—and to the Sopel project!

----

Remaining tasks:

  - [x] Restructure the documentation to prevent duplication in man page 1 and 3
  - [x] Do something about the old Willie man page
  - [x] Finalize page names (e.g. @dgw already started on documenting some of the CLI stuff, but called the file `cli.rst` instead of `cmd.rst`)
  - [x] Update documentation of logging after #1512 is merged
  - [x] Update documentation of auth after #1513 is merged
  - [x] Update documentation of flood burst after #1518 is merged
  - [x] Update documentation of logging *again* since #1547 has been merged

Other PRs being newer than this one, they will have to adapt after this is merged.

PR #1588, opened 2019-05-06
cli: sopel-plugins (basic version)

### LINKED ISSUE #244: Make module whitelisting setup easier
Currently when you want to have a module whitelist you need to type the names of the modules you want to load.

It would be awesome if instead we could show a list of installed modules and have users check the ones they want to load (using arrow keys to navigate, and enter or space to toggle selection)

### LINKED ISSUE #1519: Enhancement: Configuring individual modules only?
There should be a way to only configure individual modules only.

Thank you in advance. :)

Exirel:
Should fix #244 and fix #1519 at the same time.

This PR adds a new command line to Sopel: `sopel-plugins`.

This command line supports 4 actions:

* `list`, to list available plugins (all by default, or just enabled/disabled plugins)
* `show`, to show detailed information about a plugin
* `configure`, to run a config wizard for a plugin
* `enable`, to enable a plugin
* `disable`, to disable a plugin

In order to implement it, I had to update the plugin handlers, adding methods to get meta description. I think we could built from that to include other meta-data (such as version, author, etc.).

This is the first version of the command. I think it's better to stop here for now, let users test it, play with it, etc. and see what's asked.

In the future, I think we will need to rework how a plugin can expose the commands, URL callback, rules, and jobs to the outside world, but I don't want to do that in that PR, which is big enough as it is, and fairly usable to me. 

## sopel-plugins list

```
usage: sopel-plugins list [-h] [-c filename] [-C] [-e | -d] [-n]

List available Sopel plugins from all possible sources: built-in, from
``sopel_modules.*``, from ``sopel.plugins`` entry points, or Sopel's plugin
directories. Enabled plugins are displayed in green; disabled, in red.

optional arguments:
  -h, --help            show this help message and exit
  -c filename, --config filename
                        Use a specific configuration file
  -C, --no-color        Disable colors
  -e, --enabled-only    Display only enabled plugins
  -d, --disabled-only   Display only disabled plugins
  -n, --name-only       Display only plugin names
```

## sopel-plugins show <name>

```
usage: sopel-plugins show [-h] [-c filename] name

Show detailed information about a plugin.

positional arguments:
  name                  Plugin name

optional arguments:
  -h, --help            show this help message and exit
  -c filename, --config filename
                        Use a specific configuration file
```

## sopel-plugins configure <name>

```
usage: sopel-plugins configure [-h] [-c filename] name

Run a config wizard to configure a plugin. This can be used whether the plugin
is enabled or not.

positional arguments:
  name                  Plugin name

optional arguments:
  -h, --help            show this help message and exit
  -c filename, --config filename
                        Use a specific configuration file
```

## sopel-plugins enable

```
usage: sopel-plugins enable [-h] [-c filename] [-a] name

Enable a Sopel plugin by its name, no matter where it comes from. The
``coretasks`` plugin is always enabled. By default, a plugin that is not
excluded is enabled, unless at least one plugin is defined in the
``core.enable`` list. In that case, Sopel uses a "allow-only" policy for
plugins, and enabled plugins must be added to this list.

positional arguments:
  name                  Name of the plugin to enable

optional arguments:
  -h, --help            show this help message and exit
  -c filename, --config filename
                        Use a specific configuration file
  -a, --allow-only      Enforce allow-only policy, adding the plugin to the
                        ``core.enable`` list.
```

## sopel-plugins disable

```
usage: sopel-plugins disable [-h] [-c filename] [-f] [-r] name

Disable a Sopel plugin by its name, no matter where it comes from. It is not
possible to disable the ``coretasks`` plugin.

positional arguments:
  name                  Name of the plugin to disable

optional arguments:
  -h, --help            show this help message and exit
  -c filename, --config filename
                        Use a specific configuration file
  -f, --force           Force exclusion of the plugin. When ``core.enable`` is
                        defined, a plugin may be disabled without being
                        excluded. In this case, use this option to force its
                        exclusion.
  -r, --remove          Remove from ``core.enable`` list if applicable.
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -63,6 +63,7 @@ setup(
         'console_scripts': [
             'sopel = sopel.cli.run:main',
             'sopel-config = sopel.cli.config:main',
+            'sopel-plugins = sopel.cli.plugins:main',
         ],
     },
     python_requires='>=2.7, !=3.0.*, !=3.1.*, !=3.2.*, <4',
--- a/sopel/cli/plugins.py
+++ b/sopel/cli/plugins.py
@@ -73,6 +73,30 @@ def build_parser():
             Remove from ``core.enable`` list if applicable.
         """)
 
+    # sopel-plugin enable
+    enable_parser = subparsers.add_parser(
+        'enable',
+        help="Enable a Sopel plugins",
+        description="""
+            Enable a Sopel plugin by its name, no matter where it comes from.
+            The ``coretasks`` plugin is always enabled. By default, a plugin
+            that is not excluded is enabled, unless at least one plugin is
+            defined in the ``core.enable`` list. In that case, Sopel uses
+            a "allow-only" policy for plugins, and enabled plugins must be
+            added to this list.
+        """)
+    utils.add_common_arguments(enable_parser)
+    enable_parser.add_argument('name', help='Name of the plugin to enable')
+    enable_parser.add_argument(
+        '-a', '--allow-only',
+        dest='allow_only',
+        action='store_true',
+        default=False,
+        help="""
+            Enforce allow-only policy, adding the plugin to the ``core.enable``
+            list.
+        """)
+
     return parser
 
 
@@ -151,6 +175,59 @@ def handle_disable(options):
     print('Plugin %s disabled' % plugin_name)
 
 
+def handle_enable(options):
+    """Enable a Sopel plugin"""
+    plugin_name = options.name
+    settings = utils.load_settings(options)
+    usable_plugins = plugins.get_usable_plugins(settings)
+    enabled = settings.core.enable
+    excluded = settings.core.exclude
+
+    # coretasks is sacred
+    if plugin_name == 'coretasks':
+        tools.stderr('Plugin coretasks is always enabled')
+        return 0
+
+    # plugin does not exist
+    if plugin_name not in usable_plugins:
+        tools.stderr('No plugin named %s' % plugin_name)
+        return 1
+
+    # is it already enabled, but should we enforce anything?
+    is_enabled = usable_plugins[plugin_name][1]
+    if is_enabled and not options.allow_only:
+        # already enabled, and no allow-only option: all good
+        if plugin_name not in enabled:
+            tools.stderr(
+                'Plugin %s is enabled; '
+                'use option -a/--allow-only to enforce allow only policy'
+                % plugin_name)
+        return 0
+
+    # not enabled, or options.allow_only to enforce
+    if plugin_name in excluded:
+        # remove from excluded
+        settings.core.exclude = [
+            name
+            for name in settings.core.exclude
+            if name != plugin_name
+        ]
+    elif plugin_name in enabled:
+        # not excluded, and already in enabled list: all good
+        tools.stderr('Plugin %s is already enabled' % plugin_name)
+        return 0
+
+    if plugin_name not in enabled:
+        if enabled or options.allow_only:
+            # not excluded, but not enabled either: allow-only mode required
+            # either because of the current configuration, or by request
+            settings.core.enable = enabled + [plugin_name]
+
+    settings.save()
+    tools.stderr('Plugin %s enabled' % plugin_name)
+    return 0
+
+
 def main():
     """Console entry point for ``sopel-plugins``"""
     parser = build_parser()
@@ -165,3 +242,5 @@ def main():
         return handle_list(options)
     elif action == 'disable':
         return handle_disable(options)
+    elif action == 'enable':
+        return handle_enable(options)
--- a/sopel/cli/plugins.py
+++ b/sopel/cli/plugins.py
@@ -190,9 +190,11 @@ def handle_show(options):
     }
 
     # option meta description from the plugin itself
+    loaded = False
     try:
         plugin.load()
         description.update(plugin.get_meta_description())
+        loaded = True
     except Exception as error:
         label = ('%s' % error) or 'unknown loading exception'
         error_status = 'error'
@@ -208,6 +210,15 @@ def handle_show(options):
     print('Type:', description['type'])
     print('Source:', description['source'])
     print('Label:', description['label'])
+
+    if not loaded:
+        print('Loading failed')
+        return 1
+
+    print('Loaded successfully')
+    print('Setup:', 'yes' if plugin.has_setup() else 'no')
+    print('Shutdown:', 'yes' if plugin.has_shutdown() else 'no')
+    print('Configure:', 'yes' if plugin.has_configure() else 'no')
 
 
 def handle_disable(options):
--- a/sopel/cli/plugins.py
+++ b/sopel/cli/plugins.py
@@ -4,7 +4,7 @@ from __future__ import unicode_literals, absolute_import, print_function, divisi
 
 import argparse
 
-from sopel import plugins
+from sopel import plugins, tools
 
 from . import utils
 
@@ -50,6 +50,29 @@ def build_parser():
         action='store_true',
         default=False)
 
+    # sopel-plugin disable
+    disable_parser = subparsers.add_parser(
+        'disable',
+        help="Disable a Sopel plugins",
+        description="""
+            Disable a Sopel plugin by its name, no matter where it comes from.
+            It is not possible to disable the ``coretasks`` plugin.
+        """)
+    utils.add_common_arguments(disable_parser)
+    disable_parser.add_argument('name', help='Name of the plugin to disable')
+    disable_parser.add_argument(
+        '-f', '--force', action='store_true', default=False,
+        help="""
+            Force exclusion of the plugin. When ``core.enable`` is defined, a
+            plugin may be disabled without being excluded. In this case, use
+            this option to force its exclusion.
+        """)
+    disable_parser.add_argument(
+        '-r', '--remove', action='store_true', default=False,
+        help="""
+            Remove from ``core.enable`` list if applicable.
+        """)
+
     return parser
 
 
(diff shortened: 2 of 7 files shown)


sopel/cli/plugins.py
```diff
@@ -0,0 +1,88 @@
...
+
+
+def build_parser():
+    """Configure an argument parser for ``sopel-plugins``"""
+    parser = argparse.ArgumentParser(
+        description='Sopel plugins tool')
+
+    # Subparser: sopel-plugins <sub-parser> <sub-options>
+    subparsers = parser.add_subparsers(
+        help='Actions to perform',
```
dgw: Can you actually specify more than one action at once? 😅

Because if not:

```suggestion
        help='Action to perform',
```

sopel/cli/plugins.py
```diff
@@ -0,0 +1,88 @@
...
+
+        if options.enabled_only and not is_enabled:
+            # hide disabled plugins when displaying enabled only
+            continue
+        elif options.disabled_only and is_enabled:
+            # hide enabled plugins when displaying disabled only
+            continue
+
+        if options.no_color:
+            print(name)
```
dgw: Shouldn't there be some indication of whether the plugin is enabled or not even with `--no-color`?

Exirel: Probably yes, but I ran out of ideas on the moment. :smile:

sopel/cli/plugins.py
```diff
@@ -0,0 +1,246 @@
...
+
+    # sopel-plugins list
+    list_parser = subparsers.add_parser(
+        'list',
+        help="List available Sopel plugins",
+        description="""
+            List available Sopel plugins from all possible sources: built-in,
+            from ``sopel_modules.*``, from ``sopel.plugins`` entry points,
+            or Sopel's plugin directories. Enabled plugins are displayed in
+            green, disabled in red.
```
dgw: I forgot to add this to my earlier review. >.<

```suggestion
            green; disabled, in red.
```

sopel/cli/plugins.py
Exirel: So, I settle: I'll go with an apt-like listing. And if someone want a shorter version, I'll go with options, like `--oneline` or `--short`. I'll figure out.

Exirel:
@dgw I changed the `sopel-plugins list` to have an output like this:

    admin/python-module admin.py - Sopel Admin Module (sopel.modules.admin) [enabled]
    coretasks/python-module Tasks that allow the bot to run, but aren't user-facing functionality (sopel.coretasks) [enabled]

and I use green for enabled plugins, red for disabled one, and yellow for enabled but that can't be loaded properly.

All I need now is #1564 to write the documentation.

Exirel:
Update: actually it needs #1646 to be merged so I can write the doc, *then*, it can be merged!

Exirel:
I added a `--name-only` option to `sopel-plugins list`.

Next step before I can hit the "ready for review", is to implement a `sopel-plugins show <plugin>` subcommand.

Exirel:
And now, with `sopel-plugins configure <name>` in order to fix #1519

Exirel:
And now with documentation! :man_dancing:

setup.py
```diff
@@ -63,6 +63,7 @@ def read_reqs(path):
         'console_scripts': [
             'sopel = sopel.cli.run:main',
             'sopel-config = sopel.cli.config:main',
```
dgw: Seeing this here reminded me that `sopel-config` (#1507) isn't included in `cli.rst` either.

Not really related to this PR, but we both agreed on IRC that this isn't worth a separate issue.

sopel/cli/plugins.py
```diff
@@ -0,0 +1,397 @@
...
+        action='store_true',
+        default=False)
+
+    # sopel-plugin disable
+    disable_parser = subparsers.add_parser(
+        'disable',
+        help="Disable a Sopel plugins",
+        description="""
+            Disable a Sopel plugin by its name, no matter where it comes from.
+            It is not possible to disable the ``coretasks`` plugin.
```
dgw: Could we somehow make this two "paragraphs" (single sentences are barely paragraphs 😂)?

Ignore if this would require hacky/kludgey stuff like overriding the description parser or something.

sopel/cli/plugins.py
```diff
@@ -0,0 +1,397 @@
...
+    disable_parser.add_argument(
+        '-r', '--remove', action='store_true', default=False,
+        help="""
+            Remove from ``core.enable`` list if applicable.
+        """)
+
+    # sopel-plugin enable
+    enable_parser = subparsers.add_parser(
+        'enable',
+        help="Enable a Sopel plugins",
```
dgw: You made it _this far_ without running into your nemesi**s**!

```suggestion
        help="Enable a Sopel plugin",
```

sopel/cli/plugins.py
```diff
@@ -0,0 +1,397 @@
...
+        description="""
+            Enable a Sopel plugin by its name, no matter where it comes from.
+            The ``coretasks`` plugin is always enabled. By default, a plugin
+            that is not excluded is enabled, unless at least one plugin is
+            defined in the ``core.enable`` list. In that case, Sopel uses
+            a "allow-only" policy for plugins, and enabled plugins must be
```
dgw: One grammar nitpick, one suggestion:

```suggestion
            an "allow-only" policy for plugins, and all desired plugins must be
```

Optional emphasis on "all" and/or "must", if you like.

sopel/cli/plugins.py
```diff
@@ -0,0 +1,397 @@
...
+        elif disabled_only and is_enabled:
+            # hide enabled plugins when displaying disabled only
+            continue
+
+        description = {
+            'name': name,
+            'status': 'enabled' if is_enabled else 'disabled',
+        }
+
+        # option meta description from the plugin itself
```
dgw: This is a guess; I'm not 100% sure this is what you meant, but I try anyway!

```suggestion
        # optional meta description from the plugin itself
```

sopel/cli/plugins.py
```diff
@@ -0,0 +1,397 @@
...
+        tools.stderr('No plugin named %s' % plugin_name)
+        return 1
+
+    plugin, is_enabled = usable_plugins[plugin_name]
+    description = {
+        'name': plugin_name,
+        'status': 'enabled' if is_enabled else 'disabled',
+    }
+
+    # option meta description from the plugin itself
```
dgw: Same as above.

```suggestion
    # optional meta description from the plugin itself
```

test/plugins/test_plugins_handlers.py
```diff
@@ -0,0 +1,84 @@
+# coding=utf-8
+"""Test for the ``sopel.plugins.handlers`` module."""
```
dgw:
```suggestion
"""Tests for the ``sopel.plugins.handlers`` module."""
```

sopel/cli/plugins.py
Exirel: Sadly no. I tried but empty lines are removed. :(

sopel/cli/plugins.py
Exirel: Oh, yeah, that was my intent.

sopel/cli/plugins.py
dgw: I thought so. Found a Stack Overflow question asking how to insert newlines in option `help`, which also strips out whitespace. Even though I couldn't test yet, it seemed pretty obvious that the same string-prep logic would run on both that and parser `description`.

Exirel: Ohhhh... [formatter_class](https://docs.python.org/2/library/argparse.html#formatter-class) **is a thing**.

Unresolved then.

Exirel: Done! A combination of `formatter_class=argparse.RawTextHelpFormatter` and `inspect.cleandoc` to prevent extra tabs, and the output is much *much* **much** better.

sopel/cli/plugins.py
```diff
@@ -0,0 +1,397 @@
...
+        dest='name_only',
+        action='store_true',
+        default=False)
+
+    # sopel-plugin disable
+    disable_parser = subparsers.add_parser(
+        'disable',
+        help="Disable a Sopel plugins",
+        description="""
+            Disable a Sopel plugin by its name, no matter where it comes from.
```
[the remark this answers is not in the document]
Exirel: Very easy. :)

sopel/cli/plugins.py
```diff
@@ -0,0 +1,397 @@
...
+
+def handle_list(options):
+    """List Sopel plugins"""
+    settings = utils.load_settings(options)
+    no_color = options.no_color
+    name_only = options.name_only
+    enabled_only = options.enabled_only
+    disabled_only = options.disabled_only
+
+    for name, info in plugins.get_usable_plugins(settings).items():
```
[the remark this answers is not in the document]
Exirel: Good idea.

About `coretasks`, I prefer to list it with the other plugins: it tells users that they can use `sopel show coretasks`.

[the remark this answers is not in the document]
dgw: But why should a user want to?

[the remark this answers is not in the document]
Exirel: As of today... probably for the sheer curiosity, as it's not very useful, I admit.

However, I would like very much to display a list of commands, rules, and event that a plugin uses, and in that regard, inspecting `coretasks` would be very useful, I believe.

sopel/cli/plugins.py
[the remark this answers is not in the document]
Exirel: Not so easy but done nonetheless! I'm going to improve a bit the text formatting.

sopel/cli/plugins.py
```diff
@@ -0,0 +1,506 @@
...
+        settings.save()
+    else:
+        return 0  # nothing to disable or save, but not an error case
+
+    # display plugins actually disabled by the command
+    print(utils.get_many_text(
+        actually_disabled,
+        one='Plugin {item} disabled.',
+        two='Plugin {first} and {second} disabled.',
+        many='Plugin {left}, and {last} disabled.'
```
dgw: `two` and `many` for this message should say "Plugins".

sopel/cli/plugins.py
```diff
@@ -0,0 +1,506 @@
...
+        settings.save()
+    else:
+        return 0  # nothing to disable or save, but not an error case
+
+    # display plugins actually disabled by the command
+    print(utils.get_many_text(
+        actually_enabled,
+        one='Plugin {item} enabled.',
+        two='Plugin {first} and {second} enabled.',
+        many='Plugin {left}, and {last} enabled.'
```
dgw: `two` and `many` for this message should say "Plugins".

sopel/cli/plugins.py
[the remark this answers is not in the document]
dgw: Ah, when that's implemented, showing `coretasks` will be useful, I agree.

PR #1589, opened 2019-05-08
Fix up some docstrings for 6.6.x

dgw:
While I was using the live website over the weekend to reference Sopel's API docs, some annoyances jumped out at me in the page about other API features (formatting, various tools).

Today, I decided to fix them. Most of this is touching stuff that hasn't been overhauled (#1565) yet, but I'll probably have to deal with a merge conflict or two on stuff in `tools` that I did already fix in #1563. It's worth doing so our live site can have a bit better content—especially since I fixed an actual mistake!

### DIFF
--- a/sopel/formatting.py
+++ b/sopel/formatting.py
@@ -93,9 +93,10 @@ def _get_color(color):
 def color(text, fg=None, bg=None):
     """Return the text, with the given colors applied in IRC formatting.
 
-    The color can be a string of the color name, or an integer between 0 and
-    99. The known color names can be found in the `colors` class of this
-    module."""
+    The color can be a string of the color name, or an integer between 0 and 99.
+    The known color names can be found in the :class:`colors` class of this
+    module.
+    """
     if not fg and not bg:
         return text
 
@@ -133,7 +134,8 @@ def hex_color(text, fg=None, bg=None):
 
     The color can be provided with a string of either 3 or 6 hexadecimal digits.
     As in CSS, 3-digit colors will be interpreted as if they were 6-digit colors
-    with each digit repeated (e.g. color `c90` is identical to `cc9900`)."""
+    with each digit repeated (e.g. color ``c90`` is identical to ``cc9900``).
+    """
     if not fg and not bg:
         return text
 
--- a/sopel/tools/target.py
+++ b/sopel/tools/target.py
@@ -19,11 +19,14 @@ class User(object):
         self.channels = {}
         """The channels the user is in.
 
-        This maps channel name ``Identifier``\\s to ``Channel`` objects."""
+        This maps channel name :class:`~sopel.tools.Identifier`\\s to
+        :class:`Channel` objects.
+        """
         self.account = None
         """The IRC services account of the user.
 
-        This relies on IRCv3 account tracking being enabled."""
+        This relies on IRCv3 account tracking being enabled.
+        """
         self.away = None
         """Whether the user is marked as away."""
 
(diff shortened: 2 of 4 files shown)


sopel/tools/target.py
```diff
@@ -19,11 +19,14 @@ def __init__(self, nick, user, host):
         self.channels = {}
         """The channels the user is in.
 
-        This maps channel name ``Identifier``\\s to ``Channel`` objects."""
+        This maps channel name :class:`Identifier <sopel.tools.Identifier>`\\s
```
Exirel: To achieve the same thing, you can add a `~ ` in front of the class path:
```suggestion
        This maps channel name :class:`~sopel.tools.Identifier`\\s
```

See the [sphinx documentation on cross-referencing syntax](https://www.sphinx-doc.org/en/2.0/usage/restructuredtext/roles.html#cross-referencing-syntax).

sopel/tools/target.py
```diff
@@ -52,12 +55,16 @@ def __init__(self, name):
         self.users = {}
         """The users in the channel.
 
-        This maps username ``Identifier``\\s to channel objects."""
+        This maps username :class:`Identifier <sopel.tools.Identifier>`\\s to
```
Exirel: Ditto.
```suggestion
        This maps username :class:`~sopel.tools.Identifier`\\s to
```

sopel/tools/target.py
```diff
@@ -52,12 +55,16 @@ def __init__(self, name):
...
+        :class:`User` objects.
+        """
         self.privileges = {}
         """The permissions of the users in the channel.
 
-        This maps username ``Identifier``s to bitwise integer values. This can
-        be compared to appropriate constants from ``sopel.module``."""
+        This maps username :class:`Identifier <sopel.tools.Identifier>`\\s to
```
Exirel:
```suggestion
        This maps username :class:`~sopel.tools.Identifier`\\s to
```

sopel/tools/target.py
```diff
@@ -52,12 +55,16 @@ def __init__(self, name):
...
+        bitwise integer values. This can be compared to appropriate constants
+        from :class:`sopel.module`.
```
Exirel: This isn't a class, but a module:

```suggestion
        from :mod:`sopel.module`.
```

dgw: Sphinx doesn't actually care (the cross-ref works even though it's not a class)—but you're right, we should use the correct reference domain anyway.

sopel/tools/target.py
dgw: Eh, this is what I get for inferring what `~` does instead of reading the docs. I thought it was a sort of "implied parent" like `~` in path names on \*nix, and referred to the current module. :rofl:

sopel/formatting.py
```diff
@@ -94,8 +94,9 @@ def color(text, fg=None, bg=None):
     """Return the text, with the given colors applied in IRC formatting.
 
     The color can be a string of the color name, or an integer between 0 and
-    99. The known color names can be found in the `colors` class of this
-    module."""
+    99. The known color names can be found in the :class:`~.colors` class of
```
dgw: @Exirel should I leave this dot out, then?

```suggestion
    99. The known color names can be found in the :class:`~colors` class of
```

sopel/tools/target.py
Exirel: Yeah, you have the `.` for something similar:

```
:class:`.ClassInTheCurrentModule`
```

sopel/formatting.py
Exirel: I think you can, yes. The best way to know is to run Sphinx, to be sure!

PR #1581, opened 2019-05-01
remind: use module.interval instead of custom monitoring

Exirel:
Started by the conversation in #1543, this PR removes the usage of `time.sleep()` in a rogue thread. Now that #1557 has been merged and with the future plugin interface from #1479, job scheduling is a much better option for the remind plugin.

This PR was a good occasion to clean the code, to add some tests, and to fix some PEP 8 violations (I used `pylint` for that one). However, this will cause troubles for #1162 at first, but I'll figure this out later.

PR #1162, opened 2016-12-29
remind: add support for dates in future

alanhuang122:
This pull request addresses issue #1148 by adding support for future dates, not just the next occurrence of the specified time.

PR #1590, opened 2019-05-08
remind: add support for date in .at command

### LINKED ISSUE #1148: [remind] .at does not accept a future date, limiting reminders to next 24 hours
I do not see any past (closed) issue about this, but the possible use cases beg the question: Why doesn't `.at` take a date in addition to a time?

It would allow scheduling reminders at a specific time more than 24 hours in the future, without first manually calculating the time difference and feeding it to `.in` instead.

Exirel:
**This PR is based on #1581, and it replaces PR #1162.**
Also fix #1148.

It can replace PR #1581 if someone is brave enough to do the review of all the changes in one go.
I added an extensive test-suite to make sure the remind `.at` command works as intended.

I made `.at` opinionated about how date are expressed, with these choices:

* YYYY-mm-dd is valid, as dd-mm-YYYY
* dd-mm-YY is also valid
* screw mm-dd-YY (or mm-dd-YYYY) format
* dd-mm is valid too, as mm-YYYY and YYYY-mm
* but there is no such thing as mm-YY or YY-mm
* date separator can be either `-`, `.`, or `/`, and it must be the same between each parts

Even with all the unit-tests I added, I spent a good one hour with a live Sopel instance, and everything went according to the specs.

### DIFF
--- a/sopel/modules/remind.py
+++ b/sopel/modules/remind.py
@@ -332,7 +332,7 @@ class TimeReminder(object):
     def get_duration(self, today=None):
         """Get the duration between the reminder and ``today``
 
-        :param today: aware datetime to compare to; defaults to current time in UTC
+        :param today: aware datetime to compare to; defaults to current time
         :type today: aware :class:`datetime.datetime`
         :return: The duration, in second, between ``today`` and the reminder.
         :rtype: :class:`int`
@@ -345,6 +345,12 @@ class TimeReminder(object):
         will represent it as a negative number of days, and a positive number
         of seconds: since it returns the number of seconds, any past reminder
         will be transformed into a future reminder the next day.
+
+        .. seealso::
+
+            The :mod:`datetime` built-in module's documentation explains what
+            is an "aware" datetime.
+
         """
         if not today:
             today = datetime.now(self.timezone)
@@ -373,8 +379,8 @@ def parse_regex_match(match, default_timezone=None):
     """Parse a time reminder from ``match``
 
     :param match: :obj:`~.REGEX_AT`'s matching result
-    :param default_timezone: Default trigger's timezone
-                             (from the nick's, channel's, or config's timezone)
+    :param default_timezone: timezone used when ``match`` doesn't have one;
+                             defaults to ``UTC``
     :rtype: :class:`TimeReminder`
     """
     try:
--- a/sopel/modules/remind.py
+++ b/sopel/modules/remind.py
@@ -18,7 +18,7 @@ import time
 import pytz
 
 from sopel import tools, module
-from sopel.tools.time import get_timezone, format_time
+from sopel.tools.time import get_timezone, format_time, validate_timezone
 
 
 def get_filename(bot):
@@ -96,8 +96,19 @@ def dump_database(filename, data):
                 database.write(line)
 
 
-def create_reminder(bot, trigger, duration, message, timezone):
-    """Create a reminder into the ``bot``'s database and reply to the sender"""
+def create_reminder(bot, trigger, duration, message):
+    """Create a reminder into the ``bot``'s database and reply to the sender
+
+    :param bot: the bot's instance
+    :type bot: :class:`~sopel.bot.SopelWrapper`
+    :param trigger: the object that triggered the call
+    :type trigger: :class:`~sopel.trigger.Trigger`
+    :param int duration: duration from now, in seconds, until ``message``
+                         must be reminded
+    :param str message: message to be reminded
+    :return: the reminder's timestamp
+    :rtype: :class:`int`
+    """
     timestamp = int(time.time()) + duration
     reminder = (trigger.sender, trigger.nick, message)
     try:
@@ -106,18 +117,7 @@ def create_reminder(bot, trigger, duration, message, timezone):
         bot.rdb[timestamp] = [reminder]
 
     dump_database(bot.rfn, bot.rdb)
-
-    if duration >= 60:
-        human_time = format_time(
-            bot.db,
-            bot.config,
-            timezone,
-            trigger.nick,
-            trigger.sender,
-            datetime.utcfromtimestamp(timestamp))
-        bot.reply('Okay, will remind at %s' % human_time)
-    else:
-        bot.reply('Okay, will remind in %s secs' % duration)
+    return timestamp
 
 
 def setup(bot):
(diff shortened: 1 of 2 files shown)


Exirel:
Wouhouh, tests pass with Python 2.7! That was my main concern at this point, given the extensive test suite I ran with Python 3.

sopel/modules/remind.py
```diff
@@ -226,47 +215,219 @@ def remind_in(bot, trigger):
...
+                'date2',
+                'date3',
+                'message',
+            ]
+        )
+
+    def get_duration(self, today=None):
+        """Get the duration between the reminder and ``today``
+
+        :param today: aware datetime to compare to, default to UTC now
```
dgw: "aware"?

sopel/modules/remind.py
```diff
@@ -226,47 +215,219 @@ def remind_in(bot, trigger):
...
+        timezone,
+        match.group('date1'),
+        match.group('date2'),
+        match.group('date3'),
+        match.group('message')
+    )
 
 
 @module.commands('at')
 @module.example('.at 13:47 Do your homework!')
```
dgw: This deserves more examples. With #1403 merged, more than one example can appear in the help output.

test/modules/test_modules_remind.py
```diff
@@ -0,0 +1,426 @@
...
+    ('00', '00', '0000'),
+    ('00', '00', '00'),
+    # year full but day and/or month empty
+    ('2019', '00', '00'),
+    ('2019', '00', '05'),
+    ('2019', '03', '00'),
+    ('00', '00', '2019'),
+    ('00', '05', '2019'),
+    ('03', '00', '2019'),
+    # no year provided byt day and/or month empty
```
dgw:
```suggestion
    # no year provided but day and/or month empty
```

sopel/modules/remind.py
```diff
@@ -156,74 +215,219 @@ def remind(bot, trigger):
...
+                 date2,
+                 date3,
+                 message):
+        self.hour = hour
+        self.minute = minute
+        self.second = second
+        self.timezone = pytz.timezone(timezone)
+        self.date1 = date1
+        self.date2 = date2
+        self.date3 = date3
```
dgw: This `date1`/`date2`/`date3` stuff is fine for the constructor to accept, but it should convert them into real `year`/`month`/`day` properties immediately instead of leaving the heavy lifting to `get_duration`.

sopel/modules/remind.py
Exirel: Yup, by default, a `datetime.datetime` object are called "naive" datetime, ie. they don't know their timezone. An aware datetime object knows about its timezone, represented by their `tzinfo` property. I discovered that while working on Django, and now it's in the [Python 3 documentation](https://docs.python.org/3/library/datetime.html):

> There are two kinds of date and time objects: “naive” and “aware”.

sopel/modules/remind.py
Exirel: Good point. I remembered that there was a PR in progress, but totally forgot to check if it was merged. I was way too focus on unit-testing that.

sopel/modules/remind.py
Exirel: Hmmmmm after further consideration, I think I can do something for that. Although, I won't validate the date until `get_duration` is called.

sopel/modules/remind.py
dgw: Would it, ahem, "fuck shit up", if "aware" was part of the `:type:` instead of `:param:`?

Exirel: That's a real good question. I need to check that.

dgw:
Would like to merge this hot on the heels of #1581, so I rebased it myself to get it ready for review sooner. @Exirel if you would rather I didn't do such things in the future on your PRs, just say so. :)

sopel/modules/remind.py
```diff
@@ -226,47 +226,227 @@ def remind_in(bot, trigger):
...
+        will be transformed into a future reminder the next day.
+        """
+        if not today:
+            today = datetime.now(self.timezone)
+        else:
+            today = today.astimezone(self.timezone)
+
+        year = self.year if self.year is not None else today.year
+        month = self.month if self.month is not None else today.month
+        day = self.day if self.day is not None else today.day
```
dgw: Is it not safe enough to do e.g. `year = self.year or today.year`? When might `self.year` be not `None` but still falsy?

sopel/modules/remind.py
```diff
@@ -226,47 +226,227 @@ def remind_in(bot, trigger):
...
+
+        return duration
+
+
+def parse_regex_match(match, default_timezone=None):
+    """Parse a time reminder from ``match``
+
+    :param match: :obj:`~.REGEX_AT`'s matching result
+    :param default_timezone: Default trigger's timezone
+                             (from the nick's, channel's, or config's timezone)
```
dgw: I think the Franglish is so strong here that I can't even grok enough of what was meant to suggest an improvement. :/ In general, though (and this applies to the `match` param too), your parameter descriptions tend to lean really heavily on possessives, and that tends to make them confusing (or at least, makes them read strangely). Avoiding possessives as much as possible when writing these might help you create clearer descriptions. :)

sopel/modules/remind.py
```diff
@@ -226,47 +226,227 @@ def remind_in(bot, trigger):
...
 @module.commands('at')
-@module.example('.at 13:47 Do your homework!')
+@module.example('.at 13:47 Do your homework!', user_help=True)
+@module.example('.at 03:14:07 2038-01-19 End of signed 32-bit int timestamp',
+                user_help=True)
+@module.example('.at 00:01 25/12 Open your gift!', user_help=True)
 def remind_at(bot, trigger):
-    """
-    Gives you a reminder at the given time. Takes `hh:mm:ssTimezone message`.
+    """Gives you a reminder at the given times.
```
dgw: Who let nemesi**s** in here?!

```suggestion
    """Gives you a reminder at the given time.
```

(I mean, at least it's not a [dog](https://www.youtube.com/watch?v=KYgK97aioNA).)

sopel/modules/remind.py
```diff
@@ -226,47 +226,227 @@ def remind_in(bot, trigger):
...
+                'month',
+                'day',
+                'message',
+            ]
+        )
+
+    def get_duration(self, today=None):
+        """Get the duration between the reminder and ``today``
+
+        :param today: aware datetime to compare to, default to UTC now
```
dgw: Right, I have a non-"aware"-related thing for this line, assuming you want to leave that in even though it's also in the `type` declaration now.

```suggestion
        :param today: aware datetime to compare to; defaults to current time in UTC
```

If that's *just slightly* too long (I'm not checking line widths here, cuz GitHub UI), maybe "current UTC time" would fit? Just covering bases. :grinning:

sopel/modules/remind.py
Exirel: No, because I convert beforehand `date1` to an integer, so `self.year` could be `0`, and `0 or today.year` returns `today.year`. It means that this would not raise:

    <@Exirel> .at 13:37 0000-00-00 the first message between BC and AD

and instead it would consider today's date. If user sends an invalid input, we should gently scream at them.

Quick note from the [Zen of Python](https://www.python.org/dev/peps/pep-0020/):

> In the face of ambiguity, refuse the temptation to guess.

sopel/modules/remind.py
Exirel: Hm... could you give me a description for the default timezone to use when `match` doesn't have one?

Exirel:
Uh oh... apparently I made a mistake: since you forced-push to that branch, I had merge conflict locally for some unknown reason, and I think I messed-up the squashing... :-/

sopel/modules/remind.py
Exirel: I tried to go with that:

> timezone used when `match` doesn't have one; defaults to `UTC`

Exirel:
After double check, I believe it's all good. Ready for a new (and hopefully last) round of review before merge.

sopel/modules/remind.py
```diff
@@ -226,47 +226,227 @@ def remind_in(bot, trigger):
...
+                'month',
+                'day',
+                'message',
+            ]
+        )
+
+    def get_duration(self, today=None):
+        """Get the duration between the reminder and ``today``
+
+        :param today: aware datetime to compare to; defaults to current time
```
dgw: Would it be useful to put a `.. seealso::` at the end of this linking to [the Python docs for `datetime`](https://docs.python.org/3/library/datetime.html) (which explain what "aware datetime" means)?

Exirel: Absolutely!

Exirel:
@dgw done

PR #1592, opened 2019-05-09
clock: improve timezone guessing and refactor time-related utility

### LINKED ISSUE #1591: clock: Invalid timezone name silently uses default
This is kind of a bug in `tools.time.get_timezone()`, maybe, but it manifests like this:

```
<dgw>   !time this/doesnt_exist
<Pandy> Thursday, May 09, 2019 04:15:47 (EDT)
```

That's Sopel using the `core.default_timezone` set for the whole bot, silently, even though it was passed a timezone name. It should throw the error message that's defined for timezone names that can't be found by any means of lookup (whether nickname, channel name, or tzdata lookup):

https://github.com/sopel-irc/sopel/blob/eeb48f947f0421b70bcf4e472df1acfd5e8363c4/sopel/modules/clock.py#L65

Instead of a useful error, the calling user gets a misleading false-positive response.

Exirel:
Fixes #1591 

**Edit**: after IRC conversation, this PR became bigger and bigger, so its scope is now from "fix an oddity in the clock plugin" to "these clock/time stuff are too odd to keep them like that".

It's a nice revamp that tries to keep the convenience of existing commands with less confusing or odd behaviours.

---

Instead of using the default user/channel/bot's timezone, when `.time` is used **with** a timezone, Sopel displays an error message if this timezone is invalid:

    15:12:06 <@Exirel> .time
    15:12:07 <Sopel> 2019-05-09 - 13:12:06UTC
    15:12:20 <@Exirel> .time CET
    15:12:20 <Sopel> 2019-05-09 - 15:12:20CEST
    15:12:33 <@Exirel> .time Not/A_Timezone
    15:12:33 <Sopel> Could not find timezone Not/A_Timezone.

While I was working on that plugin, I spent the extra time to clean its imports.

### DIFF
--- a/sopel/modules/clock.py
+++ b/sopel/modules/clock.py
@@ -9,16 +9,16 @@ https://sopel.chat
 """
 from __future__ import unicode_literals, absolute_import, print_function, division
 
-from sopel.module import commands, example, OP
-from sopel.tools.time import (
-    get_timezone, format_time, validate_format, validate_timezone
-)
+from sopel import module, tools
 from sopel.config.types import StaticSection, ValidatedAttribute
-
-try:
-    import pytz
-except ImportError:
-    pytz = None
+from sopel.tools.time import (
+    format_time,
+    get_channel_timezone,
+    get_nick_timezone,
+    get_timezone,
+    validate_format,
+    validate_timezone
+)
 
 
 class TimeSection(StaticSection):
@@ -26,14 +26,12 @@ class TimeSection(StaticSection):
         'tz',
         parse=validate_timezone,
         serialize=validate_timezone,
-        default='UTC'
-    )
+        default='UTC')
     """Default time zone (see https://sopel.chat/tz)"""
     time_format = ValidatedAttribute(
         'time_format',
         parse=validate_format,
-        default='%Y-%m-%d - %T%Z'
-    )
+        default='%Y-%m-%d - %T%Z')
     """Default time format (see http://strftime.net)"""
 
 
@@ -55,8 +53,10 @@ def setup(bot):
     bot.config.define_section('clock', TimeSection)
 
 
-@commands('t', 'time')
-@example('.t America/New_York')
+@module.commands('t', 'time')
+@module.example('.t America/New_York')
+@module.example('.t Exirel')
+@module.example('.t #sopel')
 def f_time(bot, trigger):
     """Returns the current time."""
     if trigger.group(2):
@@ -71,56 +71,56 @@ def f_time(bot, trigger):
     bot.say(time)
 
 
-@commands('settz', 'settimezone')
-@example('.settz America/New_York')
+@module.commands('settz', 'settimezone')
+@module.example('.settz America/New_York')
 def update_user(bot, trigger):
+    """Set your preferred timezone.
+
+    Most timezones will work, but it's best to use one from
+    <https://sopel.chat/tz>.
     """
-    Set your preferred time zone. Most timezones will work, but it's best to
-    use one from <https://sopel.chat/tz>.
-    """
-    if not pytz:
-        bot.reply("Sorry, I don't have timezone support installed.")
+    argument = trigger.group(2)
+    if not argument:
+        bot.reply("What timezone do you want to set? Try one from "
+                  "https://sopel.chat/tz")
+        return
+
+    try:
+        zone = validate_timezone(argument)
+    except ValueError:
+        bot.say(
+            'I don\'t know that timezone. Try one from https://sopel.chat/tz')
+        return
+
+    bot.db.set_nick_value(trigger.nick, 'timezone', zone)
+
+    if len(zone) < 4:
+        bot.say(
+            'Okay, %s, but you should use one from https://sopel.chat/tz '
+            'if you use DST.' % trigger.nick)
     else:
-        tz = trigger.group(2)
-        if not tz:
-            bot.reply("What timezone do you want to set? Try one from "
-                      "https://sopel.chat/tz")
-            return
-        if tz not in pytz.all_timezones:
-            bot.reply("I don't know that time zone. Try one from "
-                      "https://sopel.chat/tz")
-            return
-
-        bot.db.set_nick_value(trigger.nick, 'timezone', tz)
-        if len(tz) < 7:
-            bot.say("Okay, {}, but you should use one from https://sopel.chat/tz "
-                    "if you use DST.".format(trigger.nick))
-        else:
-            bot.reply('I now have you in the %s time zone.' % tz)
+        bot.reply('I now have you in the %s timezone.' % zone)
 
 
-@commands('gettz', 'gettimezone')
-@example('.gettz [nick]')
+@module.commands('gettz', 'gettimezone')
+@module.example('.gettz [nick]')
 def get_user_tz(bot, trigger):
     """Gets a user's preferred time zone; will show yours if no user specified."""
-    if not pytz:
-        bot.reply("Sorry, I don't have timezone support installed.")
+    nick = trigger.group(2)
+    if not nick:
+        nick = trigger.nick
+
+    nick = nick.strip()
+    zone = get_nick_timezone(bot.db, nick)
+
+    if zone:
+        bot.say('%s\'s time zone is %s.' % (nick, zone))
     else:
-        nick = trigger.group(2)
-        if not nick:
-            nick = trigger.nick
-
-        nick = nick.strip()
-
-        tz = bot.db.get_nick_value(nick, 'timezone')
-        if tz:
-            bot.say('%s\'s time zone is %s.' % (nick, tz))
-        else:
-            bot.say('%s has not set their time zone' % nick)
+        bot.say('%s has not set their time zone' % nick)
 
 
-@commands('settimeformat', 'settf')
-@example('.settf %Y-%m-%dT%T%z')
+@module.commands('settimeformat', 'settf')
+@module.example('.settf %Y-%m-%dT%T%z')
 def update_user_format(bot, trigger):
     """
     Sets your preferred format for time. Uses the standard strftime format. You
@@ -153,8 +153,8 @@ def update_user_format(bot, trigger):
               % timef)
 
 
-@commands('gettimeformat', 'gettf')
-@example('.gettf [nick]')
+@module.commands('gettimeformat', 'gettf')
+@module.example('.gettf [nick]')
 def get_user_format(bot, trigger):
     """Gets a user's preferred time format; will show yours if no user specified."""
     nick = trigger.group(2)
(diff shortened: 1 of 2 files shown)


Exirel:
> We _could_ change the "standard" function signature for `configure`, but not in this PR. The argument is called `config` by long-standing convention.

Oh boy. I sure knew I was forgetting something important in here. You're right, I'll revert that change!

Exirel:
@dgw rebased & fixed. I think you should like it way more now. And, sure thing, I didn't just revert my change, I **had to** do something about that import. :rofl:

Exirel:
Fixed something with that last commit:

```
09:02:40 <@Exirel> .setctz America/Chicago
09:02:41 <Icicle> Exirel: I now have #icicle in the America/Chicago time zone.
09:02:42 <@Exirel> .t #icicle
09:02:43 <Icicle> 2019-05-11 - 02:02:42CDT
09:02:44 <@Exirel> .t Exirel
09:02:45 <Icicle> 2019-05-11 - 09:02:46CEST
09:02:52 <@Exirel> .t Asia/Shanghai
09:02:53 <Icicle> 2019-05-11 - 15:07:26CST
09:03:03 <@Exirel> .t Israel
09:03:04 <Icicle> 2019-05-11 - 10:03:03IDT
09:03:20 <@Exirel> .t
09:03:21 <Icicle> 2019-05-11 - 09:03:19CEST
09:03:30 <@Exirel> .t Not/ATimeZone
09:03:30 <Icicle> Could not find timezone Not/ATimeZone.
```

I don't know how I could have miss that the first time. Guess I wasn't paying attention... :(

kwaaak:
I am going to hijack this PR to talk about finding the timezone and current time for a place.
There is an open API available from [Teleport](https://developers.teleport.org/api/reference/#!/cities/searchCities) that would make this very easy.

Now, if the `.t` input also includes an IRC nicknames, it would be very hard to distinguish between searches for the time in a city and users.

There already is the `.gettz` command which looks up the timezone of a user.

Exirel:
> I am going to hijack this PR to talk about finding the timezone and current time for a place.

Oh, that could be a cool feature. In case there is a confusion: both Europe/Paris, America/Chicago, and Israel, are timezone available from the IANA TZ database (which is used by `pytz` hunder the hood). So at the moment and in this PR, `.time` is not looking for the TZ of a place. It just happened that `Israel` is a deprecated (but existing) timezone for the `Asia/Jerusalem` timezone.

> Now, if the `.t` input also includes an IRC nicknames, it would be very hard to distinguish between searches for the time in a city and users.

Right now, that's what it does. I just made sure to keep this behavior in this PR. The only change here is that now, if you provide something that isn't a channel or a nick known to the bot, and if that something isn't a valid timezone, **then**, Sopel will scream about it (but not too much).

dgw:
> I am going to hijack this PR to talk about

If your comment starts with something like this, you should probably be making it on IRC or in its own issue… PR discussion should be (mostly) restricted to _that PR_ and what needs to be done to make it ready to merge, and/or what should be done after that PR is merged (documentation updates, etc.).

A "new" feature (we already support finding the time for a city, just only if it's given as a TZ database zone name) like this is great fodder for IRC discussion or a feature request issue. :smile_cat:

sopel/modules/clock.py
```diff
@@ -54,14 +54,29 @@ def setup(bot):
...
     if trigger.group(2):
-        zone_name = trigger.group(2).strip()
+        argument = trigger.group(2).strip()
+        channel_or_nick = tools.Identifier(argument)
+
         try:
-            zone = validate_timezone(zone_name)
+            if channel_or_nick in bot.users:
+                zone = get_timezone(bot.db, bot.config, None, channel_or_nick,
+                                    trigger.sender)
```
dgw: This seems dangerous. Falling back on the current channel's timezone could give misleading results if the given nick isn't in the current channel. Hell, they could be misleading if the given nick *is* in the current channel.

Actually, falling back on anything but the user's own timezone setting is probably not a good idea. I think this should emit an error if the user hasn't set a timezone, rather than give an unclear result that may or may not be in the user's own timezone.

sopel/modules/clock.py
```diff
@@ -54,14 +54,29 @@ def setup(bot):
...
+            elif channel_or_nick in bot.channels:
+                zone = get_timezone(bot.db, bot.config, None, None,
+                                    channel_or_nick)
```
dgw: Same concern here. Falling back on the bot's default timezone is potentially misleading if the channel doesn't have one set for itself. Again, I think an error is better.

sopel/modules/clock.py
Exirel: Well, it's kind of what the previous code did, with less success I believe. I think I went with a conservative approach here: I kept the same behavior as today.

I'd suggest to do another PR to raise more errors, because that means changing the behavior of (multiple) command(s).

Exirel: I'm having second thought on that one.

Let's raise an error.

Exirel:
@dgw so I've added some specific functions for channel/nick timezone, so I don't break backward compatibility for any plugin using `get_timezone`, and I use them to let the bot warns the user that something went wrong.

sopel/modules/clock.py
```diff
@@ -55,72 +57,89 @@ def setup(bot):
...
-            return
+        argument = trigger.group(2).strip()
+        channel_or_nick = tools.Identifier(argument)
+
+        if channel_or_nick in bot.users:
+            zone = get_nick_timezone(bot.db, channel_or_nick)
+            if zone is None:
+                bot.say('Could not find timezone for nick %s.' % argument)
+                return
+        elif channel_or_nick in bot.channels:
```
dgw: I have to say, this approach seems cute at first glance until you realize that it makes getting the time for a user who is offline, or a channel that the bot knows but is not currently joined to, impossible.

While it's easy to decide if the input is a channel name (cast it to `Identifier` and check `is_nick`), the line between nickname and timezone name is less clear. A non-channel input containing `/` is almost certainly a timezone name (standard IRC doesn't allow that character in nicks, but often does allow it in channel names). But things like `CDT` are valid as nicks AND timezone names, so we'd need to prioritize one or the other.

How about this resolution order?

* Input is channel name?
  * Lookup channel TZ
* Not a channel
  * Input contains slash?
    * Attempt to validate and return as zone name
  * No slash
    * Input <= 4 characters?
      * Try as zone name first, then fall back to nick lookup
    * Input > 4 characters
      * Try nick lookup

Maybe there's a less convoluted way to do it. Maybe we should do a strict channel -> nick -> zone resolution order without trying to guess based on length or presence of `/`. But I do think it's important that we _at least_ enable the (anecdotally) common use case of checking what time it is for someone who is offline, e.g. checking whether they're offline because it's the middle of the night for them.

sopel/modules/clock.py
```diff
@@ -55,72 +57,89 @@ def setup(bot):
...
     """
-    if not pytz:
-        bot.reply("Sorry, I don't have timezone support installed.")
+    tz = trigger.group(2)
+    if not tz:
+        bot.reply("What timezone do you want to set? Try one from "
+                  "https://sopel.chat/tz")
+        return
+
+    if tz not in pytz.all_timezones:
```
dgw: How about taking advantage of `tools.time.validate_timezone()` here? That will make lazy commands like `.settz america/chicago` Just Work™, reducing effort for the user. (Totally fine if you want to make that a separate PR, but this one is already touching quite a bit more than just the behavior of `.t`. Honestly, we should probably give this PR a broader title. I'll need to remember API additions for the changelog entry, as well as the bugfix/tweak.)

sopel/modules/clock.py
```diff
@@ -55,72 +57,89 @@ def setup(bot):
...
+        bot.reply("I don't know that time zone. Try one from "
+                  "https://sopel.chat/tz")
+        return
+
+    bot.db.set_nick_value(trigger.nick, 'timezone', tz)
+    if len(tz) < 7:
```
dgw: Shouldn't this be `<= 4`?

Though even that would still yield false warnings for e.g. `Iran`. The current value emits the warning for perfectly valid zone names like `Israel`, `Egypt`, and others, which is not great. (As with the above validity check, this too could be done in a subsequent PR, but we might as well do it now while we are touching the code anyway.)

sopel/modules/clock.py
```diff
@@ -55,72 +57,89 @@ def setup(bot):
...
 
-        nick = nick.strip()
+    nick = nick.strip()
 
-        tz = bot.db.get_nick_value(nick, 'timezone')
-        if tz:
-            bot.say('%s\'s time zone is %s.' % (nick, tz))
-        else:
-            bot.say('%s has not set their time zone' % nick)
+    tz = bot.db.get_nick_value(nick, 'timezone')
```
dgw: Use new `tools.time.get_nick_timezone()` function?

sopel/modules/clock.py
```diff
@@ -172,67 +191,58 @@ def get_user_format(bot, trigger):
...
+@module.require_privilege(module.OP)
 def update_channel(bot, trigger):
     """Set the preferred timezone for the channel."""
-    if bot.channels[trigger.sender].privileges[trigger.nick] < OP:
+    tz = trigger.group(2)
+    if not tz:
+        bot.reply("What timezone do you want to set? Try one from "
+                  "https://sopel.chat/tz")
+        return
+    if tz not in pytz.all_timezones:
```
dgw: Same here as for the analogous check in `.settz` command.

sopel/modules/clock.py
```diff
@@ -172,67 +191,58 @@ def get_user_format(bot, trigger):
...
+        bot.reply("I don't know that time zone. Try one from "
+                  "https://sopel.chat/tz")
         return
-    elif not pytz:
-        bot.reply("Sorry, I don't have timezone support installed.")
+
+    bot.db.set_channel_value(trigger.sender, 'timezone', tz)
+    if len(tz) < 7:
```
dgw: Same here as for the analogous check in `.settz` command.

sopel/modules/clock.py
```diff
@@ -172,67 +191,58 @@ def get_user_format(bot, trigger):
...
 
-        channel = channel.strip()
+    channel = channel.strip()
 
-        timezone = bot.db.get_channel_value(channel, 'timezone')
-        if timezone:
-            bot.say('%s\'s timezone: %s' % (channel, timezone))
-        else:
-            bot.say('%s has no preferred timezone' % channel)
+    timezone = bot.db.get_channel_value(channel, 'timezone')
```
dgw: Use new `tools.time.get_channel_timezone()` function?

sopel/modules/clock.py
Exirel: I think that we should turn to the [Zen of Python](https://www.python.org/dev/peps/pep-0020/):

> In the face of ambiguity, refuse the temptation to guess.

I'd suggest to create new commands, like `.tn <Nick>`  and ` .tc <Channel>`  instead of trying to be super smart about the input of `.t <zone>`.

* `.t`: timezone from the trigger (nick, then channel, then default)
* ` .t <zone>` : always check for a timezone. If not found, try to see if something else exist, and warn the user that there are specific commands for that
* `.tn`  get trigger's user timezone
* `.tc` get trigger's channel timezone
* `.tn <nick>` and `.tc <channel>`: nick and channel's timezone

It would be simpler, with less guessing and more accurate and predictable behaviours.

Exirel:
@dgw You made several comments about code that I didn't changed myself, even though it looks like I did. What happened is that I removed the ` if/else` blocks because of ` pytz`. By doing so, I had to re-indent blocks of code properly. Would you consider another PR for these comments?

dgw:
> Would you consider another PR for these comments?

Definitely. I think I even said as much? All I'm really worried about here is the change in `.t`'s interpretation of its input.

sopel/modules/clock.py
dgw: My gut feeling is that what we have right now is too convenient to give up that easily. If `.t` can continue trying to do the right thing, it's pretty easy to at least get the channel/nick part of things right, and let nicks shadow timezone names if they exist in Sopel's database.

I would be much happier with a single new command, `.tz`, which takes only a timezone name and does no guessing, for use when `.t`'s guessing finds a shadowing nick. (I already know you at least like this idea, from IRC. :P)

Exirel: Let's do this.

sopel/modules/clock.py
```diff
@@ -9,27 +9,29 @@
 """
 from __future__ import unicode_literals, absolute_import, print_function, division
 
-from sopel.module import commands, example, OP
+import pytz
+
+from sopel import module, tools
+from sopel.config import types
```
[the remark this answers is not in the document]
Exirel: `sopel.tools.time` is not a good example :stuck_out_tongue:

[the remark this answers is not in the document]
dgw: OK, make fun of the `time` module all you like. But **all** of our plugins import specific things `from sopel.config.types`, and let's not start being inconsistent. (It's enough that `test/test_config.py` is already the odd man out.)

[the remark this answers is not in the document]
Exirel: That's only fair, will fix that. :wink:

sopel/tools/time.py
```diff
@@ -50,40 +56,69 @@ def validate_format(tformat):
...
+        return validate_timezone(db.get_nick_value(nick, 'timezone'))
+    except ValueError:
+        return None
+
+
+def get_channel_timezone(db, nick):
+    """Get a channel's timezone from database.
+
+    :param db: Bot's database handler
+    :param nick: Channel identifier
```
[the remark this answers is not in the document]
Exirel: And replace `nick`  by `channel`.

sopel/modules/clock.py
Exirel: What do you think now?

sopel/modules/clock.py
Exirel: Yeah I went for the change you suggested. It's easier to do that in one go, in the end.

sopel/modules/clock.py
dgw: Hard-coded `.command` bad; use `help_prefix`! (I tell _everyone_ this…)

Once I thought about it a bit, the weird logic to only output an error if the zone is not found _and_ the given nick/channel is in the bot's user/channel tracking makes sense. Looks good!

sopel/modules/clock.py
```diff
@@ -157,27 +157,32 @@ def f_time_zone(bot, trigger):
 @module.commands('settz', 'settimezone')
 @module.example('.settz America/New_York')
 def update_user(bot, trigger):
+    """Set your preferred timezone.
+
+    Most timezones will work, but it's best to use one from
+    https://sopel.chat/tz
```
dgw: I think there was a reason for using `<url>`. Per 07edfa4 it was to support auto-generating the plugin docs on the website. Something about bare URLs not getting linked due to how that page gets built…

dgw:
I saw some more little annoying things that you aren't anywhere close to touching, so I'll pretend I didn't see them. 😁 They can get fixed some other time.

Looks like just the one docstring thing for you this time.

sopel/modules/clock.py
Exirel: Oh right. I'll use `help_prefix`. Totally forgot about that. :disappointed:

sopel/modules/clock.py
Exirel: Meh. I'll revert that little change then.

Exirel:
> Looks like just the one docstring thing for you this time.

If it's the <link>, it's fixed. And I fixed the hard-coded prefix, because there is no reason to keep that.

:+1:

Exirel:
@dgw rebased & history reworked as planned. The `git add -p` command was used **at lot**. I like the outcome!

PR #1596, opened 2019-05-10
Bot.say: Handle max_messages better

deathbybandaid:
What this PR aims to fix.

Currently, when there is `excess` text content, `bot.say` calls `bot.msg`, which cycles back to `bot.say`.
This is unnecessary and cyclical. This is amplified further when initially called from `bot.msg`.

These changes change `tools.get_sendable_message` to output a list of strings.
Doing so allows us to utilize `max_messages` more effectively.

### DIFF
--- a/sopel/bot.py
+++ b/sopel/bot.py
@@ -380,53 +380,50 @@ class Sopel(irc.Bot):
         message will contain the entire remainder, which may be truncated by
         the server.
         """
-        excess = ''
         if not isinstance(text, unicode):
             # Make sure we are dealing with unicode string
             text = text.decode('utf-8')
 
-        if max_messages > 1:
-            # Manage multi-line only when needed
-            text, excess = tools.get_sendable_message(text)
+        messages_list = tools.get_sendable_message(text)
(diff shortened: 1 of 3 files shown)


deathbybandaid:
Looks like I did not account for the tests.

dgw:
The "unnecessary and cyclical" `bot.msg` -> `bot.say` ping-pong game you wanted to fix can be eliminated with a ±1-line change in this code block:

https://github.com/sopel-irc/sopel/blob/a36a434fa61bb485a69c14d3bfcc4c3284616960/sopel/bot.py#L426-L429

Then `bot.msg` can only ever be called once, by old code that hasn't been updated to use `bot.say` yet, and `bot.say` will recurse back into itself instead of jumping through additional stack frames calling `bot.msg` in between useful calls.

Rewriting all of `bot.say` function, **and** `tools.get_sendable_message` besides, **and** changing the latter's return type completely, is overkill for fixing what you stated to be the issue here.

Recursion is fine. Recursion is _good_. `for`-loops aren't the only way to solve problems.

PR #1597, opened 2019-05-11
bot: new setup sequence

Exirel:
At first, I had to rework the Job Scheduler in order to test how Sopel registers and unregisters plugins, which was annoying. But then, the solution wasn't perfect: I didn't like to have to manually stop the scheduler for testing purpose—which is time consuming and a bad thing when your test suite grow a bit (a lot at the moment). Aside from that, it also means that plugins where loaded, even if I didn't want them to be, or even needed to be. **This is the first thing I fixed**, by moving the call to `Sopel.setup()` outside of `Sopel.__init__()`.

Then, I realized: wait a minute... why do we setup plugins inside Sopel, but not logging? I couldn't find a good reason, so I moved that into Sopel's setup sequence.

Eventually, I thought: why do I even bother with the scheduler? Why do we need it started before anything else is even loaded? Answer is: we don't. We don't need that, and we don't have to start it whenever we instantiate `Sopel`.

So here we go:

* setup logging
* setup plugins
* start the job scheduler

all wrapped nicely into the old but still relevant `Sopel.setup()` method.

**BONUS POINT**: now the tests are slightly faster, and it's one step closer to facilitate reusability of `sopel.bot.Sopel` class for testing purpose (and everything else).

### DIFF
--- a/sopel/bot.py
+++ b/sopel/bot.py
@@ -153,7 +153,6 @@ class Sopel(irc.Bot):
             self.config.core.nick_blocks = []
         if not self.config.core.host_blocks:
             self.config.core.host_blocks = []
-        self.setup()
 
     @property
     def hostmask(self):
--- a/sopel/bot.py
+++ b/sopel/bot.py
@@ -16,19 +16,18 @@ import sys
 import threading
 import time
 
-from sopel import irc, plugins, tools
+from sopel import irc, logger, plugins, tools
 from sopel.db import SopelDB
 from sopel.tools import stderr, Identifier, deprecated
 import sopel.tools.jobs
 from sopel.trigger import Trigger
 from sopel.module import NOLIMIT
-from sopel.logger import get_logger
 import sopel.loader
 
 
 __all__ = ['Sopel', 'SopelWrapper']
 
-LOGGER = get_logger(__name__)
+LOGGER = logger.get_logger(__name__)
 
 if sys.version_info.major >= 3:
     unicode = str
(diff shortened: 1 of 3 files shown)


HumorBaby:
> ... and it's one step closer to facilitate reusability of sopel.bot.Sopel class for testing purpose

:heart:

sopel/bot.py
```diff
@@ -179,6 +177,23 @@ def write(self, args, text=None):  # Shim this in here for autodocs
...
+
+        """
+        self.setup_logging()
+        self.setup_plugins()
+        self.scheduler.start()
+
+    def setup_logging(self):
+        logger.setup_logging(self)
+
+    def setup_plugins(self):
```
dgw: I can't add a line note below because GitHub arbitrary limitations, but I don't think this function should print "Welcome to Sopel." It should only print "Loading plugins..." (not "modules"), and any errors encountered while doing that.

A "Welcome" message should be printed elsewhere, if at all. Maybe at the beginning of `setup()`? Assuming that would work properly before `setup_logging()` runs.

Exirel: Yeah, I'd rather see no message at all. I think a rework of logging could be a great thing to consider. I'll remove that for now.

Exirel:
What's wrong with you, Scrutinizer?

dgw:
Scrutinizer is just a flake.

sopel/bot.py
dgw: After further thought, I really want to keep some kind of "welcome" message. But the one I envision includes version numbers for Sopel itself and the Python runtime, and that's an enhancement for a separate PR IMO.

Exirel:
I had to rebase because of a conflict. :(

Exirel:
> I had to rebase because of a conflict. :(

Again.

PR #1598, opened 2019-05-12
cli: add --config-dir common option

Exirel:
I used to left a comment like this:

```
# TODO: allow to use a different homedir
```

But no more. Now, you can use `--config-dir <configdir>` for that. This option provides the location to look for configuration file. Note that this does **not** replace and must **not** be confused with the settings's homedir, which is used internally by Sopel once it is started.

**Edit**: used to be `-H/--homedir <homedir>`

### DIFF
--- a/sopel/cli/config.py
+++ b/sopel/cli/config.py
@@ -62,20 +62,24 @@ def build_parser():
 
 
 def handle_list(options):
-    """Display a list of configuration available from Sopel's config directory
+    """Display a list of configurations available in Sopel's config directory.
 
     :param options: parsed arguments
     :type options: ``argparse.Namespace``
 
     This command displays an unordered list of config names from Sopel's
-    default config directory, without their extensions::
+    config directory, without their extensions::
 
         $ sopel-config list
         default
         custom
 
-    It is possible to filter by extension using the ``-e``/``--ext``/``--extension``
-    option; default is ``.cfg`` (the ``.`` prefix is not required).
+    By default, the config directory is ``~/.sopel``. To select a different
+    config directory, options ``--config-dir`` can be used.
+
+    It is possible to filter by extension using the
+    ``-e``/``--ext``/``--extension`` option; default is ``.cfg``
+    (the ``.`` prefix is not required).
     """
     configdir = options.configdir
     display_path = options.display_path
--- a/sopel/cli/config.py
+++ b/sopel/cli/config.py
@@ -5,7 +5,7 @@ from __future__ import unicode_literals, absolute_import, print_function, divisi
 import argparse
 import os
 
-from sopel import tools, config
+from sopel import tools
 from . import utils
 
 
@@ -22,11 +22,13 @@ def build_parser():
     # sopel-config list
     list_parser = subparsers.add_parser(
         'list',
-        help="List available configurations from Sopel's homedir",
+        help="List available configurations from Sopel's config directory",
         description="""
-            List available configurations from Sopel's homedir ({homedir})
-            with the extension "{ext}".
-        """.format(homedir=config.DEFAULT_HOMEDIR, ext='.cfg'))
+            List available configurations from Sopel's config directory
+            with the extension "{ext}". Use option --config-dir to use a
+            specific config directory.
+        """.format(ext='.cfg'))
+    utils.add_common_arguments(list_parser)
     list_parser.add_argument(
         '-e', '--ext', '--extension',
         dest='extension',
@@ -60,13 +62,13 @@ def build_parser():
 
 
 def handle_list(options):
-    """Display a list of configuration available from Sopel's homedir
+    """Display a list of configuration available from Sopel's config directory
 
     :param options: parsed arguments
     :type options: ``argparse.Namespace``
 
     This command displays an unordered list of config names from Sopel's
-    default homedir, without their extensions::
+    default config directory, without their extensions::
 
         $ sopel-config list
         default
@@ -75,15 +77,16 @@ def handle_list(options):
     It is possible to filter by extension using the ``-e/--ext/--extension``
     option; default is ``.cfg`` (the ``.`` prefix is not required).
     """
-    display_path = getattr(options, 'display_path', False)
-    extension = getattr(options, 'extension', '.cfg')
+    configdir = options.configdir
+    display_path = options.display_path
+    extension = options.extension
     if not extension.startswith('.'):
         extension = '.' + extension
-    configs = utils.enumerate_configs(config.DEFAULT_HOMEDIR, extension)
+    configs = utils.enumerate_configs(configdir, extension)
 
     for config_filename in configs:
         if display_path:
-            print(os.path.join(config.DEFAULT_HOMEDIR, config_filename))
+            print(os.path.join(configdir, config_filename))
         else:
             name, _ = os.path.splitext(config_filename)
             print(name)
@@ -101,9 +104,7 @@ def handle_init(options):
        extension **must be** ``.cfg``.
 
     """
-    config_filename = utils.find_config(
-        config.DEFAULT_HOMEDIR,
-        getattr(options, 'config', None) or 'default')
+    config_filename = utils.find_config(options.configdir, options.config)
     config_name, ext = os.path.splitext(config_filename)
 
     if ext and ext != '.cfg':
--- a/sopel/cli/config.py
+++ b/sopel/cli/config.py
@@ -25,7 +25,7 @@ def build_parser():
         help="List available configurations from Sopel's config directory",
         description="""
             List available configurations from Sopel's config directory
-            with the extension "{ext}". Use option --config-dir to use a
+            with the extension "{ext}". Use option ``--config-dir`` to use a
             specific config directory.
         """.format(ext='.cfg'))
     utils.add_common_arguments(list_parser)
@@ -74,7 +74,7 @@ def handle_list(options):
         default
         custom
 
-    It is possible to filter by extension using the ``-e/--ext/--extension``
+    It is possible to filter by extension using the ``-e``/``--ext``/``--extension``
     option; default is ``.cfg`` (the ``.`` prefix is not required).
     """
     configdir = options.configdir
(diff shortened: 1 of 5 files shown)


dgw:
> the settings's

Ahem, config's* (now I see a bit more inside that brain of yours, replacing `config` with `settings` all over the place even though it just makes things more confusing sometimes)

> Note that this does **not** replace and must **not** be confused with

This is a design consideration. Should we not call this a "homedir"? Should this "homedir" override the other one if given? If this is only for finding config files, maybe it should be called `--config-dir` instead…

Just thinking aloud, here. I'm very much in favor of having this option; it's just, as usual, the details. :sweat_smile:

Exirel:
> This is a design consideration. Should we not call this a "homedir"? Should this "homedir" override the other one if given? If this is only for finding config files, maybe it should be called `--config-dir` instead…

Hm. Yeah, probably. It was called "homedir" long before I added the constant `DEFAULT_HOMEDIR` (hence my choice of name), so I kept that. But you're probably right. To be honest, how configuration are found is bonkers, because it's probably the first time I ever see a software asking for the *name* of the configuration, instead of the path.

That beind said, yes, let's go for  `--config-dir`, that sound reasonable—and way better.

dgw:
I mean, we can totally throw the whole "config name" thing in the trash along with the CLI redesign. Ordinarily I'd say that should wait for Sopel 8 and not add any more to 7's plate, but the time to do it would be in 7 with the CLI changes already being made…

Is it useful enough to have a "list configs" function, that we want to keep the way things work now? That is, specifying a config directory and name separately? Or should we nix that and make the new CLI do `-c` = `--config` = `/path/to/config/file.cfg` (or `.yml`… :smirk:)?

Exirel:
Hmmmm yeah. I think I'm on board with that idea.

However, this is heading into the "hijack a PR to talk about an issue" territory, and it's a dangerous place.

dgw:
> However, this is heading into the "hijack a PR to talk about an issue" territory, and it's a dangerous place.

I guess I should at least _try_ to avoid hypocrisy, huh? Though it's relevant because if we choose to go that alternate route, this PR isn't needed and should be closed (or overwritten with that solution). :wink:

Exirel:
> if we choose to go that alternate route, this PR isn't needed and should be closed

Indeed. At the moment, to be honest, I don't have any upgrade path in mind. I wish I already knew how to migrate from `-c name` but I don't. Even if I don't like this, even if I think it's bonkers, I prefer to take a safer, longer path for the user experience's sake: I don't see how not to break the system with such a change.

Heh, I guess sometimes I can't have it all.

Exirel:
@dgw changed to `--config-dir`

Exirel:
After a second thought: support for `-c name` is unrelated to "where are the config file?" For example, we could have `-c name.cfg` and still have `--config-dir /etc/sopel`. From my point of view, I could have a set of configurations in a shared folder, and run multiple instance of Sopel using this kind of command line:

```
export SOPEL_CONFIG_DIR = "/etc/sopel"
sopel -c freenode.cfg --config-dir $SOPEL_CONFIG_DIR
sopel -c quakenet.cfg --config-dir $SOPEL_CONFIG_DIR
sopel -c private.cfg --config-dir $SOPEL_CONFIG_DIR
```

This is exactly the kind of thing I do for other services. So, heh. You can review that, it's not a feature that will disappear when we rework the `-c name` option. ;-)

sopel/cli/config.py
```diff
@@ -22,11 +22,13 @@ def build_parser():
...
     list_parser = subparsers.add_parser(
         'list',
-        help="List available configurations from Sopel's homedir",
+        help="List available configurations from Sopel's config directory",
         description="""
-            List available configurations from Sopel's homedir ({homedir})
-            with the extension "{ext}".
-        """.format(homedir=config.DEFAULT_HOMEDIR, ext='.cfg'))
+            List available configurations from Sopel's config directory
+            with the extension "{ext}". Use option --config-dir to use a
```
dgw: Should we do this to align with changes in #1646?

```suggestion
            with the extension "{ext}". Use option ``--config-dir`` to use a
```

sopel/cli/config.py
```diff
@@ -75,15 +77,16 @@ def handle_list(options):
     It is possible to filter by extension using the ``-e/--ext/--extension``
```
dgw: Hmm, spotted another of this pattern I just mentioned in my review of #1646. Not necessary to change it in this PR, but maybe you want to do so in the other. 😸

sopel/cli/run.py
```diff
@@ -131,7 +131,10 @@ def add_legacy_options(parser):
                             "use `sopel restart` instead)"))
     parser.add_argument("-l", '--list', action="store_true",
                         dest="list_configs",
-                        help="List all config files found")
+                        help=(
+                            "List all config files found"
+                            "(deprecated, and will be removed in Sopel 8; "
+                            "use `sopel-config list` instead)"))
```
dgw: Someone remind me to add this in the v7 migration guide's CLI overhaul section.

sopel/cli/utils.py
```diff
@@ -199,17 +199,21 @@ def add_common_arguments(parser):
     :type parser: argparse.ArgumentParser
 
     This functions adds the common arguments for Sopel's command line tools.
-    At the moment, this functions adds only one argument to the parser: the
-    argument used as the standard way to define a configuration filename.
+    It adds the following arguments:
+
+    * ``-c/--config``: the name of the Sopel config, or its absolute path
```
dgw:
```suggestion
    * ``-c``/``--config``: the name of the Sopel config, or its absolute path
```

sopel/cli/utils.py
```diff
@@ -199,17 +199,21 @@ def add_common_arguments(parser):
...
+    * ``--config-dir``: the directory to look for Sopel config
```
dgw:
```suggestion
    * ``--config-dir``: the directory to scan for config files
```

or

```suggestion
    * ``--config-dir``: the directory in which to look for config files
```

sopel/cli/utils.py
```diff
@@ -219,10 +223,15 @@ def add_common_arguments(parser):
...
+        default='default',
         metavar='filename',
         dest='config',
-        help='Use a specific configuration file')
+        help='Use a specific configuration file.')
+    parser.add_argument(
+        '--config-dir',
+        default=config.DEFAULT_HOMEDIR,
+        dest='configdir',
+        help='Look for configuration from this Sopel config directory.')
```
dgw:
```suggestion
        help='Look for configuration files in this directory.')
```

sopel/cli/config.py
Exirel:
```suggestion
    It is possible to filter by extension using the ``-e``/``--ext``/``--extension``
```

sopel/cli/run.py
Exirel: Ping. :)

Exirel:
@dgw All good! Also:

> Someone remind me to add this in the v7 migration guide's CLI overhaul section.

sopel/cli/run.py
```diff
@@ -291,10 +294,10 @@ def print_version():
     print('https://sopel.chat/')
 
 
-def print_config():
-    """Print list of available configurations from default homedir."""
-    configs = utils.enumerate_configs(config.DEFAULT_HOMEDIR)
-    print('Config files in %s:' % config.DEFAULT_HOMEDIR)
+def print_config(configdir):
+    """Print list of available configurations from default config directory."""
```
dgw: Here, too, it makes little sense to say "default" any more.

sopel/cli/utils.py
```diff
@@ -266,7 +278,8 @@ def load_settings(options):
     elif 'SOPEL_CONFIG' in os.environ:
         name = os.environ['SOPEL_CONFIG'] or name  # use default if empty
 
-    filename = find_config(config.DEFAULT_HOMEDIR, name)
+    filename = find_config(
+        getattr(options, 'configdir', config.DEFAULT_HOMEDIR), name)
```
dgw: Why is this using `getattr`? `options.configdir` has a default value, defined just a few dozen lines earlier.

test/cli/test_cli_utils.py
```diff
@@ -99,14 +100,29 @@ def test_add_common_arguments():
...
+    assert options.configdir == 'test-long-dir'
+
+    options = parser.parse_args(
+        ['--config', 'test-long', '--config-dir', 'test-long-dir'])
+    assert options.config == 'test-long'
+    assert options.configdir == 'test-long-dir'
+
 
 def test_add_common_arguments_subparser():
     """Assert function adds the -c/--config option on a subparser."""
```
dgw: Should update this docstring.

sopel/cli/utils.py
Exirel: Good call!

Exirel:
@dgw Aside from the random error (caused by a test that use an online resource), it's all good to a new review.

sopel/cli/run.py
dgw: Double-backtick note still applies to latest PR revision.

test/cli/test_cli_utils.py
```diff
@@ -104,32 +105,65 @@ def test_add_common_arguments():
...
+        ['--config', 'test-long', '--config-dir', 'test-long-dir'])
+    assert options.config == 'test-long'
+    assert options.configdir == 'test-long-dir'
+
 
 def test_add_common_arguments_subparser():
-    """Assert function adds the -c/--config option on a subparser."""
+    """Assert function adds the multiple options on a subparser.
+
+    The expected options are -c/--config and --config-dir.
```
dgw: Backticks for option names?

Exirel: In a test, unless we want to generate a doc from them (we could, that can have values), I'm not really for it.

dgw: I think it's safer to write docstrings in the same style no matter where they appear, whether that file is included in the generated docs or not _at present_, so we don't have to chase down bad formatting later if/when the docs start pulling in that file.

Exirel:
@dgw I fixed the double-backtick problem. All it needs now is #1686 to be merged so tests pass with Python 2.

PR #1601, opened 2019-05-12
build: modernize/expand job matrix

dgw:
Testing the full range of Python versions Sopel claims to support is probably wise. Generating releases from one of the newer runtimes is, likewise, also probably wise.

Running releases on 3.6 is a nice compromise between bleeding-edge and platform stability.

We don't test Python 3.8 yet, because its first stable release isn't scheduled until October of this year (2019). Travis supports testing on 3.8-dev or nightly versions, but that's not necessary for us.

_(This is the part where I hope that Travis picks up these settings when building the PR, so I can make extra sure they work as expected. Serve me right if it still uses the matrix from `master`…)_

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,8 +38,14 @@ script:
   - ./checkstyle.sh
   - coverage run -m py.test -v .
   - coverage report
+env:
+  - COVERALLS_PARALLEL=true
 after_success:
   coveralls
+notifications:
+  webhooks:
+    urls:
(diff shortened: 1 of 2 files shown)


dgw:
Yes, it worked exactly as expected!

@Exirel Now that we have that Makefile you added, it's silly to duplicate the individual commands in Travis's configuration. How hard would it be to have the Makefile skip `coverage html` if it's running in CI?

PR #1606, opened 2019-05-14
Deprecate usage of bot.msg to bot.say

deathbybandaid:
This PR aims to deprecate the usage of bot.msg as per:

````
    def msg(self, recipient, text, max_messages=1):
        # Deprecated, but way too much of a pain to remove.
        self.say(text, recipient, max_messages)
````

As of right now, anything that calls `bot.msg` gets passed to `bot.say`. This skips that step.

### DIFF
--- a/sopel/bot.py
+++ b/sopel/bot.py
@@ -18,7 +18,7 @@ import time
 
 from sopel import irc, plugins, tools
 from sopel.db import SopelDB
-from sopel.tools import stderr, Identifier
+from sopel.tools import stderr, Identifier, deprecated
 import sopel.tools.jobs
 from sopel.trigger import Trigger
 from sopel.module import NOLIMIT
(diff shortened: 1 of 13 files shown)


Exirel:
@dgw I would suggest to do that in a different PR, since `self.msg` is used in the `sopel.irc.Bot` class, but this class **does not** define any `msg` method, or `say` method (among with other undefined stuff).

Basically, some of `sopel.irc.Bot`'s methods can not work unless they are called from an instance of `sopel.bot.Sopel`. And that's super weird, and I would like very much to rework that.

It's already a great achievement to remove all occurrence of `bot.msg` outside of the `Bot` class, and should probably get merged right away.

dgw:
> Basically, some of `sopel.irc.Bot`'s methods can not work unless they are called from an instance of `sopel.bot.Sopel`. And that's super weird, and I would like very much to rework that.

If `msg` isn't defined but works, then `say` can be undefined but work just the same.

I'd rather replace _everything_ and then you can rework what you want without worrying about merge conflicts. Whether we leave `self.msg` or update it, it's going to stay weird either way.

I had the same reaction—that it's weird to use stuff from a child class in the parent—but we shouldn't let that stand in the way of killing off `bot.msg`. It's been deprecated for far too long already. It needs to die.

deathbybandaid:
assuming I used `@deprecate`  correctly, I've made all the changes suggested.

PR #1609, opened 2019-05-16
isup: minor cleanup with tests

Exirel:
Just A Small PR.©

I didn't want to change how the "is this an URL?" checks are performed, but at least they are tested now.

### DIFF
--- a/test/modules/test_modules_isup.py
+++ b/test/modules/test_modules_isup.py
@@ -1,0 +1,46 @@
+# coding=utf-8
+"""Tests for Sopel's ``remind`` plugin"""
+from __future__ import unicode_literals, absolute_import, print_function, division
+
+import pytest
+
+from sopel.modules import isup
+
+
+VALID_SITE_URLS = (
+    # no TLD, no scheme
+    ('example', 'http://example.com'),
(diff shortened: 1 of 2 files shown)


test/modules/test_modules_isup.py
```diff
@@ -0,0 +1,46 @@
...
+    ('www.example.io', 'http://www.example.io'),
+)
+
+
+@pytest.mark.parametrize('site, expected', VALID_SITE_URLS)
+def test_get_site_url(site, expected):
+    assert isup.get_site_url(site) == expected
+
+
+INVVALID_SITE_URLS = (
```
dgw: I didn't have as much time to review as expected, but I did have long enough to catch this.

```suggestion
INVALID_SITE_URLS = (
```

Exirel: Actual "lol wtf is this?" here. Good catch!

dgw: Did a proper once-over tonight and this turned out to be the only thing worth noting. So, fix this silly typo, and we'll move forward. 😸

Exirel:
@dgw should be OK now

PR #1613, opened 2019-05-18
tools: add optional arguments to deprecated, and deprecate Ddict

Exirel:
So, at first, I was going to deprecate the `Ddict` class, which can be replaced by `collections.defaultdict` since Python 2.5. I suspect it was very useful when the bot had to support Python version such as 2.2 or 2.4 (which used to be quite popular for a long time).

However, I could not get the proper deprecation warning: decorating the `__init__` method didn't give me what I wanted.

So I decided to allow this usage of the `@deprecated` decorator:

```
@deprecated  # as it is now
@deprecated()  # because why not
@deprecated(reason='obsolete')  # chose your message
@deprecated(version='7.0', removed_in='8.0')  # deprecated in 7, removed in 8
@deprecated(reason='obsolete', version='7.0') # deprecated in 7, no "removed in" yet
@deprecated(reason='obsolete', version='7.0', removed_in='8.0')  # everything
```

So now, `Ddict` is deprecated, and we have a better decorator!

### DIFF
--- a/sopel/modules/meetbot.py
+++ b/sopel/modules/meetbot.py
@@ -7,16 +7,18 @@ Licensed under the Eiffel Forum License 2.
 
 https://sopel.chat
 """
-from __future__ import unicode_literals, absolute_import, print_function, division
+from __future__ import absolute_import, division, print_function, unicode_literals
 
 import codecs
+import collections
 import os
 import re
-from string import punctuation, whitespace
 import time
+from string import punctuation, whitespace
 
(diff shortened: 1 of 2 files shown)


dgw:
I like this, but it has merge conflicts with `meetbot` now. :/

What I don't like are the comma splices in the messages `@deprecated` actually prints, but I can't think of a nice way to eliminate them at the moment.

Exirel:
It doesn't anymore!

Issue #1559, opened 2019-04-15
[Proposal] Update flood protection logic


HumorBaby:
Flood-prevention measures need to be refactored. 

Proposed minimum changes:
- Move logic into `irc.Bot.write`
- Calculate writing backoff (penalty) based on an adequately massaged message (e.g., post-truncation)

**Background**

RFC 1459 spec on flood control of clients:
```
8.10 Flood control of clients

   With a large network of interconnected IRC servers, it is quite easy
   for any single client attached to the network to supply a continuous
   stream of messages that result in not only flooding the network, but
   also degrading the level of service provided to others.  Rather than
   require every 'victim' to be provide their own protection, flood
   protection was written into the server and is applied to all clients
   except services.  The current algorithm is as follows:

        * check to see if client's `message timer' is less than
          current time (set to be equal if it is);

        * read any data present from the client;

        * while the timer is less than ten seconds ahead of the current
          time, parse any present messages and penalize the client by
          2 seconds for each message;

   which in essence means that the client may send 1 message every 2
   seconds without being adversely affected.
```

Current bot layout: (I may have missed some :hushed:)
```
** = flood-prevention logic is here
+----------------+    +---------------+    +-------------+
|  Sopel.action  +--->+ **Sopel.say** +--->+             |
|  Sopel.msg     |    +---------------+    |             |    +---------------+    +--------------------------+
|  Sopel.reply   |                         | Sopel.write +--->+ irc.Bot.write +--->+ asynchat.async_chat.send |
+----------------+    +---------------+    |             |    +---------------+    +--------------------------+
                      | Sopel.notice  +--->+             |            .
                      | Sopel.part    |    +-------------+           /|\
                      | Sopel.join    |                               |
                      | Sopel.kick ?  |                               |
                      +---------------+                         +-----------+ 
                                                                | CAP,NICK, |
                                                                | PASS,...  |
                                                                +-----------+
```

**Rationale**

A few reasons this change is needed:
1) The current logic calculates a penalty based on the untruncated message. This led to the fiasco of https://github.com/sopel-irc/sopel/issues/1550#issuecomment-481937234 and #1551. 
   - The penalty should be based only on the actual part of the message that is being sent to the server.
2) All the methods that don't use `Sopel.say` currently, bypass flood-prevention logic.

**Future plans**

Potential additional changes:
- Add configurable backoff
- Add a configurable maximum penalty
- Allow message burst (with a configurable number and subsequent delays)

[...]

Issue #1342, opened 2018-06-06
Configuration of Flood Protection


deathbybandaid:
I think the Flood Protection should be configurable from the config file.

I have my bot behind ZNC for multiple reasons, and prefer the adjustability of it's flood protection, so the bot doesn't get kicked.

I think adding this to the core.config would be beneficial.

dgw:
Not super high on the priority list right now, but a PR adding this would get a look.

deathbybandaid:
Do you think any of this is still relevant/usable?

https://github.com/sopel-irc/sopel/pull/908/commits/f58fa9f938efc3af6cefec79e56b6ce9ec571ce1

dgw:
As far as I can see from #908, that was probably left unmerged because Embolalia pushed back on having config settings, and then @larsks closed it when the conversation stalled.

If you want to use that commit as a starting point (presumably you've read through #908 already?), I can't think of a reason not to. I would just ask that you rebase f58fa9f onto current `master` to keep credit for that code around in the commit log (and fix the typos in the commit message while you're at it, seeing as @larsks fixed them in the top PR comment anyway).

Out of courtesy, I also pinged @larsks. I think it's nice to ask people before building on unfinished code.

dgw:
This seems like a good feature to add in the next major version, so I've assigned it to the 7.0.0 milestone.

@deathbybandaid With no answer from @larsks in the last 5 months, I think it's safe for you to have at this if you're still interested in working on it.

predakanga:
This feature is one of the last remaining roadblocks in un-forking our own copy of Sopel, so I thought I'd offer the patch myself.

Exirel:
> so I thought I'd offer the patch myself.

Oh, that's lovely, thank you! :heart: In one way or another, I hope Sopel 7 will be able to replace your fork, so you don't have to maintain it—I know how much time it can require!

Exirel:
As I see it, #1518 fixed that (and the documentation is on its way).

PR #1617, opened 2019-05-21
Configuration of Flood Protection

predakanga:
This PR aims to make both the message throttle and loop detection features optional, as per #1342.

The features were both given a simple on/off toggle, as I anticipate more thorough configuration will be planned in #1559 and don't want to pre-empt it.

### DIFF
--- a/sopel/bot.py
+++ b/sopel/bot.py
@@ -408,7 +408,7 @@ class Sopel(irc.Bot):
                 if elapsed < 3:
                     penalty = float(max(0, len(text) - 40)) / 70
                     wait = min(0.8 + penalty, 2)  # Never wait more than 2 seconds
-                    if elapsed < wait:
+                    if elapsed < wait and self.config.core.message_throttle:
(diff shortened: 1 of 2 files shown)


dgw:
Thorough configuration is already pending in #1518 (feel free to test; more testers means fewer bugs slip into the next release).

Giving the flood protection tweakable knobs like that PR does [allows essentially disabling it, too](https://github.com/sopel-irc/sopel/pull/1518#issuecomment-492835813).

I'm mostly giving you immediate feedback on where your work overlaps with stuff already in the pipeline. I won't give a knee-jerk reaction to making the repetition detection toggleable—I want to think about it properly, and I'm a bit distracted at the moment. :sweat_smile:

predakanga:
Ah, thanks - I must have only searched for issues, not existing PRs. I'll create a new issue specifically about repetition detection, since that doesn't seem to fall under that PR.

Closing in favour of #1518.

PR #1505, opened 2019-03-16
reddit: updating reddit module praw dependency

RustyBower:
- version locking praw
- updating code for new praw version
- narrowing exception

PR #1620, opened 2019-05-24
reddit: spoiler update

dgw:
A couple years ago, reddit added support for marking posts/links as "spoiler", obsoleting the hack that I guess some subs were using. No longer does the NSFW (`over_18`) flag need to do double duty!

This frees the Sopel project from having to maintain a list of spoiler-enabled subreddits (which was apparently never updated, anyway).

Channels can also be marked as "spoiler-free" now, separately from their "sfw" status. You can have a safe-for-work channel that still allows spoilers, or a spoiler-free channel that still allows lewds. Freedom!

This is a draft for now, because it's based on #1505's branch. Once that's merged, I'll promote this to "ready for review". Feel free to bikeshed in the meantime—especially command name.

Additional code cleanup could stand to happen—this plugin doesn't really make use of all the decorators that it should (such as checking for OP). I'll do a minor refactor after #1539 is merged, to modernize as much as possible all at once.

### DIFF
--- a/sopel/modules/reddit.py
+++ b/sopel/modules/reddit.py
@@ -2,6 +2,7 @@
 """
 reddit.py - Sopel Reddit Module
 Copyright 2012, Elsie Powell, embolalia.com
+Copyright 2019, dgw, technobabbl.es
 Licensed under the Eiffel Forum License 2.
 
 https://sopel.chat
@@ -37,10 +38,6 @@ domain = r'https?://(?:www\.|old\.|pay\.|ssl\.|[a-z]{2}\.)?reddit\.com'
 post_url = r'%s/r/.*?/comments/([\w-]+)' % domain
 short_post_url = r'https?://redd.it/([\w-]+)'
 user_url = r'%s/u(ser)?/([\w-]+)' % domain
-spoiler_subs = [
-    'stevenuniverse',
-    'onepunchman',
-]
 
 
 @url(post_url)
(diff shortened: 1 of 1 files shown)


sopel/modules/reddit.py
```diff
@@ -209,3 +208,49 @@ def get_channel_sfw(bot, trigger):
...
+    Defaults to false.
+    """
+    if bot.channels[trigger.sender].privileges[trigger.nick] < OP:
+        return
+    else:
+        param = 'true'
+        if trigger.group(2) and trigger.group(3):
+            param = trigger.group(3).strip().lower()
+        spoiler_free = param == 'true'
+        bot.db.set_channel_value(trigger.sender, 'spoiler_free', sfw)
```
dgw: Missed renaming one variable after pasting…

```suggestion
        bot.db.set_channel_value(trigger.sender, 'spoiler_free', spoiler_free)
```

sopel/modules/reddit.py
```diff
@@ -206,3 +208,49 @@ def get_channel_sfw(bot, trigger):
...
+        bot.db.set_channel_value(trigger.sender, 'spoiler_free', spoiler_free)
+        if spoiler_free:
+            bot.reply('Got it. %s is now flagged as spoiler-free.' % trigger.sender)
+        else:
+            bot.reply('Got it. %s is now flagged as spoilers-allowed.' % trigger.sender)
+
+
+@commands('getspoilerfree', 'getspoilfree')
+@example('.getspoilfree [channel]')
+def get_channel_spoiler_free(bot, trigger):
```
half-duplex: Is there a compelling reason this needs to be two separate commands? `.spoilerfree` could print the current state, and `.spoilerfree on|off` could set it.

dgw: All I did was mirror the sfw setting commands. Setting only works in that channel; getting works from anywhere by specifying the channel name.

PR #1621, opened 2019-05-28
db: add methods for values associated with a plugin

### LINKED ISSUE #1501: Adding plugin-based key-value store to SopelDB
I maintain a plugin that needs to persist some values between bot executions. This data is not specific to any channel or nickname, so the existing `bot.db.{set,get}_{nick,channel}_value()` functions aren't optimal, as they would require fabricating a channel or user, which has other implications including potential collisions. My interim solution of writing the values to the bot.config.pluginname is dirty since these values aren't things the bot owner should need to care about.

Proposed are the following methods within `SopelDB`:
```python
get_plugin_value(plugin, key)
    """Retrieves the value for a given key associated with a plugin name."""

set_plugin_value(plugin, key, value)
    """Sets the value for a given key to be associated with the plugin name."""
```

deathbybandaid:
This aims to resolve issue #1501. 

The one thing I was uncertain of is if we need to bother using `.lower()` like we do with channels.

----

@dgw edit: Closes #1501

### DIFF
--- a/sopel/db.py
+++ b/sopel/db.py
@@ -80,6 +80,17 @@ class ChannelValues(BASE):
     value = Column(String(255))
 
 
+class PluginValues(BASE):
+    """
+    PluginValues SQLAlchemy Class
+    """
+    __tablename__ = 'plugin_values'
+    __table_args__ = MYSQL_TABLE_ARGS
+    plugin = Column(String(255), primary_key=True)
+    key = Column(String(255), primary_key=True)
+    value = Column(String(255))
+
+
 class SopelDB(object):
     """*Availability: 5.0+*
 
@@ -423,6 +434,70 @@ class SopelDB(object):
             result = session.query(ChannelValues) \
                 .filter(ChannelValues.channel == channel)\
                 .filter(ChannelValues.key == key) \
+                .one_or_none()
+            if result is not None:
+                result = result.value
+            return _deserialize(result)
+        except SQLAlchemyError:
+            session.rollback()
+            raise
+        finally:
+            session.close()
+
+    # PLUGIN FUNCTIONS
+
+    def set_plugin_value(self, plugin, key, value):
+        """Sets the value for a given key to be associated with a plugin."""
+        plugin = plugin.lower()
+        value = json.dumps(value, ensure_ascii=False)
+        session = self.ssession()
+        try:
+            result = session.query(PluginValues) \
+                .filter(PluginValues.plugin == plugin)\
+                .filter(PluginValues.key == key) \
+                .one_or_none()
+            # PluginValue exists, update
+            if result:
+                result.value = value
+                session.commit()
+            # DNE - Insert
+            else:
+                new_pluginvalue = PluginValues(plugin=plugin, key=key, value=value)
+                session.add(new_pluginvalue)
+                session.commit()
+        except SQLAlchemyError:
+            session.rollback()
+            raise
+        finally:
+            session.close()
+
+    def delete_plugin_value(self, plugin, key):
+        """Deletes the value for a given key associated with a plugin."""
+        plugin = plugin.lower()
+        session = self.ssession()
+        try:
+            result = session.query(PluginValues) \
+                .filter(PluginValues.plugin == plugin)\
+                .filter(PluginValues.key == key) \
+                .one_or_none()
+            # PluginValue exists, update
+            if result:
+                session.delete(result)
+                session.commit()
+        except SQLAlchemyError:
+            session.rollback()
+            raise
+        finally:
+            session.close()
+
+    def get_plugin_value(self, plugin, key):
+        """Retrieves the value for a given key associated with a plugin."""
+        plugin = plugin.lower()
+        session = self.ssession()
+        try:
+            result = session.query(PluginValues) \
+                .filter(PluginValues.plugin == plugin)\
+                .filter(PluginValues.key == key) \
                 .one_or_none()
             if result is not None:
                 result = result.value
--- a/test/test_db.py
+++ b/test/test_db.py
@@ -269,3 +269,27 @@ def test_get_preferred_value(db):
     names = ['asdf', '#asdf']
     assert db.get_preferred_value(names, 'qwer') == 'poiu'
     assert db.get_preferred_value(names, 'lkjh') == '1234'
+
+
+def test_set_plugin_value(db):
+    conn = sqlite3.connect(db_filename)
+    db.set_plugin_value('plugname', 'qwer', 'zxcv')
+    result = conn.execute(
+        'SELECT value FROM plugin_values WHERE plugin = ? and key = ?',
+        ['plugname', 'qwer']).fetchone()[0]
+    assert result == '"zxcv"'
+
+
+def test_get_plugin_value(db):
+    conn = sqlite3.connect(db_filename)
+    conn.execute("INSERT INTO plugin_values VALUES ('plugname', 'qwer', '\"zxcv\"')")
+    conn.commit()
+    result = db.get_plugin_value('plugname', 'qwer')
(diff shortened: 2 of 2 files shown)


deathbybandaid:
and depending on PR #1526 , I took the liberty of writing this snippet in advance for consistency. I was not sure if I should include it here, there, or wait until this, and #1526 are merged first.

````
def delete_plugin_value(self, plugin, key):
        """Deletes the value for a given key to be associated with the plugin."""
        plugin = Identifier(plugin).lower()
        session = self.ssession()
        try:
            result = session.query(PluginValues) \
                .filter(PluginValues.plugin == plugin)\
                .filter(PluginValues.key == key) \
                .one_or_none()
            # PluginValues exists, update
            if result:
                session.delete(result)
                session.commit()
        except SQLAlchemyError:
            session.rollback()
            raise
        finally:
            session.close()
````

**Comment by deathbybandaid** [comment]
another other concern is if `Identifier()` even needs to be used here.

**Comment by deathbybandaid** [comment]
Just dropping another note: should probably add to `test_db.py`

**Comment by dgw** [comment]
In brief:

* Good start! Exciting to have an implementation of this.
* Using `.lower()` on the plugin name is probably good for consistency, yes.
* Casting the plugin name to `Identifier` is almost certainly unnecessary (it isn't one).
* Pretend #1526 will be merged before this (because it probably will be) and include a `delete_plugin_value` method in this PR.
* Adding tests is always valuable—always!
  * Per your IRC comments on this: As far as using sqlite directly goes, rewriting the DB tests to use SQLAlchemy instead is for a separate PR, and for this PR you should mimic the existing ones. (Opening a PR to switch the DB tests yourself is welcome of course!)

Longer thoughts:

Lacking any reasonable ideas for making the `plugin` argument optional (auto-detecting the calling module from within the `db` functions would require some possibly fragile introspection), I'll refrain from asking about that. I would _love_ to have e.g. `bot.db.set_plugin_value('key', True)` "just work" without having to re-specify the plugin name, but… concerns about compatibility with alternative Python implementations and non-Linux platforms make me think doing so would be far more complicated than it's worth.

Besides, the existing set/get functions already require passing the nickname or channel, so the way this works now is parallel to how those work anyway. (Yes, I have thought about making the channel-value methods "automatic", too. Not worth it either, lol.)

**Comment by deathbybandaid** [comment]
Identifier removed, lower() retained, tests added, squashed

File: test/test_db.py [hunk 1]
```diff
@@ -240,6 +240,30 @@ def test_get_channel_value(db):
     assert result == 'zxcv'
 
 
+def test_set_plugin_value(db):
+    conn = sqlite3.connect(db_filename)
+    db.set_plugin_value('#asdf', 'qwer', 'zxcv')
```
half-duplex: Plugin named `#asdf`? (repeated below)

deathbybandaid: I can make that say whatever, however,,, the dB tests still needs to be updated for sqlalchemy instead of directly accessing sqlite.

test/test_db.py
```diff
@@ -240,6 +240,30 @@ def test_get_channel_value(db):
...
+    result = conn.execute(
+        'SELECT value FROM plugin_values WHERE plugin = ? and key = ?',
+        ['#asdf', 'qwer']).fetchone()[0]
+    assert result == '"zxcv"'
+
+
+def test_get_plugin_value(db):
+    conn = sqlite3.connect(db_filename)
+    conn.execute("INSERT INTO plugin_values VALUES ('#asdf', 'qwer', '\"zxcv\"')")
```
deathbybandaid:
```suggestion
    conn.execute("INSERT INTO plugin_values VALUES ('plugname', 'qwer', '\"zxcv\"')")
```

test/test_db.py
deathbybandaid:
```suggestion
    db.set_plugin_value('plugname', 'qwer', 'zxcv')
```

test/test_db.py
```diff
@@ -240,6 +240,30 @@ def test_get_channel_value(db):
     assert result == 'zxcv'
 
 
+def test_set_plugin_value(db):
+    conn = sqlite3.connect(db_filename)
+    db.set_plugin_value('#asdf', 'qwer', 'zxcv')
+    result = conn.execute(
+        'SELECT value FROM plugin_values WHERE plugin = ? and key = ?',
+        ['#asdf', 'qwer']).fetchone()[0]
```
deathbybandaid:
```suggestion
        ['plugname', 'qwer']).fetchone()[0]
```

test/test_db.py
```diff
@@ -240,6 +240,30 @@ def test_get_channel_value(db):
...
+    assert result == '"zxcv"'
+
+
+def test_get_plugin_value(db):
+    conn = sqlite3.connect(db_filename)
+    conn.execute("INSERT INTO plugin_values VALUES ('#asdf', 'qwer', '\"zxcv\"')")
+    conn.commit()
+    result = db.get_plugin_value('#asdf', 'qwer')
```
deathbybandaid:
```suggestion
    result = db.get_plugin_value('plugname', 'qwer')
```

deathbybandaid:
as per PR #1652 this has been updated

sopel/db.py
```diff
@@ -433,6 +444,70 @@ def get_channel_value(self, channel, key):
         finally:
             session.close()
 
+    # PLUGIN FUNCTIONS
+
+    def set_plugin_value(self, plugin, key, value):
+        """Sets the value for a given key to be associated with the plugin."""
```
dgw:
```suggestion
        """Sets the value for a given key to be associated with a plugin."""
```

sopel/db.py
```diff
@@ -433,6 +444,70 @@ def get_channel_value(self, channel, key):
...
+                session.add(new_pluginvalue)
+                session.commit()
+        except SQLAlchemyError:
+            session.rollback()
+            raise
+        finally:
+            session.close()
+
+    def delete_plugin_value(self, plugin, key):
+        """Deletes the value for a given key to be associated with the plugin."""
```
dgw:
```suggestion
        """Deletes the value for a given key to be associated with a plugin."""
```

sopel/db.py
```diff
@@ -433,6 +444,70 @@ def get_channel_value(self, channel, key):
...
+        """Sets the value for a given key to be associated with a plugin."""
+        plugin = plugin.lower()
+        value = json.dumps(value, ensure_ascii=False)
+        session = self.ssession()
+        try:
+            result = session.query(PluginValues) \
+                .filter(PluginValues.plugin == plugin)\
+                .filter(PluginValues.key == key) \
+                .one_or_none()
+            # PluginValues exists, update
```
dgw:
```suggestion
            # PluginValue exists, update
```

sopel/db.py
```diff
@@ -433,6 +444,70 @@ def get_channel_value(self, channel, key):
...
+                session.add(new_pluginvalue)
+                session.commit()
+        except SQLAlchemyError:
+            session.rollback()
+            raise
+        finally:
+            session.close()
+
+    def delete_plugin_value(self, plugin, key):
+        """Deletes the value for a given key to be associated with a plugin."""
```
dgw:
```suggestion
        """Deletes the value for a given key associated with a plugin."""
```

sopel/db.py
```diff
@@ -433,6 +444,70 @@ def get_channel_value(self, channel, key):
...
+        plugin = plugin.lower()
+        session = self.ssession()
+        try:
+            result = session.query(PluginValues) \
+                .filter(PluginValues.plugin == plugin)\
+                .filter(PluginValues.key == key) \
+                .one_or_none()
+            # PluginValues exists, update
```
dgw:
```suggestion
            # PluginValue exists, update
```

Issue #1622, opened 2019-05-29
url: private-resolution inconsistencies


dgw:
- [x] Resolved hostnames are checked only for `is_private`, not `is_loopback` (bare IPs in URLs are checked for both)
- [x] No docstrings for settings added in #1439
- [x] No "magic" documentation about newly added settings in `configure()` docstring for website (it's just a Markdown table, pretty easy to fill in based on what's already there)
- [x] Also, [this typo](https://github.com/sopel-irc/sopel/pull/1439#discussion_r288448346) (`s/form/from/`)

/cc @RustyBower

RustyBower:
I'll take a peek :)

dgw:
All of these were done in #1624. Another one bites the dust!

PR #1624, opened 2019-05-29
url: minor bug fixes

RustyBower:
Fixes for https://github.com/sopel-irc/sopel/issues/1622

### DIFF
--- a/sopel/modules/url.py
+++ b/sopel/modules/url.py
@@ -52,8 +52,10 @@ class UrlSection(StaticSection):
     """If greater than 0, the title fetcher will include a TinyURL version of links longer than this many characters."""
     enable_private_resolution = ValidatedAttribute(
         'enable_private_resolution', bool, default=False)
+    """Enable URL lookups for RFC1918 addresses"""
     enable_dns_resolution = ValidatedAttribute(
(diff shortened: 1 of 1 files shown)


RustyBower:
Waiting on @dgw for an example of the magic docstrings, otherwise this should be good.

sopel/modules/url.py
```diff
@@ -63,6 +63,8 @@ def configure(config):
     | exclude | https?://git\\\\.io/.* | A list of regular expressions for URLs for which the title should not be shown. |
     | exclusion\\_char | ! | A character (or string) which, when immediately preceding a URL, will stop the URL's title from being shown. |
     | shorten\\_url\\_length | 72 | If greater than 0, the title fetcher will include a TinyURL version of links longer than this many characters. |
+    | enable\\_private\\_resolution | True | Enable DNS resolution for all domains to validate if there are RFC1918 resolutions. |
+    | enable\\_dns\\_resolution | True | Enable DNS resolution for all domains to validate if there are RFC1918 resolutions. |
```
dgw: By convention, this example usually includes the default value of each setting. (There are existing exceptions in this module.) Both of these new settings should be `False`, by that convention.

Issue #1626, opened 2019-05-30
config: replace the comma separator by new-line in ListAttribute


Exirel:
TL;DR: in Sopel 7.x, I want to detect `\n` to switch to a new mode that doesn't require a delimiter with escape character.

**Disclaimer**: an escape mechanism was brought in for Sopel 7.x by #1460 to allow comma-separated values. I think it's a great achievement but, sadly, also something that I may want to remove entirely. I'm fully aware that this may mean having to throw off the hard work put in by @HumorBaby and I want to re-assert how much respect I've for his works, past and present.

### Current state in Sopel 6.6.x

As of today's version of [Sopel 6.6.x branch](https://github.com/sopel-irc/sopel/blob/6.6.x/sopel/config/types.py#L220-L225), this config file:

```
[spam]
eggs = one, two, three, four, and a half
bacons = grilled
    burn out
    greasy, fat, and tasty
```

will be read as this:

```
>>> config.spam.eggs
['one', ' two', ' three', ' four', ' and a half']
>>> config.spam.bacons
['grilled\nburn out\ngreasy', 'fat', 'and tasty']
```

### Current state in Sopel master

In the master branch, since #1460, it is possible to escape the comma, so it would be like that:

```
[spam]
eggs = one, two, three, four, and a half
eggs_safe = one, four\, and a half
bacons = grilled
    burn out
    greasy, fat, and tasty
```

which result in `eggs_safe` to be properly handled, but leaving `bacons` still in a poor state:

```
>>> config.spam.eggs_safe
['one', 'four, and a half']
>>> config.spam.bacons
['grilled\nburn out\ngreasy', 'fat', 'and tasty']
```

And now, if we want to throw regex pattern (that need to be escaped too), what do we have?

```
[spam]
regex_pattern = ^Full stop\, Period\\.$, also:? this pattern$
regex_multilines =
    ^Full stop, Period\.$
    also:? this pattern$
```

Sure thing, regex patterns are not the easiest things to read, but I found that the multi-lines version far more readable.

### What's next? What's my plan?

If we take back the example for Sopel 6.6.x:

```
[spam]
eggs = one, two, three, four, and a half
bacons = grilled
    burn out
    greasy, fat, and tasty
```

I want that output:

```
>>> config.spam.eggs_safe
['one', 'four, and a half']
>>> config.spam.bacons
['grilled', 'burn out', 'greasy, fat, and tasty']
```

For Sopel 7.x I want to implement that so:

[...]

PR #1628, opened 2019-05-30
config: split ListAttribute values on newlines

Exirel:
See the description of #1626 for more information.

Note: I had to upgrade the `test_config.py` file to use py.test instead of unittest. We should be fine with that.

### DIFF
--- a/sopel/config/types.py
+++ b/sopel/config/types.py
@@ -221,12 +221,13 @@ class ListAttribute(BaseValidated):
         >>> config.spam.cheeses
         ['camembert', 'cheddar', 'reblochon']
 
-    which come from this configuration file::
+    which comes from this configuration file::
 
         [spam]
-        cheeses = camembert
-                  cheddar
-                  reblochon
+        cheeses =
+            camembert
+            cheddar
+            reblochon
 
     .. versionchanged:: 7.0
 
--- a/sopel/config/types.py
+++ b/sopel/config/types.py
@@ -211,12 +211,45 @@ class ValidatedAttribute(BaseValidated):
 class ListAttribute(BaseValidated):
     """A config attribute containing a list of string values.
 
-    Values are saved to the file as a comma-separated list. It does not
-    currently support commas within items in the list. By default, the spaces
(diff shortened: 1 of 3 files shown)


dgw:
> Note: I had to upgrade the `test_config.py` file to use py.test instead of unittest. We should be fine with that.

Vaguely remembered noticing that `unittest` was used in only that one file… Found where I referenced it over a year ago: https://github.com/sopel-irc/sopel/pull/1299#issue-178875969 (knew I should have made an issue; it did, indeed, get forgotten…)

So, :tada: for nixing that one last mismatched test framework. :grinning:

Exirel:
So, I had to rebase from master because there was several differences in requirements and in documentations needed... and that messed up the commit history in Github.

Talking about documentation, I'll now update it!

Exirel:
@dgw done!

sopel/config/types.py
```diff
@@ -211,12 +211,45 @@ def configure(self, prompt, default, parent, section_name):
...
-    ESCAPE_CHARACTER = '\\'
+        class SpamSection(StaticSection):
+            cheeses = ListAttribute('cheeses')
+
+    the option will be exposed as a Python :class:`list`::
+
+        >>> config.spam.cheeses
+        ['camembert', 'cheddar', 'reblochon']
+
+    which come from this configuration file::
```
```suggestion
    which comes from this configuration file::
```

Damn, I think I missed this last time. 🙁

sopel/config/types.py
```diff
@@ -211,12 +211,45 @@ def configure(self, prompt, default, parent, section_name):
...
+
+        [spam]
+        cheeses = camembert
```
dgw: Given that the updated `configuration.rst` examples all place the first list value on the next line, shouldn't the examples in this file do the same?

dgw:
Note to self: Remove #1460 from changelog/milestone when merging this for 7.0, as this PR reverts it.

sopel/config/types.py
Exirel: Hm. Probably, yes!

PR #1629, opened 2019-05-30
currency: switched crypto api and combined rates to allow cross-lookup

Duckle29:
This PR combines the fiat currency rates and the cryptocurrency rates into one global variable `rates` instead of keeping the two separate.

This was done to allow better cross-compatibility between the two sources, allowing one to look up a conversion from a currency in one table, to a currency in the other.

To do this, a change of API was also done, as the old API used a weird and inflexible way to look up conversions.

The code also changed some variable names that I didn't like. mostly the "of" and "to" variables were changed to base and target respectively. 

This change also entirely eliminates the need for a separate crypto rates function.

### DIFF
--- a/sopel/modules/currency.py
+++ b/sopel/modules/currency.py
@@ -21,7 +21,7 @@ from sopel.module import commands, example, NOLIMIT, rule
 
 FIAT_URL = 'https://api.exchangeratesapi.io/latest?base=EUR'
 FIXER_URL = 'http://data.fixer.io/api/latest?base=EUR&access_key={}'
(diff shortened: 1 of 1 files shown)


Duckle29:
This PR also applies the same fix as https://github.com/sopel-irc/sopel/pull/1627

dgw:
> This PR also applies the same fix as #1627

Doing this negates the advantage of having three-way merges, lol.

Duckle29:
Well I'm a git noob and didn't know what to do :P I can remove it and force push if you want :)

dgw:
Might as well fix the style-checker error and the overlap with #1627 at the same time.

PR #1630, opened 2019-05-30
coretasks: complete sparse User objects in _record_who()

dgw:
`User` objects created in `handle_names()` have `None` in user/host attributes because `NAMES` does not contain that information. When a `WHO` reply comes in for a user that was already added by `handle_names()`, `_record_who()` should fill in the missing properties if possible.

### DIFF
--- a/sopel/coretasks.py
+++ b/sopel/coretasks.py
@@ -750,13 +750,20 @@ def _record_who(bot, channel, user, host, nick, account=None, away=None, modes=N
     nick = Identifier(nick)
     channel = Identifier(channel)
     if nick not in bot.users:
-        bot.users[nick] = User(nick, user, host)
-    user = bot.users[nick]
-    if account == '0':
-        user.account = None
+        usr = User(nick, user, host)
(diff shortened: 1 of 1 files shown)


sopel/coretasks.py
```diff
@@ -750,13 +750,20 @@ def _record_who(bot, channel, user, host, nick, account=None, away=None, modes=N
     nick = Identifier(nick)
     channel = Identifier(channel)
     if nick not in bot.users:
-        bot.users[nick] = User(nick, user, host)
-    user = bot.users[nick]
+        usr = User(nick, user, host)
```
Exirel: Why is it `usr` instead of `user` now?

dgw:
@Exirel `user` is the name of a function argument, so assigning the `User` object to `user` would override that and make the passed-in value unusable afterward.

I initially renamed the `user` _arg_ to `username`, but didn't like it because the `User` object constructor's parameters are named `nick`, `user`, and `host` (and I wanted to leave them matching in the call here).

PR #1632, opened 2019-05-30
module: fix decorator multiple-use inconsistency

### LINKED ISSUE #1619: module: Decorator multiple-use inconsistency
In follow-up testing to #1503 and #1486, preparatory to approving #1505, I discovered this frustrating little side-effect of how the `@sopel.module.url` decorator currently works. Once I started writing an issue about that, it snowballed from "the `@url` decorator is borked" to "man, why are the decorators so inconsistent?"… and here we are.

----

Our current `master` version of `reddit.py` does the following:

https://github.com/sopel-irc/sopel/blob/a14c145c853137a88a4e9930761362fc005aeb30/sopel/modules/reddit.py#L45-L47

Only the `post_url` decorator takes effect. Links matching the `short_post_url` pattern are ignored. But if I comment out the `post_url` decorator, short links work—so it's clearly not a problem with the pattern, but something wrong with the way Sopel handles URL decorators.

Our `@url` decorator attaches the URL pattern to a wrapper function like so:

https://github.com/sopel-irc/sopel/blob/a14c145c853137a88a4e9930761362fc005aeb30/sopel/module.py#L511-L512

This implementation seems to allow only one URL pattern per callable, unlike the other decorators that add trigger conditions. [`commands`](https://github.com/sopel-irc/sopel/blob/a14c145c853137a88a4e9930761362fc005aeb30/sopel/module.py#L186), [`nickname_commands`](https://github.com/sopel-irc/sopel/blob/a14c145c853137a88a4e9930761362fc005aeb30/sopel/module.py#L218), [`rule`](https://github.com/sopel-irc/sopel/blob/a14c145c853137a88a4e9930761362fc005aeb30/sopel/module.py#L117), [`event`](https://github.com/sopel-irc/sopel/blob/a14c145c853137a88a4e9930761362fc005aeb30/sopel/module.py#L255), [`intent`](https://github.com/sopel-irc/sopel/blob/a14c145c853137a88a4e9930761362fc005aeb30/sopel/module.py#L268), and even [`interval`](https://github.com/sopel-irc/sopel/blob/a14c145c853137a88a4e9930761362fc005aeb30/sopel/module.py#L92) all allow either passing a list of arguments, or calling the same decorator multiple times—but [`url`](https://github.com/sopel-irc/sopel/blob/a14c145c853137a88a4e9930761362fc005aeb30/sopel/module.py#L511-L512) allows neither. It restricts a single function/callable to handling a single URL pattern.

**Current workarounds, in order of "spaghetti factor":**

<dl>
<dt>Make one giant URL pattern that matches everything</dt>
<dd>Messy to write, messy to read; prone to breakage due to regex is hard™ factor.</dd>
<dt>Define wrapper callables for each additional URL pattern needed, which call the main one</dt>
<dd>More flexible, since patterns with different numbers of capture groups can be massaged before calling the main handler, but clutters the plugin file with a bunch of extra callables.</dd>
<dt>Use <tt>@rule</tt> instead and manually register/unregister URL callbacks</dt>
<dd>Convoluted; requires adding <tt>setup</tt> and <tt>shutdown</tt> handlers to plugins that otherwise wouldn't need them, so the method calls to register and unregister URL handlers can be made when appropriate. Making this pattern unnecessary by fixing the problems with how the <tt>@url</tt> decorator works is one of Sopel 7's goals.</dd>
</dl>

**Immediate fix needed:**

Rework decorator and loading routines so that `@url` can be used multiple times.

**Future direction:**

I think all of these single-argument decorators should work the same way, and consistently. All should take multiple arguments that get joined into a single list, _or_ multiple calls to the same decorator, _or_ any combination thereof. The following should all be equivalent:

```py
@rule(r'simple( ?regex)')
@rule(r'complicated(?: ?regex) with (alternate) capture')
# equals
@rule(r'simple( ?regex)', r'complicated(?: ?regex) with (alternate) capture')
```

```py
@url(r'https?://example.com/(\w+)/?', r'https://example.com/post/comments/(\w+)/?')
# equals
@url(r'https?://example.com/(\w+)/?')
@url(r'https://example.com/post/comments/(\w+)/?')
```

```py
@interval(3, 100, 512)
# equals
@interval(3)
@interval(100)
@interval(512)
# equals
@interval(3, 100)
@interval(512)
# equals
@interval(3, 512)
@interval(100)
# etc.
```

Obviously, this can only apply to some decorators. Some take 0 arguments, or have one or more optional arguments, etc. Here's the list I'm talking about:

  - `commands`
  - `event`
  - `intent`
  - `interval`
  - `nickname_commands`
  - `rule`
  - `url`

Another question is why two are plural (e.g. `commands`) and the others are all singular, even though they may also take a list of arguments—but I suspect that most of this should be for Sopel 8. In Sopel 7, we really only need to fix `@url`.

Exirel:
Not much to say: fix [main part of] #1619

(editorial by @dgw so GH won't close the other issue)

### DIFF
--- a/sopel/module.py
+++ b/sopel/module.py
@@ -487,6 +487,7 @@ def url(url_rule):
         from sopel import module
 
         @module.url(r'https://example.com/bugs/([a-z0-9]+)')
+        @module.url(r'https://short.com/([a-z0-9]+)')
         def handle_example_bugs(bot, trigger, match):
             bot.reply('Found bug ID #%s' % match.group(1))
 
@@ -495,6 +496,11 @@ def url(url_rule):
 
     Under the hood, when Sopel collects the decorated handler it uses
     :meth:`sopel.bot.Sopel.register_url_callback` to register the handler.
+
+    .. versionchanged:: 7.0
+
+        The same function can be decorated multiple times with :func:`url`
+        to register different URL patterns.
 
     .. seealso::
 
@@ -508,7 +514,9 @@ def url(url_rule):
         def helper(bot, trigger, match=None):
             match = match or trigger
             return function(bot, trigger, match)
-        helper.url_regex = re.compile(url_rule)
+        if not hasattr(helper, 'url_regex'):
+            helper.url_regex = []
+        helper.url_regex.append(re.compile(url_rule))
         return helper
     return actual_decorator
 
--- a/sopel/module.py
+++ b/sopel/module.py
@@ -8,8 +8,8 @@
 
 from __future__ import unicode_literals, absolute_import, print_function, division
 
-import re
 import functools
+import re
 
 NOLIMIT = 1
 """Return value for ``callable``\\s, which suppresses rate limiting for the call.
@@ -61,7 +61,7 @@ def unblockable(function):
     return function
 
 
-def interval(*args):
+def interval(*intervals):
     """Decorates a function to be called by the bot every X seconds.
 
     This decorator can be used multiple times for multiple intervals, or all
@@ -76,10 +76,11 @@ def interval(*args):
     There is no guarantee that the bot is connected to a server or joined a
     channel when the function is called, so care must be taken.
 
-    Example:::
+    Example::
 
-        import sopel.module
-        @sopel.module.interval(5)
+        from sopel import module
+
+        @module.interval(5)
         def spam_every_5s(bot):
             if "#here" in bot.channels:
                 bot.say("It has been five seconds!", "#here")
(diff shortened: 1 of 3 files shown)


sopel/module.py
```diff
@@ -110,11 +111,20 @@ def rule(value):
     Inside the regular expression, some special directives can be used. $nick
     will be replaced with the nick of the bot and , or :, and $nickname will be
     replaced with the nick of the bot.
+
+    .. versionchanged:: 7.0
+
+        The :func:`rule` decorator can be called with many positional
```
dgw:
```suggestion
        The :func:`rule` decorator can be called with multiple positional
```

sopel/module.py
```diff
@@ -110,11 +111,20 @@ def rule(value):
...
+        arguments, each used to add a rule. This is equivalent as decorating
```
dgw:
```suggestion
        arguments, each used to add a rule. This is equivalent to decorating
```

sopel/module.py
```diff
@@ -110,11 +111,20 @@ def rule(value):
...
+        the same function many times with this decorator.
```
dgw:
```suggestion
        the same function multiple times with this decorator.
```

sopel/module.py
```diff
@@ -496,6 +518,16 @@ def handle_example_bugs(bot, trigger, match):
     Under the hood, when Sopel collects the decorated handler it uses
     :meth:`sopel.bot.Sopel.register_url_callback` to register the handler.
 
+    .. versionchanged:: 7.0
+
+        The same function can be decorated many times with :func:`url` with
+        different URL patterns. This function will be then registered as URL
+        callback for each of these patterns.
```
dgw: This is probably more verbose than it needs to be. All it really needs to say is that decorating multiple times will work as expected (add all the URL patterns). Specifying what happens with registration etc. is redundant.

Exirel:
> I'm [...] wondering if we can't support multiple decorations without adding extra stack frames.

We can. I wanted to be sure that it worked first, and I'm glad you could review this first. I'll work on that next.

Exirel:
@dgw done

PR #1639, opened 2019-06-08
db: prevent get_nick_id from breaking database

deathbybandaid:
After some extensive testing of `get_nick_id` , I found that multiple threads calling it could quickly corrupt the database my assigning multiple id numbers.

Relevant part of Traceback:

````
  File "/usr/local/lib/python3.6/dist-packages/sopel/db.py", line 177, in get_nick_id
    .filter(Nicknames.slug == slug) \
  File "/usr/local/lib/python3.6/dist-packages/sqlalchemy/orm/query.py", line 3259, in one_or_none
    "Multiple rows were found for one_or_none()"
sqlalchemy.orm.exc.MultipleResultsFound: Multiple rows were found for one_or_none()
````

This PR adds a `threading.lock` to `get_nick_id` to prevent such errors.

### DIFF
--- a/sopel/db.py
+++ b/sopel/db.py
@@ -5,6 +5,7 @@ import errno
 import json
 import os.path
 import sys
+import threading
 
 from sopel.tools import Identifier
 
@@ -174,6 +175,8 @@ class SopelDB(object):
 
         self.ssession = scoped_session(sessionmaker(bind=self.engine))
 
+        self.nick_id_lock = threading.Lock()
+
     def connect(self):
         """Return a raw database connection object."""
         return self.engine.connect()
@@ -201,6 +204,7 @@ class SopelDB(object):
         session = self.ssession()
         slug = nick.lower()
         try:
+            self.nick_id_lock.acquire()
             nickname = session.query(Nicknames) \
                 .filter(Nicknames.slug == slug) \
                 .one_or_none()
@@ -223,6 +227,7 @@ class SopelDB(object):
             raise
         finally:
             session.close()
+            self.nick_id_lock.release()
 
     def alias_nick(self, nick, alias):
         """Create an alias for a nick.

### CODE REVIEW

**Comment by Exirel** [comment]
I'm sure you can manage that with an SQL transaction instead.

**Comment by dgw** [comment]
Hi, I'm between flights. :smile_cat: I have never been very happy with the whole "dedicated `nick_ids` table" thing. Can't do too much research rn, but is there a reason the `nicknames` table can't just make its `nick_id` column `AUTOINCREMENT` and nix the extraneous one-column table?

File: sopel/db.py [hunk 1]
```diff
@@ -189,6 +194,7 @@ def get_nick_id(self, nick, create=True):
                 nickname = Nicknames(nick_id=nick_id.nick_id, slug=slug, canonical=nick)
                 session.add(nickname)
                 session.commit()
+            self.nick_id_lock.release()
```
dgw: If an uncaught exception is raised anywhere in this relatively long `try` block, the lock might never be released. That would be Very Bad™. Releasing the lock should be done in the `finally` clause, below.

deathbybandaid: moved to `finally`

sopel/db.py
```diff
@@ -201,12 +204,14 @@ def get_nick_id(self, nick, create=True):
...
         slug = nick.lower()
         try:
+            self.nick_id_lock.acquire()
             nickname = session.query(Nicknames) \
                 .filter(Nicknames.slug == slug) \
                 .one_or_none()
 
             if nickname is None:
                 if not create:
+                    self.nick_id_lock.release()
```
dgw: Leaving this in will cause a `RuntimeError` when the lock is released in `finally`.

deathbybandaid: fixed

dgw:
Ping @RustyBower for advice on getting this kind of protection from transactions, before I approve anything. (Thanks @deathbybandaid for the quick fixups!)

RustyBower:
So I've finally had a change to dive into this and do some testing.

There is documentation in the implementation of thread safe sessions (https://docs.sqlalchemy.org/en/13/orm/contextual.html) without the lock hack that is in the code here.

Additionally, we call scoped_session.close(), and looking through the docs (https://docs.sqlalchemy.org/en/13/orm/contextual.html#sqlalchemy.orm.scoping.scoped_session.remove), it sounds like we should update that to scoped_session.remove() instead.

@deathbybandaid can you test with the update for .remove() instead of .close() and see if you are still seeing this behavior? I'll also dig in on my side and see how we can make this more thread safe.

dgw:
@RustyBower @deathbybandaid Any status updates? We're about a month from target date for 7.0.

deathbybandaid:
Somewhere in my irc backlogs, @RustyBower and I discussed him openning a seperate (better) PR. I'll have to scroll through to find the reasoning we had (drawing a blank atm)

deathbybandaid:
Okay, just scrolled my logs, and it looks like he had some thoughts on being more "thread safe" with sqlalchemy as a whole.

His potential PR would affect sqlalchemy on a broader scale. He told me I should leave this open though.

RustyBower:
I'll give this a test tonight/tomorrow

dgw:
As I understand it, #1755 replaces this solution entirely.

PR #1651, opened 2019-06-28
help: add pastebin.ubuntu.com as help pastebin provider

ddstreet:


HumorBaby:
Happened to have a minute and was curious :smirk:... here's a note (not the only one, but the only one I can type quickly :grin: )

If you set the expiry of the paste to 'day', then you will need to update the caching logic which will keep providing the same paste URL as long as modules/commands aren't added/removed, or certain settings changed.

ddstreet:
> If you set the expiry of the paste to 'day', then you will need to update the caching logic

ah thanks!  did not notice that.

ddstreet:
ok, unset expiration so it should remain valid, without needing to adjust the sopel url caching logic

Issue #1650, opened 2019-06-28
Database Encoding: UnicodeEncodeError issue


deathbybandaid:
As per this snippet of IRC chat:

> 11:43 \<deathbybandaid>  sqlalchemy+mysql seems to have a problem that the sqlite database didn't have:
> 11:43 \<deathbybandaid>  'latin-1' codec can't encode character '\u2019' in position 268: ordinal not in range(256) (../MySQLdb/connections.py:245)
> 11:44 \<deathbybandaid>  only happens when I try to save stuff like    ¯\_(ツ)_/¯  to the database
> 11:45 \<deathbybandaid>  not sure if there is something that can be tweaked in db.py to fix this,, or if there is something I can do in the interum to mitigate the issue
> 11:46 \<deathbybandaid>  I have a command `.me` that users can save catchphrases to and call via `.me 1` or `.me homer`
> 11:47 \<deathbybandaid>  so if there isn't a bot command for something, they can make their own
> 11:49 <+Exirel>  deathbybandaid: hm, we could set an encoding parameter
> 11:49 <+Exirel>  but for now you can create the database yourself with the right default encoding (ie. utf8, usually utf8_general_ci)
> 11:50 <+Exirel>  because by default MySQL create tables with latin1_swedish, IIRC
> 11:51 \<deathbybandaid>  I'll open an issue then,,, just needed validation

deathbybandaid:
I'm pretty stumped with this, I've tried just about everything that I saw via stackoverflow, and still nogo.

I even went as far as to delete the database and recreate it between tests too, just in case.

dgw:
@RustyBower I think we could use your eyes on this when you have some time.

RustyBower:
Should be fixed in https://github.com/sopel-irc/sopel/pull/1652

Exirel:
Yup, it's fixed for MySQL indeed.

PR #1652, opened 2019-07-02
db: fix unicode on mysql

RustyBower:


deathbybandaid:
After some testing, I can confirm this clears issue #1650

RustyBower:
I did just configure this for a single column in nick_values, but we should chat about all the potentially places people could put unicode characters and configure appropriately.

deathbybandaid:
I would imagine this should be added to channel_values,, as well as the upcoming plugin_values in PR #1621

dgw:
What @deathbybandaid said. Fixing this for only one DB column is no good; we need to allow Unicode pretty much anywhere there's a string/text. Some IRCds do in fact allow Unicode nicknames, for example.

sopel/db.py
```diff
@@ -59,11 +59,13 @@ class NickValues(BASE):
     NickValues SQLAlchemy Class
     """
     __tablename__ = 'nick_values'
+    __table_args__ = {'mysql_engine': 'InnoDB', 'mysql_charset': 'utf8mb4', 'mysql_collate': 'utf8mb4_unicode_ci'}
     nick_id = Column(Integer, ForeignKey('nick_ids.nick_id'), primary_key=True)
     key = Column(String(255), primary_key=True)
     value = Column(String(255))
 
 
+
```
dgw: Why was this added…?

RustyBower: great question. fixing it tho lol

deathbybandaid:
Fresh database with mysql seems good with the `__table_args__` setup.
I will try testing with sqlite and postgres later today.

I think those 3 databases are the only ones we should focus on. Any other database flavors can wait until people have issues with them.

deathbybandaid:
Should we be concerned about line length with this?

RustyBower:
Fixed the line length and additional table issues. I'll continue to do some research into defaulting this for the entire table.

dgw:
(And I won't give Rusty time to "fix it anyway", either. :smiling_imp:)

Exirel:
Awesome fix! :+1:

PR #1653, opened 2019-07-04
Bump praw version

jonringer:
I've been able to successfully run all the tests with praw-6.3.1. 

Upstreaming version bump after packaging this for nixpkgs.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,6 @@
 xmltodict
 pytz
-praw>=4.0.0,<6.0.0
+praw>=4.0.0,<7.0.0
 geoip2
 aspell-python-py2; python_version < '3'
 aspell-python-py3; python_version >= '3'


jonringer:
I'm assuming the 3.3 test is failing due to an older http package, the xenial tests all pass.

dgw:
You can fix 3.3 by giving it `praw<5.4.0` (the version in which 3.3 was dropped) with an [environment marker](https://www.python.org/dev/peps/pep-0508/#environment-markers). Note that you or another contributor might also want (in a separate PR) to pin `requests` at `<2.22.0` for Python 3.3 & 3.4, since that version dropped support for 3.4 (which implies dropping 3.3 too).

jonringer:
Well, python 3.3 and 3.4 are [end of life](https://www.python.org/downloads/release/python-3410/) not sure if you plan on maintaining it as many packages have ended supporting it as well.

dgw:
We've committed to maintaining support for Python 3.3+ (and 2.7) through the lifecycle of Sopel 7, yes.

Sopel 8 will drop support for all Python versions that are EOL at the time of its release. Would you like to fix this for merging into Sopel 7, or hold the version bump of `praw` until Sopel 8? It'll be a while; we're still probably some months away from releasing version 7.

jonringer:
I'm in no rush, I submitted the PR so your version bounds could be updated. :)

On the nixpkgs end, we can modify the contents of the package before installation/building the wheel, so we were aren't dependent on your actions (you're free to take your time as you see fit :) ).

jonringer:
been 5.5 months, closing

dgw:
As I said before, it'll be a while before development starts on version 8.0. We're just now finalizing 7.0 for release later this month.

jonringer:
packaging is a moving target, I'm entirely sure if the PR is still relevant.

The dependency changes are relatively trivial, and I'm sure they are subject to change near a release.

dgw:
I get it. 😸 There's some question on our side if `praw` still will be a dependency any more in 8.0, even—our long-term development plan includes moving most of the current "built-in" plugins into separate packages for a variety of reasons (separating release cycles & maintenance burdens, primarily; also making it easier to install a Sopel that runs *only* the user's custom plugin code and nothing else), so there's a good chance this would have been closed anyway in the future.

PR #1654, opened 2019-07-05
db: add methods for values associated with a server

deathbybandaid:
I'm not assuming this PR will get much traction, as it is likely an edge-case.

Based on PR #1621, this time for storing values associated with a/the server.

I saw another python IRC bot with database entries for `server`, so I figured it couldn't hurt to toss something like this together.

With PR #1536 , this could potentially assume the `server` key as ISUPPORTs `"NETWORK": "freenode"`

### DIFF
--- a/sopel/db.py
+++ b/sopel/db.py
@@ -80,6 +80,17 @@ class ChannelValues(BASE):
     value = Column(String(255))
 
 
+class ServerValues(BASE):
+    """
+    ServerValues SQLAlchemy Class
+    """
+    __tablename__ = 'server_values'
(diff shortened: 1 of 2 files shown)


dgw:
The code looks fine at a glance, but… what is the use case for this? Sopel instances that jump between servers on a regular basis? I don't get it. 😕

deathbybandaid:
Precisely why I said "edge-case". I had noticed another bot had it, so I figured it wouldn't hurt.

dgw:
I don't think we'll do this at this time. We can definitely reconsider if an actual use-case comes up!

Issue #1658, opened 2019-07-12
Idea: module.py action_commands function decorator


deathbybandaid:
I had an idea after looking at:

https://github.com/dgw/sopel-flipper/blob/master/flipper.py

I think it would be pretty cool to consolidate

````
@sopel.module.rule('^rolls (.+)')
@sopel.module.intent('ACTION')
````

to

````
@sopel.module.action_commands('rolls')
````

and add another format for potential command structures.

dgw:
Could be useful, I guess.

Note that the plain old `@commands` decorator currently also works in `/me` actions, e.g. `* deathbybandaid .tell nick something` still runs the`.tell` command. If there is to be a dedicated decorator for commands that should trigger on actions, I wouldn't want such overlap to exist.

Actually, I wonder if any plugins out there specifically expect their commands to work in actions, or if most plugin authors are unaware that it does work. Maybe I should write up an issue about deprecating that behavior in Sopel 7 so we can clear the way for a dedicated action command decorator in Sopel 8.

dgw:
#1660 was merged a month ago. This feature request should have been closed at that time.

PR #1660, opened 2019-07-15
module: add action_commands decorator

deathbybandaid:
As per my idea in Issue #1658, this is my attempt to integrate an `action_commands` command structure to Sopel.

There are still things to work out here, I'm sure. I copied elements I saw in `intents`, `commands`, and `nickname_commands`. I am very well aware the docstrings will definitely need work, as well as tests added.

I'm also not happy with the import line adjusted, and will likely fix that with `()` later.

This is more/less a draft PR, but my brief tests were very successful.

### DIFF
--- a/sopel/loader.py
+++ b/sopel/loader.py
@@ -68,7 +68,7 @@ def clean_callable(func, config):
             func.rule = [func.rule]
         func.rule = [compile_rule(nick, rule, alias_nicks) for rule in func.rule]
 
-    if hasattr(func, 'commands') or hasattr(func, 'nickname_commands') or hasattr(func, 'action_commands'):
+    if any(hasattr(func, attr) for attr in ['commands', 'nickname_commands', 'action_commands']):
         func.rule = getattr(func, 'rule', [])
         for command in getattr(func, 'commands', []):
             regexp = get_command_regexp(prefix, command)
--- a/sopel/loader.py
+++ b/sopel/loader.py
@@ -4,7 +4,8 @@ from __future__ import unicode_literals, absolute_import, print_function, divisi
 import re
 import sys
 
-from sopel.tools import compile_rule, itervalues, get_command_regexp, get_nickname_command_regexp
+from sopel.tools import (compile_rule, itervalues, get_command_regexp,
+                         get_nickname_command_regexp, get_action_command_regexp)
 from sopel.config import core_section
 
 default_prefix = core_section.CoreSection.help_prefix.default
@@ -67,7 +68,7 @@ def clean_callable(func, config):
             func.rule = [func.rule]
         func.rule = [compile_rule(nick, rule, alias_nicks) for rule in func.rule]
 
-    if hasattr(func, 'commands') or hasattr(func, 'nickname_commands'):
+    if hasattr(func, 'commands') or hasattr(func, 'nickname_commands') or hasattr(func, 'action_commands'):
         func.rule = getattr(func, 'rule', [])
         for command in getattr(func, 'commands', []):
             regexp = get_command_regexp(prefix, command)
@@ -75,6 +76,10 @@ def clean_callable(func, config):
                 func.rule.append(regexp)
         for command in getattr(func, 'nickname_commands', []):
             regexp = get_nickname_command_regexp(nick, command, alias_nicks)
+            if regexp not in func.rule:
+                func.rule.append(regexp)
+        for command in getattr(func, 'action_commands', []):
+            regexp = get_action_command_regexp(command)
             if regexp not in func.rule:
                 func.rule.append(regexp)
         if hasattr(func, 'example'):
@@ -131,6 +136,7 @@ def is_triggerable(obj):
         'intents',
         'commands',
         'nickname_commands',
+        'action_commands',
     )
     allowed = any(hasattr(obj, attr) for attr in allowed_attrs)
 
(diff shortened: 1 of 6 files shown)


deathbybandaid:
Test file used:

````
# coding=utf-8

from __future__ import unicode_literals, absolute_import, division, print_function

import sopel.module


@sopel.module.commands('dbbtest')
def bot_command_dbbtest(bot, trigger):
    bot.say(str(trigger))
    bot.say(str(trigger.args))



@sopel.module.action_commands('rolls')
def bot_command_dbbtestb(bot, trigger):
    bot.say(str(trigger))
    bot.say(str(trigger.args))
````

Output:

````
<deathbybandaid> .dbbtest this out
<dbbbot> .dbbtest this out
<dbbbot> ['#deathbybandaid', '.dbbtest this out']


* deathbybandaid rolls dice
<dbbbot> rolls dice
<dbbbot> ['#deathbybandaid', 'rolls dice']

````

**Comment by dgw** [comment]
Looking at this led me to the discovery that inspired #1683. Specifically, the implementation here made me wonder "what other intents are there?", which led to finding out that the intent spec no longer exists. This decorator is a pretty good idea, actually, especially _because_ we'll want to replace the internal references to "intents" on callables and triggers with something else that isn't obsolete. Could be as simple as `trigger.is_action` and a `callable.on_actions` Boolean (default `False`). Further discussion and ideas very much welcome in #1683!

**Comment by deathbybandaid** [comment]
Forced Pushed to rebase with current master.

I could use some help writing the tests for `test_module.py`.

A thorough review would also be nice to clean/tidy it up. However, I feel fairly positive about what is accomplished here.

As discussed via IRC, this can be implemented, with the current `intents` code, with the intention (not a pun) of other PRs being opened later, such as:

* Re-evaluating how CTCP is handled
* Based on ^ how we want to handle something like `trigger.is_action`
* Deprecating `@intents`
* Removal of the ability to run `@commands` via a CTCP ACTION

File: sopel/module.py [hunk 1]
```diff
@@ -261,6 +261,30 @@ def add_attribute(function):
     return add_attribute
 
 
+def action_commands(*command_list):
+    """Decorate a function to trigger on lines starting with "/me command".
+    This decorator can be used multiple times to add multiple rules. The
+    resulting match object will have the command as the first group, rest of
+    the line, excluding leading whitespace, as the second group. Parameters 1
+    through 4, separated by whitespace, will be groups 3-6.
```
dgw:
```suggestion
    through 4, separated by whitespace, will be groups 3-6.

```

sopel/module.py
```diff
@@ -261,6 +261,30 @@ def add_attribute(function):
...
+    Args:
+        command: A string, which can be a regular expression.
```
dgw:
```suggestion
        command: A string, which can be a regular expression.

```

sopel/module.py
```diff
@@ -261,6 +261,30 @@ def add_attribute(function):
...
+    Returns:
+        A function with a new regular expression appended to the rule
+        attribute. If there is no rule attribute, it is added.
```
dgw:
```suggestion
        attribute. If there is no rule attribute, it is added.

```

sopel/module.py
```diff
@@ -261,6 +261,30 @@ def add_attribute(function):
...
+    Example:
+        @action_commands("hello!"):
+            Would trigger on "/me hello!"
```
dgw: I think it would be clearer to give a "proper" action as example, e.g. `/me says hello`.

The explanation could then say it would trigger "when a user 'says hello' as an action", or something. We should avoid doing `"/me hello!"` or anything else that might imply "/me" is part of the message. (This is also why I suggested an amendment to the one-line description.)

deathbybandaid:
That was really odd with the whitespace disappearing from my push. Looked normal in my editor.

Added all the requested changes though.

dgw:
Hmm, that force-push undid all the changes. You should try rebasing/squashing again from 44cd95d.

deathbybandaid:
@dgw, something is screwy on my end, and everytime I squash, those whitespaces change.
I'll look at it more when I get the chance.

deathbybandaid:
I deleted the cloned directory, recloned, squashed, and it looks like the problem went away.

Not sure what was happening before, but it looks like everything is there.

sopel/module.py
dgw: Meh. I'll leave this to be addressed when this file gets its turn for #1565.

dgw:
If I'm touching tests, I like @Exirel to look at it. 😁

dgw:
Had to rebase on `master` for that last tweak (getting logging & docstrings squared away in `bot.py`). I didn't _want_ to rebase, but the alternative was making a separate PR, and _that_ seemed silly.

dgw:
And rebased again due to conflicting changes in #1708. 😦

sopel/loader.py
```diff
@@ -67,7 +68,7 @@ def clean_callable(func, config):
             func.rule = [func.rule]
         func.rule = [compile_rule(nick, rule, alias_nicks) for rule in func.rule]
 
-    if hasattr(func, 'commands') or hasattr(func, 'nickname_commands'):
+    if hasattr(func, 'commands') or hasattr(func, 'nickname_commands') or hasattr(func, 'action_commands'):
```
deathbybandaid: I know it isn't that much shorter, but it might help line length here to do

`if all(hasattr(func, attr) for attr in ['commands', 'nickname_commands', 'action_commands']):`

dgw: Using `all()` would break it.

Exirel: Yes, `any` would be cool.

PR #1661, opened 2019-07-15
seen: Improve readability of information

kwaaak:
Now that #1560 has been merged, let's use it.

This changes the reply to "seen" commands:
before
`nickname: I last saw other_nickname at 2019-05-29 - 23:04:13UTC in here, saying gotta go`
after
`nickname: I last saw other_nickname in here 45 days, 4 hours ago, saying: gotta go`

I don't know if the timestamp should be included in the reply or not.

### DIFF
--- a/sopel/modules/seen.py
+++ b/sopel/modules/seen.py
@@ -3,6 +3,7 @@
 seen.py - Sopel Seen Module
 Copyright 2008, Sean B. Palmer, inamidst.com
 Copyright © 2012, Elad Alfassa <elad@fedoraproject.org>
+Copyright 2019, Sopel contributors
 Licensed under the Eiffel Forum License 2.
 
 https://sopel.chat
@@ -14,7 +15,7 @@ import time
 
 from sopel.module import commands, rule, priority, thread
 from sopel.tools import Identifier
-from sopel.tools.time import get_timezone, format_time
+from sopel.tools.time import get_timezone, format_time, seconds_to_human
 
 
 @commands('seen')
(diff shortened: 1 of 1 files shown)


sopel/modules/seen.py
```diff
@@ -36,17 +37,18 @@ def seen(bot, trigger):
         tz = get_timezone(bot.db, bot.config, None, trigger.nick,
                           trigger.sender)
         saw = datetime.datetime.utcfromtimestamp(timestamp)
+        delta = seconds_to_human((trigger.time - saw).total_seconds())
         timestamp = format_time(bot.db, bot.config, tz, trigger.nick,
```
Exirel: I guess this line can be removed now.

sopel/modules/seen.py
```diff
@@ -36,17 +37,18 @@ def seen(bot, trigger):
...
                                 trigger.sender, saw)
 
-        msg = "I last saw {} at {}".format(nick, timestamp)
+        msg = "I last saw " + nick
         if Identifier(channel) == trigger.sender:
             if action:
-                msg = msg + " in here, doing " + nick + " " + message
+                msg = msg + " in here " + delta + ", doing: " + nick + " " + message
```
Exirel: While we are at it, I'd say that a `.format` would look better to me:

```
msg = 'I last saw {nick} in here {since}, doing: {nick} {action}'.format(
    nick=nick,
    since=delta,
    action=message
)
```

Of course, this would require to change the other cases accordingly.

sopel/modules/seen.py
```diff
@@ -36,17 +37,18 @@ def seen(bot, trigger):
...
             else:
-                msg = msg + " in here, saying " + message
+                msg = msg + " in here " + delta + ", saying: " + message
         else:
-            msg += " in another channel."
+            msg += " in another channel " + delta + "."
         bot.say(str(trigger.nick) + ': ' + msg)
```
Exirel: If you fancy it, you can upgrade the code with `bot.reply`:

```suggestion
         bot.reply(msg)
```

kwaaak:
This change drastically simplifies the module.

Instead of doing
`bot.db.set_nick_value(trigger.nick, 'seen_timestamp', time.time())`
we can just do
`bot.db.set_nick_value(trigger.nick, 'seen_timestamp', trigger.time)`
and get rid of the `datetime` and `time` imports.

The code in this module runs a lot. The datetime object `trigger.time` would be twice as big as the float `time.time()`.

Issue #1659, opened 2019-07-14
User's "away" status always "None"


rdswift:
I'm trying to test the away status of the users in a channel, but the `away` attribute of the `user` class is always returning 'None', even when the user shows up in my IRC client as being away (and they have a current away message).  My test script is:

```
@sopel.module.require_admin()
@sopel.module.require_privmsg()
@sopel.module.commands('userlist')
def userlist(bot, trigger):
    channel = sopel.tools.Identifier('#MyTestChannel')
    try:
        privs = bot.channels[channel].users
        for username in privs:
            user = privs[username]
            text = "User: %s status is " % (user.nick)
            if (user.away):
                text += "Away (%s)" % (user.away)
            else:
                text += "Active"
            bot.notice(text)
    except Exception as ex:
        bot.notice("Exception: %s" % (ex))
```

Sopel v. 6.6.9 running on openSUSE 42.1 (x86_64).  Sopel installed using pip.

Am I doing something wrong, or perhaps misunderstanding the meaning of the `away` attribute?

dgw:
This is a good catch. On networks that do _not_ support the `away-notify` CAP, `User` objects are never getting their `away` property set. Even on networks that _do_ support `away-notify`, the property is onlyl set after the first `AWAY` message received for that user.

We've a start on fixing the initial population of values via `WHO` replies in #1663 (thank you, @rdswift!), and we'll keep working on further enhancements—e.g. #1664, which intends to add `WHO` polling for the minority of networks that still do not support `away-notify`.

Another possible enhancement would be to assign the away _message_ (if known) to `away`, rather than only a Boolean value. I haven't thought through all of the implications yet, though it should work in theory (non-empty strings are truthy, and AFAIK an empty away message isn't possible). We might also keep the Boolean in `away` and add `User.awaymsg` or similar to contain the message itself (`None` if the user is not away).

Exirel:
> Another possible enhancement would be to assign the away _message_ (if known) to `away`, rather than only a Boolean value.

Not a huge fan, since you never now what weird stuff people are doing with boolean (like `int(True)` to get 1 or 0).

In favor of `away_message` (accurate yet short enough).

dgw:
I believe this is fully resolved by #1663 and #1664, both of which are now merged.

PR #1664, opened 2019-07-15
coretasks: Periodically send WHO to keep user information up-to-date

rdswift:
Add an interval event to periodically issue a WHO request to keep user information up-to-date.

This provides part of the solution to Issue #1659 : User's "away" status always "None"

### DIFF
--- a/sopel/coretasks.py
+++ b/sopel/coretasks.py
@@ -401,15 +401,27 @@ def _send_who(bot, channel):
         # We might be on an old network, but we still care about keeping our
         # user list updated
         bot.write(['WHO', channel])
+    channel.last_who = int(time.time())
 
 
 @sopel.module.interval(30)
 def _periodic_send_who(bot):
     """Periodically send a WHO request to keep user information up-to-date."""
-    if ('away-notify' in bot.enabled_capabilities):
+    if 'away-notify' in bot.enabled_capabilities:
         return
+    # Loops through the channels to find the one that has the longest time since the last WHO request,
+    # and issues a WHO request only if the last request for the channel was more than
+    # 'minimum_seconds_between_who_requests_per_channel' seconds ago.
+    current_time = int(time.time())
+    oldest_who_channel = None
+    oldest_who_time = current_time
+    minimum_seconds_between_who_requests_per_channel = 120
     for channel in bot.channels:
-        _send_who(bot, channel)
+        if (channel.last_who is None) or (channel.last_who < oldest_who_time):
+            oldest_who_channel = channel
+            oldest_who_time = 1 if channel.last_who is None else channel.last_who
+    if oldest_who_channel is not None and (current_time - oldest_who_time) > minimum_seconds_between_who_requests_per_channel:
+        _send_who(bot, oldest_who_channel)
 
 
 @sopel.module.rule('.*')
--- a/sopel/coretasks.py
+++ b/sopel/coretasks.py
@@ -427,7 +427,7 @@ def _periodic_send_who(bot):
             selected_channel = channel
             who_trigger_time = channel.last_who
 
-    if (selected_channel is not None):
+    if selected_channel is not None:
         # selected_channel's last who is either none or the oldest valid
         _send_who(bot, selected_channel)
 
--- a/sopel/coretasks.py
+++ b/sopel/coretasks.py
@@ -406,6 +406,8 @@ def _send_who(bot, channel):
 @sopel.module.interval(30)
 def _periodic_send_who(bot):
     """Periodically send a WHO request to keep user information up-to-date."""
+    if ('away-notify' in bot.enabled_capabilities):
+        return
     for channel in bot.channels:
         _send_who(bot, channel)
 
--- a/sopel/coretasks.py
+++ b/sopel/coretasks.py
@@ -13,6 +13,7 @@ dispatch function in bot.py and making it easier to maintain.
 from __future__ import unicode_literals, absolute_import, print_function, division
 
 from random import randint
+import datetime
 import re
 import sys
 import time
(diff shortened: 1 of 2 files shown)


dgw:
Not a bad idea. We should definitely put some thought into the interval duration, based on how other bots and clients behave. More importantly, Sopel shouldn't do any `WHO` polling for away statuses when the `away-notify` capability is enabled, because `AWAY` messages from the server will take care of it: https://github.com/sopel-irc/sopel/blob/9340c674b327e0b43bdc5a0f7bac31ee1e61cad6/sopel/coretasks.py#L822-L831

Our `JOIN` handling already takes care of sending a `WHO` on joining a new channel, fortunately. On networks that support `away-notify` ([something like 84%](https://stats.ircdocs.horse/client-caps/#cap-away-notify) of them, at last survey), #1663 alone will fix away-status tracking. This is only needed for the ~16% of networks that don't support the more efficient method yet.

rdswift:
I agree.  This PR should be updated to use a reasonable interval, and to avoid sending the WHO request if the `away-notify` capability is enabled.  I assume there's an easy way to test for that?

dgw:
`'away-notify' in bot.enabled_capabilities` or something like that should do. Simplest way is to just `return` at the beginning of this added function if that's true. Big overhaul of network properties coming with #1536 (hopefully), which might change it, but that test should work right now.

As far as the interval duration itself, I've done a small survey of clients I use or have used in the past:

* Textual: [120 seconds](https://github.com/Codeux-Software/Textual/blob/a43cca55d45fc3dff33105549ab813ecc0988cc1/Sources/App/Classes/IRC/IRCClient.m#L171)
* HexChat: [update 1 channel every 30 seconds](https://github.com/hexchat/hexchat/blob/57478b65758e6b697b1d82ce21075e74aa475efc/src/common/hexchat.c#L306)
* KVIrc: [update oldest WHO list if older than 150 seconds](https://github.com/kvirc/KVIrc/blob/665fc15dde8665e3fdec83b038000952296e63e0/src/kvirc/kernel/KviIrcConnection.cpp#L1979-L1980)
* The Lounge: appears to do no polling at present, relying on `away-notify` only

HexChat's is probably the nicest approach, but also the most complex to implement and the most likely to have outdated information (since as the bot's channel count grows, it takes longer between updates of a given channel's status information). I'd say KVIrc's approach is the next-nicest.

rdswift:
The KiwiIRC client looks to be the same as HexChat (which is what I based my 30 seconds on).  I agree the KvIRC approach seems to make the most sense.  I'll try to start taking a look at this tomorrow.

rdswift:
Further to the discussion, I've included some additional logic around sending periodic `WHO` requests.  It is sort of a hybrid between the HexChat and KvIRC approaches.

This function checks every 30 seconds to see if any channel's last `WHO` request was greater than 120 seconds ago, and initiates a request for the channel with the oldest previous request time.  It does not issue more than one `WHO` request every 30 seconds, and does not issue more than one `WHO` request for a channel within 120 seconds.

The Channel class in target.py was modified to provide a class attribute to track the last time a `WHO` request was issued for the channel.

I have no idea why this would be causing one of the DuckDuckGo search engine tests to be failing under Python 2.7.  I suspect that might be an unrelated issue?

rdswift:
I tightened up (and simplified) the code a bit and got rid of the "arbitrary" comparison value.  Also changed some of the variable names to more accurately reflect their purpose (and shorten them a bit).  I think the logic is a bit clearer now.  Is this more in line with what you were looking for?

rdswift:
@dgw,

I've added the `break` from the loop as you suggested.  Good catch!  I've also broken down each test separately which should make the logic quite clear.  There are a couple of lines that could be combined to avoid some duplicate code, as:

```
    who_trigger_interval = 120
    who_trigger_time = datetime.datetime.utcnow() - datetime.timedelta(seconds=who_trigger_interval)
    selected_channel = None
    for channel in bot.channels:
        if channel.last_who is None:
            selected_channel = channel
            break
        if selected_channel is None or channel.last_who < selected_channel.last_who:
            selected_channel = channel
    if (selected_channel is not None):
        if selected_channel.last_who is None or selected_channel.last_who < who_trigger_time:
            _send_who(bot, selected_channel)
```

I think the logic is still clear, but I'm biased because I developed it.  :wink:   I'm okay with it the way it is now (each test separate), or I can make these changes if you prefer.

Exirel:
Ah! I like where you are going with that. Let me push that even further:

``` python
    who_trigger_interval = datetime.timedelta(seconds=120)
    who_trigger_time = datetime.datetime.utcnow() - who_trigger_interval
    selected_channel = None

    for channel in bot.channels:
        if channel.last_who is None:
            # WHO was never sent yet to this channel: stop here
            selected_channel = channel
            break
        if channel.last_who < who_trigger_time:
            # this channel's last who request is the most outdated one at the moment
            selected_channel = channel
            who_trigger_time = channel.last_who

     if selected_channel is not None:
             # selected_channel's last who is either none or the oldest valid
             _send_who(bot, selected_channel)
```

The `selected_channel` will always be `None` if there is no channel that match the condition: who is None, or there is an oldest one. The trick is to initialize the `who_trigger_time` to the base limit, then to update that limit whenever we found a more outdated channel.

rdswift:
@Exirel, that suggestion is brilliant!  I'll update the PR accordingly.

sopel/coretasks.py
```diff
@@ -401,6 +402,34 @@ def _send_who(bot, channel):
...
+        if channel.last_who is None:
+            # WHO was never sent yet to this channel: stop here
+            selected_channel = channel
+            break
+        if channel.last_who < who_trigger_time:
+            # this channel's last who request is the most outdated one at the moment
+            selected_channel = channel
+            who_trigger_time = channel.last_who
+
+    if (selected_channel is not None):
```
Exirel:
```suggestion
    if selected_channel is not None:
```

rdswift: Darn.  I missed that when I was doing the last clean-up.  Done now.

PR #1666, opened 2019-07-15
coretasks: Only update user 'away' value if value passed is not 'None'

rdswift:
Only update the 'away' value for a user if the value passed to the `_record_who` function is not None.

See the discussion at https://github.com/sopel-irc/sopel/pull/1663

### DIFF
--- a/sopel/coretasks.py
+++ b/sopel/coretasks.py
@@ -780,7 +780,8 @@ def _record_who(bot, channel, user, host, nick, account=None, away=None, modes=N
         usr.account = None
     else:
         usr.account = account
-    usr.away = away
+    if away is not None:
+        usr.away = away
     priv = 0
     if modes:
         mapping = {


dgw:
This shouldn't include 7bbf23f; that commit is completely independent of this change.

rdswift:
> This shouldn't include [7bbf23f](https://github.com/sopel-irc/sopel/commit/7bbf23fe9309181b5372df955ecc78fa0ced5775); that commit is completely independent of this change.

Argh!  I used the wrong source when creating this branch.  I'll get rid of that code.  Sorry about that.

dgw:
It will still merge-conflict this way. You'll need to rebase the branch to remove the unwanted commits entirely. `git rebase -i master` should give you an editor, and you'll just delete the lines representing the unwanted commits. Save, exit editor, and Git should take care of it. `git push -f` (with remote name and branch name if you want to be safe) to update the PR.

PR #1672, opened 2019-07-30
bot: only suppress echo for PRIVMSG and NOTICE

half-duplex:
thijseigenwijs on IRC reported that the bot doesn't update privilege tracking when it's the one to change a mode. This is because the server's MODE message is ignored by the echo suppression. This PR changes the echo suppression to only suppress PRIVMSG and NOTICE events.

- Can we assume the event will always be upper case?
- Any others that should be suppressed?

### DIFF
--- a/sopel/bot.py
+++ b/sopel/bot.py
@@ -665,7 +665,10 @@ class Sopel(irc.Bot):
         event = pretrigger.event
         intent = pretrigger.tags.get('intent')
         nick = pretrigger.nick
-        is_echo_message = nick.lower() == self.nick.lower()
+        is_echo_message = (
+            nick.lower() == self.nick.lower() and
+            event in ["PRIVMSG", "NOTICE"]
+        )
         user_obj = self.users.get(nick)
         account = user_obj.account if user_obj else None
 


sopel/bot.py
```diff
@@ -665,7 +665,10 @@ def dispatch(self, pretrigger):
         event = pretrigger.event
         intent = pretrigger.tags.get('intent')
         nick = pretrigger.nick
-        is_echo_message = nick.lower() == self.nick.lower()
+        is_echo_message = (
+            nick.lower() == self.nick.lower() and
+            event in ["PRIVMSG", "NOTICE"]
```
dgw: No need to wonder if the event will always be uppercase if you do this.

```suggestion
            event.upper() in ["PRIVMSG", "NOTICE"]
```

But there's plenty of code already assuming these will always be uppercase, so it's probably not worth doing only here. (FWIW, most of that code seems to use `'single quotes'`, but that's also likely not worth changing here.)

Exirel:
>     * Can we assume the event will always be upper case?

Yes.

PR #1674, opened 2019-08-03
irc: Don't write line ending twice.

Tomin1:
When echo message is simulated it is dispatched the message with '\r\n'
appended to the end. PreTrigger expects the message to not to have any
newline or carriage return characters at the end of the message.

### DIFF
--- a/sopel/trigger.py
+++ b/sopel/trigger.py
@@ -27,8 +27,9 @@ class PreTrigger(object):
 
     def __init__(self, own_nick, line):
         """own_nick is the bot's nick, needed to correctly parse sender.
-        line is the full line from the server."""
-        line = line.strip('\r')
+        line is the full line from the server or from simulated echo
+        message."""
+        line = line.strip('\r\n')
         self.line = line
 
         # Break off IRCv3 message tags, if present


dgw:
I doubt CR or LF will be allowed even with multi-line. Discussion on that has all been about using BATCH or client-side message-tags, so let's not worry about it here.

@Tomin1 I agree that `PreTrigger` itself should handle making sure the message is cleaned properly. Thanks to @HumorBaby for digging around proactively, so I didn't have to!

Tomin1:
@dgw Sure, I'm fine doing this some other way. This was easy for someone that doesn't know sopel code base very well (yet) without breaking other use cases. I just noticed that chanlogs module appends extra characters to sopel's own messages when I added echo support to it.

I think I'll make a bit different variation of that example code from @HumorBaby. It's more readable if it's not all on one line IMO.

Tomin1:
After procrastating a bit (some say it's useful to a point), I decided a bit differently. I think it's better to just add that \n to be stripped in PreTrigger. That is also something @HumorBaby suggested.

Though, I think it would be nicer to give PreTrigger message without \n character but then again this makes less copies of it. I think CR-LF adding code should belong to send instead of write probably but that's not something that would need to be done now.

dgw:
> I think CR-LF adding code should belong to send instead of write probably

Agreed. If you like, submit a PR for that. But wait until #1676 is merged unless you like getting scared by CI failures that aren't your fault. :smile_cat:

Issue #1642, opened 2019-06-23
spellcheck: remove to separate package


dgw:
Despite switching from `enchant` to `aspell` (#1164), the `spellcheck` module can still cause issues with installing Sopel on Windows (#1142). Even on *nix systems, there are system-level package requirements (libraries and dictionary files) that often trip up the installation. It's more than a little absurd for one optional plugin to potentially halt the entire install process for Sopel, when the user might not even want _any_ of the core module set.

I still want to leave most "package-ification" for a later release (see #1291), but I've become convinced over the past few months that removing `spellcheck` from the base installation package sooner is warranted. I just see way too many complaints about installation problems related to it, and `spellcheck` is the only core plugin with dependencies that `pip` can't always install.

So, if anyone wants to be co-maintainer of the soon-to-be-external `spellcheck` plugin, comment here and/or send me a PM on freenode (I'm also `dgw` there). When I return from my overseas trip in a week or so, I'll make the repository, add collaborators (if any), and get started on this.

----

Tasks:

- [x] Remove plugin and dependencies from core (PR open: #1675)
- [x] Create new package repo (done: [sopel-irc/sopel-spellcheck](https://github.com/sopel-irc/sopel-spellcheck))
- [x] Extract `spellcheck.py` into its own package and test it (done: see above repo)
- [x] Publish new package (will probably wait until at least an alpha of Sopel 7 is published, so `pip` can actually find a valid requirement candidate)

dgw:
The plugin is now published at https://pypi.org/project/sopel-spellcheck/ (or see [the repo](https://github.com/sopel-irc/sopel-spellcheck)).

As PyPI does not have any release of Sopel 7.x available yet, this requires an existing "editable" install of Sopel @ 40162a46ad20867a50162a25708248da3083d180 or newer. PyPI will not be able to satisfy the `sopel>=7.0` dependency on its own until we begin publishing alpha/beta/RC versions.

PR #1675, opened 2019-08-06
spellcheck: remove plugin (moving to separate package)

dgw:
Removed all relevant search hits for "spellcheck", "aspell", and/or "enchant".

Working on the new package locally; I plan to make it installable from PyPI as `sopel-spellcheck` when it's ready to publish. New repo will be [sopel-irc/sopel-spellcheck](https://github.com/sopel-irc/sopel-spellcheck).

See #1642. Note that this breaks the transition plan described in #1291, but adding `sopel-spellcheck` to the requirements for Sopel 7 would only perpetuate the dependency hell we're trying to solve.


dgw:
Note: The plugin is now installable from https://pypi.org/project/sopel-spellcheck/ (or see [the repo](https://github.com/sopel-irc/sopel-spellcheck)).

PR #1676, opened 2019-08-06
dev-requirements: cap flake8-future-import version

dgw:
0.4.6 added warnings for the annotations feature (we don't care about it yet), and dropped Python 3.3 support (we care about that still).

We'll make sure to use 0.4.5 or lower in tests.

There haven't been any CI builds on this repo in the last few days, which is why this caused a surprise failure on #1675. `flake8-future-import` 0.4.6 was released only 2 days ago.

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -4,7 +4,7 @@ coveralls
 flake8<3.6.0; python_version == '3.3'
 flake8>=3.7.0,<3.8.0; python_version != '3.3'
 flake8-coding
-flake8-future-import
+flake8-future-import<0.4.6
 setuptools<40.0; python_version == '3.3'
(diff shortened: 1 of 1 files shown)


dgw:
Cap the version now to both fix this and prevent future (heh) surprises, and obey the note when we drop support for py3.3. <sub>One of the reasons I didn't just update the ignores was that the newer `flake8-future-import` version also dropped their testing on 3.3, so future releases of that plugin aren't guaranteed to be supported.</sub>

dgw:
Let's drop this into `master` before I go to bed, so I can kick the stuff that failed and poke for rebases/force-pushes on anything that won't resolve by just re-triggering Travis.

PR #1677, opened 2019-08-09
etymology: handle empty argument

### LINKED ISSUE #1657: etymology: Empty argument not handled
Doing `.ety` with no or an empty argument doesn't behave properly.

On Python 2 (my old personal instance), it causes `TypeError: None object cannot be quoted (file "/usr/lib/python2.7/urllib.py", line 1255, in quote)`

On Python 3 ("official" Sopel instance on freenode), the command looks up the etymology of "None" (and gets no results, because of the capital letter).

This is a good, easy issue for first-time contributors to fix.

AlfredOnGitHub:
Fix #1657 

Changed the etymology.py file to solve the issue with no or an empty argument.

Thank you.

### DIFF
--- a/sopel/modules/etymology.py
+++ b/sopel/modules/etymology.py
@@ -40,6 +40,9 @@ def etymology(word):
     # @@ <nsh> sbp, would it be possible to have a flag for .ety to get 2nd/etc
     # entries? - http://swhack.com/logs/2006-07-19#T15-05-29
 
+    if not word:
+        raise ValueError("No word to look for.")
+
     if len(word) > 25:
         raise ValueError("Word too long: %s[…]" % word[:10])
 


sopel/modules/etymology.py
```diff
@@ -40,8 +40,11 @@ def etymology(word):
     # @@ <nsh> sbp, would it be possible to have a flag for .ety to get 2nd/etc
     # entries? - http://swhack.com/logs/2006-07-19#T15-05-29
 
+    if len(word) == 0:
+        return 'No word added.'
+
     if len(word) > 25:
-        raise ValueError("Word too long: %s[…]" % word[:10])
+        return "Word too long: %s[…]" % word[:10]
```
Exirel: I would rather keep the `ValueError` logic here. Beside, the "no word" condition above could do exactly the same.

sopel/modules/etymology.py
```diff
@@ -40,8 +40,11 @@ def etymology(word):
     # @@ <nsh> sbp, would it be possible to have a flag for .ety to get 2nd/etc
     # entries? - http://swhack.com/logs/2006-07-19#T15-05-29
 
+    if len(word) == 0:
+        return 'No word added.'
```
Exirel: I'd suggest to raise a Value error here instead:

```suggestion
        raise ValueError('No word to look for.')
```

sopel/modules/etymology.py
Exirel:
```suggestion
        raise ValueError('Word too long: %s[…]' % word[:10])
```

dgw:
Yeah, uh, this is a big patch for what I'd expect to be a two- or three-line change at most.

AlfredOnGitHub:
Hey, I'm sorry for being away.

As I said before, I'm a newbie here, so I apologize for the big patch I requested. I hope that this new pull request will fix the errors we have here.

Thanks for letting me work on this, and if you guys have something else to say to improve the code, please, let me know.

sopel/modules/etymology.py
```diff
@@ -79,8 +81,10 @@ def f_etymology(bot, trigger):
         msg = "Can't connect to etymonline.com (%s)" % (ETYURI % web.quote(word))
         bot.say(msg, trigger.sender)
         return NOLIMIT
-    except (AttributeError, TypeError):
+    except AttributeError:
         result = None
+    except TypeError:
+        result = "No word to look for."
```
dgw: Can you help me understand why you split apart this exception handling? `etymology(word)` will never explicitly raise `TypeError`, so I don't think we can equate that type (puns!) of exception to any specific error state.

Exirel: Hm... I overlooked that, but I think it's because of that:

```
>>> len(None)
TypeError: object of type 'NoneType' has no len()
```

So I think the solution is to not use `if len(word) == 0:`, but instead `if not word:` which will test for None, 0, False, and empty string, all without raising this `TypeError`.

AlfredOnGitHub: The len(word) == 0 was for .ety with a space. The TypeError is what you get when use .ety without anything else.

But I think that we don't care about it anymore. I made the changes that @Exirel said and added the blank line on the headers. Also, I think I squash the commits, but I'm not sure about it because it is my first time doing it.

I like the project so, maybe I was a bit lost like this past days, but I will work on it. :)

dgw: Squash is 👌!

Still moves the blank line from inside to outside the module docstring header, though. Other than that (we try to avoid touching random whitespace in unrelated commits/PRs), ready to 🚢.

sopel/modules/etymology.py
```diff
@@ -4,7 +4,6 @@
 Copyright 2007-9, Sean B. Palmer, inamidst.com
 Copyright 2018-9, Sopel contributors
 Licensed under the Eiffel Forum License 2.
-
```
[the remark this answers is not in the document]
AlfredOnGitHub: The thing is that I don't have that blank line inside the header, I have it outside. I can't ```git add``` anything because there's no change :/.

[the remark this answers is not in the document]
dgw: You could move the blank line, and amend your branch's commit?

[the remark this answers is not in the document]
AlfredOnGitHub: Well, there it is.

Sorry for the wait, I was travelling. As always, if you got something else to improve the code, please, let me know :dancer:

sopel/modules/etymology.py
```diff
@@ -7,6 +7,7 @@
 
 https://sopel.chat
 """
+
```
dgw: This blank line is still added here, and shouldn't be.

dgw:
@AlfredOnGitHub Why closed?

PR #1678, opened 2019-08-10
[core] Rework logging

### LINKED ISSUE #1337: Sopel abuses log levels because of the default setting
Modules like `ip.py` (which I noticed while reviewing #1336) use the wrong log level for some things because using the correct level would result in the log entry not getting printed by default. (That module spits out a `warning` when downloading the GeoIP database, which should be an `info` line, nothing more.)

Sopel's default log level should be changed to `INFO`, and modules that abuse `logger.warning()` or `logger.error()` should be corrected. This is probably going to be a Sopel 7 change.

Exirel:
Fix #1337 

Python built-in logging package makes a lot of thing way easier: no more raw log manually added to a file, no more a scaffold of try/except because opening a log file could be a problem... everything better now.

Of course, this isn't 100% ISO-functionnal, because that's not the point. And I'm also sure we could spend a ridiculous amount of time to make sure we have a 100% perfectly consistent logging system everywhere.

But I believe it's a good first step. Something that we can be proud of, and build upon. It's still better quality code than it was!

### DIFF
--- a/sopel/bot.py
+++ b/sopel/bot.py
@@ -212,7 +212,28 @@ class Sopel(irc.Bot):
         self.scheduler.start()
 
     def setup_logging(self):
-        logger.setup_logging(self)
+        logger.setup_logging(self.config)
+        base_level = self.config.core.logging_level or 'INFO'
+        base_format = self.config.core.logging_format
+        base_datefmt = self.config.core.logging_datefmt
+
+        # configure channel logging if required by configuration
+        if self.config.core.logging_channel:
+            channel_level = self.config.core.logging_channel_level or base_level
+            channel_format = self.config.core.logging_channel_format or base_format
+            channel_datefmt = self.config.core.logging_channel_datefmt or base_datefmt
+            channel_params = {}
+            if channel_format:
+                channel_params['fmt'] = channel_format
+            if channel_datefmt:
+                channel_params['datefmt'] = channel_datefmt
+            formatter = logger.ChannelOutputFormatter(**channel_params)
+            handler = logger.IrcLoggingHandler(self, channel_level)
+            handler.setFormatter(formatter)
+
+            # set channel handler to `sopel` logger
+            LOGGER = logging.getLogger('sopel')
+            LOGGER.addHandler(handler)
 
     def setup_plugins(self):
         load_success = 0
--- a/sopel/bot.py
+++ b/sopel/bot.py
@@ -11,7 +11,6 @@ from ast import literal_eval
 import collections
 import itertools
 import logging
-import os
 import re
 import sys
 import threading
@@ -19,7 +18,7 @@ import time
 
 from sopel import irc, logger, plugins, tools
 from sopel.db import SopelDB
-from sopel.tools import stderr, Identifier, deprecated
+from sopel.tools import Identifier, deprecated
 import sopel.tools.jobs
 from sopel.trigger import Trigger
 from sopel.module import NOLIMIT
@@ -240,7 +239,7 @@ class Sopel(irc.Bot):
         load_error = 0
         load_disabled = 0
 
-        stderr("Loading plugins...")
+        LOGGER.info('Loading plugins...')
         usable_plugins = plugins.get_usable_plugins(self.config)
         for name, info in usable_plugins.items():
             plugin, is_enabled = info
@@ -252,10 +251,7 @@ class Sopel(irc.Bot):
                 plugin.load()
             except Exception as e:
                 load_error = load_error + 1
-                filename, lineno = tools.get_raising_file_and_line()
-                rel_path = os.path.relpath(filename, os.path.dirname(__file__))
-                raising_stmt = "%s:%d" % (rel_path, lineno)
-                stderr("Error loading %s: %s (%s)" % (name, e, raising_stmt))
+                LOGGER.exception('Error loading %s: %s', name, e)
             else:
                 try:
                     if plugin.has_setup():
@@ -263,24 +259,20 @@ class Sopel(irc.Bot):
                     plugin.register(self)
                 except Exception as e:
                     load_error = load_error + 1
-                    filename, lineno = tools.get_raising_file_and_line()
-                    rel_path = os.path.relpath(
-                        filename, os.path.dirname(__file__)
-                    )
-                    raising_stmt = "%s:%d" % (rel_path, lineno)
-                    stderr("Error in %s setup procedure: %s (%s)"
-                           % (name, e, raising_stmt))
+                    LOGGER.exception('Error in %s setup: %s', name, e)
                 else:
                     load_success = load_success + 1
-                    print('Loaded: %s' % name)
+                    LOGGER.debug('Plugin loaded: %s', name)
 
         total = sum([load_success, load_error, load_disabled])
         if total and load_success:
-            stderr('Registered %d modules' % (load_success - 1))
-            stderr('%d modules failed to load' % load_error)
-            stderr('%d modules disabled' % load_disabled)
+            LOGGER.info(
+                'Registered %d modules, %d failed, %d disabled',
+                (load_success - 1),
+                load_error,
+                load_disabled)
         else:
-            stderr("Warning: Couldn't load any modules")
+            LOGGER.warning("Warning: Couldn't load any modules")
 
     def reload_plugin(self, name):
         """Reload a plugin
@@ -298,12 +290,12 @@ class Sopel(irc.Bot):
         # tear down
         plugin.shutdown(self)
         plugin.unregister(self)
-        print('Unloaded: %s' % name)
+        LOGGER.info('Unloaded plugin %s', name)
         # reload & setup
         plugin.reload()
         plugin.setup(self)
         plugin.register(self)
-        print('Reloaded: %s' % name)
+        LOGGER.info('Reloaded plugin %s', name)
 
     def reload_plugins(self):
         """Reload all plugins
@@ -317,14 +309,14 @@ class Sopel(irc.Bot):
         for name, plugin in registered:
             plugin.shutdown(self)
             plugin.unregister(self)
-            print('Unloaded: %s' % name)
+            LOGGER.info('Unloaded plugin %s', name)
 
         # reload & setup all plugins
         for name, plugin in registered:
             plugin.reload()
             plugin.setup(self)
             plugin.register(self)
-            print('Reloaded: %s' % name)
+            LOGGER.info('Reloaded plugin %s', name)
 
     def add_plugin(self, plugin, callables, jobs, shutdowns, urls):
         """Add a loaded plugin to the bot's registry"""
@@ -349,12 +341,13 @@ class Sopel(irc.Bot):
                     for regex in regexes:
                         if func == self.memory['url_callbacks'].get(regex):
                             self.unregister_url_callback(regex)
+                            LOGGER.debug('URL Callback unregistered %r', regex)
         except:  # noqa
-            # TODO: consider logging?
-            raise  # re-raised
+            raise
         else:
             # remove plugin from registry
             del self._plugins[name]
+            LOGGER.info('Plugin removed %s', name)
 
     def has_plugin(self, name):
         """Tell if the bot has registered this plugin by its name"""
@@ -367,6 +360,7 @@ class Sopel(irc.Bot):
         :type obj: :term:`object`
         """
         if not callable(obj):
+            LOGGER.warning('Cannot unregister obj %r: not a callabled', obj)
             return
         if hasattr(obj, 'rule'):  # commands and intents have it added
             for rule in obj.rule:
@@ -785,36 +779,32 @@ class Sopel(irc.Bot):
 
     def _shutdown(self):
         # Stop Job Scheduler
-        stderr('Stopping the Job Scheduler.')
+        LOGGER.info('Stopping the Job Scheduler.')
         self.scheduler.stop()
 
         try:
             self.scheduler.join(timeout=15)
         except RuntimeError:
-            stderr('Unable to stop the Job Scheduler.')
+            LOGGER.exception('Unable to stop the Job Scheduler.')
         else:
-            stderr('Job Scheduler stopped.')
+            LOGGER.info('Job Scheduler stopped.')
 
         self.scheduler.clear_jobs()
 
         # Shutdown plugins
-        stderr(
-            'Calling shutdown for %d modules.' % (len(self.shutdown_methods),)
-        )
+        LOGGER.info(
+            'Calling shutdown for %d modules.', len(self.shutdown_methods))
+
         for shutdown_method in self.shutdown_methods:
             try:
-                stderr(
-                    "calling %s.%s" % (
-                        shutdown_method.__module__, shutdown_method.__name__,
-                    )
-                )
+                LOGGER.debug(
+                    'Calling %s.%s',
+                    shutdown_method.__module__,
+                    shutdown_method.__name__)
                 shutdown_method(self)
             except Exception as e:
-                stderr(
-                    "Error calling shutdown method for module %s:%s" % (
-                        shutdown_method.__module__, e
-                    )
-                )
+                LOGGER.exception('Error calling shutdown method: %s', e)
+
         # Avoid calling shutdown methods if we already have.
         self.shutdown_methods = []
 
--- a/sopel/bot.py
+++ b/sopel/bot.py
@@ -10,6 +10,7 @@ from __future__ import unicode_literals, absolute_import, print_function, divisi
 from ast import literal_eval
 import collections
 import itertools
+import logging
 import os
 import re
 import sys
@@ -27,7 +28,7 @@ import sopel.loader
 
 __all__ = ['Sopel', 'SopelWrapper']
 
-LOGGER = logger.get_logger(__name__)
+LOGGER = logging.getLogger(__name__)
 
 if sys.version_info.major >= 3:
     unicode = str
--- a/sopel/bot.py
+++ b/sopel/bot.py
@@ -272,7 +272,7 @@ class Sopel(irc.Bot):
                 load_error,
                 load_disabled)
         else:
-            LOGGER.warning("Warning: Couldn't load any plugin")
+            LOGGER.warning("Warning: Couldn't load any plugins")
 
     def reload_plugin(self, name):
         """Reload a plugin
@@ -399,7 +399,7 @@ class Sopel(irc.Bot):
 
         It is possible to have a callable with rules, commands, and nick
         commands configured. It should not be possible to have a callable with
-        commands or nick commands but without rules. Callable without rules
+        commands or nick commands but without rules. Callables without rules
         are usually event handlers.
         """
         # Append module's shutdown function to the bot's list of functions to
--- a/sopel/bot.py
+++ b/sopel/bot.py
@@ -262,17 +262,17 @@ class Sopel(irc.Bot):
                     LOGGER.exception('Error in %s setup: %s', name, e)
                 else:
                     load_success = load_success + 1
-                    LOGGER.debug('Plugin loaded: %s', name)
+                    LOGGER.info('Plugin loaded: %s', name)
 
         total = sum([load_success, load_error, load_disabled])
         if total and load_success:
             LOGGER.info(
-                'Registered %d modules, %d failed, %d disabled',
+                'Registered %d plugins, %d failed, %d disabled',
                 (load_success - 1),
                 load_error,
                 load_disabled)
         else:
-            LOGGER.warning("Warning: Couldn't load any modules")
+            LOGGER.warning("Warning: Couldn't load any plugin")
 
     def reload_plugin(self, name):
         """Reload a plugin
@@ -329,25 +329,21 @@ class Sopel(irc.Bot):
         if not self.has_plugin(name):
             raise plugins.exceptions.PluginNotRegistered(name)
 
-        try:
-            # remove commands, jobs, and shutdown functions
-            for func in itertools.chain(callables, jobs, shutdowns):
-                self.unregister(func)
+        # remove commands, jobs, and shutdown functions
+        for func in itertools.chain(callables, jobs, shutdowns):
+            self.unregister(func)
 
-            # remove URL callback handlers
-            if "url_callbacks" in self.memory:
-                for func in urls:
-                    regexes = func.url_regex
-                    for regex in regexes:
-                        if func == self.memory['url_callbacks'].get(regex):
-                            self.unregister_url_callback(regex)
-                            LOGGER.debug('URL Callback unregistered %r', regex)
-        except:  # noqa
-            raise
-        else:
-            # remove plugin from registry
-            del self._plugins[name]
-            LOGGER.info('Plugin removed: %s', name)
+        # remove URL callback handlers
+        if "url_callbacks" in self.memory:
+            for func in urls:
+                regexes = func.url_regex
+                for regex in regexes:
+                    if func == self.memory['url_callbacks'].get(regex):
+                        self.unregister_url_callback(regex)
+                        LOGGER.debug('URL Callback unregistered: %r', regex)
+
+        # remove plugin from registry
+        del self._plugins[name]
 
     def has_plugin(self, name):
         """Tell if the bot has registered this plugin by its name"""
@@ -362,21 +358,27 @@ class Sopel(irc.Bot):
         if not callable(obj):
             LOGGER.warning('Cannot unregister obj %r: not a callable', obj)
             return
+        callable_name = getattr(obj, "__name__", 'UNKNOWN')
+
         if hasattr(obj, 'rule'):  # commands and intents have it added
             for rule in obj.rule:
                 callb_list = self._callables[obj.priority][rule]
                 if obj in callb_list:
                     callb_list.remove(obj)
+            LOGGER.debug(
+                'Rule callable "%s" unregistered',
+                callable_name,
+                rule.pattern)
+
         if hasattr(obj, 'interval'):
             self.scheduler.remove_callable_job(obj)
-        if (
-                getattr(obj, "__name__", None) == "shutdown" and
-                obj in self.shutdown_methods
-        ):
+            LOGGER.debug('Job callable removed: %s', callable_name)
+
+        if callable_name == "shutdown" and obj in self.shutdown_methods:
             self.shutdown_methods.remove(obj)
 
     def register(self, callables, jobs, shutdowns, urls):
-        """Register a callable.
+        """Register rules, jobs, shutdown methods, and URL callbacks.
 
         :param callables: an iterable of callables to register
         :type callables: :term:`iterable`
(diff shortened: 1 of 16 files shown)


Exirel:
@dgw once we agree on these changes, I'll update the documentation accordingly. I think this deserve some explanation for end-users, and for plugin developers.

dgw:
Hmmmm. This is a massive rewrite just to tweak some log levels for #1337. 😛 Can you implement 01feeae in a separate PR, and we can, ahem, *bikeshed* this overhaul without blocking that?

I have some design questions. Like, why remove `sopel.logger.get_logger` instead of fixing it? Of course I understand that this function duplicates stuff the same as `sopel.web.get`/`.head`/`.post` did (as they were literally just wrappers around the equivalent functions from `requests`), but at the same time I'd really like to have `sopel.logger.get_logger()` _just work_ without worrying about the argument at all. That is, the dream would be to have it automatically figure out the correct `name` based on whether it's invoked from a core file, core plugin, or custom/external package/plugin—but that might be _too_ magic. Convince me why that dream actually would be a nightmare and I'll wake up. 😁

Rotating log files is a nice idea, but I don't see where old files will be deleted. Of course, Sopel's current logging setup doesn't prune itself either, but… that's something I should have opened an issue for long ago. It's bad to have logs stick around forever.

My first look at this was split over multiple sessions, and I hope I didn't forget any big questions. Nitpicks were deliberately excluded, though. Don't worry, the usual "dgw edits Exirel's English" will happen later! I just didn't want either of us to spend time on it just yet.

Exirel:
### A Nice Conversation Starter

> Can you implement [01feeae](https://github.com/sopel-irc/sopel/commit/01feeaec6b1bfe0149c8d020f33ea18fc57aa969) in a separate PR

I could, but I think I wouldn't like it - neither would you. I've two main concerns about the current implementation: the force redirect output, and the manual writing in a file.

First, **the force redirect output**. This is a **bad** developer solution to a **100% ops problem**. As a result, properly using logs become a nightmare, because you would think that all we do is log to the console... but that's not true at all! Beside, the `stdio.log` is a mess, because it will log both `sopel start` and any kind of print that would happen in a non-"run" command. This is bonkers.

The proper way to solve that, is to use logging where it's relevant, and to use the good old fashioned `print` elsewhere. On a surface level, it looks like I changed a lot of code, but deep down, I've removed a lot of ad-hoc code by battle tested, reusable code from Python built-in.

Second, **manually writing into log files**. I think it was produced by the best intention in the world. Here, make sure we log everything critical. Also the folder creation might fail, that's only fair that we protect us from that! Sure thing... yet logging already does that, in a much better and efficient way than we do.

Except that it won't work properly if you redirect the output, and of course there is always the possibility to mess-up the configuration. But again, it looks like a huge risky change, yet it's replaced with fewer lines of code (i.e. easier to read), using battle tested Python built-in.

There are lot I don't like with the logging system in Python (I even made a talk at a Python French Conference about how ugly its documentation is), yet it's a standard, and it does its job wonderfully: it's steady, functional, and quite simple to reuse all over the place (in core, plugins, etc.).

As a result, changing the log level wouldn't do us any good, but create lots of unintended side effects that would inevitably lead us to... replace everything with logging.

### I Have Some Design Questions.

#### `sopel.logger.get_logger`

Quick note: the docstring of `get_logger` is full of lie and I hate it. Using this inside, say, the `remind` core module:

``` python
logger = get_logger(__name__)
```

Actually means that the logger will be `sopel.modules.sopel.modules.remind`. Which is bonkers. And for that, I hate you, `get_logger`.

> why remove `sopel.logger.get_logger` instead of fixing it?

My main concern is that we try to outsmart and hide a system that is otherwise pretty common and industry standard in Python application (also it's buggy).

The real fix is to allow developers to use `logging.getLogger` properly. For that, we need to create a context that allows them to do so freely and safely (no force redirect output, proper use of logging everywhere, good documentation and guidance, etc.).

> That is, the dream would be to have it automatically figure out the correct `name` based on whether it's invoked from a core file, core plugin, or custom/external package/plugin—but that might be _too_ magic.

You said it: too magic. It means looking at the stack trace, figuring out where we are. And I really don't like that approach. There is enough magic in a `logging.getLogger(__name__)`, while keeping the "explicit is better than implicit" zen of Python.

The best part would be to document our logging handlers, so that plugin developers can do that in their own code:

``` python
# file: sopel_custom/custom.py
import logging
from logging.config import dictConfig

LOGGER = logging.getLogger(__name__)  # logger=sopel_custom.custom

def setup(bot):
    dictConfig({
        'version': 1,
        'disable_existing_loggers': False,
        'loggers': {
            'sopel_custom': {
                'level': 'WARNING',
                'handlers': ['console', 'logfile'],
            }
        }
    })
```

#### Rotating log files is a nice idea

On that one, I had mixed feeling at first, just like you:

> I don't see where old files will be deleted.

Until I had the exact same realization as you:

> Of course, Sopel's current logging setup doesn't prune itself either

Yeah, Sopel does nothing to its log at the moment. It's solution is to: just open the file every single time something is output on stdout/stderr. If the file doesn't exist anymore, it's directly re-created. It makes log rotating easy, but that's also a good context for race condition and I/O congestion.

In the "rotating file" scenario, Sopel would write a log file every single day. To delete them, you can use any kind of cron tasks to remove files older than X days/weeks/months.

### Design Conclusion From A Totally Objective Person (me)

As of today, we don't let the user decided what they will get with logging. I think this PR shows that we can give them more control with a system that is already existing, battle tested code, and that is closer to industry standards.

Let people use `logging` is what I'm trying to say. :muscle: 

### "dgw edits Exirel's English"

:fr: :fr: :fr: :fr: :fr: :fr:

sopel/bot.py
```diff
@@ -211,14 +211,35 @@ def setup(self):
         self.scheduler.start()
 
     def setup_logging(self):
-        logger.setup_logging(self)
+        logger.setup_logging(self.config)
```
Exirel: @dgw the more I think of this... the more I want to move it *out* of the bot.

It's more of a "client" configuration, i.e. the tool used to launch the bot.

Do you mind if I try to put logging setup into `sopel.cli.utils`, and then I'd call it from every single command line tool?

dgw: I think Sopel should always have logs. A lot of this rework makes sense to me, but not completely decoupling the Bot object from logs.

dgw: If we decouple logging, why not decouple plugins? Why not make the launch tool set up job scheduling itself? You see how we get into the weeds just by removing one piece of the bot's startup?

Exirel: Yeah, the bot always *send* logs. But does it have to configure it itself? It's more a "load the settings" problem rather than a "load the bot itself".

And yes, I'd like very much to be able to load and pre-setup plugins without the bot, for many good reasons. :grin:

Exirel: Note that I'd keep the "add the channel logger" in the bot's setup. Because that is 100% bot-specific.

sopel/bot.py
```diff
@@ -326,12 +340,13 @@ def remove_plugin(self, plugin, callables, jobs, shutdowns, urls):
                     regex = func.url_regex
                     if func == self.memory['url_callbacks'].get(regex):
                         self.unregister_url_callback(regex)
+                        LOGGER.debug('URL Callback unregistered %r', regex)
```
```suggestion
                        LOGGER.debug('URL Callback unregistered: %r', regex)
```

Also, it's outside the range of lines that GitHub will let me add notes on, but the `self.unregister(*args)` loop above would be another good place for debug output.

sopel/bot.py
```diff
@@ -326,12 +340,13 @@ def remove_plugin(self, plugin, callables, jobs, shutdowns, urls):
...
         except:  # noqa
-            # TODO: consider logging?
-            raise  # re-raised
+            raise
         else:
             # remove plugin from registry
             del self._plugins[name]
+            LOGGER.info('Plugin removed %s', name)
```
```suggestion
            LOGGER.info('Plugin removed: %s', name)
```

sopel/bot.py
```diff
@@ -344,6 +359,7 @@ def unregister(self, obj):
         :type obj: :term:`object`
         """
         if not callable(obj):
+            LOGGER.warning('Cannot unregister obj %r: not a callabled', obj)
```
```suggestion
            LOGGER.warning('Cannot unregister obj %r: not a callable', obj)
```

sopel/bot.py
```diff
@@ -326,12 +340,13 @@ def remove_plugin(self, plugin, callables, jobs, shutdowns, urls):
                     regex = func.url_regex
                     if func == self.memory['url_callbacks'].get(regex):
                         self.unregister_url_callback(regex)
+                        LOGGER.debug('URL Callback unregistered %r', regex)
         except:  # noqa
-            # TODO: consider logging?
-            raise  # re-raised
+            raise
```
dgw: Why remove this comment?

Also, why not log the exception that happened? It could be useful information if the plugin machinery ever catastrophically fails in the wild.

sopel/cli/run.py
```diff
@@ -73,20 +75,26 @@ def signal_handler(sig, frame):
             if hasattr(signal, 'SIGILL'):
                 signal.signal(signal.SIGILL, signal_handler)
             p.setup()
+        except KeyboardInterrupt:
+            break
+        except Exception:
+            # In that case, there is nothing we can do.
+            # If the bot can't setup itself, then it won't run.
+            # This is a critical case scenario, where the user should have
+            # directly access to the exception traceback right in the console
```
dgw:
```suggestion
            # direct access to the exception traceback right in the console.
```

sopel/cli/run.py
```diff
@@ -73,20 +75,26 @@ def signal_handler(sig, frame):
...
+            # Beside, we can't know if logging has been set-up or not, so
```
dgw:
```suggestion
            # Besides, we can't know if logging has been set up or not, so
```

sopel/cli/utils.py
```diff
@@ -335,8 +335,15 @@ def redirect_outputs(settings, is_quiet=False):
     :param bool is_quiet: Optional, set to True to make Sopel's outputs quiet
 
     Both ``sys.stderr`` and ``sys.stdout`` are redirected to a logfile.
+
+    .. deprecated:: 7.0
+
+        Sopel now use the built-in logging system for its output, and this
```
dgw:
```suggestion
        Sopel now uses the built-in logging system for its output, and this
```

sopel/cli/utils.py
```diff
@@ -327,6 +326,7 @@ def load_settings(options):
     return config.Config(filename)
 
 
+@tools.deprecated(version='7.0', removed_in='8.0')
```
dgw: Could put a `reason` here, even as simple as "Obsoleted by modernized logging." or something like that. Possibly unnecessary, though, since the Sphinx docs will actually show the details. Your call!

sopel/tools/__init__.py
```diff
@@ -643,3 +628,25 @@ def contains(self, key):
...
+def get_raising_file_and_line(tb=None):
+    """Get the file and line number where an exception happened.
+
+    :param tb: the traceback (uses the most recent exception if not given)
+    :return: a tuple of the filename and line number
+    :rtype: (str, int)
+
+    .. deprecated:: 7.0
+
+        Use Python built-in logging system, with the ``logger.exception``
```
dgw:
```suggestion
        Use Python's built-in logging system, with the ``logger.exception``
```

sopel/tools/__init__.py
```diff
@@ -55,21 +55,6 @@ def get_input(prompt):
         return raw_input(prompt).decode('utf8')
 
 
-def get_raising_file_and_line(tb=None):
```
dgw: Why is this moved to the end of the file?

sopel/logger.py
```diff
@@ -30,39 +34,109 @@ def formatException(self, exc_info):
...
+                'when': 'midnight',
+                'formatter': 'raw',
+            },
+        },
+    }
+    dictConfig(logging_config)
 
 
+@deprecated(
+    'use logging.getLogger(__name__) instead',
```
dgw:
```suggestion
    'Use logging.getLogger(__name__) instead',
```

sopel/logger.py
```diff
@@ -30,39 +34,109 @@ def formatException(self, exc_info):
...
-    `get_logger(__name__)` to get a logger."""
+    This is equivalent to ``logging.getLogger('sopel.modules.' + name)`` when
+    name is given, and ``logging.getLogger('sopel')`` when it is not.
+    The latter case is intended for use in Sopel's core; modules should call
+    ``get_logger(__name__)`` to get a logger.
+
+    .. deprecated:: 7.0
+
+        Use ``logging.getLogger(__name__)`` instead, or
+        ``logging.getLogger('sopel.plugins.YOURPLUGINNAME')``.
```
dgw: What are the possible consequences if we recommend simply calling `logging.getLogger('YOURPLUGINNAME')`?

We probably don't want to encourage plugin developers to insert their log entries into an existing namespace. (This also ties into that IRC discussion we started the other day about the mess we now have with `sopel.plugins` being used for the machinery when it should ideally be the new name for `sopel.modules`…)

I like that, even though it's being deprecated, you still made an effort to fix the docstring to render inline-code properly in Sphinx. 😁

sopel/irc.py
```diff
@@ -189,11 +173,11 @@ def run(self, host, port=6667):
         try:
             self.initiate_connect(host, port)
         except socket.error as e:
-            stderr('Connection error: %s' % e)
+            LOGGER.exception('Connection error: %s', e)
             self.handle_close()
 
     def initiate_connect(self, host, port):
-        stderr('Connecting to %s:%s...' % (host, port))
+        LOGGER.info('Connecting to: %s:%s', host, port)
```
dgw: I think this was more correct without the added `:`. I also rather like the trailing `...` because registering an IRC connection is a process that may take some time, and this log line is only the start of it.

```suggestion
        LOGGER.info('Connecting to %s:%s...', host, port)
```

sopel/irc.py
```diff
@@ -202,13 +186,14 @@ def initiate_connect(self, host, port):
             self.send = self._ssl_send
             self.recv = self._ssl_recv
         elif not has_ssl and self.config.core.use_ssl:
-            stderr('SSL is not avilable on your system, attempting connection '
-                   'without it')
+            LOGGER.warning(
+                'SSL is not avilable on your system, '
```
dgw: Let's fix this ancient grammatical mess as long as it's already getting touched.

```suggestion
                'SSL is not available on your system; '
```

sopel/irc.py
```diff
@@ -300,7 +286,7 @@ def handle_connect(self):
         self.write(('USER', self.user, '+iw', self.nick), self.name)
 
         # maintain connection
-        stderr('Connected.')
+        LOGGER.info('Connected.')
```
dgw: This is why the `...` in the "Connecting" log entry above is appropriate. This is the end of the possibly-lengthy process started there.

sopel/irc.py
```diff
@@ -332,7 +318,10 @@ def _get_cnames(self, domain):
     def _timeout_check(self):
         while self.connected or self.connecting:
             if (datetime.now() - self.last_ping_time).seconds > int(self.config.core.timeout):
-                stderr('Ping timeout reached after %s seconds, closing connection' % self.config.core.timeout)
+                LOGGER.warning(
+                    'Ping timeout reached after %s seconds, '
```
dgw: Another old grammatical mess, but less bad (because it isn't accompanied by a spelling error).

```suggestion
                    'Ping timeout reached after %s seconds; '
```

sopel/irc.py
```diff
@@ -419,83 +408,54 @@ def found_terminator(self):
...
+
+        LOGGER.exception(message)
+
+        if trigger and self.config.core.reply_errors and trigger.sender is not None:
+            self.say(message, trigger.sender)
 
     def handle_error(self):
         """Handle any uncaptured error in the core.
 
         Overrides asyncore's handle_error.
```
dgw:
```suggestion
```

sopel/irc.py
```diff
@@ -419,83 +408,54 @@ def found_terminator(self):
...
```
dgw: Links to Python docs for `asyncore` and `asyncore.dispatcher.handle_error` would be sugar on top of this. 😁 (Starting point: https://docs.python.org/3/library/asyncore.html#asyncore.dispatcher.handle_error)

sopel/bot.py
Exirel: Oh, that's a dumb typo of mine.

sopel/logger.py
Exirel: Say that `__name__` is `sopel_modules.myplugin`, then it won't pollute the namespace. Same for `sopel_myplugin`.

I'm still thinking about tools to make plugin logging configuration easier, though.

sopel/tools/__init__.py
Exirel: Because it's deprecated and I wanted everything deprecated at the bottom of the file, together. I can totally revert that.

dgw: It's mostly unnecessary diff noise, but it also affects blame output. Neither of which is a major concern for something deprecated, but still. I don't think moving deprecated code is a habit we should start. The decorator is enough to find it all. :)

Exirel: Fair enough, will revert then.

sopel/bot.py
Exirel: If the plugin machinery catastrophically fails in the wild, then a higher level of the bot will handle the error and log it anyway.

By the way, this try/except/else doesn't have a purpose anymore. We could just let the code fail without enclosing it into that, since we don't do anything in the `except`, and we don't rely on a `finally` clause.

sopel/tools/__init__.py
Exirel: Oh. Except that... I can't: the `deprecated` function is defined later in the file, so `get_raising_file_and_line` **must** be declared after.

Moving `deprecated` is even bigger. I'm not sure this is worth it either. :-/

sopel/logger.py
Exirel: Also I realized that I misunderstood your comment.

The consequence of `logging.getLogger('YOURPLUGINNAME')`  is that, as it's not in the same namespace as ` sopel.*`, it won't use the sopel's logging handler, so it might prevent it from logging into the console/the log file.

For that, a plugin's `setup` function will have to configure manually a log handler for the plugin. This is not ideal, and I need to craft some helper for that.

sopel/bot.py
dgw: Hmm, might as well remove it then.

sopel/bot.py
dgw: Looking at the current patch, this colon is still missing somehow.

sopel/tools/__init__.py
dgw: Well, then it's not "unnecessary diff noise". 😁

Python needs forward declarations. 😛

sopel/logger.py
dgw: I'd guess that was ultimately the idea behind Sopel's `get_logger` function, even if the implementation was a mess: Give plugins a way to get a logger that always outputs to where the bot's logs should go, regardless of package naming, because the function handled figuring out the namespace.

I hope you're still thinking to keep plugin logs in the same place(s) as Sopel's main logs? Otherwise the log directory becomes pretty messy if there are potentially up to two files for _each plugin_ in addition to Sopel itself.

sopel/bot.py
Exirel: Done.

Exirel:
The live test for Google Translate is broken. Of course it is.

sopel/logger.py
Exirel: I just realized two things:

* `sopel.logger`  is undocumented, which means that `sopel.logger.get_logger` is undocumented; we might consider changing its behavior without a warning
* we could log everything into one logging file, AND into a dedicated file for every plugins, i.e. log twice everything that come from a plugin in, say, `default.plugins.log`

My idea with `get_logger` would be to create a new function `get_plugin_logger`, that would return a logger for `sopel.externals.<plugin-name>`, and `get_logger` would call that function. Internally, we can configure Sopel logging with:

* log everything under `sopel.*` into `default.sopel.log`
* log everything under `sopel.externals.*`  into `default.plugins.log`  AND `default.sopel.log`

And even better than that: we can have a ListAttribute to deactivate logging for a list of plugins, something like `logging_plugin_disable`, and all I'd have to do is to configure `sopel.externals.<plugin_name>` with "don't log anything".

That could be cool.

Another option would be to allow a sub-section of the configuration for logging, where each key is a plugin-name, and each value is a log level, something likes this:

```
[core]
# ...
[core:logging]
plugin_name_1 = INFO
plugin_name_2 = DEBUG
plugin_name_3 = ERROR
```

sopel/bot.py
```diff
@@ -386,32 +388,84 @@ def register(self, callables, jobs, shutdowns, urls):
         :type shutdowns: :term:`iterable`
         :param urls: an iterable of functions to call when matched against a URL
         :type urls: :term:`iterable`
+
+        The ``callables`` argument contains a list of "callable objects", i.e.
+        object for witch :func:`callable` will return ``True``. They can be:
```
dgw: "objects for which" :)

sopel/bot.py
```diff
@@ -386,32 +388,84 @@ def register(self, callables, jobs, shutdowns, urls):
...
         self.shutdown_methods += shutdowns
+        match_any = re.compile('.*')
         for callbl in callables:
-            if hasattr(callbl, 'rule'):
-                for rule in callbl.rule:
+            callable_name = getattr(callbl, "__name__", 'UNKNOWN')
+            rules = getattr(callbl, 'rule', [])
+            commands = getattr(callbl, 'commands', [])
+            nick_commands = getattr(callbl, 'nickname_commands', [])
+            is_rule_only = rules and not commands and not nick_commands
```
dgw: As far as this logic goes, what about callables with `urls` (or `url_regex`, from the decorator)?

I'm glancing at this quickly while waiting for Travis to make sure my fix for the broken `translate` test comes through into this PR, so maybe I'm missing something.

Exirel: They are covered line 461 with the `for func in urls:`: URL Callbacks are not in the "callables" list. Same for jobs and shutdown functions, which are in separated lists.

dgw: I knew I had to be missing something in my rush. :)

dgw:
From an unrelated IRC question, I found https://github.com/sopel-irc/sopel/blob/c752b10707e5ffcecc8814e315f31a8271937e76/sopel/coretasks.py#L218 which should be a debug message.

Can you change that in this PR so we don't need to open a separate one? 😸

sopel/bot.py
```diff
@@ -230,35 +251,28 @@ def setup_plugins(self):
...
-            stderr('%d modules failed to load' % load_error)
-            stderr('%d modules disabled' % load_disabled)
+            LOGGER.info(
+                'Registered %d plugins, %d failed, %d disabled',
+                (load_success - 1),
+                load_error,
+                load_disabled)
         else:
-            stderr("Warning: Couldn't load any modules")
+            LOGGER.warning("Warning: Couldn't load any plugin")
```
```suggestion
            LOGGER.warning("Warning: Couldn't load any plugins")
```

sopel/logger.py
```diff
@@ -30,40 +34,134 @@ def formatException(self, exc_info):
...
+        LOGGER = logger.get_plugin_logger('my_custom_plugin')
+
+    is equivalent to this::
+
+        import logging
+        LOGGER = logging.getLogger('sopel.externals.my_custom_plugin')
+
+    Internally, Sopel configures logging for the ``sopel`` namespace, so
+    external plugins can't benefit from it with ``logging.getLogger(__name__)``
+    as they won't be in the same namespace. This function use the
```
dgw:
```suggestion
    as they won't be in the same namespace. This function uses the
```

sopel/bot.py
```diff
@@ -370,32 +388,98 @@ def register(self, callables, jobs, shutdowns, urls):
...
+        objects for which :func:`callable` will return ``True``. They can be:
+
+        * a callable with rules (will match triggers with a regex pattern)
+        * a callable without rules (will match any triggers, such as events)
+        * a callable with commands
+        * a callable with nick commands
+
+        It is possible to have a callable with rules, commands, and nick
+        commands configured. It should not be possible to have a callable with
+        commands or nick commands but without rules. Callable without rules
```
```suggestion
        commands or nick commands but without rules. Callables without rules
```

Exirel:
> One part of me wants to keep the simple name `get_logger` going forward

My main concern is that its docstring promoted `get_logger(__name__)` which is very bad because the end result was `sopel.modules.sopel_modules.my_module` when called from the `sopel_modules.my_module` custom plugin. But I totally get what you say. What I want to do is to:

* keep `get_logger` where it is for now
* change it so it can detect if the argument is a fully qualified name or not
* if so, use `logging.getLogger()`
* otherwise, use `get_plugin_logger` (or something similar)

> Another part of me thinks that `get_plugin_logger` belongs somewhere in `sopel.tools`

What about `sopel.plugins.get_logger`? That could make sense. It would allow to slowly deprecate `sopel.logger.get_logger` (doc in 7.0, warning in 8.0, removed in 9.0), while having a "plugin" specific version that we could promote and work with.

Exirel:
@dgw I made the changes I suggested above. I think it's a sane approach, and I'm glad you made me reconsider this. :smiley: 

With that, internal plugins will have to use:

``` python
# in sopel/modules/clock.py
import logging
```

And it'll be under the `sopel` namespace, which is perfect.

For an external plugin, such as `sopel_modules.youtube`  or any new fancy entry-point based plugins for Sopel 7.x:

``` python
# in modsopelunker.py
from sopel import plugins
LOGGER = plugins.get_logger('sopelunker')
```

Which will log everything with the `sopel.externals.sopelunker` logger, which is under the right logging namespace. So that's perfect too!

sopel/logger.py
```diff
@@ -119,49 +119,23 @@ def setup_logging(settings):
...
 
     """
-    if name:
-        return logging.getLogger('sopel.modules.' + name)
-    else:
+    if not name:
         return logging.getLogger('sopel')
 
+    parts = name.strip().split('.')
+    if len(parts) > 1 or parts[0] in ['sopel', 'sopel_modules']:
```
Exirel: Note that this will catch either `get_logger('my.custom.pymod')` as `get_logger('sopel')` and `get_logger('sopel_modules')`; but **not** `get_logger('sopelunker')`. Which is exactly what we want!

dgw:
We had a very productive discussion on IRC today that touched on this. Key points were:

  1. `get_logger()` probably belongs in `tools`, long-term
      - Putting it in `plugins` encourages plugin authors to import from a package that otherwise provides only internal/loader utilities, whereas `tools` already contains a bunch of things that are (mostly) useful for plugin code.
        - Yes, there is stuff in `tools` that is only useful for core code. That's a separate issue that should also be addressed at some point.
  2. The log namespace for external plugins may change to `sopel.external`, either preemptively in this PR or in a future patch, to match a proposed move of core plugins from `sopel.modules` to `sopel.internal` (or `sopel.core`) and sound more grammatically correct

I've also just thought of a couple other considerations:

  1. If we follow our current plans, there will be no "internal" modules at some future point (in 8.0 or 9.0 at the latest). At that point, having logs labeled `sopel.external.*` won't make as much sense.
  2. While we _do_ still have "internal"/"core" plugins, having "external" hard-coded in `get_logger()`'s generated namespace means our core plugins can't use the same logging tools as third-party plugins without log lines becoming misleading.

Thing is, these last couple things don't have to be solved in this PR. We should just keep them in mind while rearranging the plugin/module stuff as discussed today, and not be afraid to make further tweaks to things when it makes sense.

Exirel:
@dgw I moved `get_logger`  into `sopel.tools`. For everything else, I think the way forward is "another PR".

sopel/logger.py
```diff
@@ -30,40 +34,108 @@ def formatException(self, exc_info):
...
-    `get_logger(__name__)` to get a logger."""
-    if name:
-        return logging.getLogger('sopel.modules.' + name)
-    else:
+    .. deprecated:: 7.0
+
+        Use ``logging.getLogger(__name__)`` in Sopel's code instead, and
+        :func:`sopel.tools.get_logger` for external plugins.
+
+        This will warn a deprecation warning in Sopel 8.0 then removed in 9.0.
```
dgw: Why not add `@deprecated` (the warning) now, and remove it in 8.0? Use this suggested change if appropriate:

```suggestion
        Will be removed in Sopel 8.0.
```

sopel/irc.py
```diff
@@ -419,83 +408,55 @@ def found_terminator(self):
...
         """
-        trace = traceback.format_exc()
-        stderr(trace)
-        LOGGER.error('Fatal error in core, please review exception log')
-        # TODO: make not hardcoded
-        logfile = codecs.open(
-            os.path.join(self.config.core.logdir, self.config.basename + '.exceptions.log'),
-            'a',
-            encoding='utf-8'
+        LOGGER.error('Fatal error in core, please review exceptions logs.')
```
dgw:
```suggestion
        LOGGER.error('Fatal error in core; please review exception log.')
```

May or may not be useful enough to put the actual log filename here, if we can get it.

sopel/irc.py
```diff
@@ -419,83 +408,55 @@ def found_terminator(self):
...
+
+        err_log = logging.getLogger('sopel.exceptions')
+        err_log.error(
+            'Fatal error in core, handle_error() was called.\n'
```
dgw:
```suggestion
            'Fatal error in core; handle_error() was called.\n'
```

sopel/irc.py
```diff
@@ -419,83 +408,55 @@ def found_terminator(self):
...
-        logfile.write('----------------------------------------\n\n')
-        logfile.close()
+        err_log.exception('Fatal error traceback')
+        err_log.error('----------------------------------------')
+
         if self.error_count > 10:
+            # quit if too many errors
             if (datetime.now() - self.last_error_timestamp).seconds < 5:
-                stderr("Too many errors, can't continue")
+                LOGGER.error('Too many errors, can\'t continue')
```
dgw:
```suggestion
                LOGGER.error('Too many errors; can\'t continue')
```

sopel/cli/run.py
```diff
@@ -49,7 +51,7 @@ def run(settings, pid_file, daemon=False):
 
     def signal_handler(sig, frame):
         if sig == signal.SIGUSR1 or sig == signal.SIGTERM or sig == signal.SIGINT:
-            tools.stderr('Got quit signal, shutting down.')
+            LOGGER.warning('Got quit signal, shutting down.')
```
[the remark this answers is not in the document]
Exirel: In my mind, a quit signal is out of the ordinary for the bot, which is supposed to spend its life running. I'd consider any signal as a warning by the way. (quit, sighup, sigusr1, sigusr2, etc.)

sopel/cli/run.py
```diff
@@ -49,7 +51,7 @@ def run(settings, pid_file, daemon=False):
...
             p.quit('Closing')
         elif sig == signal.SIGUSR2 or sig == signal.SIGILL:
             tools.stderr('Got restart signal.')
```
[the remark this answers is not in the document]
Exirel: I think I missed that one!

sopel/logger.py
Exirel: Because I assume a lot of external plugins are using it right now, so it'd raise a *lot* of warnings. I'm not sure it's a good idea, but I don't have a strong opinion.

If that was me, deprecation warning would be all over the place... :stuck_out_tongue:

sopel/irc.py
Exirel: Hm. That's an idea. I'd suggest to improve that in another PR.

sopel/irc.py
Exirel: I edited your review comment (:open_mouth:) to replace the single quotes by double quotes.

sopel/logger.py
dgw: Still gonna mention it specifically in the migration guide for 7, then.

Exirel:
:tada:

PR #1679, opened 2019-08-12
dice, choose: extract `.choose` command into own module

dgw:
Could have moved it to `rand.py`, but it doesn't *really* fit in any better there. There's very little cost to having a standalone file for this command aside from possibly breaking exclude settings in existing Sopel installations (and we're already adding other new modules in 7.0).

Made sure this new file credits all the right people. Namely, I added Exirel to the header, because he did a significant chunk of work on this function last year.

### DIFF
--- a/sopel/modules/choose.py
+++ b/sopel/modules/choose.py
@@ -1,0 +1,54 @@
+# coding=utf-8
+"""
+choose.py - Sopel Choice Module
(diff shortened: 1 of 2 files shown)


Exirel:
Oh yeah I remember that one. It's a good idea!

PR #1669, opened 2019-07-26
Document `tools.web` package

dgw:
Draft because it's based on the #1616 branch, which isn't yet merged.

See 6d53ced for the relevant changes. I'll keep an eye on this after merging its parent.

Issue #1346, opened 2018-06-23
Give web.quote an inverse function


dgw:
Whatever happens with #1318, assuming the `web.quote` functionality is kept around somewhere, it really should have a sibling that does the opposite. We could call it… `web.unquote`. Imaginative, no? :laughing:

This is _totally_ not inspired by the multiple Unicode bugs we currently have to work around (#860, #1345, etc.) because Python's own `unquote` function is inconsistent across versions… Not at all. /s

Not into putting API changes in the 6.x series at this point, so this likely won't actually happen until Sopel 7.

dgw:
Working on this. I'm even adding tests for `quote()` and `unquote()`.

PR #1681, opened 2019-08-18
tools.web: add `unquote()` function

dgw:
Also added tests for the old `quote()` function/shim and the new `unquote()` thing. They're even using `@pytest.mark.parametrize`! (And also using a cheeky `zip()` trick that @Exirel might think is "not dumb enough" for using in a test file. 🤐)

Aside from the obvious hope that anyone else still thinks this is a good idea (it's been around for a _while_ in #1346), I also hope that this merges cleanly with #1669, without any need to rebase crap later. It's not a big deal, but I do like the `gitk` graphs we get from merging un-rebased PRs that were developed in staggered parallel and not necessarily merged in order. 😼

### DIFF
--- a/sopel/modules/calc.py
+++ b/sopel/modules/calc.py
@@ -14,14 +14,7 @@ from requests import get
 
 from sopel.module import commands, example
 from sopel.tools.calculation import eval_equation
-
-if sys.version_info.major < 3:
-    from urllib import quote as _quote
-
-    def quote(s):
-        return _quote(s.encode('utf-8')).decode('utf-8')
-else:
-    from urllib.parse import quote
+from sopel.tools.web import quote
 
 if sys.version_info.major >= 3:
     unichr = chr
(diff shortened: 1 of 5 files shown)


dgw:
> The tests asserts that anything non-ASCII (plus `,` and `_`) is properly quoted/unquoted. But what about `+/?%`, and a space? As a web developer, I know these can be a very bad surprise when nobody checks for them.

Added cases for these. `quote()` leaves `/` alone by default, so the test case for `/` reflects that. Could certainly pass `safe=None` to the test call, though, if you think it would be better. Or add a separate test for the `safe` param… So many options!

> as `quote` and `unquote` may or will appear in the documentation, I wonder if we should not add a `.. note:: this is python 2 and 3 compatibility`?

I put a note on `unquote()` in this PR. Once we bikeshed it (or like it without bikeshedding), I'll port the same thing over to `quote()` in #1669.

Exirel:
> I put a note on `unquote()` in this PR. Once we bikeshed it (or like it without bikeshedding), I'll port the same thing over to `quote()` in #1669.

Looks like 1669 didn't introduce any merge conflict _and_ you can add that `note` block. 😀

Issue #1640, opened 2019-06-09
aspell problems [How to solve them]


ThijsEigenwijs:
Dear all,

If during upgrading you find yourself facing this issue:
`aspell.c:53:10: fatal error: aspell.h: No such file or directory
     #include <aspell.h>
              ^~~~~~~~~~
    compilation terminated.
    error: command 'x86_64-linux-gnu-gcc' failed with exit status 1

    ----------------------------------------
^CCommand "/usr/bin/python3 -u -c "import setuptools, tokenize;__file__='/tmp/pip-build-h511kldh/aspell-python-py3/setup.py';f=getattr(tokenize, 'open', open)(__file__);code=f.read().replace('\r\n', '\n');f.close();exec(compile(code, __file__, 'exec'))" install --record /tmp/pip-yq7v_yrd-record/install-record.txt --single-version-externally-managed --compile --user --prefix=" failed with error code 1 in /tmp/pip-build-h511kldh/aspell-python-py3/`

then this is for you!

The problem is that your linux machine is missing some libraries (sorry windows, dont know how to fix for you!)

If you execute this script

`sudo apt-get install libaspell-dev
pip3 install -r requirements.txt
python3 ./sopel.py`

It will install the needed libraries, install all the python things and start up sopel.

If you get an error at an other place search for this on google
"{module} not found python3 {distro}"

9/10 time google knows how to solve your issue!

PR #1684, opened 2019-08-24
ipython: remove plugin (moving to separate package)

dgw:
IPython as a dependency isn't as fiddly as aspell (see discussion in #1291, #1640, #1642), but it's still unnecessary install-bloat that applies only to a small portion of the user base. `ipython` and `spellcheck` should be the only exceptions to the package-ization plan outlined in #1291, which targets most removals for Sopel 8 (our _next_ next major release).

Removed plugin file, requirements, test exclusions (???), and contrib script for hiding warnings generated by IPython. (I don't know if that contrib script even still does anything.)

New package will live at [sopel-irc/sopel-ipython](https://github.com/sopel-irc/sopel-ipython), and will be published on PyPI as `sopel-ipython`. @sopel-irc/rockstars, I'd like your opinions on whether the new package should be added as a requirement for `sopel`, as dictated by the transition plan in #1291. I'm on the fence, but given its low utility and the ease of reinstalling the plugin if desired, I'm leaning toward not bothering.

Merging logging changes like #1678 may require applying the `ipython.py` changes from ~~30a597f01386453bfe7f830ef875e9c5601c3ebb~~ e18681b64f5ffa644a9cb1b69ad8a98350994994 (@Exirel's sometimes-unnecessary rebasing strikes again!) to the new package.


dgw:
The plugin will be installable shortly from https://pypi.org/project/sopel-ipython/ (or see [the repo](https://github.com/sopel-irc/sopel-ipython)).
I forgot to publish a package last night. :sweat_smile:

Note: As PyPI does not have any release of Sopel 7.x available yet, this requires an existing "editable" install of Sopel @ 40162a46ad20867a50162a25708248da3083d180 or newer. PyPI will not be able to satisfy the `sopel>=7.0` dependency on its own until we begin publishing alpha/beta/RC versions.

PR #1686, opened 2019-09-02
test: fix py2 inconsistency with regex ID

Exirel:
For some reason [assertions about compiled regex fail with Python 2](https://github.com/sopel-irc/sopel/commit/8eb80dd39fa703cc8d0379573aa1adc552a95fa0#commitcomment-34920303).

We don't know why, but at least now we make assertions about the patterns, and not the compiled regex objects. This should prevent the false-negative on Travis CI (I hope).

### DIFF
--- a/test/test_module.py
+++ b/test/test_module.py
@@ -2,8 +2,6 @@
 """Tests for message formatting"""
 from __future__ import unicode_literals, absolute_import, print_function, division
 
-import re
-
 import pytest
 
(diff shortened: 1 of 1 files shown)


Exirel:
Travis seems happy about these tests with Python 2.7.

dgw:
@Exirel Thanks for putting this together so quickly! 👌

PR #1688, opened 2019-09-07
config: do not serialize empty ListAttribute to blank line

dgw:
Tweaks the serialization behavior of empty `ListAttribute` to eliminate unnecessary whitespace in the saved config file after Sopel writes it out.

Before:

```
[section]
list_attribute = 
	
next_thing = foo bar baz
```

After:

```
[section]
list_attribute = 
next_thing = foo bar baz
```


dgw:
Not needed any more; this fix was included in #1690:

https://github.com/sopel-irc/sopel/blob/5755a11b8acc836553dd8017cd705ab4570a1668/sopel/config/types.py#L329-L331

PR #1687, opened 2019-09-07
config: No more '#' or ';' comments; use '//' instead

dgw:
Any alternative idea that would let us keep using `#` for comments is welcome, but as described in the relevant commit message I don't think that's possible without some very dirty monkey-patching.

'Documentation' label because I'll have to make _damn sure_ I include very prominent notes about this in the changelog and migration guide. If someone loads an old config with now-unrecognized comments, Sopel will explode with a very unhelpful traceback:

```
$ sopel
Traceback (most recent call last):
  File "/mnt/c/Users/dgw/github/sopel/sopel/config/types.py", line 55, in __init__
    getattr(self, value)
  File "/mnt/c/Users/dgw/github/sopel/sopel/config/types.py", line 384, in __get__
    this_section = getattr(main_config, instance._section_name)
  File "/mnt/c/Users/dgw/github/sopel/sopel/config/__init__.py", line 218, in __getattr__
    section = self.ConfigSection(name, items, self)  # Return a section
  File "/mnt/c/Users/dgw/github/sopel/sopel/config/__init__.py", line 187, in __init__
    value = item[1].strip()
AttributeError: 'NoneType' object has no attribute 'strip'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/local/bin/sopel", line 11, in <module>
    load_entry_point('sopel', 'console_scripts', 'sopel')()
  File "/mnt/c/Users/dgw/github/sopel/sopel/cli/run.py", line 680, in main
    return command(opts)
  File "/mnt/c/Users/dgw/github/sopel/sopel/cli/run.py", line 573, in command_legacy
    config_module = get_configuration(opts)
  File "/mnt/c/Users/dgw/github/sopel/sopel/cli/run.py", line 329, in get_configuration
    settings = utils.load_settings(options)
  File "/mnt/c/Users/dgw/github/sopel/sopel/cli/utils.py", line 338, in load_settings
    return config.Config(filename)
  File "/mnt/c/Users/dgw/github/sopel/sopel/config/__init__.py", line 119, in __init__
    validate=validate)
  File "/mnt/c/Users/dgw/github/sopel/sopel/config/__init__.py", line 173, in define_section
    setattr(self, name, cls_(self, name, validate=validate))
  File "/mnt/c/Users/dgw/github/sopel/sopel/config/types.py", line 65, in __init__
    value)
ValueError: Missing required value for core.ca_certs
```

[...]

Issue #864, opened 2015-08-18
Error on saftey module Configuration


Serubin:
Encountered this error on setup

```
Configure safety module (y/n)? [n]y
Encountered an error while writing the config file. This shouldn't happen. Check permissions.
Traceback (most recent call last):
  File "/usr/local/bin/sopel", line 11, in <module>
    sys.exit(main())
  File "/usr/local/lib/python3.4/dist-packages/sopel/run_script.py", line 133, in main
    _create_config(configpath)
  File "/usr/local/lib/python3.4/dist-packages/sopel/config/__init__.py", line 254, in _create_config
    config._modules()
  File "/usr/local/lib/python3.4/dist-packages/sopel/config/__init__.py", line 207, in _modules
    module.configure(self)
  File "/usr/local/lib/python3.4/dist-packages/sopel/modules/safety.py", line 49, in configure
    config.safety.configure_setting('enabled_by_default')
  File "/usr/local/lib/python3.4/dist-packages/sopel/config/types.py", line 96, in configure_setting
    value = clazz.configure(prompt, default)
  File "/usr/local/lib/python3.4/dist-packages/sopel/config/types.py", line 205, in configure
    prompt += ' (y/n)'
TypeError: unsupported operand type(s) for +=: 'NoneType' and 'str'
```

There are no permission issues and other modules setup fine

PR #1689, opened 2019-09-08
config: support quoting ListAttribute values

dgw:
Likely a better alternative to #1687. Suggested by @Exirel a little while ago as he was heading to bed. I felt like doing something code-y, so here we are.

Most of the time I spent on this actually went into the docstring updates, which are separated into their own commit. That will make it easier to pull them out and (with minor tweaks) apply the same fixes in a separate PR if we ultimately go with a different solution to the `#` problem.

Of note: I removed a reference to some feature of `StaticSection.configure_setting()` that never worked (having `prompt` be an optional argument). See #864 & ab83a1f7.


dgw:
Decided to go with #1690's solution. The doc updates will become a separate PR.

PR #1690, opened 2019-09-10
config: quote ListAttribute items when required

Exirel:
In a ListAttribute, a value can start with a #, which is then considered as a comment by the Config reader. This is bad news, as channel names usually start with a # (they can start with a & too, which is less common on freenode).

To prevent this, what I do is:

* quote the item only if necessary
* unquote it if it correspond to a very specific pattern `"#<name>"`

This way, we can have this ListAttribute:

    channels =
        "#sopel"
        &private

Edge case like `"&something"` won't be unquoted, because they don't match the exact pattern!

@dgw it replaces your PR #1689 because my first idea wasn't really on point. It would lead to a myriad of edge cases and unforeseen consequences. This one is, I believe, better, and doesn't change how current config are written.

### DIFF
--- a/sopel/config/types.py
+++ b/sopel/config/types.py
@@ -27,6 +27,7 @@ As an example, if one wanted to define the ``[spam]`` section as having an
 from __future__ import unicode_literals, absolute_import, print_function, division
 
 import os.path
+import re
 import sys
 from sopel.tools import get_input
 
@@ -221,7 +222,7 @@ class ListAttribute(BaseValidated):
     the option will be exposed as a Python :class:`list`::
 
         >>> config.spam.cheeses
-        ['camembert', 'cheddar', 'reblochon']
+        ['camembert', 'cheddar', 'reblochon', '#brie']
 
     which comes from this configuration file::
 
@@ -230,6 +231,12 @@ class ListAttribute(BaseValidated):
             camembert
             cheddar
             reblochon
+            "#brie"
+
+    Note that the ``#brie`` item starts with a ``#``, hence the double quote:
+    without these quotation marks, the config parser would think it's a
+    comment. The quote/unquote is managed automatically by this field, and
+    if and only if it's necessary (see :meth:`parse` and :meth:`serialize`).
 
     .. versionchanged:: 7.0
 
@@ -254,6 +261,13 @@ class ListAttribute(BaseValidated):
         the list.
     """
     DELIMITER = ','
+    QUOTE_REGEX = re.compile(r'^"(?P<value>#.*)"$')
+    """Regex pattern to match value that requires quotation marks.
+
+    This pattern matches values that start with ``#`` inside quotation marks
+    only: ``"#sopel"`` will match, but ``"sopel"`` won't, and neither will any
+    variant that doesn't conform to this pattern.
+    """
 
     def __init__(self, name, strip=True, default=None):
         default = default or []
@@ -276,11 +290,11 @@ class ListAttribute(BaseValidated):
             multi-line string.
         """
         if "\n" in value:
-            items = [
+            items = (
                 # remove trailing comma
                 # because `value,\nother` is valid in Sopel 7.x
                 item.strip(self.DELIMITER).strip()
-                for item in value.splitlines()]
+                for item in value.splitlines())
         else:
             # this behavior will be:
             # - Discouraged in Sopel 7.x (in the documentation)
@@ -288,21 +302,52 @@ class ListAttribute(BaseValidated):
             # - Removed from Sopel 9.x
             items = value.split(self.DELIMITER)
 
-        value = list(filter(None, items))
-        if self.strip:  # deprecate strip option in Sopel 8.x
-            return [v.strip() for v in value]
-        else:
-            return value
+        items = (self.parse_item(item) for item in items if item)
+        if self.strip:
+            return [item.strip() for item in items]
+
+        return list(items)
+
+    def parse_item(self, item):
+        """Parse one ``item`` from the list.
+
+        :param str item: one item from the list to parse
+        :rtype: str
+
+        If ``item`` matches the :attr:`QUOTE_REGEX` pattern, then it will be
+        unquoted. Otherwise it's returned as-is.
+        """
+        result = self.QUOTE_REGEX.match(item)
+        if result:
+            return result.group('value')
+        return item
 
     def serialize(self, value):
         """Serialize ``value`` into a multi-line string."""
         if not isinstance(value, (list, set)):
             raise ValueError('ListAttribute value must be a list.')
+        elif not value:
+            # return an empty string when there is no value
+            return ''
 
         # we ensure to read a newline, even with only one value in the list
         # this way, comma will be ignored when the configuration file
         # is read again later
-        return '\n' + '\n'.join(value)
+        return '\n' + '\n'.join(self.serialize_item(item) for item in value)
+
+    def serialize_item(self, item):
+        """Serialize an ``item`` from the list value.
+
+        :param str item: one item of the list to serialize
+        :rtype: str
+
+        If ``item`` starts with a ``#`` it will be quoted in order to prevent
+        the config parser from thinking it's a comment.
+        """
+        if item.startswith('#'):
+            # we need to protect item that would otherwise appear as comment
+            return '"%s"' % item
+        return item
 
     def configure(self, prompt, default, parent, section_name):
         each_prompt = '?'
--- a/test/test_config.py
+++ b/test/test_config.py
@@ -2,6 +2,7 @@
 from __future__ import unicode_literals, division, print_function, absolute_import
 
 import os
+import sys
 
 import pytest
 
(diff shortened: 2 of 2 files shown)


dgw:
`./sopel/config/types.py:307:39: F812 list comprehension redefines 'value' from line 277`, but only on Py2.7, because some scoping rule changed from 2 to 3.

Exirel:
@dgw fixed, no problem. :)

Exirel:
> @dgw fixed, no problem. :)

Apparently there is something wrong with it: tests are failing for 2.7 only.

lol, well done Python 2.7, you're officially the worst.

Exirel:
OK I found the problem, it's from Python 2.7's ConfigParser built-in module:

```
if line.strip() == '' or line[0] in '#;':
    continue
```

So, if the line is empty: skip it.
If the line *starts with* # or ;: skip it.
If the line *starts with* anything else: **do not skip**.

On the other hand in Python 3, and [I quote the documentation](https://docs.python.org/3.6/library/configparser.html#configparser.ConfigParser):

> Comments can be indented.

My mood right now:

ezgif com-video-to-gif

(This is a gif by [Boulet on Twitter](https://twitter.com/Bouletcorp/status/1171056137057329152), a French artist that also provides [an English version of his blog](http://english.bouletcorp.com/))

Exirel:
NOW it works.

@dgw I waited for the tests this time.

sopel/config/types.py
```diff
@@ -254,6 +261,13 @@ class SpamSection(StaticSection):
         the list.
     """
     DELIMITER = ','
+    QUOTE_REGEX = re.compile(r'^"(?P<value>#.*)"$')
+    """Regex pattern to match value that requires quotation mark.
```
```suggestion
    """Regex pattern to match value that requires quotation marks.
```

sopel/config/types.py
```diff
@@ -254,6 +261,13 @@ class SpamSection(StaticSection):
...
+
+    This pattern matches values that starts with a ``#`` inside quotation marks
```
```suggestion
    This pattern matches values that start with ``#`` inside quotation marks
```

sopel/config/types.py
```diff
@@ -254,6 +261,13 @@ class SpamSection(StaticSection):
...
+    only: ``"#sopel"`` will match, but ``"sopel"`` won't, neither will any
```
```suggestion
    only: ``"#sopel"`` will match, but ``"sopel"`` won't, and neither will any
```

(probably +wrap)

sopel/config/types.py
```diff
@@ -276,33 +290,64 @@ def parse(self, value):
             multi-line string.
         """
         if "\n" in value:
-            items = [
+            items = (
                 # remove trailing comma
                 # because `value,\nother` is valid in Sopel 7.x
                 item.strip(self.DELIMITER).strip()
-                for item in value.splitlines()]
+                for item in value.splitlines())
```
dgw: Why change this away from list, only to convert it to list anyway below, in the `return`?

The fallback logic uses `.split()`, which already returns a list. Just seems weirdly redundant.

sopel/config/types.py
```diff
@@ -276,33 +290,64 @@ def parse(self, value):
...
             # - Deprecated in Sopel 8.x
             # - Removed from Sopel 9.x
             items = value.split(self.DELIMITER)
 
-        value = list(filter(None, items))
-        if self.strip:  # deprecate strip option in Sopel 8.x
-            return [v.strip() for v in value]
-        else:
-            return value
+        items = (self.parse_item(item) for item in items if item)
```
dgw: Same as above re: tuple/list.

sopel/config/types.py
```diff
@@ -276,33 +290,64 @@ def parse(self, value):
...
+        return list(items)
+
+    def parse_item(self, item):
+        """Parse one ``item`` from the list.
+
+        :param str item: one item from the list to parse
+        :rtype: str
+
+        If ``item`` matches the :attr:`QUOTE_REGEX` pattern, then it will be
+        unquoted. Otherwise it's returned as it is.
```
```suggestion
        unquoted. Otherwise it's returned as-is.
```

Maybe this is a colloquialism? 🤔

sopel/config/types.py
```diff
@@ -276,33 +290,64 @@ def parse(self, value):
...
+        return '\n' + '\n'.join(self.serialize_item(item) for item in value)
+
+    def serialize_item(self, item):
+        """Serialize an ``item`` from the list value.
+
+        :param str item: one item of the list to serialize
+        :rtype: str
+
+        If ``item`` starts with a ``#`` it will be quoted in order to prevent
+        the config parser to think it's a comment.
```
```suggestion
        the config parser from thinking it's a comment.
```

sopel/config/types.py
Exirel: A `(... for ... in ...)` syntax is a generator-expression, it returns a generator (no execution yet, so iteration count is 0), while a `[... for ... in ...]` is a list-expression which creates a list (there was an execution, so the iteration count is N, the length of the list).

I haven't look a the code around, but I assume I reuse `items` for another transformation. If so, we can keep that. If not, then we can safely use a list-expression.

sopel/config/types.py
Exirel: No idea, my English is tainted by school, and we didn't learn about urban dictionary.

sopel/config/types.py
Exirel: So now that I've read the code, this version does:

```
items = (item.strip(...) for item in value.splitlines())  # 0 iteration
items = (self.parse_item(item) for item in items if item)  # 0 iteration
if self.strip:
    return [item.strip() for item in items]  # N iterations
# else:
return list(items) # N iterations
```

So in total: N iterations, N being the number of channels.

On the other side, if we change these generator-expression into list-expression, we will have:


```
items = [item.strip(...) for item in value.splitlines()]  # N iterations
items = [self.parse_item(item) for item in items if item]  # +N iterations
if self.strip:
    return [item.strip() for item in items]  # +N iterations
# else:
return items # 0 iteration
```

At best, we have 2N iterations, at worst 3N iterations. It's nothing, in term of optimization, because the data is so small. Yet, I've this habit of using generator-expression and `yield` syntax everywhere I can.

Exirel:
@dgw will squash when approved! :)

sopel/config/types.py
dgw: Fair enough. Guess I've outed myself through this conversation, revealing that I never use generators. 😁

sopel/config/types.py
dgw: Oh, it's not slang (which is what UD tends to cover). But it _is_ an idiomatic expression, which is just as likely to be left out of formal language education. 😆

Exirel:
Waiting for Travis to do its job after the rebase & squash. :+1:

test/test_config.py
```diff
@@ -26,8 +27,43 @@
...
+    "#startquote
+    &endquote"
+    "&quoted"
 """  # noqa (trailing whitespaces are intended)
 
+TEST_CHANNELS = [
+    '#sopel',
+    '&peculiar',
+    '#private',
+    '"#startquote',  # start quote withtout end quote: kept
```
dgw:
```suggestion
    '"#startquote',  # start quote without end quote: kept
```

test/test_config.py
```diff
@@ -26,8 +27,43 @@
...
+    '&endquote"',
+    '"&quoted"',  # quoted, but no #: quotes kept
+]
+
+if sys.version_info.major < 3:
+    # Python 2.7's ConfigParser interpret as comments
```
dgw:
```suggestion
    # Python 2.7's ConfigParser interprets as comment
```

test/test_config.py
```diff
@@ -26,8 +27,43 @@
...
+    # a line that starts with # or ;.
+    # Python 3 on the other hand, allow comments to be indented.
```
dgw:
```suggestion
    # Python 3, on the other hand, allows comments to be indented.
```

test/test_config.py
```diff
@@ -26,8 +27,43 @@
...
+    # As a result, the same config file will result in a different
+    # config file depending on the Python version used.
```
dgw:
```suggestion
    # config object depending on the Python version used.
```

test/test_config.py
```diff
@@ -26,8 +27,43 @@
...
+    # TODO: Deprecated with Python 2.7.
+    TEST_CHANNELS = [
+        '#sopel',
+        '&peculiar',
+        '# python 3 only comment',  # indented line are not comment for Py2
```
dgw:
```suggestion
        '# python 3 only comment',  # indented lines cannot be comments in Py2
```

test/test_config.py
```diff
@@ -287,3 +335,11 @@ def test_save_modified_config(multi_fakeconfig):
         'ListAttribute with one line only, with commas, must *not* be split '
         'differently from what was expected, i.e. into one (and only one) value'
     )
+    assert saved_config.spam.channels == [
+        '#sopel',
+        '#private',
+        '&peculiar',
+        '"#startquote',  # start quote withtout end quote: kept
```
dgw:
```suggestion
        '"#startquote',  # start quote without end quote: kept
```

dgw:
@Exirel Looks good. Squash again and I'll merge before I have any time to look for typos a third time. 😝

Exirel:
Done! :shipit:

Issue #1683, opened 2019-08-20
Message intents are dead


dgw:
The `intents` feature is long dead, having been pulled from IRCv3 almost three years ago (ircv3/ircv3-specifications#271). This affects Sopel in a few ways.

First, we have a `@module.intent` decorator that is now obsolete, named for a dead specification that nobody is going to implement.

Second, Sopel internally handles CTCP ACTIONs (the only use case I was ever aware of for intents) by stripping the CTCP bytes and adding the "ACTION" intent to the trigger instead. That's why plugin functions wanting to act only upon ACTION messages must use `@module.intent('ACTION')`.

Third, obviously the `intents` attribute added to Sopel callables with intent filters is also obsolete.

In short, there's a fair bit of restructuring to do here. A quick chat with the IRCv3 folks on freenode tells me that CTCP ACTIONs themselves aren't going anywhere, so we should be fine designing an API around those as they currently exist. Maybe we add `@module.action`, like `@module.echo`, and kill off regular `@module.commands`-decorated callables receiving ACTION messages at all. (That would take care of an oddity that exists in current Sopel 6.x, wherein one can do `/me .command` and Sopel will act on it just like a regular message, which can be unexpected.)

That same chat on IRC also told me that the direct replacement for intents is message-tags. We don't support those at all, at least not in a way that plugins can use.

I'd like to get this done for Sopel 7.0, but only deprecating `@module.intent` and replacing it with something else is truly important. Opening message-tags up in the plugin API can wait until 7.1+.

deathbybandaid:
Started digging in the code regarding this since I am actively interested in #1660 . Trying to wrap my head around what happens here for 'intents' currently.

Warning: This may turn into a rant/rambling.

Note: testing was done on UnrealIRCd-4.2.2

**Obviously, whatever direction this goes, we may want to retain some backwards compatibility for older ircd servers. I'm fairly certain that unrealircd is a participant in ircv3, so I'm wondering what other tag ACTION would fall under besides 'intent'.

looks like Message Tags is somewhat already there as of 4.1.0?

https://github.com/sopel-irc/sopel/blob/master/NEWS#L1209


Looks like this is where we parse the tags into a usable format:
https://github.com/sopel-irc/sopel/blob/master/sopel/trigger.py#L34-L43

At a glance, the code looks like it should be making a reference of all the tags a message contains even if we aren't looking for it. Curious what tags are possible here.

[...]

PR #1693, opened 2019-09-17
loader: default rule for events

dgw:
Helped @deathbybandaid with `@event` decorators not working and realized that this specific problem comes up _a lot_, relatively speaking, when Sopel users have trouble writing custom plugins.

Obviously our documentation doesn't do a great job of making sure plugin devs know that `@rule` is required along with `@event`. But rather than fix the documentation, I decided to do the user-friendly (developer-friendly?) thing and make it so `event` is considered triggerable just like a command or rule, and so it will also get the default `.*` rule applied if necessary once registered.

An early version of this patch (pre-publication) also applied this default rule to intents, but:

  * There's only one meaningful intent value (`'ACTION'`), which almost always requires an explicit rule anyway.
  * We're probably going to deprecate and/or remove `@intents` pretty soon because it's obsolete (#1683).

Other rejected ideas were applying this to `@url` (not needed) and to `@echo` (immediately determined to be _the exact opposite_ of intuitive behavior once I gave it ten seconds of thought).

So here we are: A ±0 line patch (if we ignore tests & code formatting) to the loader that saves us almost 20 real-code lines of boilerplate. Not bad for a bit of afternoon hacking, if I do say so myself.

### DIFF
--- a/sopel/coretasks.py
+++ b/sopel/coretasks.py
@@ -62,7 +62,6 @@ def auth_after_register(bot):
 
 
 @sopel.module.event(events.RPL_WELCOME, events.RPL_LUSERCLIENT)
-@sopel.module.rule('.*')
 @sopel.module.thread(False)
 @sopel.module.unblockable
 def startup(bot, trigger):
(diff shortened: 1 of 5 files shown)


dgw:
@Exirel Merged so you can continue #1678, as you mentioned on IRC. 🚀

PR #1694, opened 2019-09-26
tell: use config to send reminders as private messages

### LINKED ISSUE #1151: Add private option to .tell
Or at least add a setting or something that way the channel doesn't get spammed by people telling each other stuff.

Exirel:
This fix #1151 

* `config.tell.use_private_reminder` (y/n) [n]: send tell/ask reminders as private messages
* `config.tell.maximum_public` [4]: maximum number of tell/ask reminders to send as public messages before sending them as private messages

As a bonus point: minor refactoring with unit tests.

### DIFF
--- a/sopel/modules/tell.py
+++ b/sopel/modules/tell.py
@@ -130,55 +130,88 @@ def f_remind(bot, trigger):
         bot.say("Hey, I'm not as stupid as Monty you know!")
 
 
-def getReminders(bot, channel, key, tellee):
+def get_nick_reminders(reminders, nick):
     lines = []
     template = "%s: %s <%s> %s %s %s"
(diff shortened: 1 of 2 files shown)


Exirel:
> Why bring `io` into this? The only thing I see used from that module, `io.open()`, is [just an alias](https://docs.python.org/3/library/io.html#io.open) to the built-in `open()`.

Because [Python 2.7's open function](https://docs.python.org/2/library/functions.html#open) doesn't know about "encoding". Since I wanted to deal with unicode string only, I switched to `io.open`.

Exirel:
> Now I just hope this doesn't cause conflicts with #1699.

I think it'll. But that's okay, the change aren't that big, and you now have unit-tests to help you!

PR #1697, opened 2019-09-27
sopel: refactor how the bot manages its connection

Exirel:
This PR started as a proof of concept, and now that it has been tested with a secure connection (using the freenode IRC network) for several hours, and given the amount of added tests, I feel confident enough to suggest this PR for Sopel 7.0.0.

The main idea is to separate the bot instance, which handles the plugins, the setup, and the bot's life cycle; from the IRC connection, which handles the socket, the SSL config, and the connection's life cycle.

Future PR for future version of Sopel might include 2 backends:

* one using `asynchat.async_chat` and `asyncore`, both from the Python 2.7-3.5 era
* and the other using `asyncio`, stable in Python 3.6 and above

Technically speaking, it's compatible with Sopel 6.x, and it's compatible with the current set of internal plugins. The interface remains the same as before. This is particularly visible with the added test on `sopel.bot.SopelWrapper`.

I know it looks risky at first, and it won't be easy to review because I moved a lot of code. I tried to keep things as simple as possible, given the scope in features and behaviours.

### For Reviewvers

Here is a little tour of what has been modified and what you have to review.

There are 4 big parts:

* `sopel.irc.AbstractBot`: the base bot interface, between the IRC Backend and the outside world
* `sopel.bot.Sopel`: is now a subclass of `sopel.irc.AbstractBot`, and handle the dispatch of messages, the plugin management, and so on
* `sopel.irc.abstract_backends`: this module define the base interface for IRC Backend, which is how the backend should interact with the bot
* `sopel.irc.backends`: this module contains an SSL IRC Connection backend and a non-SSL one using `asynchat` (which depends on `asyncore` itself)

The new interface between the IRC Connection and the outside world lies in `sopel.irc.AbstractBot`: this is your old `sopel.irc.Bot` without the `asynchat` base and implementation. Both `collect_incoming_data` and `found_terminator`  have been moved to backend implementation based on `asynchat`. This interface exposes the following behaviours for the Bot:

* `bot.on_connect` is called by the backend when the socket connection is established; should be called only once
* `bot.on_close` is called by the backend when the socket connection must be closed
* `bot.on_error` is called when the IRC server sends an error message
* `bot.on_message`  is called when the IRC server sends a message

The backend interface is defined in `sopel.irc.abstract.backends.AbstractIRCBackend`: it defines the method that must be implemented by any IRC Connection backend. Basically, it defines a `send_command` method which requires:

* an implementation of the `send`  method in the subclass
* a method called `on_message_sent` on the bot's side

This abstract backend doesn't define how the PING and Server Timeout is managed, and at the moment I managed all of that in the `sopel.irc.backends.AsynchatBackend`

This backend (and its subclass for SSL connection) are now responsible for the direct interaction with the IRC Server. You will notice that these backend implementations are now responsible for all the `handle_*` methods, which are the methods the `asynchat` package uses for its interface with the socket.


Exirel:
@dgw you should have a look when you have an hour or so. I think it's quite promising!

dgw:
I'm optimistically slotting this into 7.1. If it's stable, and doesn't change the API or requirements, there's no reason not to ship a transparent rewrite like that mid-cycle.

Exirel:
It works! It works on Freenode with a secure connection (with verify SSL), and it works on a private server without a secure connection (which isn't available). I find it pretty stable, I was able to run it for 2h today without a single crash or timeout. I tried various commands, various URL callbacks, etc. and jobs seems fine too - beside, I'm even using the Job Scheduler for ping/timeout check.

I'm really happy to see how it turned out well so quickly. My main concern now is to have merge and rebase conflicts as other PRs will get merged first. In theory, it shouldn't have a big impact: anything that is related to a plugin or a tool will be fine. Everything that touch `sopel.irc`  or `sopel.bot` **will be a problem**. That's also why I based this PR off #1678 in the first place.

@dgw I believe this could be available for Sopel 7.0, provided that #1678 is merged as soon as possible so I can install a version of this bot to run on a server for a week or so, making sure it's still stable after 2h.

Exirel:
Ok, thanks to #1678 being merged, this PR is now way easier* to review. Yet, I'd like to add more tests, because we can always do more tests.

*: _give or take_

dgw:
Hmm, you're really pushing to get this in 7.0. 😏

Maybe, since we're only a couple months from the release target anyway (sometime in December), I should start cutting alphas of 7.0.0 after this is merged? Publishing alphas on PyPI would make it easier to crowdsource the task of finding out what's been broken by all the refactoring. 😂

Exirel:
Hm, yes, could be a great idea indeed!

> Hmm, you're really pushing to get this in 7.0. 😏

I really don't want to fix the merge conflicts that will inevitably come if I don't. 😁

Exirel:
Here is a list of PR that have been published before that one and that *may* generate a merge conflict:

* #1536: maybe, but since it adds something to `sopel.bot` it shouldn't be a problem
*  #1672: might not have a conflict, because the dispatch method shouldn't be a problem

And... that's it for now. I counted a dozen other PRs in the FIFO order that won't be a problem.

test/test_irc.py
```diff
@@ -1,146 +1,324 @@
...
+        'PRIVMSG #sopel :hello',
+        'PRIVMSG #sopel :hello',
+        'PRIVMSG #sopel :hello',
+        'PRIVMSG #sopel :...',
+        # the new ones are also replaced by '...'
+        'PRIVMSG #sopel :...',
+        'PRIVMSG #sopel :...',
+    )
+
+    # but at some point it just stop talking
```
dgw:
```suggestion
    # but at some point it just stops talking
```

sopel/irc/backends.py
```diff
@@ -0,0 +1,252 @@
...
+            if why[0] in (asyncore.EWOULDBLOCK, errno.ESRCH):
+                return 0
+            else:
+                raise why
+            return 0
+
+    def recv(self, buffer_size):
+        """Replacement for self.recv() during SSL connections.
+
+        From a (now deleted) blog post by Evan "K7FOS" Fosmark.
```
dgw: I'd leave the URL even if it errors, in case an archive of it turns up.

https://k7fos.com/2010/09/ssl-support-in-asynchatasync_chat

test/test_irc.py
```diff
@@ -1,146 +1,324 @@
...
+def test_say_safe(bot):
+    bot.say('Hello\r\nworld!\r\n', '#sopel')
+
+    assert bot.backend.message_sent == rawlist(
+        'PRIVMSG #sopel :Helloworld!',
+    )
+
+
+def test_say_long_fit(bot):
+    """Test a long message that fit into the 512 bytes limit."""
```
dgw:
```suggestion
    """Test a long message that fits into the 512 bytes limit."""
```

Aha! Found your nemesi**S** lurking!

sopel/bot.py
```diff
@@ -823,6 +615,40 @@ def dispatch(self, pretrigger):
...
+        """Called when a job from the Job Scheduler fails.
+
+        .. seealso::
+
+            :meth:`error`
+        """
+        self.error(exception=exc)
+
+    def error(self, trigger=None, exception=None):
+        """Called internally when a module causes an error."""
```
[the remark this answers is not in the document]
Exirel: Hm... I think it's a plugin here. Good catch.

sopel/irc/backends.py
```diff
@@ -0,0 +1,252 @@
...
+
+    def on_scheduler_error(self, scheduler, exc):
+        """Called when the Job Scheduler fails."""
+        LOGGER.exception('Error with the timeout scheduler: %s', exc)
+        self.handle_close()
+
+    def on_job_error(self, scheduler, job, exc):
+        """Called when a job from the Job Scheduler fails."""
+        LOGGER.exception('Error with the timeout scheduler: %s', exc)
+        self.handle_close()
```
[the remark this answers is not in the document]
Exirel: Hm. Indeed that's weird. Why did I do that?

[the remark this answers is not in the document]
Exirel: Oh, I know. It's because on the backend, there is one and only one job scheduler: the timeout one. So when there is a timeout error, the bot force-closes the connection.

This is the current behavior in 6.x & master.

sopel/irc/backends.py
```diff
@@ -0,0 +1,252 @@
...
+                            "using {0} instead of {1} for TLS connection"
+                            .format(hostname, self.host))
+                        has_matched = True
+                        break
+                    except ssl.CertificateError:
+                        pass
+
+                if not has_matched:
+                    # everything is broken
+                    LOGGER.error("invalid certificate, no hostname matches")
```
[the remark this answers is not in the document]
Exirel: Do I add periods, too?

sopel/irc/abstract_backends.py
```diff
@@ -0,0 +1,179 @@
...
+        :param str text: optional text for leaving the channel
+        """
+        self.send_command('PART', safe(channel), text=reason)
+
+    def send_quit(self, reason=None):
+        """Send a ``QUIT`` command.
+
+        :param str reason: optional text for leaving the server
+
+        This won't send anything if the backend isn't connected.
```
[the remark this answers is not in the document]
Exirel: That's a good question, let me think about it.

sopel/irc/abstract_backends.py
```diff
@@ -0,0 +1,179 @@
...
+        """
+        if self.connected:  # TODO: refactor for a method instead of attribute
+            self.send_command('QUIT', text=reason)
+
+    def send_kick(self, channel, nick, reason=None):
+        """Send a ``KICK`` command for ``nick`` in ``channel`` .
+
+        :param str channel: Channel name from which to kick ``nick``
+        :param str nick: Nick to kick from the ``channel``
+        :param str reason: optional reason for the kick
```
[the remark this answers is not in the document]
Exirel: Oh! Very good point, thank you. I'll fix that!

sopel/irc/backends.py
```diff
@@ -0,0 +1,252 @@
...
+import sys
+
+from sopel.tools.jobs import JobScheduler, Job
+from .abstract_backends import AbstractIRCBackend
+from .utils import get_cnames
+
+try:
+    import ssl
+    if not hasattr(ssl, 'match_hostname'):
+        # Attempt to import ssl_match_hostname from python-backports
```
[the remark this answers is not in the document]
Exirel: Nice catch!

sopel/irc/backends.py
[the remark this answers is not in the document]
Exirel:
```suggestion
                    LOGGER.error("Invalid certificate, no hostname matches.")
```

Exirel:
Note: the failing test on 51548ba553ff8734a2727c2794ea18a233428755 is due to a timeout error of an external service.

sopel/irc/abstract_backends.py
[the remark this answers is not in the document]
Exirel: In the current 6.x/master version, `quit`  is the only case where we fail silently: in all other cases, we try to send our message (using `write`, mostly). The code you have here does the same thing, so it's "good enough" for now, IMHO.

Exirel:
@dgw I did squash review commits into 3 commits, but didn't touch commits existing before your review, nor did I rebase upon upstream/master so I think you shouldn't have problem with the re-review.

sopel/irc/abstract_backends.py
```diff
@@ -0,0 +1,179 @@
...
+        """
+        self.send_command('NICK', safe(nick))
+
+    def send_user(self, user, mode, nick, name):
+        """Send a ``USER`` command with a ``user``.
+
+        :param str user: IRC username
+        :param str mode: List of mode to send for the user
+        :param str nick: Nick associated with this user
+        :param str name: Name for the user
```
dgw:
```suggestion
        :param str name: "real name" for the user
```

(That is what this field's for, right? I'm kinda lost in grammar-land rn. 😅)

sopel/irc/abstract_backends.py
```diff
@@ -0,0 +1,179 @@
...
+
+        :param str nick: Nick to take
+        """
+        self.send_command('NICK', safe(nick))
+
+    def send_user(self, user, mode, nick, name):
+        """Send a ``USER`` command with a ``user``.
+
+        :param str user: IRC username
+        :param str mode: List of mode to send for the user
```
dgw:
```suggestion
        :param str mode: mode(s) to send for the user
```

Calling it a "list of modes" could be confusing, since the expected type is `str`.

sopel/irc/abstract_backends.py
```diff
@@ -0,0 +1,179 @@
...
+        :param str nick: Nick associated with this user
+        :param str name: Name for the user
+        """
+        self.send_command('USER', safe(user), mode, safe(nick), text=name)
+
+    def send_pass(self, password):
+        """Send a ``PASS`` command with a ``password``.
+
+        :param str password: Password for authentication
```
dgw:
```suggestion
        :param str password: password for authentication
```

sopel/irc/abstract_backends.py
```diff
@@ -0,0 +1,179 @@
...
+        """
+        self.send_command('PASS', safe(password))
+
+    def send_join(self, channel, password=None):
+        """Send a ``JOIN`` command to ``channel`` with optional ``password``.
+
+        :param str channel: Channel to join
```
dgw:
```suggestion
        :param str channel: channel to join
```

sopel/irc/abstract_backends.py
```diff
@@ -0,0 +1,179 @@
...
+        """
+        if password is None:
+            self.send_command('JOIN', safe(channel))
+        else:
+            self.send_command('JOIN', safe(channel), safe(password))
+
+    def send_part(self, channel, reason=None):
+        """Send a ``PART`` command to ``channel``.
+
+        :param str channel: Channel to part
```
dgw:
```suggestion
        :param str channel: the channel to part
```

sopel/irc/abstract_backends.py
```diff
@@ -0,0 +1,179 @@
...
+
+        This won't send anything if the backend isn't connected.
+        """
+        if self.connected:  # TODO: refactor for a method instead of attribute
+            self.send_command('QUIT', text=reason)
+
+    def send_kick(self, channel, nick, reason=None):
+        """Send a ``KICK`` command for ``nick`` in ``channel`` .
+
+        :param str channel: Channel name from which to kick ``nick``
```
dgw:
```suggestion
        :param str channel: the channel from which to kick ``nick``
```

sopel/irc/abstract_backends.py
```diff
@@ -0,0 +1,179 @@
...
+        :param str nick: Nick to kick from the ``channel``
```
dgw:
```suggestion
        :param str nick: nickname to kick from the ``channel``
```

sopel/irc/abstract_backends.py
```diff
@@ -0,0 +1,179 @@
...
+        :param str reason: optional reason for the kick
+        """
+        self.send_command('KICK', safe(channel), safe(nick), text=reason)
+
+    def send_privmsg(self, dest, text):
+        """Send a ``PRIVMSG`` command to ``dest`` with ``text``.
+
+        :param str dest: Channel name or Nick name
```
dgw:
```suggestion
        :param str dest: nickname or channel name
```

sopel/irc/abstract_backends.py
```diff
@@ -0,0 +1,179 @@
...
+        :param str text: Text to send
```
dgw:
```suggestion
        :param str text: the text to send
```

sopel/irc/abstract_backends.py
```diff
@@ -0,0 +1,179 @@
...
+        """
+        self.send_command('PRIVMSG', safe(dest), text=text)
+
+    def send_notice(self, dest, text):
+        """Send a ``NOTICE`` command to ``dest`` with ``text``.
+
+        :param str dest: Channel name or Nick name
```
dgw:
```suggestion
        :param str dest: nickname or channel name
```

sopel/irc/abstract_backends.py
```diff
@@ -0,0 +1,179 @@
...
+        :param str text: Text to send
```
dgw:
```suggestion
        :param str text: the text to send
```

sopel/test_tools.py
```diff
@@ -41,6 +43,30 @@
     basestring = str
 
 
+def rawlist(*args):
+    """Build a list of IRC raw messages from the lines given as ``*args``.
```
dgw:
```suggestion
    """Build a list of raw IRC messages from the lines given as ``*args``.
```

sopel/test_tools.py
```diff
@@ -41,6 +43,30 @@
...
+
+    :return: a list of IRC raw messages as seen by the bot
```
dgw:
```suggestion
    :return: a list of raw IRC messages as seen by the bot
```

sopel/irc/__init__.py
```diff
@@ -0,0 +1,547 @@
...
+
+import sys
+import time
+import os
+import logging
+
+try:
+    import ssl
+    if not hasattr(ssl, 'match_hostname'):
+        # Attempt to import ssl_match_hostname from python-backports
```
dgw: Here, too, needs a py2 note like the other one you added earlier.

sopel/irc/__init__.py
```diff
@@ -0,0 +1,547 @@
...
+        self.backend.send_join(channel, password=password)
+
+    def kick(self, nick, channel, text=None):
+        """Kick a ``nick`` from a ``channel``.
+
+        :param str nick: Nick to kick out of the ``channel``
+        :param str channel: Channel to kick ``nick`` from
+        :param str text: Optional text for the kick
+
+        The bot must be a channel operator in specified channel for this to
```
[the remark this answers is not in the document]
Exirel: How did I miss these Nick/Channel/Optional? :(

sopel/irc/__init__.py
```diff
@@ -411,11 +412,11 @@ def join(self, channel, password=None):
...
-        :param str nick: Nick to kick out of the ``channel``
-        :param str channel: Channel to kick ``nick`` from
-        :param str text: Optional text for the kick
+        :param str nick: nick to kick out of the ``channel``
+        :param str channel: channel to kick ``nick`` from
+        :param str text: optional text for the kick
 
-        The bot must be a channel operator in specified channel for this to
+        The bot must be operator in the specified channel for this to work.
         work.
```
dgw: This line is redundant now. 😁

PR #1698, opened 2019-10-04
bot: fix arg order in say() recursion

dgw:
Mixed up by @deathbybandaid in #1606. For shame that the rest of us missed it in code review, too!

Thanks to @cottongin for reporting.


dgw:
**Note:** The very same bug is fixed in #1697's refactor. If that goes in soon, we don't need this.

PR #1699, opened 2019-10-06
remind, tell: Use config name in fake `.db` files

### LINKED ISSUE #1359: [tell/remind] Switch to config-name for "db" files
It's kind of silly that the "database" files `tell` and `remind` write use `bot.config.core.host` in the filename. We have a config name, which is already used for the main SQLite database. They should just use it so the core stuff is a bit more predictable.

Will need a migration routine to rename the existing file at module load. That's a prerequisite for merging, whether I create this patch or someone else does.

dgw:
Instead of using the bot's nick and configured host (to connect to), let's use the config `basename`. This matches the default SQLite DB filename and (as of #1547) the names used for Sopel's log files.

Closes #1359. Labeling "Documentation" as I need to write a new section covering this for the upgrade guide in sopel-irc/sopel.chat#17 (already referenced here in the help links given if the migration fails).

### DIFF
--- a/sopel/modules/remind.py
+++ b/sopel/modules/remind.py
@@ -2,6 +2,7 @@
 """
 remind.py - Sopel Reminder Module
 Copyright 2011, Sean B. Palmer, inamidst.com
+Copyright 2019, dgw, technobabbl.es
 Licensed under the Eiffel Forum License 2.
 
 https://sopel.chat
@@ -11,6 +12,7 @@ from __future__ import unicode_literals, absolute_import, print_function, divisi
 import codecs
 import collections
 from datetime import datetime
+import logging
 import os
 import re
 import time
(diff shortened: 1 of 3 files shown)


kwaaak:
Should we use a separate file at all? Now that there is #1621, this information could live in the big database.

dgw:
@kwaaak It could, but the `bot.db` API isn't really designed to handle object dumps. We technically only "support" up to 255 characters each for `key` and `value`.

These plugins could use custom tables, true, but their dump files are not _that_ different from the GeoIP and malwaredomains files we already cache on disk. And those _definitely_ aren't going into `bot.db`!

dgw:
(Wrong browser tab, touched milestones I shouldn't have. Whoops.)

PR #1701, opened 2019-10-06
module, bot: add `output_prefix` decorator & feature

### LINKED ISSUE #1352: Plugin API: Allow declaring output prefix
This has come up enough times in just my own work on modules (most recently sopel-wolfram & sopel-github) that I think it's time to open a proper issue.

I think it would be useful for Sopel's module API to support declaring an output prefix, such that any calls to `bot.say()` or `bot.reply()` automatically get prefixed with the module's choice of string. (`bot.msg()` and other lower-level calls, maybe not. I'm still thinking about that.)

Why? It would eliminate having to hard-code, for example, `[GitHub]` (see sopel-irc/sopel-github#9) or `[W|A]` all over the place. The prefix string would be defined only once.

Modules can, of course, roll their own version of this. But just like with #1346 (proposing the addition of `web.unquote()` to mirror `web.quote()`), there are some things that happen often enough across modules to warrant inclusion in Sopel's library. I think prefixing the output is one of those things.

Obviously the idea that this API functionality should exist is the easy part. The hard part is figuring out how it should work, but there's plenty of time between now and when Sopel 6.6.0 drops to think, plus all the time between then and Sopel 7 (when I'd like to actually release this) to iterate on it.

Considerations I've thought of so far:

  * Should square brackets be added automatically, for consistency, or should modules be allowed to define the entire prefix string? Including the trailing space between it and the rest of the message?
  * If prefixes get bracket-wrapped automatically, should there be an override mechanism?
  * Where should this be set up? The module `setup()` function is a logical place, I guess, though it would mean setting the prefix would need to be a function of the `bot` object. Hmm…

(As I said, the idea is the easy part!)

dgw:
Decorated functions will have anything they output (via the `say` or `notice` wrapper methods; it doesn't make sense for `action` or `reply`) prefixed with the given string.

Closes #1352.

### DIFF
--- a/sopel/bot.py
+++ b/sopel/bot.py
@@ -717,7 +717,8 @@ class Sopel(irc.Bot):
                         continue
 
                     # call triggered function
-                    wrapper = SopelWrapper(self, trigger)
+                    wrapper = SopelWrapper(
+                        self, trigger, output_prefix=func.output_prefix)
                     if func.thread:
                         targs = (func, wrapper, trigger)
                         t = threading.Thread(target=self.call, args=targs)
@@ -990,12 +991,16 @@ class SopelWrapper(object):
     the sender (either a channel or in a private message) and even to
     :meth:`reply to someone<reply>` in a channel.
     """
-    def __init__(self, sopel, trigger):
+    def __init__(self, sopel, trigger, output_prefix=''):
+        if not output_prefix:
+            # Just in case someone passes in False, None, etc.
+            output_prefix = ''
         # The custom __setattr__ for this class sets the attribute on the
         # original bot object. We don't want that for these, so we set them
         # with the normal __setattr__.
         object.__setattr__(self, '_bot', sopel)
         object.__setattr__(self, '_trigger', trigger)
+        object.__setattr__(self, '_out_pfx', output_prefix)
 
     def __dir__(self):
         classattrs = [attr for attr in self.__class__.__dict__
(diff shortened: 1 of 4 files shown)


sopel/module.py
```diff
@@ -674,3 +675,22 @@ def __call__(self, func):
...
+
+def output_prefix(prefix):
+    """Decorate a function to add a prefix on its output.
+
+    :param str prefix: the prefix to add (must include trailing whitespace if
+                       desired; Sopel does not assume it should add anything)
+
+    Prefix will be added to text sent through:
+
+      * :meth:`sopel.bot.SopelWrapper.say<bot.say>`
```
Exirel: I think it's over-indented here. Not sure restructuredtext likes it.

sopel/bot.py
```diff
@@ -716,8 +716,11 @@ def dispatch(self, pretrigger):
                     if is_echo_message and not func.echo:
                         continue
 
+                    # Handle output prefix feature
+                    pfx = getattr(func, 'output_prefix', '')
```
Exirel: [nitpick] Not sure I like the `pfx` variable name.

sopel/module.py
dgw: From a test `make html` run in `/docs/`:

image

Looks OK, but removing the leading spaces does change the output, so I'll take those out in a future PR update. I also need to fix the `:meth:` references; they're broken.

sopel/bot.py
dgw: Fixed by eliminating the intermediate variable. 😝

PR #1707, opened 2019-10-10
tools.time: fully utilize pytz to validate timezones

dgw:
Instead of trying to coerce zone names all the way into something we can throw into an `in pytz.all_timezones` statement, all we have to do is rearrange the bits just enough to let `pytz.timezone()` take over. Its built-in normalization routines are plenty comprehensive, and if we make use of them there's no need to reinvent the wheel.

Updated the docstring for `validate_timezone()` to match. We aren't expecting that zone names like "EDT" will actually work. Neither 6.6.x nor the code this replaces will match that example, so we're not losing any functionality. Most of the three- and four-letter zone abbreviations I tried, in fact, fail because they aren't specific enough (might refer to one of multiple zones around the world). This is why we have the `tz` page users can visit if they don't know their local zone name, anyway.

Bonus spelling fixes included because if I didn't make them now, I'd forget and possibly miss them when going over this module's docs later.

### DIFF
--- a/sopel/tools/time.py
+++ b/sopel/tools/time.py
@@ -35,8 +35,7 @@ def validate_timezone(zone):
         tz = pytz.timezone(zone)
     except pytz.exceptions.UnknownTimeZoneError:
         raise ValueError('Invalid time zone.')
-    else:
-        return tz.zone
+    return tz.zone
 
 
 def validate_format(tformat):
@@ -44,7 +43,7 @@ def validate_format(tformat):
     try:
         time = datetime.datetime.utcnow()
         time.strftime(tformat)
-    except Exception:  # TODO: Be specific
+    except (ValueError, TypeError):
         raise ValueError('Invalid time format.')
     return tformat
 
--- a/sopel/tools/time.py
+++ b/sopel/tools/time.py
@@ -39,7 +39,12 @@ def validate_timezone(zone):
 
 
 def validate_format(tformat):
-    """Returns the format, if valid, else None"""
+    """Validate a time format string.
+
+    :param string tformat: the format string to validate
+    :return: the format string, if valid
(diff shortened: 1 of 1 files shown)


sopel/tools/time.py
```diff
@@ -20,30 +20,23 @@ def validate_timezone(zone):
...
-        zone = zone.upper()
-    else:
-        zone = zone.title()
-    if zone in pytz.all_timezones:
-        return zone
+    try:
+        tz = pytz.timezone(zone)
+    except pytz.exceptions.UnknownTimeZoneError:
+        raise ValueError('Invalid time zone.')
     else:
```
Exirel: You don't need the else here, because the `except` re-raise an exception.

sopel/tools/time.py
```diff
@@ -52,7 +45,7 @@ def validate_format(tformat):
         time = datetime.datetime.utcnow()
         time.strftime(tformat)
     except Exception:  # TODO: Be specific
```
Exirel: While we are here, I tried several format and the expected errors are ValueError (invalid format) and TypeError (not a string):

```suggestion
     except (ValueError, TypeError):
```

sopel/tools/time.py
```diff
@@ -93,7 +86,7 @@ def get_channel_timezone(db, channel):
 
 
 def get_timezone(db=None, config=None, zone=None, nick=None, channel=None):
-    """Find, and return, the approriate timezone
+    """Find, and return, the appropriate timezone
```
Exirel: I'm afraid to git blame this file but I'm pretty sure that's on me... 😁

dgw: Nope, not this time! It was Embo: a3e6653f2730f7b808f194e02f17c8f691f0955f

PR #1708, opened 2019-10-11
loader: is_triggerable rejects jobs

Exirel:
When `clean_callable` runs on a callable, it automatically adds an `event` attribute: until now, the `is_triggerable` would return `True` for that callable after that. The problem is that, upon reloading the bot (after a timeout, a KILL message, etc.), the `event` attribute would cause job callables to be wrongly counted as triggerable. Thus, they would be registered with a 'match-any' rule, and throw errors on every message thereafter.

This bug is the result of a 5+ year old behavior (adding default `event` to all callables), plus a few-week-old change that would allow `event` to be considered as triggerable by default, without `rule`. We caught it because of a lucky Sopel Rockstar whose bot running from `master` happened to receive a KILL message.

This bug was only in master and didn't affect user of any official release. I'd still suggest to break the FIFO order for that one.

### DIFF
--- a/sopel/loader.py
+++ b/sopel/loader.py
@@ -104,10 +104,37 @@ def clean_callable(func, config):
 
 
 def is_triggerable(obj):
-    return any(
-        hasattr(obj, attr)
-        for attr in ('rule', 'event', 'intents', 'commands',
-                     'nickname_commands'))
+    """Check if ``obj`` can handle the bot's triggers.
+
+    :param obj: any :term:`function` to check
+    :return: ``True`` if ``obj`` can handle the bot's triggers
+
+    A triggerable is a callable that will be used by the bot to handle a
+    particular trigger (i.e. an IRC message): it can be a regex rule, an event,
+    an intent, a command, or a nickname command. However, it must not be a job
+    or a URL callback.
+
+    .. seealso::
+
+        Many of the decorators defined in :mod:`sopel.module` make the
+        decorated function a triggerable object.
+    """
+    forbidden_attrs = (
+        'interval',
+        'url_regex',
+    )
+    forbidden = any(hasattr(obj, attr) for attr in forbidden_attrs)
+
+    allowed_attrs = (
+        'rule',
+        'event',
+        'intents',
+        'commands',
+        'nickname_commands',
+    )
+    allowed = any(hasattr(obj, attr) for attr in allowed_attrs)
+
+    return allowed and not forbidden
 
 
 def clean_module(module, config):
--- a/test/test_loader.py
+++ b/test/test_loader.py
@@ -74,14 +74,34 @@ def tmpconfig(tmpdir):
     return config.Config(conf_file.strpath)
 
 
-def test_clean_module_commands(tmpdir, tmpconfig):
+@pytest.fixture
+def testplugin(tmpdir):
     root = tmpdir.mkdir('loader_mods')
     mod_file = root.join('file_mod.py')
     mod_file.write(MOCK_MODULE_CONTENT)
 
-    plugin = plugins.handlers.PyFilePlugin(mod_file.strpath)
-    plugin.load()
-    test_mod = plugin._module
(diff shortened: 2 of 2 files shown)


sopel/loader.py
```diff
@@ -104,10 +104,35 @@ def clean_callable(func, config):
...
-        for attr in ('rule', 'event', 'intents', 'commands',
-                     'nickname_commands'))
+    """Check if ``obj`` can handle the bot's triggers.
+
+    :param obj: any :term:`function` to check.
+    :return: ``True`` if ``obj`` can handle the bot's triggers.
+
+    A triggerable is a callable that will be used by the bot to handle a
+    particular trigger (i.e. an IRC message): it can be a regex rule, an event,
+    an intent, a command or nickname command. However, it must not be a job
```
dgw:
```suggestion
    an intent, a command, or a nickname command. However, it must not be a job
```

Note to self: Will require another rebase of #1660.

sopel/loader.py
```diff
@@ -104,10 +104,35 @@ def clean_callable(func, config):
...
+    or an URL callback.
```
dgw:
```suggestion
    or a URL callback.
```

sopel/loader.py
```diff
@@ -104,10 +104,35 @@ def clean_callable(func, config):
...
+
+    .. seealso::
+
+        The :mod:`sopel.module` defines decorators to make a function a
```
dgw:
```suggestion
        Many of the decorators defined in :mod:`sopel.module` make the decorated function a
```
(+ wrap)

Ultimately this is just to get away from saying "The sopel.module defines", because that shouldn't start with "The", but I got carried away making the whole sentence sound nicer. 😸

sopel/loader.py
```diff
@@ -104,10 +104,35 @@ def clean_callable(func, config):
...
+    )
+    must_not = not any(hasattr(obj, attr) for attr in forbidden)
+
+    allowed = (
+        'rule',
+        'event',
+        'intents',
+        'commands',
+        'nickname_commands',
+    )
```
dgw: Note to self: Will require another rebase and update of #1660.

sopel/loader.py
```diff
@@ -104,10 +104,35 @@ def clean_callable(func, config):
...
+    return must_not and any(hasattr(obj, attr) for attr in allowed)
```
dgw: I think it would be clearer if the above could be refactored so `forbidden` and `allowed` were the names of variables used in the return here, so this could become:

```suggestion
    return allowed and not forbidden
```

Mostly, I just don't like how the var name `must_not` reads here. 😁

sopel/loader.py
```diff
@@ -104,10 +104,35 @@ def clean_callable(func, config):
...
 
 def is_triggerable(obj):
-    return any(
-        hasattr(obj, attr)
-        for attr in ('rule', 'event', 'intents', 'commands',
-                     'nickname_commands'))
+    """Check if ``obj`` can handle the bot's triggers.
+
+    :param obj: any :term:`function` to check.
+    :return: ``True`` if ``obj`` can handle the bot's triggers.
```
dgw: I _think_ we've been omitting trailing `.` on param/return lines.

Exirel:
```suggestion
    :return: ``True`` if ``obj`` can handle the bot's triggers
```

sopel/loader.py
```diff
@@ -104,10 +104,35 @@ def clean_callable(func, config):
 
 
 def is_triggerable(obj):
-    return any(
-        hasattr(obj, attr)
-        for attr in ('rule', 'event', 'intents', 'commands',
-                     'nickname_commands'))
+    """Check if ``obj`` can handle the bot's triggers.
+
+    :param obj: any :term:`function` to check.
```
Exirel:
```suggestion
    :param obj: any :term:`function` to check
```

sopel/loader.py
Exirel: I'll find a way!

PR #1709, opened 2019-10-13
loader: don't add defaults to non-triggerables in clean_callable

dgw:
It turns out there was also a mistake in the `clean_callable_defaults` test (checking `func.rate` again instead of `func.channel_rate` the second time).

This feels messy because of all the duplicated assertions, but I'm too dumb (today?) to think up a better way. Hoping that @Exirel (who is _far_ more familiar with `pytest`) might have a suggestion or two.

(I'm half expecting to completely overwrite this PR at a later date with some much-improved alternative implementation, honestly.)

### DIFF
--- a/sopel/loader.py
+++ b/sopel/loader.py
@@ -47,10 +47,17 @@ def clean_callable(func, config):
     doc = trim_docstring(func.__doc__)
     examples = []
 
+    func.thread = getattr(func, 'thread', True)
+
+    if not is_triggerable(func):
+        # Rate-limiting, priority, etc. doesn't apply to non-triggerable functions.
+        # Adding the default attributes below is a waste of memory, as well as
+        # potentially confusing to other code.
+        return
+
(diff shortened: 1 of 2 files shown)


Exirel:
> This feels messy because of all the duplicated assertions, but I'm too dumb (today?) to think up a better way. Hoping that @Exirel (who is _far_ more familiar with `pytest`) might have a suggestion or two.

No, it's fine: tests *are* dumb. They are required to be dumb in order to do as little as possible. An assertion checks one and only one thing. That's also why the more state you have, the harder it is to test, because you have to repeat so many "dumb" assertions.

It's ok. I'm glad you did it!

dgw:
> I haven't run a bot with that to see if it doesn't have side-effects.

FWIW I did, but it was a pretty short test. If this breaks anything, I'm sure we'll hear from @half-duplex about five minutes after it's merged. 😝

PR #1710, opened 2019-10-14
calc: extract .py command to its own file

dgw:
This is mostly so it's easier to disable `.py` (which relies on an external web service and may encourage skiddies) while keeping the basic calculation command available.

### DIFF
--- a/sopel/modules/calc.py
+++ b/sopel/modules/calc.py
@@ -8,19 +8,8 @@ https://sopel.chat
 """
 from __future__ import unicode_literals, absolute_import, print_function, division
 
-import sys
-
-from requests import get
-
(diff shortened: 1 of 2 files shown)


dgw:
> dgw force-pushed the `py-cmd-own-file` branch from 98c3ab5 to c07bf12 20 seconds ago

\*grumbles about text editors making stupid default line-ending choices for new files\*

Exirel:
@dgw I had a very quick look at the conflict, and I think it has something to do with the quote/unquote thing you refactored few weeks ago. I'm not sure, but I believe you will need to rework a bit that part.

dgw:
Another fairly easy rebase. 😁

PR #1712, opened 2019-10-17
py: configurable endpoint

dgw:
Adds an option to configure the [Oblique](https://github.com/sopel-irc/oblique) endpoint Sopel should use for `.py` commands.

Based on #1710 branch, so will begin its life as a draft.

Note to self: Rescue the good bits from your last time implementing this (before deciding to split `.py` into its own file). See 0d4fc85a9d98327b6844565373b01a26ce43ca16, e16ae1fb01779f9807ebd0d2e38bc405fb74cb59, & abda3a61549f251b86a5b72d44b8709e8f156c10 from the forgotten `calc-config` branch.

### DIFF
--- a/sopel/modules/py.py
+++ b/sopel/modules/py.py
@@ -11,10 +11,40 @@ from __future__ import unicode_literals, absolute_import, print_function, divisi
 from requests import get
 
 from sopel import module
+from sopel.config.types import StaticSection, ValidatedAttribute
 from sopel.tools.web import quote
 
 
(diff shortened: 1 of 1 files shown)


dgw:
> (even if I still hate the `.py` command)

You're not alone. 😹 Script kiddies finding ways to corrupt the App Engine instance state are a pain, but we only need to get through 7.x. Then this will become an external package for 8.0, and no more random CI failures because App Engine's state got corrupted!

PR #1713, opened 2019-10-18
search: don't unquote DDG result

dgw:
Bing doesn't need it. Unquoting breaks URLs that contain spaces.

#1265 and associated changes later in the 6.x cycle introduced this unquoting, without any clear reason why. It doesn't appear to fix anything (not that my suite of test queries is _that_ exhaustive).

As a bonus, we shed the entire block of py2/py3 shimming for `unquote`.

Also of note: This reduces the core plugin set's need for #1681, but it'll still be good to have.

### DIFF
--- a/sopel/modules/search.py
+++ b/sopel/modules/search.py
@@ -59,8 +59,7 @@ def duck_search(query):
         bytes = bytes.split('web-result')[1]
     m = r_duck.search(bytes)
     if m:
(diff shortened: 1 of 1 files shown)


Exirel:
Another conflict! I believe this is also related to quote/unquote refactor, and nothing to worry about that much.

dgw:
Easy rebase. 😸

PR #1714, opened 2019-10-21
config: add a default `logging_format`

dgw:
Without this option specified, we get the Python default format, which is (among other problems) completely devoid of timestamp information.

This format, devised by @Exirel while refactoring the logging mechanism, outputs the timestamp, the package that generated the log line, the log level of the line, and (finally) the actual message.

We'll go from:

    Registered 42 plugins, 0 failed, 0 disabled

to:

    [2019-10-21 12:47:44,272] sopel.bot            INFO     - Registered 42 plugins, 0 failed, 0 disabled

Once this hits the stable release, it will become quite a bit easier to follow along with bug reports!

### DIFF
--- a/sopel/config/core_section.py
+++ b/sopel/config/core_section.py
@@ -254,10 +254,19 @@ class CoreSection(StaticSection):
         default='[%(asctime)s] %(name)-20s %(levelname)-8s - %(message)s')
     """The logging format string to use for logs.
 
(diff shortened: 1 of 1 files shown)


Exirel:
You might want to add that information to the documentation page!

dgw:
Updating with more documentation. Rendered preview:

PR #1716, opened 2019-10-22
test: URL callbacks (register, search, unregister)

Exirel:
This will probably generate a very small merge conflict with #1697 in a way that is not a problem, since it adds tests and an import.

My main concern at this point is: will it work as intended with Python 2.7?

Also: yay, tests! :unicorn:


Exirel:
> My main concern at this point is: will it work as intended with Python 2.7?


And they do!

Exirel:
Oh, no merge conflict. That's nice.

PR #1723, opened 2019-10-26
irc.backends: fix behavior of `core.verify_ssl` setting

dgw:
Actual behavior was reversed from before: `verify_ssl = True` ignored things like self-signed certs, and `verify_ssl = False` threw errors for self-signed certs instead.

Requesting review directly from @Exirel because he literally wrote the entire affected file.

### DIFF
--- a/sopel/irc/backends.py
+++ b/sopel/irc/backends.py
@@ -175,7 +175,7 @@ class SSLAsynchatBackend(AsynchatBackend):
 
     def handle_connect(self):
         # handle potential TLS connection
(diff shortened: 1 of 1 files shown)


dgw:
Don't look at the force-pushed diff; it'll just confuse you. The actual patch is a one-line change, matching the new condition with its ancestor: https://github.com/sopel-irc/sopel/blob/40162a46ad20867a50162a25708248da3083d180/sopel/irc.py#L240

PR #1726, opened 2019-10-26
tell: add relative timestamp format

cottongin:
This adds the option for switching between relative and absolute timestamp formats for messages sent via `.tell` or `.ask` and also cleans up the output of each.

Users can set their own preference via `.settellformat`, channel operators can set a channel preference via `.setchanneltellformat`, and the bot admin can set a global preference via the configuration file / option.

Demo:
```
16:30:26 <@cottongin> `tell nottongin this is a tell message
16:30:26 <bot> cottongin: I'll tell nottongin that when they're around.
16:30:33 <@cottongin> `ask nottongin did the tell message work
16:30:33 <bot> cottongin: I'll ask nottongin that when they're around.
16:30:45 <nottongin> .
16:30:45 <bot> nottongin: 2019-10-26 - 21:30:26UTC <cottongin> this is a tell message
16:30:45 <bot> nottongin: 2019-10-26 - 21:30:33UTC cottongin asked did the tell message work
16:31:08 <nottongin> `tell cottongin yes the tell message worked
16:31:08 <bot> nottongin: I'll tell cottongin that when they're around.
16:31:21 <nottongin> bot: ask cottongin if that worked
16:31:21 <bot> nottongin: I'll ask cottongin that when they're around.
16:31:39 <@cottongin> .
16:31:39 <bot> cottongin: <nottongin> yes the tell message worked (sent 31 seconds ago)
16:31:39 <bot> cottongin: nottongin asked if that worked (sent 18 seconds ago)
```

### DIFF
--- a/sopel/modules/tell.py
+++ b/sopel/modules/tell.py
@@ -8,14 +8,16 @@ https://sopel.chat
 """
 from __future__ import unicode_literals, absolute_import, print_function, division
 
+import datetime
 import os
 import time
 import threading
 import sys
 
-from sopel.module import commands, nickname_commands, rule, priority, example
+from sopel.module import commands, nickname_commands, rule, priority, example, require_privilege, OP
+from sopel.config.types import StaticSection, ValidatedAttribute
 from sopel.tools import Identifier, iterkeys
-from sopel.tools.time import get_timezone, format_time
+from sopel.tools.time import get_timezone, format_time, seconds_to_human
 
 
 MAXIMUM = 4
@@ -32,10 +34,10 @@ def loadReminders(fn, lock):
                 line = line.decode('utf-8')
             if line:
                 try:
-                    tellee, teller, verb, timenow, msg = line.split('\t', 4)
+                    tellee, teller, verb, timenow, timestamp, msg = line.split('\t', 5)
                 except ValueError:
                     continue  # @@ hmm
-                result.setdefault(tellee, []).append((teller, verb, timenow, msg))
+                result.setdefault(tellee, []).append((teller, verb, timenow, timestamp, msg))
         f.close()
     finally:
         lock.release()
@@ -65,7 +67,26 @@ def dumpReminders(fn, data, lock):
     return True
 
 
+class FormatSection(StaticSection):
+    tell_format = ValidatedAttribute(
+        'tell_format',
+        default='relative')
+    """Show relative timestamps by default"""
+
+
+def configure(config):
+    """
+    | name        | example  | purpose                                       |
+    | ----------- | -------- | --------------------------------------------- |
+    | tell_format | absolute | Show absolute timestamp; defaults to relative |
+    """
+    config.define_section('tell', FormatSection)
+    config.tell.configure_setting(
+        'tell_format', 'Default timestamp format for messages sent later via tell/ask')
+
+
 def setup(bot):
+    bot.config.define_section('tell', FormatSection)
     fn = bot.nick + '-' + bot.config.core.host + '.tell.db'
     bot.tell_filename = os.path.join(bot.config.core.homedir, fn)
     if not os.path.exists(bot.tell_filename):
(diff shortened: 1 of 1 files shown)


sopel/modules/tell.py
```diff
@@ -8,14 +8,16 @@
...
 from __future__ import unicode_literals, absolute_import, print_function, division
 
+import datetime
 import os
 import time
 import threading
 import sys
 
-from sopel.module import commands, nickname_commands, rule, priority, example
+from sopel.module import commands, nickname_commands, rule, priority, example, require_privilege, OP
```
dgw: Honestly, at this point, might as well switch to `from sopel import module` and update all the usage below. This list gets absurdly long if a plugin uses even a third of the whole plugin API. 😅

sopel/modules/tell.py
```diff
@@ -122,16 +143,18 @@ def f_remind(bot, trigger):
...
+        timestamp = datetime.datetime.timestamp(datetime.datetime.now())
+        timestamp = str(int(timestamp))
         bot.memory['tell_lock'].acquire()
         try:
             if tellee not in bot.memory['reminders']:
-                bot.memory['reminders'][tellee] = [(teller, verb, timenow, msg)]
+                bot.memory['reminders'][tellee] = [(teller, verb, timenow, timestamp, msg)]
             else:
-                bot.memory['reminders'][tellee].append((teller, verb, timenow, msg))
+                bot.memory['reminders'][tellee].append((teller, verb, timenow, timestamp, msg))
```
dgw: Refactoring opportunity: Use the `if thing not in memory: initialize it` pattern to avoid duplicating the tuple format. Pseudocode:

```
if tellee not in reminder-memory:
    reminder-memory[tellee] = []
reminder-memory[tellee].append((reminder, data, tuple))
```

sopel/modules/tell.py
```diff
@@ -65,7 +67,26 @@ def dumpReminders(fn, data, lock):
     return True
 
 
+class FormatSection(StaticSection):
+    tell_format = ValidatedAttribute(
```
dgw: This should be a `ChoiceAttribute`.

sopel/modules/tell.py
```diff
@@ -65,7 +67,26 @@ def dumpReminders(fn, data, lock):
...
+        'tell_format',
+        default='relative')
+    """Show relative timestamps by default"""
```
dgw:
```suggestion
    """Which type of timestamp to use when delivering messages."""
```

I don't necessarily mean for this to be the exact wording, but something closer to this or to the prompt used in `configure_setting()` below would be much better.

sopel/modules/tell.py
```diff
@@ -65,7 +67,26 @@ def dumpReminders(fn, data, lock):
...
+
+
+def configure(config):
+    """
+    | name        | example  | purpose                                       |
+    | ----------- | -------- | --------------------------------------------- |
+    | tell_format | absolute | Show absolute timestamp; defaults to relative |
```
dgw: Adapt the `purpose` column here based on whatever changes you settle on above.

sopel/modules/tell.py
```diff
@@ -174,7 +220,10 @@ def message(bot, trigger):
         return
 
     reminders = []
-    remkeys = list(reversed(sorted(bot.memory['reminders'].keys())))
+    try:
+        remkeys = list(reversed(sorted(bot.memory['reminders'].keys())))
+    except KeyError:
+        return
```
dgw: Why would this ever throw `KeyError`?

sopel/modules/tell.py
```diff
@@ -32,10 +34,10 @@ def loadReminders(fn, lock):
                 line = line.decode('utf-8')
             if line:
                 try:
-                    tellee, teller, verb, timenow, msg = line.split('\t', 4)
+                    tellee, teller, verb, timenow, timestamp, msg = line.split('\t', 5)
                 except ValueError:
                     continue  # @@ hmm
-                result.setdefault(tellee, []).append((teller, verb, timenow, msg))
+                result.setdefault(tellee, []).append((teller, verb, timenow, timestamp, msg))
```
dgw: I'm very curious what you expect to happen with old `tell` files that contain only 4 columns of data.

Aside from backward-compatibility with existing Sopel installations, let's talk about both `timenow` _and_ `timestamp` exist, since they're basically the same data in different formats.

`timenow` is defined during reminder creation as a formatted string based on the current `datetime.datetime.utcnow()` (which will be used [if no `time` is passed](https://github.com/sopel-irc/sopel/blob/ff09966ebf2d49197fd65aa55f105a38196522a2/sopel/tools/time.py#L174-L175) to `tools.time.format_time()`, as below).

`timestamp`, meanwhile, is a `datetime.datetime.timestamp()` created from `datetime.datetime.now()` (note: not `utcnow()`), cast to an int, then a string.

I honestly don't understand why there need to be two timestamps stored with the reminder, when a single UTC timestamp would suffice (and could be converted to any desired relative or absolute output format, in whatever timezone is desired, later). And the apparent lack of support for existing `tell` "database" files bothers me a lot.

dgw:
Per IRC discussion, let's put this on the back burner until after 7.0 is done.

kwaaak:
Maybe you can get around the `datetime` import by using [`trigger.time`](https://github.com/sopel-irc/sopel/blob/master/sopel/trigger.py#L123)

Very pleased to see that someone tackled this! I have thought about this functionality but not of a way to do it with backwards compatibility.

dgw:
Re-evaluating this as we gear up for 7.1, I don't think we want to get into it. The original author has vanished, and the demand for this feature hasn't come up from anyone else. Add to that all the concerns (noted in my review above) about backward compatibility/migration, and storage format.

PR #1727, opened 2019-10-26
module: fix `__all__` nonsense & docs for `event`

dgw:
Apparently `event` was left out of the `__all__` list. I question whether we should keep it, since it exports literally every name in the module anyway, but that's for another time.

Right now, the important thing is that `event` is now exported properly, and the docstring has been updated to account for the fact that a default `rule` is now set if an event callable does not have one.

How this didn't come up before, I have no idea. Being left out of `__all__` means that `event` wasn't included in the Sphinx docs, and things "should" break when trying to use e.g. `from sopel import module; @module.event('FOO')`… But Python's import machinery is fickle and likes to fill in module dictionaries based on other imports in the program. (Can't quickly find a reference page about this for linking, but I remember reading about it when the transition to using `__all__` came up.)

### DIFF
--- a/sopel/module.py
+++ b/sopel/module.py
@@ -18,6 +18,7 @@ __all__ = [
     'action_commands',
     'commands',
     'echo',
+    'event',
     'example',
     'intent',
     'interval',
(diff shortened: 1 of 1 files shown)


dgw:
Merging this "out-of-order" for selfish reasons, namely that this bug breaks existing plugin code on my own Sopel instance and I'd rather fix the bug than refactor the code so it doesn't import `*`.

PR #1729, opened 2019-10-29
db: add pool_recycle to db creation

RustyBower:
This should fix the issue that some users are having with DB timeouts when there is no activity for 8 hours.


deathbybandaid:
I've experienced the need for this using mysql,, fixing it here is a much better solution than modifying the timeout settings via mysql config.

`/etc/mysql/mysql.conf.d/mysqld.cnf`      `[mysqld] wait_timeout    = 50000`

PR #1730, opened 2019-10-29
cleanup: remove things marked for version 7.0

deathbybandaid:
I decided to do a search of the code for things marked to remove for version 7.0

I don't see anywhere else in the repo where this .stats dictionary is used, so it may be as simple as killing this single line?

I also intend on going through the code to see if there are other things that should have been removed in prior releases (I doubt it though)

### DIFF
--- a/sopel/bot.py
+++ b/sopel/bot.py
@@ -69,7 +69,6 @@ class Sopel(irc.AbstractBot):
         self._command_groups = collections.defaultdict(list)
         """A mapping of module names to a list of commands in it."""
 
-        self.stats = {}  # deprecated, remove in 7.0
         self._times = {}
         """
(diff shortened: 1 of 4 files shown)


deathbybandaid:
Okay, so some digging here:

https://github.com/sopel-irc/sopel/commit/526bdc62a771f3b0bb313cd6f47afdc63fdbfe33#diff-5ca66e1c69a6af630a3309cc3385d2d8L75-L81

@Exirel did some really cool stuff here, the above lines in the diff suggest to me that these following lines in the tests need to be removed as well

https://github.com/sopel-irc/sopel/blob/e3ccd19f945b50c0758e48b01599fec76839c150/sopel/test_tools.py#L100-L102

PR #1728, opened 2019-10-26
bot: fix race-condition in dispatch

Exirel:
The scenario is:

* dispatch is called for a pretrigger
* it looks into the list of callable by priority
* it triggers one of the first callables, which wants to remove a callable of the same priority
* because Python must not allow a dict to be modified while it is iterated over, Python raises an exception

If we put a lock here, Python won't raise the exception... but we'll get a deadlock. To prevent that, instead, we copy the list of callables, and return the ones that are triggered. If any of them wants to change the callables of the same priority, they won't raise an error, because we are not iterating anymore on the dict.

Hoo-ray!

@dgw have fun with my miserable grammar at 2am.

Issue #1355, opened 2018-07-17
[Proposal] Better ignore system


dgw:
I noticed this weekend that `.blocks add hostmask <regex>` actually adds a host block, and doesn't work as expected if given an actual hostmask, e.g. `*!*@bad-domain.tld`. Ignoring the fact that I gave an invalid regular expression (starts with a quantifier), `.*!.*@bad-domain\.tld` won't work as expected either.

What `.blocks add hostmask` actually does is add to the `host_blocks`, which does exactly what it says on the tin: It blocks based on the line originator's host. It's a bit confusing, due to the command naming.

The simple solution would be to rename the `.blocks ___ hostmask` commands to `.blocks ___ host`, but I'd like to do more than that:

  - [ ] Rename `.blocks ___ hostmask` commands to `.blocks ___ host`
  - [ ] Implement a real `.blocks ___ hostmask` command set that actually works on hostmasks
  - [ ] Validate ignore masks before adding them (at present, adding an invalid entry can crash Sopel, but without saving the bad entry to the config file; implementing #1354's config auto-save would make bad ignore entries a serious pitfall for new and fat-fingered bot owners alike)
  - [ ] Make regex syntax optional for ignore-list entries, e.g. `/nick-?regex/` vs. `nickname` (`/` will never appear in a valid hostmask, hostname, or nickname, so it's a logical choice)
    - [ ] Ideally, the new, default, non-regex syntax would support conventional wildcards identical to how ban masks are interpreted by IRC servers (`*` and `?`)

The first three items definitely can be done in Sopel 6.x, as the only users who can actually use the ignore-system commands are bot owners/admins who read the changelogs before upgrading. (You _do_ read changelogs before upgrading, _riiight?_ :stuck_out_tongue:). Validation would be a nice non-breaking enhancement, too.

I'm less inclined to change how config file entries are interpreted in a point update, so making regex syntax optional might be a Sopel 7 thing. But if there's a clean way to just quietly tweak existing configs on first start and make all existing ignore entries into the regex variety (however that's represented in the file), it doesn't _have_ to wait. The biggest barrier to that is that Sopel's config file isn't versioned—but determining the config format version is easy if it doesn't contain a version! :grin:

Self-assigned this because I already started thinking about the implementation, but no code has been committed Yet™. Therefore, it's also Patches Welcome until I actually prototype (some of) the changes.

alanhuang122:
[...]

PR #1732, opened 2019-11-02
bot: keep track of trigger processing threads

Exirel:
Based on #1728 to avoid future merge conflict.

My initial idea was to join trigger processing threads directly in the `dispatch` method. For several reasons (including the echo-message polyfill), that's not doable.

Instead, since all I want is a way to know, access, and join processing threads in automated tests, I make sure the bot keeps track of its threads. This way, I'll be able to run a dispatch, going through all the bot's internals and machineries, and then wait for the end result in my tests.

It was a bit convoluted, but in the end I think it turned out great!

Quick note about memory consumption: this information generate a very low overhead, since running threads are kept in memory anyway by Python. All that is stored is a reference to the thread object, and every time dispatch is called, the list is flushed out of obsolete running threads.

### DIFF
--- a/sopel/bot.py
+++ b/sopel/bot.py
@@ -44,6 +44,8 @@ class Sopel(irc.AbstractBot):
         super(Sopel, self).__init__(config)
         self._daemon = daemon  # Used for iPython. TODO something saner here
         self.wantsrestart = False
+        self._running_triggers = []
+        self._running_triggers_lock = threading.Lock()
 
         # `re.compile('.*') is re.compile('.*')` because of caching, so we need
         # to associate a list with each regex, since they are unexpectedly
@@ -650,6 +652,8 @@ class Sopel(irc.AbstractBot):
             :meth:`~get_triggered_callables`. This method is also responsible
             for telling ``dispatch`` if the function is blocked or not.
         """
+        # list of commands running in separate threads for this dispatch
+        running_triggers = []
         # nickname/hostname blocking
         nick_blocked = host_blocked = self._is_pretrigger_blocked(pretrigger)
         blocked = bool(nick_blocked or host_blocked)
@@ -659,11 +663,9 @@ class Sopel(irc.AbstractBot):
         for priority in ('high', 'medium', 'low'):
             items = self.get_triggered_callables(priority, pretrigger, blocked)
             for func, trigger, is_blocked in items:
+                function_name = "%s.%s" % (func.__module__, func.__name__)
                 # skip running blocked functions, but track them for logging
                 if is_blocked:
-                    function_name = "%s.%s" % (
-                        func.__module__, func.__name__
-                    )
                     list_of_blocked_functions.append(function_name)
                     continue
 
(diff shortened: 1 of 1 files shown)


sopel/bot.py
```diff
@@ -660,37 +664,72 @@ def dispatch(self, pretrigger):
...
             )
 
+        # update currently running triggers
+        self._update_running_triggers(running_triggers)
+
+    @property
+    def running_triggers(self):
+        """Current active threads for triggers.
+
+        This read-only list contains the currently running thread processing
```
dgw: This being "read-only", shouldn't it be named with a leading `_` by convention? The actual attribute can be named with leading `__`, which prevents directly accessing it from outside this object via name mangling.

sopel/bot.py
```diff
@@ -660,37 +664,72 @@ def dispatch(self, pretrigger):
...
+        with self._running_triggers_lock:
+            return [t for t in self._running_triggers if t.is_alive()]
+
+    def _update_running_triggers(self, running_triggers):
+        """Update list of running triggers.
+
+        :param list running_triggers: new started threads
+
+        We want to keep track of running triggers, mostly for testing and
+        debugging purpose. For instance, it'll help make sure, in test, that
```
dgw: "purposes", "tests"

Because I'm reviewing this as a subset, GitHub won't let me make proper suggestions. Sorry! Though the reason I'm reviewing as a subset is actually your fault. 😛

sopel/bot.py
```diff
@@ -660,37 +664,72 @@ def dispatch(self, pretrigger):
...
+        a bot's plugin finished processing of a trigger, by manually joining
+        all running threads.
+
+        This is kept private, as this is pure internal machinery and isn't made
```
dgw: s/made/meant/

See above re: suggestions.

sopel/bot.py
```diff
@@ -660,37 +664,72 @@ def dispatch(self, pretrigger):
...
+    def running_triggers(self):
+        """Current active threads for triggers.
+
+        This read-only list contains the currently running thread processing
+        a trigger. This is for testing and debugging purpose only.
+        """
+        with self._running_triggers_lock:
+            return [t for t in self._running_triggers if t.is_alive()]
+
+    def _update_running_triggers(self, running_triggers):
```
dgw: This, too, could be named `__update_running_triggers` and thus completely inaccessible from outside this object, unless you need it to be accessible.

sopel/bot.py
Exirel: Prefix `_` is the Python convention for the "protected" attribute from OOP, which doesn't stand for read-only. These are two different concepts.

sopel/bot.py
Exirel: Another trick of OOP & Python convention: `_` is for protected which means access is restricted to the class and its subclasses, while `__` is for private which means access is restricted to the class but **not** its subclasses - which makes it very inconvenient.

Exirel:
I rebased to fix the conflict, but I'll wait until #1728 is merged before fixing the remaining issues.

sopel/bot.py
```diff
@@ -659,38 +663,73 @@ def dispatch(self, pretrigger):
...
                     continue
 
                 # call triggered function
                 wrapper = SopelWrapper(
                     self, trigger, output_prefix=func.output_prefix)
                 if func.thread:
+                    # run in a separate thread
                     targs = (func, wrapper, trigger)
                     t = threading.Thread(target=self.call, args=targs)
+                    t.name = '%s-%s' % (t.name, function_name)
```
Exirel: Just a quick note on that line: this will be of tremendous help when debugging threads, because now we'll be able to print the current thread's name, and it will contain the function triggered!

Or even, if someone is fond of debug tools/memory inspection/threading inspection, they'll be able to know exactly what's going on at any given time.

This is just the little thing that help with observability.

sopel/bot.py
```diff
@@ -659,38 +663,73 @@ def dispatch(self, pretrigger):
...
             if nick_blocked and host_blocked:
                 block_type = 'both'
             elif nick_blocked:
                 block_type = 'nick'
             else:
                 block_type = 'host'
-            LOGGER.info(
-                "[%s]%s prevented from using %s.",
+            LOGGER.debug(
+                "[%s] %s prevented from using %s.",
```
dgw: This honestly confused the #*!$ out of me when I was testing something else earlier, so a PR where the line is already touched is a perfect place to suggest a different way of putting it.

```suggestion
                "%s prevented from using %s by %s.",
```

Unfortunately this isn't useful as a suggestion because it'll require changes to several surrounding lines:

* The if/elif/else of block types needs to change the strings:
  * `block_type = 'both blocklists'`
  * `block_type = 'nick blocklist'`
  * `block_type = 'host blocklist'`
* The order of parameters passed to the function needs to be updated below

I don't think this will make the eventual ignore system improvements (#1355) any harder to implement.

sopel/bot.py
```diff
@@ -659,38 +663,73 @@ def dispatch(self, pretrigger):
...
 
+        # update currently running triggers
+        self._update_running_triggers(running_triggers)
+
+    @property
+    def running_triggers(self):
+        """Current active threads for triggers.
+
+        This read-only list contains the currently running thread processing
+        a trigger. This is for testing and debugging purpose only.
```
```suggestion
        a trigger. This is for testing and debugging purposes only.
```

Did I miss this in my last review? Damn.

sopel/bot.py
```diff
@@ -659,38 +663,73 @@ def dispatch(self, pretrigger):
...
+            return [t for t in self._running_triggers if t.is_alive()]
+
+    def _update_running_triggers(self, running_triggers):
+        """Update list of running triggers.
+
+        :param list running_triggers: new started threads
+
+        We want to keep track of running triggers, mostly for testing and
+        debugging purposes. For instance, it'll help make sure, in tests, that
+        a bot's plugin finished processing of a trigger, by manually joining
```
```suggestion
        a bot plugin has finished processing a trigger, by manually joining
```

sopel/bot.py
```diff
@@ -659,38 +663,73 @@ def dispatch(self, pretrigger):
...
+        all running threads.
+
+        This is kept private, as this is pure internal machinery and isn't
+        meant to be manipulated by an outside code.
```
```suggestion
        meant to be manipulated by outside code.
```

Did I miss this too? 🙄

sopel/bot.py
Exirel: That's all fair to me, will do. :)

sopel/bot.py
Exirel: I surely did.

Exirel:
All fixed (again). Waiting for a go to squash.

Exirel:
Done.

PR #1735, opened 2019-11-04
treewide: fix typos

torstehu:


deathbybandaid:
I'm not familiar with treewide, had to google it, but this looks fairly thorough.

Crazy how many typos it found though!

Exirel:
I love this. 👍 
Thank you very much.

sopel/tools/calculation.py
```diff
@@ -146,7 +146,7 @@ def guarded_pow(left, right):
     elif not isinstance(right, numbers.Integral):
         pass
     elif pow_complexity(left, right) < 0.5:
-        # Value 0.5 is arbitary and based on a estimated runtime of 0.5s
+        # Value 0.5 is arbitrary and based on a estimated runtime of 0.5s
```
dgw: Can we fix this too as long as the line is being touched anyway?

```suggestion
        # Value 0.5 is arbitrary and based on an estimated runtime of 0.5s
```

torstehu: Will fix

PR #1739, opened 2019-11-10
cli: Fix plugin config duplicate y/n prompt

half-duplex:
Was:
```
Would you like to see if there are any modules that need configuring (y/n)? [n] y
Configure safety.py - Alerts about malicious URLs (y/n)? [n] (y/n)? [n]  
```
Now:
```
Would you like to see if there are any modules that need configuring (y/n)? [n] y
Configure safety.py - Alerts about malicious URLs (y/n)? [n] 
```

### DIFF
--- a/sopel/cli/utils.py
+++ b/sopel/cli/utils.py
@@ -167,7 +167,7 @@ def _plugins_wizard(settings):
 
 def _plugin_wizard(settings, plugin):
     plugin.load()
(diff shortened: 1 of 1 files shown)


Exirel:
There are battles not worth winning.

dgw:
It's all fun and games until someone enables `flake8-quotes`. 😆

PR #1744, opened 2019-11-13
tools: Fix Identifier lowercase

### LINKED ISSUE #1389: Identifier lowercase is not right?
Poking around in `Identifier`'s class definition for other reasons, I saw that `Identifier._lower()` makes the following replacements after lowercasing the input:

| Character | Replacement |
|:---------:|:-----------:|
|    `{`    |     `[`     |
|    `}`    |     `]`     |
|    `\|`   |     `\\`    |
|    `^`    |     `~`     |

RFC 2812 [defines](https://tools.ietf.org/html/rfc2812#section-2.2) the characters in the left column as "lowercase" and the characters in the right column as "uppercase", though, so it seems `Identifier._lower()` has been doing case conversion of these symbols in the wrong direction for years. :joy_cat:

Now, I don't think that changing this will actually affect anything at runtime. `bot.users` (and friends) gets populated at connection time, and will just use whatever lowercase representation that's generated.

The most likely wrinkle is from the bot database, which stores both the `canonical` nickname (capitalized as seen on IRC) and a `slug` representation (in lowercase). The generated `slug` might change for some nicks in the database after updating to a version of Sopel that changes `Identifier._lower()`, and so we might need a database migration script to fix that up. It's likely that changing the lowercase representation of `Identifier`s would result in some users with an affected character in their nick losing their settings unless a migration is run. (A unique `slug`—the primary key for the `nicknames` table—would generate a new `nick_id`, which wouldn't have any matching rows in `nick_values`.)

Changing how `Identifier`s work could also potentially break a few third-party modules—especially if they store `Identifier` values as lowercase in a custom database table or external file. For that reason, even though I would really rather just fix this for **6.6.0**, I think the changelog will simply include a notice that this change is planned for Sopel 7 so end users and module developers have time to plan ahead. (See task list below for why the milestone is not currently set to **7.0.0**.)

----

Task List:

  - [x] Note in all future Sopel 6.x release notes that this change is planned for Sopel 7.
    - Update milestone as appropriate to make sure this doesn't get forgotten.
  - [x] Actually fix this bug in Sopel 7. (#1744)

dgw:
Resolves #1389.

@RustyBower do you have any thoughts on a better migration scheme? Doing it opportunistically, as needed, seems elegant—but I wonder if it might cause problems somewhere.

### DIFF
--- a/sopel/db.py
+++ b/sopel/db.py
@@ -400,9 +400,47 @@ class SopelDB(object):
 
     # CHANNEL FUNCTIONS
 
+    def get_channel_slug(self, chan):
+        """Return the case-normalized representation of ``channel``.
+
+        :param str channel: the channel name to normalize, with prefix
+                            (required)
+        :return str: the case-normalized channel name (or "slug"
+                     representation)
+
+        This is useful to make sure that a channel name is stored consistently
+        in both the bot's own database and third-party plugins'
+        databases/files, without regard for variation in case between
+        different clients and/or servers on the network.
+        """
+        chan = Identifier(chan)
+        slug = chan.lower()
+        session = self.ssession()
+        try:
+            count = session.query(ChannelValues) \
+                .filter(ChannelValues.channel == slug) \
+                .count()
+
+            if count == 0:
+                # see if it needs case-mapping migration
+                old_rows = session.query(ChannelValues) \
+                    .filter(ChannelValues.channel == Identifier._lower_swapped(chan))
+                old_count = old_rows.count()
+                if old_count > 0:
+                    # it does!
+                    old_rows.update({ChannelValues.channel: slug})
+                    session.commit()
+
+            return slug
+        except SQLAlchemyError:
+            session.rollback()
+            raise
+        finally:
+            self.ssession.remove()
+
     def set_channel_value(self, channel, key, value):
         """Sets the value for a given key to be associated with the channel."""
-        channel = Identifier(channel).lower()
+        channel = self.get_channel_slug(channel)
         value = json.dumps(value, ensure_ascii=False)
         session = self.ssession()
         try:
@@ -427,7 +465,7 @@ class SopelDB(object):
 
     def delete_channel_value(self, channel, key):
         """Deletes the value for a given key associated with a channel."""
-        channel = Identifier(channel).lower()
+        channel = self.get_channel_slug(channel)
         session = self.ssession()
         try:
             result = session.query(ChannelValues) \
@@ -446,7 +484,7 @@ class SopelDB(object):
 
     def get_channel_value(self, channel, key):
         """Retrieves the value for a given key associated with a channel."""
-        channel = Identifier(channel).lower()
+        channel = self.get_channel_slug(channel)
         session = self.ssession()
         try:
             result = session.query(ChannelValues) \
--- a/sopel/db.py
+++ b/sopel/db.py
@@ -206,6 +206,16 @@ class SopelDB(object):
                 .one_or_none()
 
             if nickname is None:
+                # see if it needs case-mapping migration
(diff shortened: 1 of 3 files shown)


dgw:
Was about to merge this, and realized we all forgot about something important—channel names: https://github.com/sopel-irc/sopel/blob/60f298b4f79c1acb0a282fb0f5a01ab8d08533ad/sopel/db.py#L397

The same pattern is used in the channel-value versions of get, set, and delete. Can't do it at this moment for want of time (have a flight in the morning), but the channel stuff probably needs an analogue for `get_nick_id()` so the same "try again with the old case mapping" logic needn't repeat thrice.

Thankfully, plugin values use a regular old string as the plugin name, so no migration necessary.

Unrelated to finishing this PR, but worth thinking about: honestly, we could consider further tweaks to how `Identifier` calculates its lowercase equivalent once ISUPPORT parsing is done. There's [a CASEMAPPING field](http://www.irc.org/tech_docs/005.html) that will tell us exactly which sets of characters the IRC server uses to compare nicks & channel names case-insensitively, if using that value seems wise. (There are edge cases to consider, like if a bot is moved after a network ceases operating and the new network has a different CASEMAPPING setting than the original…)

Exirel:
@dgw tests disagree with your code. :(

dgw:
@Exirel don't mind me, just learning SQLAlchemy by trial-and-error 😁

sopel/db.py
```diff
@@ -390,9 +400,47 @@ def merge_nick_groups(self, first_nick, second_nick):
...
+        """
+        chan = Identifier(chan)
+        slug = chan.lower()
+        session = self.ssession()
+        try:
+            count = session.query(ChannelValues) \
+                .filter(ChannelValues.channel == slug) \
+                .count()
+
+            if count == 0:
```
Exirel: I think a `# TODO: remove for Sopel 8` might be useful here, because I don't think we should handle this case this far in the future.

sopel/db.py
```diff
@@ -206,6 +206,16 @@ def get_nick_id(self, nick, create=True):
                 .one_or_none()
 
             if nickname is None:
+                # see if it needs case-mapping migration
```
Exirel: Same `# TODO: remove for Sopel 8.0` here.

sopel/db.py
dgw: I'd rather a follow-up issue, or adding a task to #1389.

Honestly, later in 7.x, I think we should just add a database field somewhere that stores a version number we can use to run migrations like this all at once, if needed. Removing these migration checks without first making sure that _all_ of a Sopel instance's values have been fixed could lead to some really bad debugging times—probably involving one or both of us scratching our heads, thinking "but the code doesn't work that way!" while an innocent user anxiously waits for help. 😅

dgw: (Alternatively, we can pause this change until 7.1, and do a more comprehensive "all at once" migration strategy in the first place.)

Exirel:
Commenting here instead because I think it's more global:

> (Alternatively, we can pause this change until 7.1, and do a more comprehensive "all at once" migration strategy in the first place. This is another thing that's been broken For So Long™ that another point release is not a huge deal.)

Yes! I think we can see into the Sopel 7.x's lifetime what kind of migrations we can use. I know alembic can do the trick for most cases, so this might be an option.

sopel/db.py
```diff
@@ -206,6 +206,16 @@ def get_nick_id(self, nick, create=True):
...
+                nickname = session.query(Nicknames) \
+                    .filter(Nicknames.slug == Identifier._lower_swapped(nick)) \
```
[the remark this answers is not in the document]
dgw: We figured this out on IRC. It was a weird GitHub display issue that confused poor Rusty. 😁

PR #1748, opened 2019-11-14
plugins: add licence and author

Exirel:
I forgot to add these the first time. There are probably more here and there, but these ones bothered me.


Exirel:
Fair enough! :+1:

Issue #1069, opened 2016-04-19
Large list of channels cause Quit due to "excess flood"


victorclf:
When joining a large number of channels (e.g. 40) in freenode, the bot is disconnected because of excess flood. This happens even when setting the 'throttle_join=1'.

It seems that the 'throttle_join' command does not help because it just limits the JOIN commands. After all channels are joined, sopel sends TOPIC and WHO commands for each channel joined which seem to cause the excess flood.

A workaround is to use a custom module that joins each channel after a few seconds. In this way, the TOPIC and WHO commands are called right after joining the channel instead of being sent in batch at the end.

victorclf:
Only after exiting 'coretasks.startup()', sopel processes the JOIN confirmations from the server and these confirmations trigger a flood of TOPIC and WHO commands. The 'time.sleep()' call for implementing 'throttle_join' seems to be problem.

elad661:
So this happens because we only rate limit on `.msg()` (which is PRIVMSG) and not on `.write()` which we use for everything else.

While the bot still needs a way to send arbitrary IRC commands without passing through the rate-limiter, it makes sense to have `.write()` rate-limited by default as well.

This is not just a matter of moving the current rate-limiter code to `.write()`, because it works on a per-destination (usually channel) basis, and not globally.

Exirel:
That's an interesting bug. A user recently come up on IRC asking for a solution.

At first, one could think the solution would be to flood-protect everything. However, that would be rather difficult, given the nature of the problem: everything is asynchronous.

The root cause of the problem is *not* in an excess of JOIN (even though it doesn't help), but whenever the bot received a JOIN event from the server, if it's for itself, it'll send a TOPIC request and a WHO request (see [track_join](https://github.com/sopel-irc/sopel/blob/v6.6.9/sopel/coretasks.py#L407-L431) in `sopel.coretasks`).

The `TOPIC` should probably be removed, because server are supposed to send a `RPL_TOPIC`, per [RFC 1459 Section 4.2.1](https://tools.ietf.org/html/rfc1459#section-4.2.1). This event is the exact same event as a result of a `TOPIC` command (see [RFC 1459 Section 4.2.4](https://tools.ietf.org/html/rfc1459#section-4.2.4)). Removing that command would greatly limit the flood.

[...]

PR #1749, opened 2019-11-14
core: do not send TOPIC on join

Exirel:
It's a very small change: as discussed in https://github.com/sopel-irc/sopel/issues/1069#issuecomment-548617872 the `TOPIC` command is not required since the server is supposed to send the required information when the bot joins a channel.

This won't fix the bigger problem, but it'll help in some cases.


dgw:
Note: #1751 keeps this TOPIC command, and will need to be updated after this is merged. (It'll probably merge-conflict anyway, to be fair.)

PR #1751, opened 2019-11-14
core: throttle JOIN and WHO commands

Exirel:
(based on #1750 for documentation) (may conflict with #1749)

This PR adds two things:

* `core.throttle_wait`: time waiting between JOIN/WHO batches
* a queue to process JOIN events by batches

This reuse the `_send_who` function which updates the `last_who` flag for the channels, so it fits nicely with the existing code.

That being said, I think I'll take a mental note to refactor the whole start-up sequence, because it's a nightmare to debug it.

### DIFF
--- a/sopel/coretasks.py
+++ b/sopel/coretasks.py
@@ -9,21 +9,27 @@ dispatch function in bot.py and making it easier to maintain.
 # (yanovich.net)
 # Copyright © 2012, Elad Alfassa <elad@fedoraproject.org>
 # Copyright 2012-2015, Elsie Powell embolalia.com
+# Copyright 2019, Florian Strzelecki <florian.strzelecki@gmail.com>
+#
 # Licensed under the Eiffel Forum License 2.
 from __future__ import unicode_literals, absolute_import, print_function, division
 
 import logging
 from random import randint
+import collections
 import datetime
 import re
 import sys
 import time
+
 import sopel
 import sopel.module
 import sopel.tools.web
+from sopel import loader
 from sopel.irc.utils import CapReq
 from sopel.tools import Identifier, iteritems, events
 from sopel.tools.target import User, Channel
+from sopel.tools.jobs import Job
 import base64
 
 if sys.version_info.major >= 3:
@@ -33,6 +39,46 @@ LOGGER = logging.getLogger(__name__)
 
 batched_caps = {}
 who_reqs = {}  # Keeps track of reqs coming from this module, rather than others
+
+
+def setup(bot):
+    bot.memory['join_events_queue'] = collections.deque()
+
+    # Manage JOIN flood protection
+    if bot.settings.core.throttle_join:
+        wait_interval = max(bot.settings.core.throttle_wait, 1)
+
+        @sopel.module.interval(wait_interval)
+        def processing_job(bot):
+            _join_event_processing(bot)
+
+        loader.clean_callable(processing_job, bot.settings)
+        bot.scheduler.add_job(Job(wait_interval, processing_job))
+
+
+def shutdown(bot):
+    try:
+        bot.memory['join_events_queue'].clear()
+    except KeyError:
+        pass
+
+
+def _join_event_processing(bot):
+    """Process a batch of JOIN event from the ``join_events_queue`` queue.
+
+    Every time this function is executed, it processes at most
+    ``throttle_join`` JOIN events. For each JOIN, it sends a WHO request to
+    know more about the channel. This will prevent an excess of flood when
+    there are too many channels to join at once.
+    """
+    batch_size = max(bot.settings.core.throttle_join, 1)
+    for _ in range(batch_size):
+        try:
+            channel = bot.memory['join_events_queue'].popleft()
+        except IndexError:
+            break
+        LOGGER.debug('Sending WHO after channel JOIN: %s', channel)
+        _send_who(bot, channel)
 
 
 def auth_after_register(bot):
@@ -83,6 +129,7 @@ def execute_perform(bot, trigger):
     _execute_perform(bot)
 
 
+@sopel.module.priority('high')
 @sopel.module.event(events.RPL_WELCOME, events.RPL_LUSERCLIENT)
 @sopel.module.thread(False)
 @sopel.module.unblockable
@@ -109,15 +156,33 @@ def startup(bot, trigger):
 
     bot.memory['retry_join'] = dict()
 
-    if bot.config.core.throttle_join:
+    channels = bot.config.core.channels
+    if not channels:
+        LOGGER.info('No initial channels to JOIN.')
+    elif bot.config.core.throttle_join:
         throttle_rate = int(bot.config.core.throttle_join)
+        throttle_wait = max(bot.config.core.throttle_wait, 1)
         channels_joined = 0
-        for channel in bot.config.core.channels:
+
+        LOGGER.info(
+            'Joining %d channels (with JOIN throttle ON); '
+            'this may take a moment.',
+            len(channels))
+
+        for channel in channels:
             channels_joined += 1
             if not channels_joined % throttle_rate:
-                time.sleep(1)
+                LOGGER.debug(
+                    'Waiting %ds before next JOIN batch.',
+                    throttle_wait)
+                time.sleep(throttle_wait)
             bot.join(channel)
     else:
+        LOGGER.info(
+            'Joining %d channels (with JOIN throttle OFF); '
+            'this may take a moment.',
+            len(channels))
+
         for channel in bot.config.core.channels:
             bot.join(channel)
 
@@ -158,11 +223,10 @@ def enable_service_auth(bot, trigger):
 @sopel.module.event(events.ERR_NOCHANMODES)
 @sopel.module.priority('high')
 def retry_join(bot, trigger):
-    """Give NickServer enough time to identify on a +R channel.
+    """Give NickServ enough time to identify on a +R channel.
 
     Give NickServ enough time to identify, and retry rejoining an
     identified-only (+R) channel. Maximum of ten rejoin attempts.
-
     """
     channel = trigger.args[1]
     if channel in bot.memory['retry_join'].keys():
@@ -290,6 +354,7 @@ def track_modes(bot, trigger):
         # Something fucky happening, like unusual batching of non-privilege
         # modes together with the ones we expect. Way easier to just re-WHO
         # than try to account for non-standard parameter-taking modes.
+        LOGGER.debug('Sending WHO for channel: %s', channel)
         _send_who(bot, channel)
         return
 
@@ -447,6 +512,7 @@ def _periodic_send_who(bot):
 
     if selected_channel is not None:
         # selected_channel's last who is either none or the oldest valid
+        LOGGER.debug('Sending WHO for channel: %s', selected_channel)
         _send_who(bot, selected_channel)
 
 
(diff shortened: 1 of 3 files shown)


Exirel:
Also this will help with #1069. It might even fix it.

sopel/coretasks.py
```diff
@@ -35,6 +41,46 @@
...
+    try:
+        bot.memory['join_events_queue'].clear()
+    except KeyError:
+        pass
+
+
+def _join_event_processing(bot):
+    """Process a batch of JOIN event from the ``join_events_queue`` queue.
+
+    Every time this function is executed, it process at most ``throttle_join``
```
dgw:
```suggestion
    Every time this function is executed, it processes at most ``throttle_join``
```

sopel/coretasks.py
```diff
@@ -35,6 +41,46 @@
...
+    JOIN event: for each, it sends a WHO request to know more about the
```
dgw:
```suggestion
    JOIN events. For each JOIN, it sends a WHO request to know more about the
```

sopel/coretasks.py
```diff
@@ -35,6 +41,46 @@
...
+    channel. This will prevent an excess of flood when there are too many
+    channels to join at once.
+    """
+    batch_size = max(bot.settings.core.throttle_join, 1)
+    for _ in range(batch_size):
+        try:
+            channel = bot.memory['join_events_queue'].popleft()
+        except IndexError:
+            break
+        LOGGER.debug('Send WHO after JOIN channel: %s', channel)
```
dgw:
```suggestion
        LOGGER.debug('Sending WHO after JOIN channel: %s', channel)
```

sopel/coretasks.py
```diff
@@ -90,11 +137,15 @@ def startup(bot, trigger):
 
     if bot.config.core.throttle_join:
         throttle_rate = int(bot.config.core.throttle_join)
+        throttle_wait = max(bot.config.core.throttle_wait, 1)
```
dgw: Do you intend to prevent someone from using `throttle_wait = 0` if they explicitly set it to that value?

sopel/coretasks.py
```diff
@@ -90,11 +137,15 @@ def startup(bot, trigger):
...
         channels_joined = 0
         for channel in bot.config.core.channels:
             channels_joined += 1
             if not channels_joined % throttle_rate:
-                time.sleep(1)
+                LOGGER.info(
```
dgw:
```suggestion
                LOGGER.debug(
```

sopel/coretasks.py
```diff
@@ -90,11 +137,15 @@ def startup(bot, trigger):
...
+                    'Wait %ds before next batch of JOIN.',
```
dgw:
```suggestion
                    'Waiting %ds before next JOIN batch.',
```

sopel/coretasks.py
```diff
@@ -267,6 +318,7 @@ def track_modes(bot, trigger):
         # Something fucky happening, like unusual batching of non-privilege
         # modes together with the ones we expect. Way easier to just re-WHO
         # than try to account for non-standard parameter-taking modes.
+        LOGGER.debug('Send WHO for channel: %s', channel)
```
dgw:
```suggestion
        LOGGER.debug('Sending WHO for channel: %s', channel)
```

sopel/coretasks.py
```diff
@@ -424,6 +476,7 @@ def _periodic_send_who(bot):
 
     if selected_channel is not None:
         # selected_channel's last who is either none or the oldest valid
+        LOGGER.debug('Send WHO for channel: %s', selected_channel)
```
dgw:
```suggestion
        LOGGER.debug('Sending WHO for channel: %s', selected_channel)
```

sopel/coretasks.py
```diff
@@ -432,20 +485,32 @@ def _periodic_send_who(bot):
...
-        bot.privileges[trigger.sender] = dict()
-        bot.channels[trigger.sender] = Channel(trigger.sender)
-        _send_who(bot, trigger.sender)
+    # is it a new channel?
+    if channel not in bot.channels:
+        LOGGER.info('Channel joined: %s', channel)
+        bot.privileges[channel] = dict()
+        bot.channels[channel] = Channel(channel)
+
+    # did we just joined?
```
dgw:
```suggestion
    # did *we* just join?
```

Though I'm not sure how a joined channel could be `not in bot.channels` with a user who is not Sopel. If Sopel has joined, it *must* be a new channel, unless we're leaving remnants of privilege tracking around after Sopel PARTs/QUITs.

sopel/coretasks.py
```diff
@@ -9,21 +9,27 @@
...
+import collections
 import datetime
 import re
 import sys
 import time
+
 import sopel
 import sopel.module
 import sopel.tools.web
+from sopel import loader
```
dgw: Adding this causes LGTM to [flag line R25](https://lgtm.com/projects/g/sopel-irc/sopel/snapshot/5f682a51ca40fac982830cd8429d512c35f7bb92/files/sopel/coretasks.py?sort=name&dir=ASC&mode=heatmap#x59bd4334b139ad7b:1) (`import sopel`) as a duplicate import.

At least, I assume it's this change. It couldn't really be anything else in this patch.

sopel/coretasks.py
Exirel: Yup, hence the `max(setting, 1)`: there is no point having a `throttle_join` with 0s of wait time.

sopel/coretasks.py
Exirel: In my usage of this feature, I find it quite confusing not to have this information: if you have a lot of channels to join, it feels like the bot is blocked doing nothing. This INFO log can help with that, making sure the user knows there is something going on, and that the wait time is actually expected!

What are your thought on that point?

sopel/coretasks.py
Exirel: I had some issue with the MODE event being processed first on some very rare case - maybe it was when I played with the `thread=True` and nothing else - and I don't want to assume that the event sequence will always be the same.

In that particular case, I feel like it's better to be less strict: in the end, a WHO request is sent, privilege are reset to 0, and if the server starts acting weird, at least it won't break Sopel with something that we can handle fine.

sopel/coretasks.py
dgw: My thoughts are: Output one INFO log line if throttling is enabled saying so, before JOIN processing starts, and make this one a DEBUG.

If you want to output more INFO-level stuff at startup, output to the log when the bot joins a channel. 😝 It's true that right now it just says "Connected." and shuts up. Maybe take inspiration [from Bucket](https://github.com/zigdon/xkcd-Bucket/blob/50a9f13b4c875aa7385144ba27df04877c7aaef0/bucket.pl#L2529-L2532)? (Yes, I know it's "ew, Perl", but it's sensible log output.)

sopel/coretasks.py
dgw: Fair. I'll leave this unresolved until you've applied (or discarded) the suggestion.

sopel/coretasks.py
Exirel: > My thoughts are: Output one INFO log line if throttling is enabled saying so, before JOIN processing starts, and make this one a DEBUG.

That's clever and better. I'll do that.

sopel/coretasks.py
```diff
@@ -432,20 +485,32 @@ def _periodic_send_who(bot):
...
-        _send_who(bot, trigger.sender)
+    # is it a new channel?
+    if channel not in bot.channels:
+        LOGGER.info('Channel joined: %s', channel)
+        bot.privileges[channel] = dict()
+        bot.channels[channel] = Channel(channel)
+
+    # did we just joined?
+    if trigger.nick == bot.nick:
+        bot.write(('TOPIC', trigger.sender))
```
dgw: This should be removed, per #1749.

Exirel:
I had to rebase to fix the conflict, but I forgot about your review. As a result, I had to manually fix as you suggested. I tried to keep the fix from your suggestion in a single commit, except for the import thing, which I fixed later.

sopel/coretasks.py
```diff
@@ -135,19 +135,31 @@ def startup(bot, trigger):
...
+    channels = bot.config.core.channels
     if bot.config.core.throttle_join:
         throttle_rate = int(bot.config.core.throttle_join)
         throttle_wait = max(bot.config.core.throttle_wait, 1)
         channels_joined = 0
-        for channel in bot.config.core.channels:
+
+        LOGGER.info(
+            'Joining %d channels, this may take a moment. '
+            '(throttle JOIN is ON)',
```
dgw: For this and the below "OFF" variant, I propose this tweaked wording (you can wrap it to pep8 line lengths at will):

```text
Joining %d channels (with JOIN throttle ON); this may take a moment.
```

sopel/coretasks.py
```diff
@@ -135,19 +135,31 @@ def startup(bot, trigger):
...
-                    'Wait %ds before next batch of JOIN.',
+                LOGGER.debug(
+                    'Waiting %ds before next JOIN batch.',
                     throttle_wait)
                 time.sleep(throttle_wait)
             bot.join(channel)
     else:
+        LOGGER.info(
+            'Joining %d channels, this may take a moment. '
+            '(throttle JOIN is OFF)',
```
dgw: See above suggestion at the ON log line for wording change.

sopel/coretasks.py
```diff
@@ -35,6 +36,46 @@
...
+    know more about the channel. This will prevent an excess of flood when
+    there are too many channels to join at once.
+    """
+    batch_size = max(bot.settings.core.throttle_join, 1)
+    for _ in range(batch_size):
+        try:
+            channel = bot.memory['join_events_queue'].popleft()
+        except IndexError:
+            break
+        LOGGER.debug('Sending WHO after JOIN channel: %s', channel)
```
dgw: I don't like this new wording either, even though it's my own. Sorry! 😹

```suggestion
        LOGGER.debug('Sending WHO after channel JOIN: %s', channel)
```

sopel/coretasks.py
```diff
@@ -132,8 +190,8 @@ def enable_service_auth(bot, trigger):
             'owner.')
 
 
-@sopel.module.event(events.ERR_NOCHANMODES)
-@sopel.module.priority('high')
+@module.event(events.ERR_NOCHANMODES)
+@module.priority('high')
 def retry_join(bot, trigger):
     """Give NickServer enough time to identify on a +R channel.
```
dgw: This annoys me, even though it's not part of this patch. (The typo is quite literally [6 years old](https://github.com/sopel-irc/sopel/commit/0540f998596d5e041e35fbedbba7d48bf669c46d).) I don't mind an unrelated tweak if you don't!

```suggestion
    """Give NickServ enough time to identify on a +R channel.
```

sopel/coretasks.py
```diff
@@ -131,14 +131,16 @@ def startup(bot, trigger):
     bot.memory['retry_join'] = dict()
 
     channels = bot.config.core.channels
+    if not channels:
+        LOGGER.info('No initial channels to JOIN.')
```
dgw: Shouldn't this `return`? If there are no channels, it's silly to log this _and_ "Joining 0 channels" later.

Exirel: Oh. I forgot to use a `elif` after that... >.<

Exirel:
Done.

Also I'm going to hate these merge conflict... x)

Exirel:
Let's go whenever you want.

PR #1753, opened 2019-11-16
safety: rework cache cleanup logic

### LINKED ISSUE #1572: safety: _clean_cache() doesn't, not really
Behold, the `safety` module's cache-cleaning function:

https://github.com/sopel-irc/sopel/blob/39e3680db18c4bed9801d040f7486c655b95a9a0/sopel/modules/safety.py#L207-L220

It's called once every 24 hours, and by `url_handler()` when the cache has too many entries. It seems to remove precisely _one_ cache entry—the oldest—every time it's called. [To quote @HumorBaby](https://github.com/sopel-irc/sopel/pull/1569#discussion_r278521202), "Not really a `_clean`'ing if you ask me :stuck_out_tongue_closed_eyes:"

Ideally, it would:

  * Remove any entries older than some reasonable threshold (a week?)
  * If there are still too many entries, continue removing the oldest one until below the limit (presently 1024)

Improvements on this algorithm are, as always, welcome. I'm just tossing out a hastily thrown-together idea for fixing this.

_Discovered in unrelated review: https://github.com/sopel-irc/sopel/pull/1569#discussion_r278354791_

dgw:
Details in the commit message, but the gist is implementing exactly what I described previously. This resolves #1572, assuming I haven't made any silly mistakes.

### DIFF
--- a/sopel/modules/safety.py
+++ b/sopel/modules/safety.py
@@ -12,6 +12,7 @@ import logging
 import os.path
 import re
 import sys
+import threading
 import time
 
 import requests
@@ -43,6 +44,7 @@ LOGGER = logging.getLogger(__name__)
 vt_base_api_url = 'https://www.virustotal.com/vtapi/v2/url/'
 malware_domains = set()
 known_good = []
+cache_limit = 512
 
 
 class SafetySection(StaticSection):
@@ -83,6 +85,8 @@ def setup(bot):
 
     if 'safety_cache' not in bot.memory:
         bot.memory['safety_cache'] = sopel.tools.SopelMemory()
+    if 'safety_cache_lock' not in bot.memory:
+        bot.memory['safety_cache_lock'] = threading.Lock()
     for item in bot.config.safety.known_good:
         known_good.append(re.compile(item, re.I))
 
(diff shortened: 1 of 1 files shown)


sopel/modules/safety.py
```diff
@@ -216,13 +218,31 @@ def toggle_safety(bot, trigger):
...
+        if overage > 0:
+            extra_keys = sorted(
+                (data.fetched, key)
+                for (key, data)
+                in bot.memory['safety_cache'].items())[:overage]
+            for (_, key) in extra_keys:
+                bot.memory['safety_cache'].pop(key, None)
+
+        LOGGER.info('Safety cache cleanup finished.')
+        bot.memory['safety_cache_lock'].release()
```
Exirel: What if any of the above code raise an exception and `release` is never called? This might need a try/except block.

Also, if a job is running twice at the same time with a 24h interval, we have a bigger problem than having to wait for the lock to be available.

dgw: It's also run directly if the cache gets too big. The interval is only one way it can get called.

Fair point about `release`. I probably should've gone with my initial impulse and used a `with`.

dgw: Actually I remember now why I didn't use `with`… I need the return value from `acquire` in non-blocking mode to log that debug entry. Which isn't _strictly_ necessary, but…

dgw: I used a try/finally instead.

PR #1754, opened 2019-11-17
qa: fix additional LGTM alerts

dgw:
Since it seems running LGTM locally isn't a thing, I have to open this PR and hope that it does everything I think it should. I _think_ this fixes 18 alerts. If the actual results don't match, I'll figure it out Soon™ (but probably not within the next 24 hours).

**Edit:** Looks like easy fixes (only the exception-comment changes failed, and it seems to be an inconsistency in LGTM's documentation about whether a space is allowed).

### DIFF
--- a/sopel/modules/remind.py
+++ b/sopel/modules/remind.py
@@ -353,6 +353,25 @@ class TimeReminder(object):
             ]
         )
 
+    def __ne__(self, other):
+        return any(
+            getattr(self, attr) != getattr(other, attr, None)
+            for attr in [
+                'hour',
+                'minute',
+                'second',
+                'timezone',
+                'year',
+                'month',
+                'day',
+                'message',
+            ]
+        )
+
+    # Mutable objects probably shouldn't be made hashable
+    # https://docs.python.org/3/reference/datamodel.html#object.__hash__
+    __hash__ = None
+
     def get_duration(self, today=None):
         """Get the duration between the reminder and ``today``
 
--- a/contrib/rpm/makerpm.py
+++ b/contrib/rpm/makerpm.py
@@ -6,7 +6,6 @@ import os
 import os.path
 import time
 from subprocess import *
-import shutil
 
 repo = git.Repo(os.getcwd())
 head_hash = repo.head.commit.hexsha[:7]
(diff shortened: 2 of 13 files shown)


dgw:
So, exclusion comments don't work, either with or without a space between `lgtm` and the `[specific/rule]` to ignore for that line/block. Great.

I'll come back to this another day.

sopel/modules/remind.py
```diff
@@ -353,6 +353,33 @@ def __eq__(self, other):
...
+                'second',
+                'timezone',
+                'year',
+                'month',
+                'day',
+                'message',
+            ]
+        )
+
+    def __hash__(self):
```
Exirel: I've seen this warning, and I did some research on `__hash__` because I remembered something very important about it. From the [python documentation about `__hash__`](https://docs.python.org/3/reference/datamodel.html#object.__hash__) (emphasis is mine):

> If a class defines mutable objects and implements an `__eq__()` method, it **should not** implement `__hash__()`, since the implementation of hashable collections requires that a key’s hash value is immutable (if the object’s hash value changes, it will be in the wrong hash bucket).

Since you can change `TimeReminder`'s attributes, you **should not** implement the `__hash__` magic method.

dgw: LGTM probably doesn't care whether the method is "implemented" or just set to `None`, indicating the object isn't hashable. I'd like to give them a few days (since it's the weekend) for looking into semmle/ql#2361 though, before I come back and tweak this.

dgw:
@Exirel Assuming the exclusion comments can be made to work, you're OK with this as amended?

dgw:
I'm not waiting 15 minutes for LGTM to re-analyze this with the suppression-comment changes dropped (I moved them to another branch; no reason to keep them unless they _work_). Presumably it'll pass the analysis with (still) 14 alerts fixed, and I'll merge it tomorrow afternoon-ish (currently 02:35 😅).

dgw:
Right, so LGTM apparently doesn't apply suppression comments in PRs _on purpose_. I put them back, and will merge this Very Soon™. Gonna find something else to do while the analysis runs. 😁

PR #1756, opened 2019-11-21
Prepare Sopel to adapt its output when py2 officially dies

dgw:
Kind of magic-number-y, but I'd rather this than duplicating `datetime` code in two places.

No, I do not feel like optimizing this out into a function that we'll remove as soon as we start on 8.0.


half-duplex:
You could also just write "Python 2 will stop receiving updates on Mon Day 2020. Sopel will stop supporting it in 8.0" and skip the dependency and magic.

dgw:
@half-duplex but I _like_ magic 😝

Issue #1305, opened 2018-04-04
Expose NICKLEN to plugins somewhere


dgw:
I think `NICKLEN` (from the `005 RPL_ISUPPORT` server property list) would be a useful constant to have available for modules. For example, the `tell` module could use it instead of the hardcoded constant last updated for #1304, so it never accepts messages for nicks that are too long for the current network.

dgw:
Available through `bot.isupport.NICKLEN` (or `bot.isupport['NICKLEN']`) as of #1758.

PR #1758, opened 2019-11-21
core: handle 005 RPL_ISUPPORT event

Exirel:
WIP to show a bit of the ISupport thing: my idea is to have a storage structure with the basic parsing tools at first. Then, I'll implement an actual callable, triggered by the event `RPL_ISUPPORT`, with the proper machinery to parse, update, and remove ISupport data.

This Draft version aims to open a small conversation about the design and the usability of the ISupport class.

Quick note: the ISupport class will generate immutable object. There is no good reason to mess with that, as every bit of information comes from the server itself.

### DIFF
--- a/sopel/coretasks.py
+++ b/sopel/coretasks.py
@@ -24,8 +24,10 @@ import time
 from random import randint
 
 from sopel import loader, module
+from sopel.irc import isupport
 from sopel.irc.utils import CapReq
 from sopel.tools import Identifier, events, iteritems, jobs, target, web
+
 
 if sys.version_info.major >= 3:
     unicode = str
@@ -193,6 +195,25 @@ def startup(bot, trigger):
         bot.say(msg, bot.config.core.owner)
 
     _execute_perform(bot)
+
+
+@module.priority('high')
+@module.event(events.RPL_ISUPPORT)
+@module.thread(False)
+@module.unblockable
+@module.rule('are supported by this server')
+def handle_isupport(bot, trigger):
+    """Handle ``RPL_ISUPPORT`` events."""
+    parameters = {}
+    for arg in trigger.args:
+        try:
+            key, value = isupport.parse_parameter(arg)
+            parameters[key] = value
+        except ValueError:
+            # ignore malformed parameter: log a warning and continue
+            LOGGER.warning('Unable to parse ISUPPORT parameter: %r', arg)
+
+    bot._isupport = bot._isupport.apply(**parameters)
 
 
 @module.require_privmsg()
--- a/sopel/irc/isupport.py
+++ b/sopel/irc/isupport.py
@@ -222,13 +222,16 @@ class ISupport(object):
             False
 
         """
-        kwargs_ci = dict((key.upper(), value) for key, value in kwargs.items())
+        kwargs_upper = dict(
+            (key.upper(), value)
+            for key, value in kwargs.items()
+        )
         kept = (
             (key, value)
             for key, value in self.__isupport.items()
-            if ('-%s' % key) not in kwargs_ci
+            if ('-%s' % key) not in kwargs_upper
         )
-        updated = dict(itertools.chain(kept, kwargs_ci.items()))
+        updated = dict(itertools.chain(kept, kwargs_upper.items()))
 
         return self.__class__(**updated)
 
--- a/sopel/irc/isupport.py
+++ b/sopel/irc/isupport.py
@@ -2,7 +2,7 @@
 """IRC Tools for ISUPPORT management.
 
 When a server wants to advertise its features and settings, it can use the
-``RPL_ISUPPORT`` command with a list of arguments.
+``RPL_ISUPPORT`` command (``005`` numeric) with a list of arguments.
 
 .. seealso::
 
@@ -62,7 +62,7 @@ def _parse_chanmodes(value):
     if len(items) < 4:
         raise ValueError('Not enough channel types to unpack from %r.' % value)
 
-    # add extra channel type's modes to their own tuple
+    # add extra channel mode types to their own tuple
     # result in (A, B, C, D, (E, F, G, H, ..., Z))
     # where A, B, C, D = result[:4]
     # and extras = result[4]
@@ -140,7 +140,7 @@ def parse_parameter(arg):
         # ignore value for removed parameters
         return (key, None)
 
-    parser = ISUPPORT_PARSER.get(key, str)
+    parser = ISUPPORT_PARSER.get(key, _optional(str))
     return (key, parser(value))
 
 
@@ -186,6 +186,23 @@ class ISupport(object):
 
     def __contains__(self, key):
         return key.upper() in self.__isupport
+
+    def __getattr__(self, name):
+        if name not in self.__isupport:
+            raise AttributeError(name)
+
+        return self.__isupport[name]
+
+    def __setattr__(self, name, value):
+        # make sure you can't set the value of any ISUPPORT attribute yourself
+        if name == '_ISupport__isupport':
+            # allow to set self.__isupport inside of the class
+            super(ISupport, self).__setattr__(name, value)
+        elif name in self.__isupport:
+            # reject any modification of __isupport
+            raise AttributeError("Can't set value for %r" % name)
+        elif name not in self.__dict__:
+            raise AttributeError('Unknown attribute')
 
     def apply(self, **kwargs):
         """Build a new instance of :class:`ISupport`.
@@ -249,7 +266,7 @@ class ISupport(object):
     def CHANMODES(self):
         """Expose ``CHANMODES`` as a dict, if advertised by the server.
 
-        This exposes information about 4 types of channel::
+        This exposes information about 4 types of channel modes::
 
             >>> isupport.CHANMODES
             {
@@ -277,7 +294,7 @@ class ISupport(object):
     def MAXLIST(self):
         """Expose ``MAXLIST`` as a dict, if advertised by the server.
 
-        This exposes information about maximum combination of modes::
+        This exposes information about maximums for combinations of modes::
 
             >>> isupport.MAXLIST
             {
@@ -304,7 +321,8 @@ class ISupport(object):
     def PREFIX(self):
         """Expose ``PREFIX`` as a dict, if advertised by the server.
 
-        This exposes information about prefix used for user privileges::
+        This exposes information about the modes and nick prefixes used for
+        user privileges in channels::
 
             >>> isupport.PREFIX
             {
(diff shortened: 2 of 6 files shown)


Exirel:
I'm getting closer to the end: now the ISupport class is able to build a new instance of ISupport by applying changes!

What's left: adding the right rule callables.

Exirel:
By exposing `bot.isupport.NICKLEN`, this closes #1305 and by parsing `005 RPL_ISUPPORT`, this also closes #1082

sopel/irc/isupport.py
```diff
@@ -0,0 +1,380 @@
+# coding=utf-8
+"""IRC Tools for ISUPPORT management.
+
+When a server wants to advertise its features and settings, it can use the
+``RPL_ISUPPORT`` command with a list of arguments.
```
dgw:
```suggestion
``RPL_ISUPPORT`` command (``005`` numeric) with a list of arguments.
```

This might or might not be important enough to mention.

sopel/irc/isupport.py
```diff
@@ -0,0 +1,380 @@
...
+#
+# Licensed under the Eiffel Forum License 2.
+from __future__ import unicode_literals, absolute_import, print_function, division
+
+import functools
+import itertools
+import re
+
+
+def _optional(parser, default=None):
```
dgw: For this and all of the "private" parser methods, it would be more consistent with the overhauled `irc` submodule docs (see #1772) to have proper docstrings as if Sphinx will generate docs based on them, even though they aren't presently included in any public pages.

sopel/irc/isupport.py
```diff
@@ -0,0 +1,380 @@
...
+    return wrapped
+
+
+def _parse_chanmodes(value):
+    items = value.split(',')
+
+    if len(items) < 4:
+        raise ValueError('Not enough channel types to unpack from %r.' % value)
+
+    # add extra channel type's modes to their own tuple
```
dgw:
```suggestion
    # add extra channel mode types to their own tuple
```

sopel/irc/isupport.py
```diff
@@ -0,0 +1,380 @@
...
+    modes = result.group('modes')
+    prefixes = result.group('prefixes')
+
+    if len(modes) != len(prefixes):
+        raise ValueError('Mode list does not match for PREFIX: %r' % value)
+
+    return tuple(sorted(zip(modes, prefixes)))
+
+
+ISUPPORT_PARSER = {
```
dgw: Causes ripple-effect changes elsewhere, I'm aware. If you don't want to do this, meh. I'm just suggesting.

```suggestion
ISUPPORT_PARSERS = {
```

sopel/irc/isupport.py
```diff
@@ -0,0 +1,380 @@
...
+
+            >>> updated = {'-AWAYLEN': None, 'NICKLEN': 25, 'CHANNELLEN': 10}
+            >>> new = isupport.apply(**updated)
+            >>> 'CHANNELLEN' in new
+            True
+            >>> 'AWAYLEN' in new
+            False
+
+        """
+        kwargs_ci = dict((key.upper(), value) for key, value in kwargs.items())
```
dgw: Like the tests named with `_ci`, I think we can be clearer here.

```suggestion
        kwargs_upper = dict((key.upper(), value) for key, value in kwargs.items())
```

sopel/irc/isupport.py
```diff
@@ -0,0 +1,380 @@
...
+        if 'CHANLIMIT' not in self:
+            raise AttributeError('CHANLIMIT')
+
+        return dict(self['CHANLIMIT'])
+
+    @property
+    def CHANMODES(self):
+        """Expose ``CHANMODES`` as a dict, if advertised by the server.
+
+        This exposes information about 4 types of channel::
```
dgw:
```suggestion
        This exposes information about 4 types of channel modes::
```

Also, what about the allowed additional types that may follow in a tuple at position 5? Are those only to be accessible through subscripting the `ISupport` instance?

sopel/irc/isupport.py
```diff
@@ -0,0 +1,380 @@
...
+        if 'CHANMODES' not in self:
+            raise AttributeError('CHANMODES')
+
+        return dict(zip('ABCD', self['CHANMODES'][:4]))
+
+    @property
+    def MAXLIST(self):
+        """Expose ``MAXLIST`` as a dict, if advertised by the server.
+
+        This exposes information about maximum combination of modes::
```
dgw:
```suggestion
        This exposes information about maximums for combinations of modes::
```

sopel/irc/isupport.py
```diff
@@ -0,0 +1,380 @@
...
+        if 'MAXLIST' not in self:
+            raise AttributeError('MAXLIST')
+
+        return dict(self['MAXLIST'])
+
+    @property
+    def PREFIX(self):
+        """Expose ``PREFIX`` as a dict, if advertised by the server.
+
+        This exposes information about prefix used for user privileges::
```
dgw:
```suggestion
        This exposes information about the modes and nick prefixes used for
        user privileges in channels::
```

**Also** + re-wrap

sopel/coretasks.py
```diff
@@ -25,6 +25,7 @@
 from sopel.tools import Identifier, iteritems, events
 from sopel.tools.target import User, Channel
 import base64
+from sopel.irc import isupport
```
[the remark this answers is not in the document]
Exirel: I know. It's ok.

test/test_coretasks.py
```diff
@@ -111,3 +111,75 @@ def test_mode_colon(mockbot, ircfactory):
...
+    assert 'CHANNELLEN' in mockbot.isupport
+    assert 'TOPICLEN' in mockbot.isupport
+    assert 'DEAF' in mockbot.isupport
+    assert 'FNC' in mockbot.isupport
+    assert 'TARGMAX' in mockbot.isupport
+    assert 'EXTBAN' in mockbot.isupport
+    assert 'CLIENTVER' in mockbot.isupport
+    assert 'ETRACE' in mockbot.isupport
+    assert 'WHOX' in mockbot.isupport
+    assert 'KNOCK' in mockbot.isupport
```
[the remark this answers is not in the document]
Exirel: I trust my fellow Sopel rockstart comrades not to forget tests when touching ISUPPORT parsing.

test/irc/test_irc_isupport.py
```diff
@@ -0,0 +1,508 @@
...
+    assert 'AWAYLEN' not in instance
+    assert 'AWAYLEN' in new
+    assert 'AWAYLEN' not in new_removed
+    assert new['AWAYLEN'] == 50, 'original instance must not be modified'
+
+    assert 'NICKLEN' in new_removed
+    assert new_removed['NICKLEN'] == 31
+
+
+def test_isupport_apply_ci():
```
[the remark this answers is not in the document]
Exirel: lol, this is such a convention that I didn't realized it could be misleading.

sopel/irc/isupport.py
Exirel: As I was reading the doc, they talk about "channel type" having "modes". I'm not sure `channel mode types` convey the same meaning. How can I be sure?

sopel/irc/isupport.py
Exirel: Yup. `other_chantype_modes = isupports['chanmodes'][5]` where `other_chantype_modes` is now a tuple of everything else.

test/test_coretasks.py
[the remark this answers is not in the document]
dgw: Lol, trusting people to update tests. 😹 But reading between the lines, this sounds like a "no, there's no chance of that", so good enough.

sopel/irc/isupport.py
dgw: I'm not sure exactly what documentation you used for reference, but the four standard categories you get from the CHANMODES parameter _are_ different types of _mode_:

<ol type="A">
  <li>Modes that add or remove an address from a list (MUST have a parameter unless the client is sending them without one to obtain the current list)</li>
  <li>Modes that change a channel setting (MUST have a parameter, always)</li>
  <li>Modes that change a channel setting (MUST be set with a parameter, and MUST NOT have a parameter when unset</li>
  <li>Modes that change a channel setting (MUST NOT have a parameter)</li>
</ol>

Neither [the Internet draft](https://tools.ietf.org/html/draft-hardy-irc-isupport-00#section-4.3) I checked (linked from my go-to [list of numerics](https://www.alien.net.au/irc/irc2numerics.html), which always comes up in my searches for "irc numerics") nor [modern.ircdocs.horse](https://modern.ircdocs.horse/#chanmodes-parameter) says anything about "channel type" in their respective sections about the CHANMODES parameter. They are categories/types of _mode_ only.

sopel/irc/isupport.py
dgw: I'm not sure I like `isupport['CHANMODES']` to return different information than `isupport.CHANMODES`. Shouldn't they present the same data, with the only difference being how it was obtained?

sopel/irc/isupport.py
Exirel: Ooooh okay now I get it. My mistake!

sopel/irc/isupport.py
Exirel: See `isupport['key']`  as a low level API with raw data, and `isupport.SPECIFIC_KEY` as a high level API with structured and documented data.

Since the modern ircdoc tend to discourage from using extra types:

> However, server authors SHOULD NOT extend this parameter without good reason, and SHOULD CONSIDER whether their mode would work as one of the existing types instead.
> **The behaviour of any additional types is undefined.**

And since the behaviour of additional types is undefined, I'd prefer not to expose them in the high level API.

Exirel:
Rebased, conflict resolved, and squashed.

dgw:
#1751 why? You betrayed me. ☹️

sopel/irc/isupport.py
```diff
@@ -0,0 +1,384 @@
...
+
+    parser = ISUPPORT_PARSERS.get(key, _optional(str))
+    return (key, parser(value))
+
+
+class ISupport(object):
+    """Storage class for IRC's ``ISUPPORT`` feature.
+
+    An instance of ``ISupport`` can be used as a read-only dict, to store
+    feature advertised by the IRC server::
```
dgw:
```suggestion
    features advertised by the IRC server::
```

I missed this before! Caught while drafting changelog entry and building the docs to make sure my cross-links would be correct. 😁

Exirel:
@dgw ready to merge. I'll rebase #1769 upon this one, so you shouldn't have problems when merging both back to back.

PR #1722, opened 2019-10-25
reddit: handle subreddits, and `u/` & `r/` references

deathbybandaid:
This PR adds to the functionality of `reddit.py` by adding:

* Add a subreddit command as well as url pattern match for subreddits

```
13:32 <@deathbybandaid> https://reddit.com/r/plex
13:32 <reddit-test> [REDDIT] https://www.reddit.com/r/plex | subscribers (124733) | Created at 2008-08-16 - 06:54:02UTC | For questions and comments about the Plex Media Server. The Plex Media Server is smart software that makes playing Movies, TV Shows and other media on your computer simple.
13:32 <@deathbybandaid> .subreddit plex
13:32 <reddit-test> [REDDIT] https://www.reddit.com/r/plex | subscribers (124733) | Created at 2008-08-16 - 06:54:02UTC | For questions and comments about the Plex Media Server. The Plex Media Server is smart software that makes playing Movies, TV Shows and other media on your computer simple.
```

* add `r/` and `u/` matching. 

```
13:31 <@deathbybandaid> r/plex
13:31 <reddit-test> [REDDIT] https://www.reddit.com/r/plex | subscribers (124732) | Created at 2008-08-16 - 06:54:02UTC | For questions and comments about the Plex Media Server. The Plex Media Server is smart software that makes playing Movies, TV Shows and other media on your computer simple.
13:31 <@deathbybandaid> u/plex
13:31 <reddit-test> [REDDITOR] plex | https://reddit.com/u/PleX | Link: 1 | Comment: 0
```

PR #1760, opened 2019-11-23
reddit: cleanup & various fixes

dgw:
This is a regular laundry list, honestly.

* `post_url` was end-anchored to prevent matching the same links as `comment_url`, but didn't allow for the `/post_title_slug/` usually included after the submission ID.
* `comment_url` & `post_url` used `.*` where they should have had `\S+` (whitespace is never present in reddit URLs, nor will any compenent of the link ever contain zero characters).
* `user_url` unnecessarily captured "ser" (or "") as group 1.
* So many unescaped `.`s everywhere!

Fixes post links failing to match and falling back to `url.py` title fetching, and possibly some other silly bugs. I take full responsibility for breaking `post_url` in #1720. Whoops.

Also adds error handling in `video_info()`, in case the hacky hack it uses to get the post URL fails. The errors I saw in testing that led to this whole patch were probably just because the `post_url` pattern was broken (see above), but... why not handle the error case anyway?

**Edit:** Also fixed some image-handler stuff (it couldn't find NSFW source posts).

### DIFF
--- a/sopel/modules/reddit.py
+++ b/sopel/modules/reddit.py
@@ -81,7 +81,11 @@ def image_info(bot, trigger, match):
         .subreddit('all')
         .search('url:{}'.format(url), sort='new', params={'include_over_18': 'on'})
     )
-    oldest = results[-1]
+    try:
+        oldest = results[-1]
+    except IndexError:
+        # Fail silently if the image link can't be mapped to a submission
+        return NOLIMIT
     return say_post_info(bot, trigger, oldest.id)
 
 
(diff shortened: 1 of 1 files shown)


dgw:
If whoever double-checks this for me thinks it's worthwhile, we can definitely be more specific than `\S` for path components, for example. This just seems Good Enough™.

@deathbybandaid Also, this confirms that the link-handling problem we both saw while testing #1722 was completely unrelated to that patch. Unfortunately, our two PRs will merge-conflict. 🙁

dgw:
@sopel-irc/rockstars Final few tweaks added, rebased on top of #1722, and ready to review. 🚀

PR #1762, opened 2019-11-23
bot, reload: log/show plugin source when reloading

dgw:
This would have saved me a good half-hour of confusion tonight, by showing me that my bot was using an (outdated) plugin file in `.sopel/modules` that shadowed the built-in plugin from `sopel.modules.*`.

I think the added `bot.get_plugin_meta()` function is helpful, and it was necessary to implement the output I wanted in `reload`, but there could be a reason I haven't thought of that makes it a bad idea. Definitely tell me if so!

Even just having the reloaded plugin's metadata in logs, and not on IRC, would have helped. I added it back to `reload` because 1) it used to show the file path before the plugin system rework, and 2) I like not having to check my terminal/log-files for things. 😛

Note: I'm putting this in 7.0 because let's face it: a lot of people are going to make updates to their plugins for this release, to take advantage of new features and/or fix compatibility problems. We should make it easy for them to verify that the correct file/module is being loaded during development, even if `.reload` is still buggy for some plugin types.

### DIFF
--- a/sopel/bot.py
+++ b/sopel/bot.py
@@ -311,6 +311,20 @@ class Sopel(irc.AbstractBot):
         """Tell if the bot has registered this plugin by its name"""
         return name in self._plugins
 
+    def get_plugin_meta(self, name):
+        """Get info about a registered plugin by its name.
+
+        :param str name: name of the plugin about which to get info
+        :return: the plugin's metadata
+                 (see :meth:`~.plugins.handlers.AbstractPluginHandler.get_meta_description`)
+        :rtype: :class:`dict`
+        :raise PluginNotRegistered: when there is no ``name`` plugin registered
+        """
+        if not self.has_plugin(name):
+            raise plugins.exceptions.PluginNotRegistered(name)
+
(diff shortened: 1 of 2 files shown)


sopel/bot.py
```diff
@@ -311,6 +315,19 @@ def has_plugin(self, name):
         """Tell if the bot has registered this plugin by its name"""
         return name in self._plugins
 
+    def get_plugin_meta(self, name):
+        """Get info about a registered plugin by its name
+
+        :param str name: name of the plugin about which to get info
+        :return: the plugin's metadata
```
half-duplex: It may be good to refer to where the contents of the metadata dict are defined/documented (anywhere other than https://github.com/sopel-irc/sopel/blob/e9ca7e1b163375aef62e8345276f12213eb29165/sopel/plugins/handlers.py#L93 ?)

dgw: There's no other place (plugin metadata is new in 7.0/master), but I can link to the abstract handler's method's docstring (done in force-push).

PR #1763, opened 2019-11-23
reddit: subreddit handling for r/all and r/popular

deathbybandaid:
As per my comment in #1722 here's handling for the two non-subreddit subreddits!

### DIFF
--- a/sopel/modules/reddit.py
+++ b/sopel/modules/reddit.py
@@ -205,6 +205,23 @@ def comment_info(bot, trigger, match):
 
 def subreddit_info(bot, trigger, match, commanded=False):
     """Shows information about the given subreddit"""
+    match_lower = match.lower()
+    if match_lower in ['all', 'popular']:
+        message = ('[REDDIT] {link}{nsfw} | {public_description}')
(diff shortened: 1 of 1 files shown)


sopel/modules/reddit.py
```diff
@@ -195,6 +195,22 @@ def comment_info(bot, trigger, match):
 
 def subreddit_info(bot, trigger, match, commanded=False):
     """Shows information about the given subreddit"""
+    if match.lower() in ['all', 'popular']:
```
Exirel: Way too many `match.lower()` in here. You can work with an intermediate variable for this one I believe.

deathbybandaid: assuming #1764 gets merged before this, I can rebase over that when the time comes.

sopel/modules/reddit.py
```diff
@@ -205,6 +205,23 @@ def comment_info(bot, trigger, match):
 
 def subreddit_info(bot, trigger, match, commanded=False):
     """Shows information about the given subreddit"""
+    match = match.lower()
```
Exirel: I'm not sure this'll play well with this line: `r.subreddits.search_by_name(match, exact=True)`

Maybe use a different name for that?

PR #1764, opened 2019-11-24
reddit: Reduce try/except size for redditor_info

deathbybandaid:
This PR does:
* severely reduce the size of the giant try/except that is `redditor_info`
* move the `is_cakeday` logic to it's own function for potential reusability in future PRs

While `u.id` is not a used variable for `redditor_info`, this PR uses Reddits ID system to verify that a user does exist. Simply put, if a username doesn't have an ID it won't have `.gold` or `.link_karma` or other attributes that would be associated with a redditor.

This also adds a `u.is_employee` check for the message to display when the user is a Reddit employee.

### DIFF
--- a/sopel/modules/reddit.py
+++ b/sopel/modules/reddit.py
@@ -73,6 +73,24 @@ def get_time_created(bot, trigger, entrytime):
     return created
 
 
+def get_is_cakeday(entrytime):
+    now = dt.datetime.utcnow()
+    cakeday_start = dt.datetime.utcfromtimestamp(entrytime)
+    cakeday_start = cakeday_start.replace(year=now.year)
+    day = dt.timedelta(days=1)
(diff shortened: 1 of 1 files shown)


deathbybandaid:
I will note, that I was originally trying to also determine the difference between accounts that don't exist, and accounts that have been permanently banned.

The praw documentation says that it's possible, but I wasn't able to get it to actually work.
If I get more time, since this will be slated for 7.1, I may try again.

dgw:
> This also adds a `u.is_employee` check for the message (willing to switch color from `PURPLE`) to display when the user is a Reddit employee.

reddit itself uses red to denote site admins speaking officially, FWIW.

deathbybandaid:
Easy enough change when I get a minute

deathbybandaid:
Updated the color

PR #1771, opened 2019-12-03
tools: subpackage doc overhaul

dgw:
Tin. See #1565.

I was surprised that `tools.calculation` took the most time, which was probably mostly wasted on docstrings that don't appear in the public docs anyway, but pretty tables are always worth it… right? 😁

### DIFF
--- a/sopel/tools/jobs.py
+++ b/sopel/tools/jobs.py
@@ -59,7 +59,7 @@ class JobScheduler(threading.Thread):
         self.stopping.set()
 
     def remove_callable_job(self, callable):
-        """Removes specific callable from job queue"""
+        """Remove ``callable`` from the job queue."""
         with self._mutex:
             self._jobs = [
                 job for job in self._jobs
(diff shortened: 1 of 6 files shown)


sopel/tools/jobs.py
```diff
@@ -125,17 +125,19 @@ def _call(self, job):
 
 
 class Job(object):
-    """Hold information about when a function should be called next.
+    """Holds information about when a function should be called next.
```
Exirel: Wait, I'm confused. "Remove", but "Holds"? Shouldn't it be "Hold" here too?

dgw: Subtle difference in what sounds right (to me) for an object vs. function/method. Though you could make the argument that functions should use the present tense too…

Exirel: Well, I've learn, a long time ago, to always use the infinitive for function and method, but nothing about class. I know that Panda/Nympy uses this style (infinitive), and my friends working on Django too.

So... I'd definitively keep the infinitive on functions/methods, and I don't have a strong opinion about classes. This just seems weird to me.

dgw: For what it's worth, this isn't the only place we use similar verb tense for classes. One example is `config.Config.ConfigSection`: https://github.com/sopel-irc/sopel/blob/c43fe42685b1d35b63562e943bb4733fa3d286c7/sopel/config/__init__.py#L180 (which remains unchanged in #1178).

Goodness knows we can leave this as imperative, but then we go from seeming weird to _you_ to seeming weird to _me_, and the net result is basically nothing. (Also, I don't want to amend this PR just to remove that single `s`. 😝)

Exirel: Well, we could argue that functions and classes have very different meanings and purposes so we can't apply the same rules. The more I think about it, the more I believe the problem is not the imperative vs verb tense, and more "how to write the synopsis of a class?".

So, you know what? Without a second thought, I hereby declare this resolved.

PR #1772, opened 2019-12-04
bot, irc: doc overhaul

dgw:
Overhauled docs for both of these modules at once because they're rather intertwined.

Touches the `trigger` docs a bit because some cross-references weren't working without `PreTrigger` being published in the built docs.

Fixed a mistake in the CAP REQ docs telling users to use the wrong prefix for required capabilities.

Even tweaked the code for loading block lists from the config, because it annoyed me that we weren't using the `ListAttribute` type's `default` kwarg. (This change is easy to back out, in case I overlooked a good reason things worked the way they did, as it's a separate commit.)

Of course, lots of other tweaks & fixes, too!

### DIFF
--- a/docs/source/trigger.rst
+++ b/docs/source/trigger.rst
@@ -1,5 +1,15 @@
 Triggers
 ========
 
+A :class:`Trigger` is the main type of user input plugins will see.
+
+Sopel uses :class:`PreTrigger`\s internally while processing incoming IRC
+messages. Plugin authors can reasonably expect that their code will never
+receive one. They are documented here for completeness, and for the aid of
+Sopel's core development.
+
 .. autoclass:: sopel.trigger.Trigger
+    :members:
+
+.. autoclass:: sopel.trigger.PreTrigger
     :members:
--- a/sopel/bot.py
+++ b/sopel/bot.py
@@ -131,13 +131,6 @@ class Sopel(irc.AbstractBot):
         self.scheduler = sopel.tools.jobs.JobScheduler(self)
         """Job Scheduler. See :func:`sopel.module.interval`."""
 
-        # Set up block lists
-        # Default to empty
-        if not self.settings.core.nick_blocks:
-            self.settings.core.nick_blocks = []
-        if not self.settings.core.host_blocks:
-            self.settings.core.host_blocks = []
-
     @property
     def command_groups(self):
         """A mapping of plugin names to a list of commands in it."""
--- a/sopel/bot.py
+++ b/sopel/bot.py
@@ -69,7 +69,7 @@ class Sopel(irc.AbstractBot):
         """
 
         self._command_groups = collections.defaultdict(list)
-        """A mapping of plugin names to a list of commands in it."""
+        """A mapping of plugin names to lists of their commands."""
 
         self._times = {}
         """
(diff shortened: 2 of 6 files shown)


docs/source/trigger.rst
```diff
@@ -1,5 +1,17 @@
...
 ========
 
+A ``Trigger`` is the main type of user input plugins will see.
+
 .. autoclass:: sopel.trigger.Trigger
     :members:
+
+
+PreTriggers
+===========
```
Exirel: Even though Sphinx is usually smart with hierarchy, I'd recommend against using the same level as the main title.

Here, I'd see:

```
========
Triggers
========

.. autoclass:: sopel.trigger.Trigger

PreTriggers
===========

.. autoclass:: sopel.trigger.PreTrigger
```

sopel/config/core_section.py
```diff
@@ -198,7 +198,7 @@ def homedir(self):
     host = ValidatedAttribute('host', default='irc.dftba.net')
     """The server to connect to."""
 
-    host_blocks = ListAttribute('host_blocks')
+    host_blocks = ListAttribute('host_blocks', default=[])
```
Exirel: Do we need that `default`?

docs/source/trigger.rst
dgw: Does this hierarchy look fine to you? It seems OK to me, unless we're to consider a `PreTrigger` as subordinate to a `Trigger`. Which it is, kind of…but only functionally—the types are siblings in the code.

image

Put another way, I think we either keep both headings at the same level (optionally with a new top-level heading unifying the two, if your worry is actually about Sphinx getting confused somehow), or eliminate the "PreTriggers" heading and have them both under "Triggers".

sopel/config/core_section.py
dgw: Oh, technically not. `ListAttribute`'s constructor does it if no default is specified: https://github.com/sopel-irc/sopel/blob/185686f92d299f95027f4dccf40ffd994e7807fd/sopel/config/types.py#L317-L318

I'll revert this and the identical change in `nick_blocks` when we reach a consensus on the rST headings and I apply it, to save Travis & LGTM the trouble of running excessive jobs.

docs/source/trigger.rst
dgw: For completeness: The proposed heading structure makes "PreTriggers" a second-level heading, subordinate to the "Triggers" at page top.

image

In both cases, the page title remains "Triggers".

dgw: Found a tiny problem with having two top-level headings on the page while working on `trigger`'s doc overhaul (still WIP): Sphinx generates navigation links for both "Trigger" and "PreTrigger". Didn't notice that before.

But my opinion stands: Either we add a new top-level heading above the two, or put both classes under the single top-level "Triggers" heading. (I'm open to other suggestions, of course. I've been 200% distracted during all Sopel stuff I did today, including this!)

Exirel: I'd say you can leave it as it was: Trigger is the page's title (until we find something better?), and both classes are under this title, without any subdivision. So:

```
Trigger
=======

.. class:: Trigger

.. class:: PreTrigger
```

Exirel:
By the way: great work here. It's not an easy task, I know it, and I'm glad you're the kind of project maintainer that take it seriously. ❤️

dgw:
@Exirel Removed the `default` kwargs from `host_blocks`/`nick_blocks`, and tweaked `trigger.rst`, as discussed. (That's so you know what to look at, without having to re-review this whole patch. 😸)

Issue #1623, opened 2019-05-29
test_db.py needs updating


deathbybandaid:
test_db.py currently tests directly with sqlite. `sqlite3.connect(db_filename)` is heavily used in these tests. These tests should reflect the new database system via sqlalchemy.

The default DB for sopel is sqlite, and we only NEED to test a mockconfig set for sqlite.

dgw:
It's late in the cycle for 7.0, so this is totally puntable if @RustyBower hasn't the time to work on it in the next several weeks.

PR #1777, opened 2019-12-06
tests: migrating db tests to sqlalchemy

RustyBower:
This should fix https://github.com/sopel-irc/sopel/issues/1623

### DIFF
--- a/test/test_db.py
+++ b/test/test_db.py
@@ -8,13 +8,12 @@ from __future__ import unicode_literals, absolute_import, print_function, divisi
 
 import json
 import os
-import sqlite3
 import sys
 import tempfile
 
 import pytest
 
-from sopel.db import SopelDB
+from sopel.db import ChannelValues, PluginValues, Nicknames, NickValues, SopelDB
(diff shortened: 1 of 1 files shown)


dgw:
@Exirel I'd still like you to look at these changes, but I also want to merge this and #1781 together, in the next couple days, since they're _kinda_ related…ish.

test/test_db.py
```diff
@@ -35,6 +34,7 @@
 def db():
     config = MockConfig()
     config.core.db_filename = db_filename
+    config.core.db_type = 'sqlite'
```
Exirel: I need to update #1781 so it will uses `sqlite` in its new config.

dgw: `sqlite` is already the default: https://github.com/sopel-irc/sopel/blob/185686f92d299f95027f4dccf40ffd994e7807fd/sopel/config/core_section.py#L133-L135 Did you need to update something else besides the fake config?

Exirel: Nothing else. If it's already the default, then I've nothing to update!

PR #1776, opened 2019-12-06
db, docs: overhaul docstrings & add SQLAlchemy to intersphinx mappings

dgw:
Some of the stuff in `SopelDB` needed to reference SQLAlchemy docs, and adding those intersphinx mappings was trivially easy, so it was done.

PR #1778, opened 2019-12-07
config: doc overhaul

dgw:
Besides the absolute ton of new documentation stuff, this deprecates `config.ConfigSection.get_list()` (which is ancient, unused, and made obsolete by the `ListAttribute` type).

The same intersphinx mapping addition had to be made here as in #1776 to enable easy links to the SQLAlchemy documentation. I'll deal with the merge conflict when the time comes, if Git isn't smart enough to realize that both patches have the same result.

This also enables Sphinx's built-in `autosectionlabel` extension, which turns every section heading into a `:ref:` name in exchange for keeping every section title unique across the whole documentation project. I think it's more than a fair trade.

**Related:** #1565.

**Potential conflict:** #1751 — @Exirel, if you would, please tweak the format of those CoreSection docstring updates when you have time to address the pending change requests on that PR! 😸)

----

As an aside, I'm pretty proud of how this turned out. It's almost certainly my longest documentation overhaul to date, if memory serves, clocking in at over 4 hours spent messing with these two files to get them looking _just right_ and double-check all the inline references.

### DIFF
--- a/sopel/config/core_section.py
+++ b/sopel/config/core_section.py
@@ -14,7 +14,8 @@ from sopel.tools import Identifier
 def _find_certs():
     """Find the TLS root CA store.
 
-    :returns: str (path to file)
+    :returns: path to CA store file
+    :rtype: str
     """
     # check if the root CA store is at a known location
     locations = [
@@ -34,6 +35,11 @@ def _find_certs():
 
 
 def configure(config):
+    """Interactively configure the bot's ``[core]`` config section.
+
+    :param config: the bot's config object
+    :type config: :class:`~.config.Config`
+    """
     config.core.configure_setting('nick', 'Enter the nickname for your bot.')
     config.core.configure_setting('host', 'Enter the server to connect to.')
     config.core.configure_setting('use_ssl', 'Should the bot connect with SSL?')
@@ -56,7 +62,11 @@ def configure(config):
 
 
 class CoreSection(StaticSection):
-    """The config section used for configuring the bot itself."""
+    """The config section used for configuring the bot itself.
+
+    .. important::
+        All **Required** values must be specified, or Sopel will fail to start.
+    """
 
     admins = ListAttribute('admins')
     """The list of people (other than the owner) who can administer the bot."""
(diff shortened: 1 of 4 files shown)


sopel/config/core_section.py
```diff
@@ -111,120 +139,271 @@ class CoreSection(StaticSection):
...
+    * ``postgres``
+    * ``sqlite``
+    * ``sybase``
+
+    Here are the additional PyPI packages you may need to install to use one of
+    the most commonly requested alternatives:
+
+    mysql
+      ``pip install mysql-python`` (Python 2)
+
```
Exirel: Does this empty line is required?

sopel/config/core_section.py
```diff
@@ -111,120 +139,271 @@ class CoreSection(StaticSection):
...
+    """The default timezone to use for time in messages.
+
+    :default: ``UTC``
+
+    Used when plugins format times with :func:`sopel.tools.time.format_time`.
+    """
 
     enable = ListAttribute('enable')
-    """A whitelist of the only plugins you want to enable."""
+    """A whitelist of the only plugins you want to enable.
```
Exirel: It's not a whitelist, it's a simple list. A whitelist (or, as I prefer, a "safelist") is a list of allowed. In our case, it's a list of enabled and/or activated. Something can be allowed but not enabled, and something can be enabled but not allowed.

If you follow what I mean.

sopel/config/core_section.py
dgw: Unfortunately, yes. Without it, Sphinx considers this as a single wrapped paragraph:

image

I wasn't able to find a way of forcing a line-break in rST. :/

Would you rather see each term (`mysql`, `postgres`, `mssql`) preceded by a blank line? That still generates a clean `<dl>` but looks slightly better in source-code form.

sopel/config/core_section.py
dgw: Obviously you can see I didn't write this line myself. 😁 I'll `s/white//` it when I'm applying whatever you say to change (if anything) on the `<dl>` nitpick above.

sopel/config/core_section.py
Exirel: The more you know. Thanks for your answer! :+1:

dgw:
@Exirel During rebase, I touched docstrings for `commands_on_connect`, `throttle_join`, and `throttle_wait` only. If you want to re-review those, it shouldn't take more than a minute or two.

Exirel:
> If you want to re-review those, it shouldn't take more than a minute or two.

Looks great!

PR #1779, opened 2019-12-08
find_updates: error handling

### LINKED ISSUE #1643: find_updates: No error handling on JSON fetch
See this code:

https://github.com/sopel-irc/sopel/blob/b105fe4aaa6c1cd258337e60a4f17c1a0751ecb5/sopel/modules/find_updates.py#L49

There's no error-handling at all. If the JSON doesn't parse, Sopel will spit out an exception. If the JSON URL won't load for some reason (times out, connection gets reset, domain name expires, etc.), Sopel will spit out an exception. These are just examples.

This code really needs to be rewritten with a robust `try`/`catch` structure to catch as many possible error conditions as possible. It probably wouldn't have prevented e.g. #1433, but we still should gracefully handle failures in the update checker. At present the `latest.json` file Sopel checks is hosted on Netlify, with very good uptime, but the site is still (very) rarely unavailable.

Bonus points for implementing some kind of logic to detect several failed update checks in a row and also alert the bot's owner to that issue, in case there's a networking issue on Sopel's host machine, or a problem with the update endpoint.

dgw:
So I wrote this late one night, forgot to make a pull request for it, and here we are two weeks later. Maybe I held off making a PR, planning to come back and improve something first? Who knows.

Fixes #1643, at least in theory.

### DIFF
--- a/sopel/modules/find_updates.py
+++ b/sopel/modules/find_updates.py
@@ -39,13 +39,45 @@ def startup_version_check(bot, trigger):
         check_version(bot)
 
 
+def _check_succeeded(bot):
+    bot.memory['update_failures'] = 0
+
+
+def _check_failed(bot):
+    bot.memory['update_failures'] = 1 + bot.memory.get('update_failures', 0)
+
+
 @sopel.module.interval(wait_time)
 def check_version(bot):
     version = sopel.version_info
+    success = False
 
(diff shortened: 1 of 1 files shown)


dgw:
I forgot that `JSONDecodeError` isn't a normal built-in exception. 😟

Issue #946, opened 2015-11-22
Ensure running with a config file outside ~/.sopel works as expected


embolalia:
- [x] Config wizard should not create ~/.sopel if the specified config is not in that directory
- [x] Existing code should be checked to ensure nothing else assumes things in ~/.sopel
- [x] How to specify a config outside of ~/.sopel should be documented

Exirel:
I should have a look while working on #1429

dgw:
This issue gets a specific callout in [irctest](https://github.com/ProgVal/irctest)'s readme, and I think we would be well served to solve whatever remains of this issue for the Sopel 7.0 release. I know we've already done some homedir-related stuff, and if my hunch is correct we're already most of the way to closing this.

dgw:
Preliminary list of related PRs: #1404 (merged), #1574 (awaiting final review from yours truly), & #1598 (awaiting final tweaks from @Exirel). About the only unaddressed reference to `~/.sopel` I could find in my brief survey was a hard-coded reference in `test_tools` that should be updated to use the `config.DEFAULT_HOMEDIR` constant added way back in #1429. https://github.com/sopel-irc/sopel/blob/3404aa29f6395952d5f984c5fa2183a3cf6936a4/sopel/test_tools.py#L97

Exirel:
I need to double check on my end, but as @dgw said, it's almost done yes.

Exirel:
Fun fact, the rework of test tools in #1731 make the hard-coded home_dir unnecessary, because we can use a real `Sopel` instance instead of the `MockSopel`. So it's almost done: all that is required now is to 1) merge #1731 then 2) use the new tools in every tests.

dgw:
> So it's almost done: all that is required now is to 1) merge #1731 then 2) use the new tools in every tests.

FWIW, `test_tools` still has the same old hard-coded `home_dir` value. @Exirel is tweaking that even worth blocking close of this issue for? I feel like it isn't.

Exirel:
I'm working on that last part with #1781 and I'm probably going to remove both `MockSopel` and `MockSopelWrapper` (and also `MockConfig`).

Exirel:
> is tweaking that even worth blocking close of this issue for? I feel like it isn't.

Agree, we can totally close that issue once #1781 is merged, because Sopel won't use the deprecated code itself.

dgw:
Checked off the "existing code" task now that #1781 is merged. We should be golden… FINALLY! 🎉

PR #1781, opened 2019-12-09
test: test_tools's mocks are obsolete

Exirel:
This PR makes test_tools's mocks obsolete, adding deprecation warning on `MockConfig`, `MockSopel`, and `MockSopelWrapper` - replaced by the `configfactory` fixture, the `botfactory` fixture, and by the `sopel.bot.SopelWrapper` class respectively.

As far as I'm concerned, Sopel's codebase doesn't use `MockSopel` nor `MockConfig`, making sure that no part of Sopel assumes `~/.sopel/` as the homedir. We may close issue #946 once this is merged.

### DIFF
--- a/sopel/test_tools.py
+++ b/sopel/test_tools.py
@@ -39,6 +39,7 @@ if sys.version_info.major >= 3:
 
 
 class MockConfig(sopel.config.Config):
+    @sopel.tools.deprecated('use configfactory fixture instead', '7.0', '8.0')
     def __init__(self):
         self.filename = tempfile.mkstemp()[1]
         self.parser = ConfigParser.RawConfigParser(allow_no_value=True)
@@ -54,6 +55,7 @@ class MockConfig(sopel.config.Config):
 
 
 class MockSopel(object):
+    @sopel.tools.deprecated('use botfactory fixture instead', '7.0', '8.0')
     def __init__(self, nick, admin=False, owner=False):
         self.nick = nick
         self.user = "sopel"
@@ -115,7 +117,17 @@ class MockSopel(object):
 
 
 class MockSopelWrapper(SopelWrapper):
-    pass
+    @sopel.tools.deprecated('use sopel.bot.SopelWrapper instead', '7.0', '8.0')
+    def __init__(self, *args, **kwargs):
+        super(MockSopelWrapper, self).__init__(*args, **kwargs)
+
+
+TEST_CONFIG = """
+[core]
+nick = {name}
+owner = {owner}
+admin = {admin}
+"""
 
 
 def get_example_test(tested_func, msg, results, privmsg, admin,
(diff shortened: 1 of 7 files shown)


dgw:
I'm not a huge fan of needing to put `$nickname: ` into every example that uses `bot.reply()`. 🙁

Dropping `test_tools` without warning isn't entirely safe, FYI. It's not in the API docs, but it _is_ referenced in [our help tutorial](https://github.com/sopel-irc/sopel.chat/blob/a438b6d8b860530c71696186311f3e9c5547f546/_tutorials/part-4-the-help-command.md). That's basically "documented", and a quick code search of GitHub for "sopel test_tools" shows a number of third-party packages that use our test tools for their own plugin tests.

Exirel:
Ok, I'll do two thing:

* be more flexible with `$nickname` (and revert my changes about that) 
* add some deprecation warnings here and there

And I think we'll be good to go. As mentioned on IRC, internal plugins will become externals by Sopel 8.0, and we have a year to craft better testing tools. I'll remember that:

> 09:45:20 <+dgw> (I would love a testing framework that just auto-tests plugins without having to do the `__name__ == 'main'` crap though :D)

Exirel:
@dgw and done! It's way easier to review now I think.

sopel/test_tools.py
```diff
@@ -152,38 +172,51 @@ def test():
...
-            for result, output in zip(results, wrapper.output):
-                if type(output) is bytes:
-                    output = output.decode('utf-8')
+
+            output_triggers = (
+                sopel.trigger.PreTrigger(bot.nick, message.decode('utf-8'))
+                for message in wrapper.backend.message_sent
+            )
+            output_texts = (
+                # substract "Sopel: " when necessary
```
dgw:
```suggestion
                # subtract "Sopel: " when necessary
```

sopel/test_tools.py
```diff
@@ -152,38 +172,51 @@ def test():
...
+                pattern.sub('', output_trigger.args[-1])
+                for output_trigger in output_triggers
+            )
+            outputs = [text for text in output_texts if isnt_ignored(text)]
+
+            # output lenght
```
dgw:
```suggestion
            # output length
```

test/modules/test_modules_remind.py
```diff
@@ -8,17 +8,18 @@
...
-    bot.config.basename = 'default'
-    bot.config.core.owner = 'Admin'
-    bot.config.core.host = 'chat.freenode.net'
-    return bot
+TMP_CONFIG = """
+[core]
+owner = Admin
+nick = Sopel
+enable =
+    coretasks
```
dgw: "Explicit is better than implicit", or would `coretasks` not be loaded in the test without this?

test/test_db.py
```diff
@@ -15,7 +15,6 @@
 import pytest
 
 from sopel.db import SopelDB
-from sopel.test_tools import MockConfig
```
dgw: I requested your review on #1777 mostly because you're the testing expert, but it also works to have you look at that and determine if anything there will need adjusting after the changes here.

(code line unrelated; I just needed an anchor)

test/modules/test_modules_remind.py
Exirel: It's a way to prevent anything else from being loaded by mistake. It's not necessary right now, since `bot.setup` isn't called - but what if it is? But yes, this is not strictly necessary.

test/test_db.py
Exirel: Oh. I theory no: there might be a merge conflict but nothing I can't fix. The main change here affects the `db` fixture, so it's relatively small.

dgw:
@Exirel Oh wait, this still has a "post-review fix" commit that should be squashed out before merge. I guess you can take care of that _and_ glance at #1777 at the same time. 😁

Exirel:
lol

Exirel:
Done.

PR #1783, opened 2019-12-15
db: migrate to connectionless execution for raw sql

RustyBower:
I really hate being able to execute raw SQL against the database, but this should fix the issue for anyone needing to do it


dgw:
@RustyBower I ❤️ that you fixed this thing for me anyway, despite disagreeing with the feature, for the sake of API compatibility. Much appreciated!

Exirel:
Yeah good work here @RustyBower 

For Sopel 7.1 we might want to reconsider how Sopel exposes its access to the DB into something different, but we'll see that later.

PR #1795, opened 2019-12-30
setup: fix/update python version checks

half-duplex:
The check for python 3.3+ was broken, and python 2's last maintenance release is [now](https://mail.python.org/archives/list/python-dev@python.org/thread/APWHFYQDKNVYQAK3HZMBGQIZHAVRHCV2/) [scheduled](https://www.python.org/dev/peps/pep-0373/#maintenance-releases) for "mid-april" 2020.

### Checklist
- [x] No (related) issues are reported by `make qa`

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -34,16 +34,22 @@ else:
         sys.exit(1)
 
 if sys.version_info < (2, 7) or (
-        sys.version_info[0] > 3 and sys.version_info < (3, 3)):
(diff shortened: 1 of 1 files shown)


half-duplex:
Actually, what case is the manual python version check necessary in? `python sopel.py` wouldn't hit it, and any pip or setuptools installation should hit the python_requires.

PR #1800, opened 2020-01-08
setup, cli.run: parity between py2 checks & code cleanup

dgw:
This was @half-duplex's suggestion, so I'll specifically ask him to review. I ended up doing it because I got impatient, and the number 1800 was calling to me. 😛

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -43,10 +43,7 @@ if sys.version_info < (2, 7) or (
 # Py2 EOL: https://www.python.org/dev/peps/pep-0373/#maintenance-releases
 if sys.version_info.major == 2:
     now = time.time()
-    if now >= 1577836800:  # 2020-01-01 00:00:00 UTC
-        state = "is near end of life"
-    else:
-        state = "has reached end of life"
+    state = "has reached end of life"
     if now >= 1588291200:  # 2020-05-01 00:00:00 UTC
         state += " and will receive no further updates"
     print(
--- a/setup.py
+++ b/setup.py
@@ -33,6 +33,8 @@ else:
         )
         sys.exit(1)
 
+# We check Python's version ourselves in case someone installed Sopel on an
+# old version of pip (<9.0.0), which doesn't know about `python_requires`.
 if sys.version_info < (2, 7) or (
     sys.version_info.major >= 3 and sys.version_info < (3, 3)
 ):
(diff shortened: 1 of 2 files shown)


sopel/cli/run.py
```diff
@@ -21,15 +21,20 @@
 from sopel import bot, config, logger, tools, __version__
 from . import utils
 
+# We check Python's version ourselves in case someone installed Sopel on an
+# old version of pip (<9.0.0), which doesn't know about `python_requires`.
 if sys.version_info < (2, 7):
     tools.stderr('Error: Requires Python 2.7 or later. Try python2.7 sopel')
```
half-duplex: Not a problem introduced here, but backticks or words to indicate that `python2.7 sopel` is a command might be good.
Though, does that actually work, if sopel is installed to eg. python2.5's site_packages by pip? Or should it be changed to `python2.7 sopel.py`?

setup.py
```diff
@@ -33,18 +33,21 @@
...
     sys.version_info.major >= 3 and sys.version_info < (3, 3)
 ):
     # Maybe not the best way to do this, but this question is tiring.
     raise ImportError('Sopel requires Python 2.7+ or 3.3+.')
 # Py2 EOL: https://www.python.org/dev/peps/pep-0373/#maintenance-releases
 if sys.version_info.major == 2:
-    if time.time() >= 1577836800:  # 2020-01-01 00:00:00 UTC
-        state = 'is near end of life'
+    now = time.time()
+    if now >= 1577836800:  # 2020-01-01 00:00:00 UTC
```
half-duplex: Not introduced by this PR, but I think this logic is inverted, though since we're past the date already this conditional could be made static.

sopel/cli/run.py
```diff
@@ -21,15 +21,20 @@
...
     sys.exit(1)
 if sys.version_info.major == 2:
-    if time.time() >= 1577836800:  # 2020-01-01 00:00:00 UTC
+    now = time.time()
+    if now >= 1577836800:  # 2020-01-01 00:00:00 UTC
```
half-duplex: Same as other file - inverted? remove conditional?

sopel/cli/run.py
dgw: I'm not going to worry about it too much. Rather than try to figure out the correct command on behalf of the user, I'd just drop the "Try" bit and let them work it out.

sopel/cli/run.py
```diff
@@ -21,15 +21,20 @@
 from sopel import bot, config, logger, tools, __version__
 from . import utils
 
+# We check Python's version ourselves in case someone installed Sopel on an
+# old version of pip (<9.0.0), which doesn't know about `python_requires`.
```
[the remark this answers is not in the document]
dgw: Amended.

Issue #1804, opened 2020-01-17
Update URL callback registration mechanism


dgw:
Currently, only one callback can be stored per pattern in `bot.memory['url_callbacks']`. We should give serious thought to reworking this in a future version, 8.0 or 9.0. At the same time, we can officially remove the memory dict.

PR #1805, opened 2020-01-17
bot: Support multiple handlers for a URL pattern

half-duplex:
### Description
Implements #1804

This introduces `Sopel._url_callbacks`, changes the *_url_callback functions to use it, and returns both Sopel._url_callbacks and bot.memory["url_callbacks"] results in a search, maintaining the legacy functionality.

Because Sopel doesn't have a SopelMemory-type object for lists, this may have minor concurrency issues, but url_callbacks updates are infrequent enough I don't think it's deadly.

### Checklist

### DIFF
--- a/sopel/bot.py
+++ b/sopel/bot.py
@@ -57,6 +57,11 @@ class Sopel(irc.AbstractBot):
         }
         self._plugins = {}
 
+        self._url_callbacks = tools.SopelMemory()
+        """A mapping of compiled regexes to lists of callables.
+        Regexes are matched against found URLs, and each callable called.
+        """
+
         self.doc = {}
         """A dictionary of command names to their documentation.
 
@@ -325,13 +330,20 @@ class Sopel(irc.AbstractBot):
             self.unregister(func)
 
         # remove URL callback handlers
+        for func in urls:
+            regexes = func.url_regex
+            for regex in regexes:
+                if func in self._url_callbacks.get(regex, []):  # can .get ever fail?
+                    self.unregister_url_callback(regex, func)
+                    LOGGER.debug('URL Callback unregistered: %r', regex)
+        # and the old way, until 8.0
         if "url_callbacks" in self.memory:
             for func in urls:
                 regexes = func.url_regex
                 for regex in regexes:
                     if func == self.memory['url_callbacks'].get(regex):
                         self.unregister_url_callback(regex, func)
-                        LOGGER.debug('URL Callback unregistered: %r', regex)
+                        LOGGER.warning('Legacy URL Callback unregistered: %r', regex)
 
         # remove plugin from registry
         del self._plugins[name]
@@ -945,13 +957,14 @@ class Sopel(irc.AbstractBot):
         It's recommended you completely avoid manual management of URL
         callbacks through the use of :func:`sopel.module.url`.
         """
-        if 'url_callbacks' not in self.memory:
-            self.memory['url_callbacks'] = tools.SopelMemory()
-
         if isinstance(pattern, basestring):
             pattern = re.compile(pattern)
 
-        self.memory['url_callbacks'][pattern] = callback
+        if pattern not in self._url_callbacks:
+            # TODO: Ideally this would be a list version of SopelMemory
+            self._url_callbacks[pattern] = []
+
+        self._url_callbacks[pattern].append(callback)
 
     def unregister_url_callback(self, pattern, callback):
         """Unregister the callback for URLs matching the regex ``pattern``.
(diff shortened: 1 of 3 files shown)


Exirel:
I like the idea, and the code is quite good. Yet, I'd like to point out that we planned on removing `memory['url_callbacks']`, to prevent any manual access to this feature - as mentioned in the issue:

> At the same time, we can officially remove the memory dict.

For that, I think you can add a new attribute to `Sopel`, something like `Sopel.url_callbacks` maybe? It could be a `SopelMemory` object, where keys are compiled regex. Alternatively, it could be a protected attribute, like `_url_callbacks`.

half-duplex:
Thanks. Yeah, this was originally aimed for 7.0 which needs to keep manual management. Mostly redone now to be fully compatible with no API changes at all, so this maintains exact compatibility with the manual `memory["url_callbacks"]` way.
I like it more as a private attribute, since I can't think of any valid reason to directly work with it from a plugin, and when private we don't need to worry about compatibility if it requires any further changes.

sopel/bot.py
```diff
@@ -324,13 +329,20 @@ def remove_plugin(self, plugin, callables, jobs, shutdowns, urls):
             self.unregister(func)
 
         # remove URL callback handlers
+        for func in urls:
+            regexes = func.url_regex
+            for regex in regexes:
+                if func in self._url_callbacks.get(regex, []):  # can .get ever fail?
```
Exirel: The only reason this line could fail is if someone replace `_url_callbacks[regex]` by something that isn't a list/tuple:

```python console
>>> d = {}
>>> d['a'] = 2  # will fail because int
>>> d['b'] = []  # will work because list
>>> d['c'] = 'some str'  # will fail because iterable BUT wrong type
>>> None in d['a']
TypeError: argument of type 'int' is not iterable
>>> None in d['b']
False
>>> None in d['c']
TypeError: 'in <string>' requires string as left operand, not NoneType
```

sopel/bot.py
```diff
@@ -324,13 +329,20 @@ def remove_plugin(self, plugin, callables, jobs, shutdowns, urls):
...
+                    LOGGER.debug('URL Callback unregistered: %r', regex)
+        # and the old way, until 8.0
         if "url_callbacks" in self.memory:
             for func in urls:
                 regexes = func.url_regex
                 for regex in regexes:
                     if func == self.memory['url_callbacks'].get(regex):
                         self.unregister_url_callback(regex, func)
-                        LOGGER.debug('URL Callback unregistered: %r', regex)
+                        LOGGER.warning('Legacy URL Callback unregistered: %r', regex)
```
Exirel: I believe this is the right place and the right amount of warning. Nothing more needed.  :+1:

sopel/bot.py
half-duplex: Oh, I mean more "should it ever be possible for `regex` to not be in `self._url_callbacks` unless we've got a bug in some plugin handling code.

Exirel: I see. Well, that's why you can use `.get` with a default value. However, `[]` could totally fail indeed.

dgw:
I'm in favor of doing this for 7.1. Where are we on addressing Exi's feedback?

sopel/bot.py
half-duplex: `regex not in self._url_callbacks` is probably not a case we need to handle better than "don't crash", so resolving

sopel/test_tools.py
```diff
@@ -107,15 +107,17 @@ def register_url_callback(self, pattern, callback):
         if isinstance(pattern, basestring):
             pattern = re.compile(pattern)
 
-        self.memory['url_callbacks'][pattern] = callback
+        if pattern not in self._url_callbacks:
+            self._url_callbacks[pattern] = []
+        self._url_callbacks[pattern].append(callback)
```
Exirel, line 112: You should check if `callback` is not yet registered for that pattern.

sopel/test_tools.py
```diff
@@ -107,15 +107,17 @@ def register_url_callback(self, pattern, callback):
...
 
     def unregister_url_callback(self, pattern, callback):
         if isinstance(pattern, basestring):
             pattern = re.compile(pattern)
 
         try:
-            del self.memory['url_callbacks'][pattern]
-        except KeyError:
+            self._url_callbacks[pattern].remove(callback)
+        except ValueError:
```
Exirel, line 120: You might want to catch a `KeyError` here too.

sopel/bot.py
```diff
@@ -1016,10 +1037,16 @@ def search_url_callbacks(self, url):
...
+            match = regex.search(url)
+            if match:
+                for function in functions:
+                    yield function, match
+
+        # legacy, remove in 8.0
         if 'url_callbacks' not in self.memory:
             # nothing to search
             return
-
```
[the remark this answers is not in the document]
dgw: I think it was to make clear that the whole block should be removed in 8.0 per the above comment. I'm fine with or without the blank line here, but I understand the (suspected) reasoning behind deleting it.

dgw:
I suspect #1904 will replace this. Even if not, this will need some refactoring after the recent massive changes to how plugin rules work.

PR #1808, opened 2020-01-18
bot: require callback to unregister URL handler

half-duplex:
### Description
Introduces the API change #1805 will require. The test diff isn't particularly minimal because this is a reduction of that patchset for 7.0.

### Checklist

### DIFF
--- a/test/test_bot.py
+++ b/test/test_bot.py
@@ -8,6 +8,7 @@ import pytest
 
 from sopel import bot, plugins
 from sopel.tests import rawlist
+from sopel.tools import SopelMemory
 
 
 TMP_CONFIG = """
@@ -308,7 +309,7 @@ def test_search_url_callbacks_not_found(tmpconfig):
     assert not list(results), 'URL must not match any pattern'
 
 
-def test_register_url_callback_twice(tmpconfig):
+def test_register_url_callback_multiple(tmpconfig):
     """Test register_url_callback replace URL callbacks for a pattern."""
     test_pattern = r'https://(www\.)?example\.com'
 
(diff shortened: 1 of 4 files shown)


test/test_bot.py
```diff
@@ -308,51 +309,52 @@ def url_handler(*args, **kwargs):
...
         return None
 
-    def url_handler_replacement(*args, **kwargs):
+    def url_handler2(*args, **kwargs):
         return None
 
     sopel = bot.Sopel(tmpconfig, daemon=False)
-    sopel.register_url_callback(test_pattern, url_handler)
 
+    sopel.register_url_callback(test_pattern, url_handler)
```
dgw: Why?

test/test_bot.py
```diff
@@ -308,51 +309,52 @@ def url_handler(*args, **kwargs):
     assert not list(results), 'URL must not match any pattern'
 
 
-def test_register_url_callback_twice(tmpconfig):
-    """Test register_url_callback replace URL callbacks for a pattern."""
+def test_register_url_callback_multiple(tmpconfig):
+    """Test register_url_callback adding multiple callbacks for a pattern."""
```
dgw: This patch doesn't enable multiple callbacks for a pattern yet.

half-duplex: It doesn't, but the test still tests it. This test's behavior was reverted (in this branch) to the old test's behavior of ensuring that the second-added clobbered the first-added.

dgw: Fair enough on future-proofing the test name, I suppose. But there are a lot of changes below given that no behavior has changed. I don't want to waste time now reviewing stuff that doesn't apply to what we're shipping in 7.0.

test/test_bot.py
```diff
@@ -340,21 +341,53 @@ def url_handler(*args, **kwargs):
...
+    def url_handler(*args, **kwargs):
+        return None
+
+    sopel = bot.Sopel(tmpconfig, daemon=False)
+    sopel.unregister_url_callback(test_pattern, url_handler)
+    # no exception implies success
+
+
+def test_unregister_url_callback_manual(tmpconfig):
+    """Test unregister_url_callback removes a specific callback that was added manually"""
```
Exirel: Here, I'd add a note saying that this can be removed once we deprecate the usage of callback added manually - which will happen soon enough.

Because otherwise, the next dev will wonder why we test the internal behavior of (un)register URL callback when it's clearly not our business.

Exirel: And by that I mean that this test is very good and add value for backward compatibility purpose, but let's not forget why it's good.

dgw: Oh, agreed. Let's actually plan on moving the URL callback registry to a private member (`bot._url_callbacks`?) in 8.0, and stop supporting `bot.memory['url_callbacks']` at that time, along with removing this test.

dgw:
I'm glad of doing one last check in the code…

Highly unlikely that any code actually calls `MockSopel.unregister_url_callback()`, but let's not allow the signature to be wrong just because the class is deprecated.

dgw:
Overwrote the original merge commit with a new one after fixing the unrelated `ip` test failure (presumably caused by a change in MaxMind's latest database version).

PR #1809, opened 2020-01-22
bot: warn on deprecated memory["url_callbacks"] use

half-duplex:
### Description
Adds `SopelMemory.deprecated_keys` and uses it to warn when something accesses `bot.memory["url_callbacks"]`. This is the least gross way I could come up with.

Not sure how to document deprecated_keys, and not sure how to (or if we should bother to) suppress the deprecation warnings for normal operation of the core bot (many of these are removed by #1805). Also, should deprecated_keys be private?

### Checklist


half-duplex:
Well, uh. Is caplog not available on 3.3? Can I just skip the test on 3.3 or something? =\

test/test_bot.py
```diff
@@ -366,6 +366,29 @@ def url_handler(*args, **kwargs):
...
+    sopel = bot.Sopel(tmpconfig, daemon=False)
+    sopel.memory["url_callbacks"] = SopelMemory()
+
+    # register a callback manually
+    sopel.memory["url_callbacks"][re.compile(test_pattern)] = url_handler
+    results = list(sopel.search_url_callbacks("https://www.example.com"))
+    assert results[0][0] == url_handler, "Callback must be present"
+
+    for record in caplog.records:
+        if "url_callbacks" in record.text and "deprecated" in record.text:
```
dgw, line 386: This should (I didn't try it) fix the test on Python != 3.3:

```suggestion
        if "url_callbacks" in record.message and "deprecated" in record.message:
```

dgw:
So, now that I think about it, the "Legacy URL Callback" warnings in #1805 kinda obviate any need for this. What if that patch just got a bit more aggressive with the wording, and we hold off on this until 8.0 (and then remove even this warning in 9.0)?

I'm honestly on the fence about whether we should actually do this, since `bot.memory` is supposed to be for plugins' use only. That `url_callbacks` ever ended up in there is an historical accident. But it was also the "official" way to register URL handlers for quite a while, unfortunately.

half-duplex:
I don't think the warning in #1805 is sufficient, since it's only shown when things are unloaded, and no better way than this PR comes to mind for warnings when things are registered, which is when chances are best that a plugin dev sees it.
I also just realized I'm not sure a legacy plugin would even trigger 1805's warning, since unregister_url_callback() would never be called by a plugin using only bot.memory[]

dgw:
#1805 also logs a warning in `remove_plugin()`. (No easy way to link it that won't break if the PR is updated, that I can see.) The plugin doesn't have to call any of the new functions to trigger that.

Still true that it only fires when things are being unloaded, though. Maybe we do need this.

half-duplex:
The remove_plugin() warning depends on two things:
- shutdown() is called _after_ remove_plugin() or fails to unregister the url_callback bot.memory
- remove_plugin() has a way to know the registered callback belongs to the unloading plugin

I can't actually hit that warning right now even when I'm trying... I think # 2 is failing, that code might be dead?

dgw:
So, `bot.remove_plugin()` appears to be called only from the code paths related to reloading plugins (via `PyModulePlugin.unregister()`). Not dead, but irrelevant to this PR.

@Exirel can we talk you into looking at both this and #1805 as kind of a unit? 🤞 I hope this month.

Exirel:
I'm reading #1805 and it seems to me that it handle the right amount of warning, making this PR #1809 not that useful. It seems to me that this PR is a lot of work for something that is, ultimately, not that important.

dgw:
Let's have @Exirel look at this once we're all done with #1904. It's relevant.

dgw:
Is there still interest in tweaking this to address suggested tweaks, or should I close it?

half-duplex:
1: changed
2: I can imagine cases where different plugins by different authors would have issues that way, but we can cross that bridge if we get there. changed
3: I think I tried last time and gave up. I can't find a a way to get a tb without raising an exception, and even when I do, it doesn't give me a useful tb. Help?

dgw:
Requiring `pytest-capturelog` still won't work on py3.3. It turns out my next suggestion would solve that anyway: How about using [`warnings.warn()`](https://docs.python.org/3/library/warnings.html#warnings.warn) and the `FutureWarning` [category](https://docs.python.org/3/library/warnings.html#warning-categories)? (Note: `DeprecationWarning` would make sense but it's ignored by default.)

As for point 3, maybe this snippet of `tools.deprecated()` will help? https://github.com/sopel-irc/sopel/blob/005a53452fdd96dde1dd266b1587d67170665317/sopel/tools/__init__.py#L160-L162 Might have to use a different stack frame, though; I haven't tested anything on top of this PR, but Python's traceback very well might put `SopelMemory.__setitem__()` on there, which wouldn't be helpful to the user.

dgw:
There's still some cleanup to do (updating tests & requirements), but I wanted to push what I have so far, 98094f7.

I tweaked this to use `warnings`. The emitted log entry looks like (hand-wrapped for GH display):
```
[2020-10-25 14:11:52,027] py.warnings          WARNING  - /path/to/sopel/modules/bugzilla.py:45: FutureWarning:
Key "url_callbacks" is deprecated: Manual management of URL callbacks has been replaced with @url and the
register_url_callback() functions
  bot.memory['url_callbacks'] = tools.SopelMemory()
```

I've not (yet?) figured out a way to reformat the message, e.g. so the file and line number would appear above the offending code snippet (instead of at the beginning of the text), but the basic functionality is there.

Exirel (changes requested):
I think you've got the core idea here, and that's really great!

Now, to the nitpick phase: I think we should also add version + removed in arguments, to reflect the behavior of `tools.deprecated`. What do you think?

dgw:
> I think we should also add version + removed in arguments, to reflect the behavior of `tools.deprecated`. What do you think?

Agreed in principle. My evil side even thinks that it would be cool if `SopelMemory.__setitem__()` failed when Sopel's current version is at least the deprecated key's `removed_in` version. 😈 (With an exception, not silently. I'm not _that_ evil!)

half-duplex:
pytest shows warnings generated by bot.py's initialization of memory["url_callbacks"]. I don't see a way to only ignore the warning in that one spot.

Also, hold the phone. @Exirel, legacy plugins use this pattern:
```py
if "url_callbacks" not in bot.memory:
    bot.memory["url_callbacks"] = SopelMemory()
```
And we initialize url_callbacks in bot.py, so plugins will never hit a warning in `__setitem__()`. So we need to either warn on access (maybe directly call `memory.__getitem__(key, warn=False)` from inside sopel to denoise it?) or set url_callbacks to a SopelMemory with _all keys_ deprecated.

Exirel:
> And we initialize url_callbacks in bot.py, so plugins will never hit a warning in `__setitem__()`. So we need to either warn on access (maybe directly call `memory.__getitem__(key, warn=False)` from inside sopel to denoise it?) or set url_callbacks to a SopelMemory with _all keys_ deprecated.

Yup, that's why I don't think deprecating *keys* is the path to successfully warns plugin author about the problem, and that it should be more like a check done after a plugin is loaded, saying "heh, you tried to register some URLs the **old** way, and you should not".

dgw:
I'd like to get Exi's input on the current approach before I dive in. The real test will be whether this works on all the Python releases supported by 7.1, though. The last push skipped CI, so we're missing that data for now.

half-duplex:
Exi suggested on IRC that it might be more straightforward to just check the contents of url_callbacks before and after each plugin setup() call. That'll miss callbacks added outside there, but is probably sufficient.

dgw:
This looks replaced by #2033.

PR #1811, opened 2020-01-24
db: raw sql functionality and session object api

RustyBower:


dgw:
Right, I spammed warnings all over the place. Hopefully Exirel will have time to look in the next few days.

sopel/db.py
```diff
@@ -180,9 +185,44 @@ def __init__(self, config):
...
+
+        """
+        if self.type != 'sqlite':
+            # log non-sqlite uses of raw connections for troubleshooting, since
+            # unless the developer had a good reason to use this instead of
+            # `session()`, it indicates the plugin was written before Sopel 7.0
+            # and might not work right when connected to non-sqlite DBs
+            LOGGER.info(
+                "Raw connection requested when db_type is not sqlite\n%s",
+                traceback.format_list(traceback.extract_stack()[:-1])[-1][:-1])
```
Exirel: I was about to say "use `stack_info`", but sadly, this feature has been added in Python 3.2 and is not available in Python 2.7 (yet another reason to complain about it).

So, I'd suggest to add a `# TODO: replace by 'stack_info=True' when Python 2.7 is dropped` above that line.

sopel/db.py
```diff
@@ -180,9 +185,44 @@ def __init__(self, config):
...
+           build (or convert) your plugin to use the SQLAlchemy ORM.
+
+        """
+        if self.type != 'sqlite':
+            # log non-sqlite uses of raw connections for troubleshooting, since
+            # unless the developer had a good reason to use this instead of
+            # `session()`, it indicates the plugin was written before Sopel 7.0
+            # and might not work right when connected to non-sqlite DBs
+            LOGGER.info(
+                "Raw connection requested when db_type is not sqlite\n%s",
```
Exirel: We should add: `Please use 'db.session()' instead to use an SQL Alchemy session`.

sopel/db.py
```diff
@@ -109,10 +114,10 @@ class SopelDB(object):
     def __init__(self, config):
         # MySQL - mysql://username:password@localhost/db
         # SQLite - sqlite:////home/sopel/.sopel/default.db
-        db_type = config.core.db_type
+        self.type = config.core.db_type
```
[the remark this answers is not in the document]
Exirel: And also the config option is `db_type`, which is a fine name.

sopel/db.py
```diff
@@ -180,9 +185,44 @@ def __init__(self, config):
...
+           one type of database, you probably want to use :meth:`session` and
+           build (or convert) your plugin to use the SQLAlchemy ORM.
+
+        """
+        if self.type != 'sqlite':
+            # log non-sqlite uses of raw connections for troubleshooting, since
+            # unless the developer had a good reason to use this instead of
+            # `session()`, it indicates the plugin was written before Sopel 7.0
+            # and might not work right when connected to non-sqlite DBs
+            LOGGER.info(
```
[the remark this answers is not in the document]
dgw: We started with `warning` when discussing this and wound up at `info`.

sopel/db.py
dgw: I stole it from here: https://github.com/sopel-irc/sopel/blob/c5cffdb5443c5b8b3eda99a8c17fe1fdd3d3314d/sopel/tools/__init__.py#L350-L351

Using `stack_info=True` would give us the whole stack. We don't want the whole stack. We want just the most recent call, which is what this snippet yields.

sopel/db.py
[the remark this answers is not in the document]
dgw: An object attribute can be named `type`, however, without conflicting anywhere. It will always be accessed as `self.type` or `db.type` or `bot.db.type`. Writing `bot.db.db_type` looks redundant.

sopel/db.py
dgw: Rewriting the message rn.

dgw:
Updated the warning in `db.connect()`; it now looks like this:

```
[2020-01-24 16:47:33,258] sopel.db             INFO     - Raw connection requested when 'db_type' is not 'sqlite':
Consider using 'db.session()' to get a SQLAlchemy session instead here:
  File "/usr/local/lib/python3.6/dist-packages/sopel_modules/github/webhook.py", line 49, in setup_webhook
    conn = sopel.db.connect()
```

sopel/db.py
Exirel: Well, it's hard to know in what context the call is made when you don't have the full context - that's what my experience tells me. But heh. I know you don't like traceback as much as I do. :p

sopel/db.py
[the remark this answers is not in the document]
Exirel: Oh well, I tried.

PR #1815, opened 2020-02-05
allow commas when converting currency

rj1:
### Description

when using currency converter, users naturally like to type commas in the amount (e.g. .cur 1,000,000 USD to CAD) but this causes an error. this pull request simply removes commas before sending it to currency APIs

### Checklist

### DIFF
--- a/sopel/modules/currency.py
+++ b/sopel/modules/currency.py
@@ -106,7 +106,7 @@ def update_rates(bot):
 
 
 def get_rate(base, target):
-    base = base.upper()
+    base = base.upper().replace(',', '')
     target = target.upper()
 
     if base not in rates:
--- a/sopel/modules/currency.py
+++ b/sopel/modules/currency.py
@@ -23,7 +23,7 @@ FIAT_URL = 'https://api.exchangeratesapi.io/latest?base=EUR'
 FIXER_URL = 'http://data.fixer.io/api/latest?base=EUR&access_key={}'
 CRYPTO_URL = 'https://api.coingecko.com/api/v3/exchange_rates'
 EXCHANGE_REGEX = re.compile(r'''
-    ^(\d+(?:\.\d+)?)                                            # Decimal number
+    ^[+-]?[0-9]{1,3}(?:,?[0-9]{1})*(?:\.[0-9]{2})?             # Decimal number
     \s*([a-zA-Z]{3})                                            # 3-letter currency code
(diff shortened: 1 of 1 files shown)


Exirel:
Hi!

Thanks for your interest in Sopel. I've read your PR but I believe there is a problem.

> this pull request simply removes commas before sending it to currency APIs

The diff I've just read tries to replace the comma in the *base*, which isn't the amount, but the currency used for the amount:

* `100 EUR in USD`
* amount = 100
* base = EUR
* target = USD

After further reading, I believe the `EXCHANGE_REGEX` pattern need to be updated to fit your need.

rj1:
hey there, I've made the changes now and verified that is indeed working

dgw:
Well, it _looks_ functional. I have reservations about doing this, though. If we support one alternative format, that could lead to a snowball effect of requests to support yet more (making the regex ever more complicated and/or necessitating the use of multiple patterns).

Some applications/languages use whitespace, or `_`, or `'`, or [the list goes on] for digit grouping. Sometimes `,` is used as radix point instead of `.`, and `.` (or something else) is used for digit grouping. It gets complicated trying to correctly handle everyone's local language.

I'm actually surprised @Exirel didn't mention this, given how numbers are written in French.

Maybe we'd be better off supporting both `.` and `,` as radix points instead, like `calc` does? (Maybe without the snarky comment? 😹) https://github.com/sopel-irc/sopel/blob/85c6a58aa8923922b125e0f65eb0ba5f8c196dd5/sopel/modules/calc.py#L27-L28

half-duplex:
Maybe set `[,.]\d\d$` (or `\d\d?`? to use a period and only remove other `[,.]`s? No idea how e.g. arabic works though, and this would also fall apart for cryptocurrencies... And if anywhere writes eg. $5.5
Do we support locales at all? That or a config option for the module seem like the only way to speak 1.000,00 somewhat reliably.

Exirel:
What you really need is [Babel's decimal parsing](http://babel.pocoo.org/en/latest/api/numbers.html#babel.numbers.parse_decimal).

On the other hand, I more or less agree with @dgw about not trying to manage every possible cases. At least, not right now, not until the currency plugin is extracted from Sopel (as it is the plan for plugins).

dgw:
I still haven't come around on this. The more variants we support, the more people will expect this plugin to "just work" no matter what they put in. It's better to have one "right way" to input numbers, and just teach users what that is.

PR #1827, opened 2020-03-30
reddit: Allow `%` in post URLs

### LINKED ISSUE #1826: reddit: URL-encoded characters don't match in links
In the "post" pattern, for example, `\w` won't match the `%` in a URL-encoded character:
https://github.com/sopel-irc/sopel/blob/7cdefb91597cd7f995bd59eac3c7bbcc52da1964/sopel/modules/reddit.py#L42

Of course, anywhere a post's "slugified" title (not sure what else to call it) should allow `%`, too. We should be fine with other stuff. IDs only contain alphanumeric characters, subreddit names must be alphanumeric (plus `_`), and usernames are alphanumeric plus `-` and `_` (already covered).

dgw:
Fixes #1826

### DIFF
--- a/sopel/modules/reddit.py
+++ b/sopel/modules/reddit.py
@@ -39,7 +39,7 @@ else:
 
 domain = r'https?://(?:www\.|old\.|pay\.|ssl\.|[a-z]{2}\.)?reddit\.com'
 subreddit_url = r'%s/r/([\w-]+)/?$' % domain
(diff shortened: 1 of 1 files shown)


dgw:
[Travis builds all passed](https://travis-ci.org/github/sopel-irc/sopel/builds/668680799). Something's broken in the status-reporting pipeline, apparently.

PR #1840, opened 2020-04-08
bot: fix per-channel configuration by plugin names

### LINKED ISSUE #1839: Per-channel configuration doesn't work as expected/documented
### Description
It is not possible (any longer?) to use just the plugin name in per-channel configuration.

### Reproduction steps
1. Set up a per-channel configuration as documented, for example:
   ```
   [#channel]
   disable_plugins = emoticons
   disable_commands = {"url": ["title_auto"]}
   ```
2. Start bot, and make sure it is in `#channel`
3. Nothing in `#channel` is any different from before.

### Expected behavior
Per the intended functionality of this feature, and according to the website documentation, this configuration should prevent the use of all `emoticons` commands and block automatic link title fetching in the `#channel` channel.

### Environment
- Sopel `.version`: 7.0.1.dev0
- Sopel installed via: current `7.0.x` branch
- Python version: 3.6.9
- Operating system: Ubuntu 18.04 (WSL)
- IRCd `/version`: n/a
- Relevant plugins: n/a; this is a core issue

### Notes
The problem stems from use of `func.__module__` in the code that applies these restrictions: https://github.com/sopel-irc/sopel/blob/c705f450e56b9ddded10226c640a607249e33710/sopel/bot.py#L567-L587 In the past, I believe that `func.__module__` used to be just `pluginname` for plugin callables, but that is clearly no longer the case. Now, `func.__module__` for a core plugin like `url` is the full module path, `sopel.modules.url`. (It probably changed with #1479 and associated work, but I haven't verified that with a `git bisect` or anything. Circumstantially, #1235 implemented this long before #1479 altered the plugin mechanics, and multiple reports confirm that this *used to* work as documented.)

### Initial implementation ideas
The immediately obvious solution is to add an attribute to each plugin callable during loading, containing the plugin name as expected by this feature's code. Presumably, being able to tell which plugin a callable came from would be useful for future features too.

Alternatively, the bot's internal tracking of loaded plugins could include a way to translate module names into plugin names. I'm partial to the first solution because retrieving an attribute of the function currently being processed is probably cheaper from a performance perspective, and logically clearer than performing a lookup on some (probably) dict stored in the bot's global state.

But either idea would work for the current version. I really don't want to wait to fix this until @Exirel's very admirable goal of refactoring plugins from "bunch of callables with attributes" to "actual Plugin object with state & behavior" happens in 8.0 or 9.0. 😅

dgw:
### Description
Annotate callables with the source plugin name during plugin registration. Use that name when applying per-channel configuration settings instead of the Python module name.

Old solution: ~~Translates plugin names given in the config to the module names that would/will be returned by `func.__module__`, and uses those when filtering per-channel configuration.~~

Resolves #1839.

### Checklist
  - I can't reasonably run `make quality` locally due to other long-term work on adding new flake8 checks that currently generate hundreds of warnings, so I'll leave this unchecked even though `make test` does pass.
  - ~~Most of it. I made sure it works for core plugins and entry-point plugins, and the debug log output of module names I used to get this far tells me it _should_ also work for `sopel_modules` plugins and single-file plugins (a few of which I have on my test install). The only wildcard should be folder plugins.~~ This disclaimer does not apply to the replacement `plugin_name` attribute solution.

### Other
_The below comments are no longer relevant, but are kept for posterity._

~~I fully expect at least one method name to get bikeshedded. That's fine. My primary concern is making sure that the base principle of this change isn't complete lunacy. 😹~~

~~I'm also fully aware that @Exirel disagrees with this approach on principle—but I'm hoping that having a PR in which to discuss it will at least push us toward either accepting this one (perhaps with modifications) or developing a different kind of fix that we can ship relatively soon.~~

### DIFF
--- a/sopel/bot.py
+++ b/sopel/bot.py
@@ -574,16 +574,28 @@ class Sopel(irc.AbstractBot):
 
                 # if "*" is used, we are disabling all plugins on provided channel
                 if '*' in disabled_plugins:
+                    LOGGER.debug(
+                        "All plugins disabled in %s; skipping execution of %s.%s",
+                        trigger.sender, func.plugin_name, func.__name__
+                    )
                     return
-                if func.__module__ in disabled_plugins:
+                if func.plugin_name in disabled_plugins:
+                    LOGGER.debug(
+                        "Plugin %s is disabled in %s; skipping execution of %s",
+                        func.plugin_name, trigger.sender, func.__name__
+                    )
                     return
 
             # disable chosen methods from plugins
             if 'disable_commands' in channel_config:
                 disabled_commands = literal_eval(channel_config.disable_commands)
 
-                if func.__module__ in disabled_commands:
-                    if func.__name__ in disabled_commands[func.__module__]:
+                if func.plugin_name in disabled_commands:
+                    if func.__name__ in disabled_commands[func.plugin_name]:
+                        LOGGER.debug(
+                            "Skipping execution of %s.%s in %s: disabled_commands matched",
+                            func.plugin_name, func.__name__, trigger.sender
+                        )
                         return
 
         try:
--- a/sopel/plugins/handlers.py
+++ b/sopel/plugins/handlers.py
@@ -36,6 +36,7 @@ from __future__ import unicode_literals, absolute_import, print_function, divisi
 import inspect
 import imp
 import importlib
+import itertools
 import os
 
 from sopel import loader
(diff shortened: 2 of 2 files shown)


Exirel:
I found what I disliked here, because the code is good but I wasn't happy about it. For me the problem is that it creates a hard coupling between a Plugin and a Python module, which is exactly what I wanted to avoid with entrypoint. It looks fine for now, because a) it fixes a bug that is properly annoying and b) it works fine with the current code without having to break anything. However, it has the side effect of blocking future options.

So that's a no for me. I've added more information on the bug #1839 about the context, and my suggestion solution.

Sorry @dgw to delay the resolution further. :(

dgw:
@Exirel I think you'll be much happier with the replacement solution I just force-pushed. Or at least I hope you are, because the new approach is based directly on your suggestions! 😛

Issue #1824, opened 2020-03-25
tools.time feature: "Raw" mode for seconds_to_human()


dgw:
It's cute that `tools.time.seconds_to_human()` takes care of the future/past grammar for you (adding "in&nbsp;" to the front or "&nbsp;ago" to the back, depending on the delta's sign). But it makes using that function really annoying in some cases, when it should be the perfect tool.

For the use cases I'm looking at, the most useful thing would be a "raw" mode that returns the intermediate list of units, rather than the fully-formatted string. In implementing this, we could actually simplify the code around here and eliminate the `split(', ')` below it: https://github.com/sopel-irc/sopel/blob/3959137b8e91ab392d7dccecec5cebd5765b8f0d/sopel/tools/time.py#L243-L250

Related: #1798, #1799. Not all of these will necessarily be implemented. Right now they're all just ideas, and picking the right one or two could make the API flexible and intuitive enough that we don't need to bake every little feature into it.

dgw:
This actually should be low-hanging fruit, and I'll self-assign it for the last two weeks of Hacktoberfest to (hopefully) motivate myself to take care of it.

Exirel:
I did that for you.

dgw:
Aww, I forgot about this one after all. 😩

Issue #1799, opened 2020-01-08
Feature: Abbreviated humanized time output


dgw:
It'd be pretty damn nice if `tools.time.seconds_to_human()` could output abbreviated units on demand. If I think about it for more than a few seconds, though, it seems more likely that we'll need a full-on format specification to accommodate different preferences—whether the abbreviations should contain `.`, whether to use e.g. `sec` or `secs` (or just `s`)…

Tangentially related to #1798, if only because we'll need support in this feature for both "fuzzy" and exact output modes.

half-duplex:
Do we actually need more flexibility than smallest_unit = s/m/h/d/y? Months are dubious because of differing day counts, weeks are of debatable utility, and I don't want to have to build/maintain a formatter.

Exirel:
Or maybe what we could split `seconds_to_human` in two: the first part would give a tuple of (y, m, d, h, min, s, ms), and the second part would take such a tuple and return the "human readable" version.

For instance, this is the first part:

```
    years = secs // 31536000
    months = (secs - years * 31536000) // 2635200
    days = (secs - years * 31536000 - months * 2635200) // 86400
    hours = (secs - years * 31536000 - months * 2635200 - days * 86400) // 3600
    minutes = (secs - years * 31536000 - months * 2635200 - days * 86400 - hours * 3600) // 60
    seconds = secs - years * 31536000 - months * 2635200 - days * 86400 - hours * 3600 - minutes * 60
```

And this is the second part:

```
    result = ", ".join(filter(lambda x: bool(x), [
        "{0} {1}".format(years, years_text) if years else "",
        "{0} {1}".format(months, months_text) if months else "",
        "{0} {1}".format(days, days_text) if days else "",
        "{0} {1}".format(hours, hours_text) if hours else "",
        "{0} {1}".format(minutes, minutes_text) if minutes else "",
        "{0} {1}".format(seconds, seconds_text) if seconds else ""
    ]))
```

RustyBower:
I've pretty successfully used https://arrow.readthedocs.io/en/latest/#humanize to create "humanized" output from a datetime object

half-duplex:
Exirel - Years and months are still not consistent lengths, so unless we know a start or end date, they cannot be correct.

Exirel:
> Years and months are still not consistent lengths

What do you mean?

dgw:
[...]

Issue #1798, opened 2020-01-03
Feature: Higher precision mode for `tools.time.seconds_to_human()`


dgw:
What it says on the tin. This is one of the missing features that prevents me from using the new function in some of my plugins.

dgw:
I have updated the OP with more details because "What it says on the tin" was absolutely _not_ clear to me after 9 months. But I really don't think this is happening for 7.1. Let's just push it to 8.0.

PR #1843, opened 2020-04-13
tools.time: zero is a special case for `seconds_to_human()`

### LINKED ISSUE #1841: tools.time: `seconds_to_human()` doesn't handle 0 well
### Description
`tools.time.seconds_to_human()` returns just `" ago"` when passed something that `int()` parses to 0 (i.e. -0.999... through 0.999...)

### Expected behavior
A sensible return value would be `"0 seconds ago"`.

Other values might also make sense (`"just now"`, `"very recently"`, etc.) but they don't necessarily work as well in the same contexts as the expected return format for non-zero inputs.

### Environment
- Sopel `.version`: 7.0.1

dgw:
Patch looks big, but that's because Python doesn't have braces. 😏 (Just kidding, I'd update the indentation even if it did. I'm not a monster.)

### Description
Fixes #1841.

Setup:
```
>>> from sopel.tools.time import seconds_to_human
```

Before:
```
>>> seconds_to_human(0)
' ago'
```

After:
```
>>> seconds_to_human(0)
'0 seconds ago'
``` 

### Checklist
  - Caveats apply as with all my recent PRs: `make quality` is currently useless in my local environment

### DIFF
--- a/sopel/tools/time.py
+++ b/sopel/tools/time.py
@@ -226,30 +226,35 @@ def seconds_to_human(secs):
     secs = int(secs)
     secs = abs(secs)
 
-    years = secs // 31536000
-    months = (secs - years * 31536000) // 2635200
-    days = (secs - years * 31536000 - months * 2635200) // 86400
-    hours = (secs - years * 31536000 - months * 2635200 - days * 86400) // 3600
-    minutes = (secs - years * 31536000 - months * 2635200 - days * 86400 - hours * 3600) // 60
-    seconds = secs - years * 31536000 - months * 2635200 - days * 86400 - hours * 3600 - minutes * 60
+    if secs == 0:
+        # zero is a special case that the algorithm below won't handle correctly (#1841)
+        result = "0 seconds"
+    else:
+        years = secs // 31536000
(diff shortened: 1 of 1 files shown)


Exirel:
What about this instead?

```python
if secs == 0:
    return 'just now'
```

dgw:
@Exirel Yes in theory, but in practice it doesn't work in many of the places I and others are using this function. It's expected to give time units, and the language around it tends is often written in such a way that "just now" would be grammatically or syntactically wrong. Not sure what to do about those.

Exirel:
> the language around it tends is often written in such a way that "just now" would be grammatically or syntactically wrong

Then I'd say help yourself and make it a function's argument with a default value (I'm suggesting this "just now" but I don't have your use-case in mind, any default will be fine by me):

```python
def fn(secs, right_now='just now'):
    if secs == 0:
        return right_now

    # here, you can do whatever you need to do when time isn't up
```

dgw:
I'll have to think about that in relation to the other future ideas (#1798, #1799, #1824) for this function.

Exirel:
That's only fair then. Let's keep my suggestion for later.

Issue #1812, opened 2020-01-25
coretasks: make URL handling respect per-channel config


dgw:
We removed the `unblockable` status from `handle_url_callbacks()` in #1806, so it will respect the ignored status of users matching entries in the block list (`.blocks` command set). We should further enhance this function so certain URL handlers (plugins or specific callables) can be disabled using the per-channel settings introduced by #1235.

URL callable handling directly invokes the matching function: https://github.com/sopel-irc/sopel/blob/c5cffdb5443c5b8b3eda99a8c17fe1fdd3d3314d/sopel/coretasks.py#L1009-L1012

Normal callable handling goes through `bot.call()`, which enforces per-channel restrictions: https://github.com/sopel-irc/sopel/blob/c5cffdb5443c5b8b3eda99a8c17fe1fdd3d3314d/sopel/bot.py#L565-L585

dgw:
Closing as fixed by #1845.

PR #1845, opened 2020-04-14
bot: route URL callbacks through `bot.call()` too

dgw:
### Description
With this change, URL callbacks are no longer called outside the enforcement area of rate-limits, per-channel configuration, etc.

See #1812.

I'm initially ignoring the checklist below because I expect @Exirel will look at this Soon™. He'll either use it as a starting point and add tests, or propose an alternative implementation (and add tests).

### Checklist

### DIFF
--- a/sopel/bot.py
+++ b/sopel/bot.py
@@ -567,6 +567,10 @@ class Sopel(irc.AbstractBot):
         # if channel has its own config section, check for excluded plugins/plugin methods
         if trigger.sender in self.config:
             channel_config = self.config[trigger.sender]
+            LOGGER.debug(
+                "Evaluating configuration for %s.%s in channel %s",
+                func.plugin_name, func.__name__, trigger.sender
+            )
 
             # disable listed plugins completely on provided channel
             if 'disable_plugins' in channel_config:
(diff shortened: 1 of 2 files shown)


sopel/coretasks.py
```diff
@@ -1009,4 +1009,4 @@ def handle_url_callbacks(bot, trigger):
         for function, match in bot.search_url_callbacks(url):
             # trigger callback defined by the `@url` decorator
             if hasattr(function, 'url_regex'):
-                function(bot, trigger, match=match)
+                bot.call(function, bot, trigger, match=match)
```
Exirel, line 1012: You can remove the `match` argument by doing this instead:

```
@wraps(function)
def decorated(bot, trigger):
    return function(bot, trigger, match=match)
```

And voilà. Nothing to do in `bot.call` now.

sopel/bot.py
```diff
@@ -567,6 +575,10 @@ def call(self, func, sopel, trigger):
         # if channel has its own config section, check for excluded plugins/plugin methods
         if trigger.sender in self.config:
             channel_config = self.config[trigger.sender]
+            LOGGER.debug(
```
Exirel, line 578: Keep that log though. 👍

dgw:
@Exirel you earned yourself a co-author tag for that wrapping decorator suggestion. _Magnifique!_

PR #1848, opened 2020-04-20
xkcd: suppress link if triggered by a URL match

### LINKED ISSUE #1847: xkcd outputs link even if triggered by one
The `xkcd` plugin shouldn't output a link to the comic if it was triggered by a link.

See `wikipedia`, for example. This is an anti-cycling measure in case multiple bots handling the same link(s) are present in the channel.

dgw:
This is the smallest patch I could devise, at the expense of being kinda messy. Ideally only one function would need to handle the `commanded` argument, but due to how the code is already structured it has to get passed around a bit first.

We could probably do with a small rewrite of this plugin, but not for a bugfix release. It might happen later if someone feels like it.

Closes #1847.

### Checklist
  - [Insert my recent standard disclaimer about local flake8 warnings]

### DIFF
--- a/sopel/modules/xkcd.py
+++ b/sopel/modules/xkcd.py
@@ -96,7 +96,7 @@ def xkcd(bot, trigger):
     say_result(bot, requested)
 
 
-def numbered_result(bot, query, latest):
+def numbered_result(bot, query, latest, commanded=True):
     max_int = latest['num']
     if query > max_int:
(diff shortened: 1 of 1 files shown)


dgw:
/me grumbles about Travis-CI being inconsistent about whether it picks up base branch changes before processing a rebuild request

I'll tinker with it. If I can get it to go green without touching the PR branch, I'll be happy. If not… I'll give up eventually. 😝

PR #1852, opened 2020-04-27
search: some Bing fixes (UA spoofing, example test, regex)

dgw:
### Description
Bing results look different than they used to; the HTML has changed a little. Seems like Bing also sends fake "No results" pages to the default requests user-agent, so we now spoof for both DDG and Bing.

Not knowing how long this was broken, I added an example test for Bing. Hopefully it will help us catch future breakage, while not becoming an annoyance. If it starts failing at random too often, I'll nuke it.

This should also fix the xkcd plugin not being able to find comics by keyword, since it relies on the search plugin's Bing function.

### Checklist
  - I ran tests with `coverage3` and `-k 'test_example_bing'`, so I only paid attention to tests directly related to this change.

### DIFF
--- a/sopel/modules/search.py
+++ b/sopel/modules/search.py
@@ -18,6 +18,11 @@ from sopel.module import commands, example
 from sopel.tools import web
 
 
+header_spoof = {
+    'User-Agent': 'Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/55.0.2883.87 Safari/537.36'
+}
+
+
 def formatnumber(n):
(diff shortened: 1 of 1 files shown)


dgw:
I already had to kick two of the seven CI builds for this, but only one of those was actually Bing. (The other was some weird-ass mismatch in a `rand` test, which is a whole other can of worms.)

Still, though, I'm not sure if this test is a great idea, given my comments in 484aa3867b205537e78f7ed72aa8d7a943304b05 (Aug 2018).

dgw:
Hoping for #1853 to go in first, in which case I'll update this to take advantage of recorded HTTP requests and bypass the random "No results" issue entirely.

PR #1854, opened 2020-05-01
search: Bing regex needs to be less specific

dgw:
### Description
The OL element isn't always in the same place. Sometimes there are other things between it and the first LI.

### Checklist

### DIFF
--- a/sopel/modules/search.py
+++ b/sopel/modules/search.py
@@ -31,7 +31,7 @@ def formatnumber(n):
     return ''.join(parts)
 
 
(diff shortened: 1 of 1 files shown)


dgw:
> In the long run, there is no way to know if this will work forever. We have to accept that it might fail sometimes, or at some point in the future.

Indeed, it's hard to test this reliably without some layer to insulate our tests from the frequent errors/blocks Bing sends back when hit from a datacenter IP. (I'm speculating this is why our Bing tests are so flaky, but I can't ever trigger the flakiness from home.)

You know I'm thinking about how to make such a layer as in #1853 work while still catching actual upstream changes, and I hope you'll participate in that—on a day when you have more motivation, maybe. 😸

PR #1855, opened 2020-05-01
loader: make sure URL callbacks are ready to pass through `bot.call()`

dgw:
### Description
Non-admins hit an AttributeError on URL callbacks: 'function' object has no attribute 'unblockable'. The loader wasn't passing them through `clean_callable` to apply default attributes during `clean_module`.

This could be done without a new "limitable" category (just let URL callbacks live under the "triggerable" umbrella), but... it's more fun to micro-optimize memory usage for no reason.

Also: Fixed a stray assertion in loader tests that was checking a variable's identity property instead of equality.

### Checklist

### DIFF
--- a/test/test_loader.py
+++ b/test/test_loader.py
@@ -3,6 +3,7 @@
 from __future__ import unicode_literals, absolute_import, print_function, division
 
 import inspect
+import re
 
 import pytest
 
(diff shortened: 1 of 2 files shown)


test/test_loader.py
```diff
@@ -156,7 +173,7 @@ def test_clean_module_idempotency(testplugin, tmpconfig):
     assert new_callables == callables
     assert new_jobs == jobs
     assert new_shutdowns == shutdowns
-    assert new_urls == new_urls
+    assert new_urls == urls
```
Exirel, line 176: LOL

Issue #1001, opened 2016-01-16
Half of the user commands lacks argument checking


mossarelli:
And there's no guidelines for how each function ends. So I have to add:

```
if not trigger.group(2):
        bot.say("Use the command like this: .kickban [#chan] user1 user!*@* message")
        return
```

To every command in the standard module's library.

Test: .movie
Bot says nothing.

Todo:
Standardize argument checking. Use @example as return message or use the example in the comment section of the commands. Every command that takes argument should tell the user when there is no arguments.

Thank you.

thomastanck:
Perhaps a @arguments decorator that modifies the way rules are processed?
```python
@commands('test')
@arguments('\W*', '\W*')
def fancy_test(bot, trigger):
    print('%s tests %s' % (trigger.group(1), trigger.group(2))
```
See ```sopel.tools.get_command_regex``` in tools/__init__.py to see the current commands-->rules conversion that's done internally.

embolalia:
I've been thinking for a while of doing something similar to [docopt](http://docopt.org/), which is fairly similar to your suggestion @thomastanck. You would only need to specify the usage, and it'd figure out the command and arguments from there. Parsing it and designing good semantics for this context is non-trivial, but it's on the list of things to do.

thomastanck:
Why not actually use docopt? We can pass in the args into the docopt function
```python
docopt(usagestr, argv=args_here)
```

```python
# will create a callable similar to
# @rule('\.test (.*)')
@usage("""Usage:
.test <tester> <testee>...
.test vigorously <tester> <testee>""")
def fancy_test(bot, trigger):
    # if it doesn't match the usage, fancy_test doesn't get called.
    tester = trigger.docopt.tester
    testee = trigger.docopt.testee # testee is a list because of the ellipsis in the first command
    if trigger.docopt.vigorously == True:
        print('%s tests %s vigorously' % (tester, testee[0])
    else:
        for t in testee:
            print('%s tests %s' % (tester, t))
```

thomastanck:
Got a mock working

```python
# coding=utf-8

from __future__ import unicode_literals, absolute_import, print_function, division

from docopt import docopt

from sopel.module import rule

@rule('\.test.*')
def test(bot, trigger):
    print('matched')
    print(trigger.group(0))
    print(trigger.group(0).split(' '))
    usage = str('''Usage:
    .test vigorously <tester> <testee>
    .test <tester> <testee>...''')
    try:
[...]

PR #1859, opened 2020-05-09
announce: send announces to TARGMAX channels at once, if advertised

### LINKED ISSUE #1306: announce: Batch up to MAXTARGETS channels at a time if available
The simple loop in `announce.py` is kind of...meh. The number of lines it actually sends to the server could be reduced as much as 75% depending on how many channels Sopel is in when `.announce` is called.

Some IRC servers (Hybrid, for example) send a `MAXTARGETS` field in the `ISUPPORT` properties when a client connects. If this is present, it would be useful for `announce.py` to have that information and send its announcement messages to `MAXTARGETS` channels at a time.

Long-term thing, more of an optimization than anything.

dgw:
### Description
Tin. Closes #1306.

### Checklist
  - Not against a live server. I started writing unit tests but got lost in @Exirel's mocks. 😅 Maybe he'll have time to help me finish later.

### DIFF
--- a/sopel/modules/announce.py
+++ b/sopel/modules/announce.py
@@ -29,7 +29,11 @@ def _chunks(items, size):
 @plugin.require_admin('Sorry, I can\'t let you do that', reply=True)
 @plugin.output_prefix('[ANNOUNCEMENT] ')
 def announce(bot, trigger):
-    """Send an announcement to all channels the bot is in"""
+    """Send an announcement to all channels the bot is in."""
+    if trigger.group(2) is None:
+        bot.reply('Announce what? I need a message to say.')
+        return
+
     size = bot.isupport.get('TARGMAX', {}).get('PRIVMSG', 1)
     channels = _chunks(bot.channels.keys(), size)
     for cgroup in channels:
--- a/sopel/modules/announce.py
+++ b/sopel/modules/announce.py
@@ -12,12 +12,26 @@ from __future__ import absolute_import, division, print_function, unicode_litera
 from sopel import plugin
 
 
+def _chunks(items, size):
+    """Break a list of items into groups.
+
+    :param items: the collection of items to chunk
+    :type items: :term:`iterable`
+    :param int size: the size of each chunk
+    """
+    # from https://stackoverflow.com/a/312464/5991 with modified names for readability
+    for delim in range(0, len(items), size):
+        yield items[delim:delim + size]
+
+
 @plugin.command('announce')
 @plugin.example('.announce Some important message here')
 @plugin.require_admin('Sorry, I can\'t let you do that', reply=True)
 @plugin.output_prefix('[ANNOUNCEMENT] ')
 def announce(bot, trigger):
     """Send an announcement to all channels the bot is in"""
-    for channel in bot.channels:
-        bot.say(trigger.group(2), channel)
+    size = bot.isupport.get('TARGMAX', {}).get('PRIVMSG', 1)
+    channels = _chunks(bot.channels.keys(), size)
+    for cgroup in channels:
+        bot.say(trigger.group(2), ','.join(cgroup))
     bot.reply('Announce complete.')


sopel/modules/announce.py
```diff
@@ -9,14 +9,24 @@
 """
 from __future__ import unicode_literals, absolute_import, print_function, division
 
+from builtins import range
+
 from sopel.module import commands, example, require_admin
 
 
+# from https://stackoverflow.com/a/312464/5991 with modified names for readability
```
Exirel, line 17: Put that in a proper docstring and I'm sold.

sopel/modules/announce.py
```diff
@@ -9,14 +9,24 @@
...
+def chunks(iterbl, size):
```
Exirel, line 18: There is something important that is missing here: `iterbl` **must have** a length, which isn't the case for a generator.

In that regard, I suggest two actions:

* rename from "iterbl" (which I assume is a shortcut (which is not good to me) for "iterable") to "items"
* add a docstring that explains this

I know, 2 lines of code, *so many* consequences!

Exirel: Seconde note: this could be in tools/utils.

sopel/modules/announce.py
dgw: Why would you want this in a docstring? I can understand wanting **a** proper docstring for the function, but nobody needs to see the SO source in a hypothetical documentation page (not that `modules` gets doc'd).

sopel/modules/announce.py
dgw: The original was built to work with lists, and named the argument `lst`. This is kind of trivial to put in `tools`, IMO. It's just meant to be a helper (and I considered simplifying this to a comprehension expr, but made it separate for eventual unit testing).

I don't really care about the parameter name(s), but I object to excessively bulletproofing a one-off helper that nobody should use because it's not part of a documented API… I'd sooner specialize it to take `bot` as an argument instead, and rename the function `get_channel_groups` or something.

dgw:
The big issue right now, besides lack of tests, is that py2 can't import from `builtins`.

dgw:
Re the new commit: "Thanks, I hate it." Don't worry, I don't want to merge that kludge.

dgw:
Well-taken point from @Exirel applied: Performance concerns in this case are not significant enough to worry about what underlying implementation of `range()` we get.

sopel/modules/announce.py
Exirel: Oh, I think we can come with a good solution here: make it a bot's method instead! After all, the bot knows about targmax, it knows about channels, and it can do everything asked!

sopel/modules/announce.py
Exirel: Yeah, it's me, I always assume I shouldn't have to read the source code to get everything I want from a documentation. In that case, maybe we don't need a docstring.

But sometimes, I have to read the source code. In that case, I tend to scan for `def`, then I read under it, not above it. So, middle ground here: can you put that under the `def chuncks` line instead of above?

sopel/modules/announce.py
dgw: What are you proposing as the API feature here, exactly? `bot.group_max_channels(*list)`?

OK, so naming things is hard. But we probably want to give serious thought to actually _designing_ API features around the basic ISUPPORT/MYINFO data Sopel tracks, rather than tossing them in at random based on trivial use cases like this.

Exirel: Yeah it is.

As "trivial" as it sound, this is exactly the right use-case for TARGMAX: take a list of targets, and yields each group of targets according to TARGMAX!

I'd do a method with 2 arguments:

```
def lets_find_a_name(self, *targets, max_target=None):
    """Yield group of ``targets`` according to ISUPPORT's TARGMAX or up to ``max_target``.

    <more details>
    """
    # 1. targmax = (TARGMAX or 1) if max_target is None else min(TARGMAX, max_target)
    # 2. chunk according to this targmax
```

No name yet, but I think you get the general idea. I think it's a good interface.

Exirel:
@dgw I still like this PR. If you have time to touch it... :+1:

sopel/modules/announce.py
dgw: Let's do this separately, and put some real discussion into it. I agree that it would be useful for the `bot` object or `tools` module to have something for this, but implementing this for `announce` is so trivial that it shouldn't be held up by a larger discussion about new API behavior/features.

dgw:
@Exirel I updated it a bit, but have not rebased yet. Want to know first if it's good enough (obviously not the perfect solution discussed in line notes in that prior review) to ship, while we work on the larger API question.

sopel/modules/announce.py
Exirel: I now realize that this syntax: `(self, *targets, max_target=None)` ; is not supported in Py2. This will have to wait Sopel 8.

Exirel:
@dgw all good for a rebase & squash.

dgw:
@Exirel This is now the squishiest PR you ever saw (not really, but I wanted to say that anyway)

Exirel (approved):
👍

sopel/modules/announce.py
```diff
@@ -12,12 +12,26 @@
...
 @plugin.require_admin('Sorry, I can\'t let you do that', reply=True)
 @plugin.output_prefix('[ANNOUNCEMENT] ')
 def announce(bot, trigger):
     """Send an announcement to all channels the bot is in"""
-    for channel in bot.channels:
-        bot.say(trigger.group(2), channel)
+    size = bot.isupport.get('TARGMAX', {}).get('PRIVMSG', 1)
+    channels = _chunks(bot.channels.keys(), size)
+    for cgroup in channels:
+        bot.say(trigger.group(2), ','.join(cgroup))
```
Exirel, line 36: If you could check that there is a message here, that would do it for #1001 as well!

Exirel (changes requested):
Nitpick for argument checking, and we are all good!

sopel/modules/announce.py
dgw: Should be satisfactorily implemented in ce10d0a

Exirel (approved):

PR #1866, opened 2020-05-20
doc: add more examples to core_section

Exirel:
### Description

I'm tired of user asking "how to provide multiple values?" for list, and I think our users are probably tired of asking, too! So I tried to add more examples whenever possible.

I'm sure I've miss a lot, and @dgw will probably have to fight my Nemesi**S**.

Oh well.

### Checklist

### DIFF
--- a/docs/source/configuration.rst
+++ b/docs/source/configuration.rst
@@ -30,7 +30,21 @@ This file can be generated with ``sopel configure``.
 
 .. seealso::
 
-    The :doc:`cli` chapter for ``sopel``'s subcommands.
+    The :doc:`cli` chapter for ``sopel configure`` subcommand.
+
+.. note::
+
+    In this document, when a config option is a "list", it means you can
+    provide more than one value, each on its own line, like this::
+
+        [core]
+        single_option = only one value
+        multi_option =
+            first value
+            second value
+            "# escape value starting with # using double quotes"
+            # this is a comment and won't be used as value
+            yet another value
 
 
 .. contents::
--- a/docs/source/configuration.rst
+++ b/docs/source/configuration.rst
@@ -32,24 +32,42 @@ This file can be generated with ``sopel configure``.
 
     The :doc:`cli` chapter for ``sopel configure`` subcommand.
 
-.. note::
-
-    In this document, when a config option is a "list", it means you can
-    provide more than one value, each on its own line, like this::
-
-        [core]
-        single_option = only one value
-        multi_option =
-            first value
-            second value
-            "# escape value starting with # using double quotes"
-            # this is a comment and won't be used as value
-            yet another value
-
-
 .. contents::
     :local:
     :depth: 2
+
+
+INI file structure
+==================
+
+Sopel uses an INI file structure, and uses `Python's ConfigParser`__ for that
+purpose. Note that the exact behavior of the parser depends on the version of
+Python you are running your instance of Sopel with (indentation and comments
+in particular are handled differently between Python 2.7 and Python 3+).
+
+On top of that, Sopel uses its own configuration class to handle certain types
+of options, such as choice, integer, float, list, and boolean.
+
+In this document, when a config option is a "list", it means you can provide
+more than one value, each on its own line, like this::
+
+    [core]
+    single_option = only one value
+    multi_option =
+        first value
+        second value
+        "# escape value starting with # using double quotes"
+        # this is a comment and won't be used as value
+        yet another value
+
+When a config option is a boolean, it means you can provide one of these
+case-insensitive values as "true": ``1``, ``yes``, ``y``, ``true``, ``on``. Any
+other value will be considered as "false".
+
+When a config option is a choice, it means you can provide only one of the
+values defined as valid for this option, or Sopel will complain about it.
+
+.. __: https://docs.python.org/3/library/configparser.html#supported-ini-file-structure
 
 
 Identity & Admins
--- a/sopel/config/core_section.py
+++ b/sopel/config/core_section.py
@@ -74,8 +74,8 @@ class CoreSection(StaticSection):
 
     .. note::
 
-        You can use the ``sopel configure`` to generate a config file with
-        the minimal required options.
+        You can use the command ``sopel configure`` to generate a config file
+        with the minimal required options.
 
     """
 
@@ -189,7 +189,8 @@ class CoreSection(StaticSection):
     Used by the URL callbacks feature to call plugins when links are posted in
     chat; see the :func:`sopel.module.url` decorator.
 
-    This is equivalent to the default values:
+    The default value allows ``http``, ``https``, and ``ftp``. It is equivalent
+    to this configuration example:
 
     .. code-block:: ini
 
@@ -242,9 +243,10 @@ class CoreSection(StaticSection):
 
     .. important::
 
-        When you edit the config file manually, make sure to escape each line
-        starting with a ``#`` with double quotes (as shown in the example
-        above), otherwise Python will ignore them, thinking they are comments.
+        If you edit the config file manually, make sure to wrap each line
+        starting with a ``#`` in double quotes, as shown in the example above.
+        An unquoted ``#`` denotes a comment, which will be ignored by Sopel's
+        configuration parser.
 
     """
 
@@ -252,7 +254,7 @@ class CoreSection(StaticSection):
     """A list of commands to send upon successful connection to the IRC server.
 
     Each line is a message that will be sent to the server once connected,
-    in that order:
+    in the order they are defined:
 
     .. code-block:: ini
 
@@ -394,7 +396,7 @@ class CoreSection(StaticSection):
 
     Used when plugins format times with :func:`sopel.tools.time.format_time`.
 
-    Example with the British time::
+    For example, to make Sopel fall back on British time::
 
         default_timezone = Europe/London
 
@@ -417,7 +419,7 @@ class CoreSection(StaticSection):
             xkcd
             help
 
-    In that case, only ``url``, ``xkcd``, and the ``help`` plugins will be
+    In that case, only the ``url``, ``xkcd``, and ``help`` plugins will be
     enabled and loaded by Sopel.
 
     To load *all* available plugins, clear this setting by removing it, or
@@ -883,7 +885,7 @@ class CoreSection(StaticSection):
     It is a regular expression (so the default, ``\\.``, means commands start
     with a period), though using capturing groups will create problems.
 
-    With the default value, user will invoke commands like this::
+    With the default value, users will invoke commands like this::
 
         <nick> .help
 
@@ -895,7 +897,7 @@ class CoreSection(StaticSection):
             prefix = \\?
             help_prefix = ?
 
-        In that example, user will invoke commands like this::
+        In that example, users will invoke commands like this::
 
             <nick> ?help
     """
@@ -992,10 +994,11 @@ class CoreSection(StaticSection):
 
     :default: ``120``
 
-    You can increase the timout (for example up to 200s) like this:
+    You can change the timeout like this:
 
     .. code-block:: ini
 
+        # increase to 200 seconds
         timeout = 200
 
     """
--- a/sopel/config/core_section.py
+++ b/sopel/config/core_section.py
@@ -578,18 +578,24 @@ class CoreSection(StaticSection):
     """A list of hostnames which Sopel should ignore.
 
     Messages from any user whose connection hostname matches one of these
-    values will be ignored. Both :ref:`regular expression syntax <re-syntax>`
-    and exact match are supported:
+    values will be ignored. :ref:`Regular expression syntax <re-syntax>`
+    is supported, so remember to escape special characters:
 
     .. code-block:: ini
 
         host_blocks =
-            exact-match.com
-            .*\\.regex-domain\\.com
+            (.+\\.)*domain\\.com
 
     .. seealso::
 
         The :attr:`nick_blocks` list can be used to block users by their nick.
+
+    .. note::
+
+        We are working on a better block system; see `issue #1355`__ for more
+        information and update.
+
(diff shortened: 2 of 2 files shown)


sopel/config/core_section.py
```diff
@@ -152,33 +189,72 @@ class CoreSection(StaticSection):
...
+            "#channel"
+            "#logs"
+            &rare_prefix_channel
+            "#private password"
+
+    .. important::
+
+        When you edit the config file manually, make sure to escape each line
+        starting with a ``#`` with double quotes (as shown in the example
+        above), otherwise Python will ignore them, thinking they are comments.
```
dgw, line 247: I tried to improve this, with mixed results.

```suggestion
        If you edit the config file manually, make sure to wrap each line
        starting with a ``#`` in double quotes, as shown in the example above.
        An unquoted ``#`` denotes a comment, which will be ignored by Sopel's
        configuration parser.
```

Would a link to [the ConfigParser syntax docs](https://docs.python.org/3/library/configparser.html#supported-ini-file-structure) be overkill here?

sopel/config/core_section.py
```diff
@@ -152,33 +189,72 @@ class CoreSection(StaticSection):
...
+
     """
 
     commands_on_connect = ListAttribute('commands_on_connect')
     """A list of commands to send upon successful connection to the IRC server.
 
-    Each line is a message that will be sent to the server once connected.
-    Example::
+    Each line is a message that will be sent to the server once connected,
+    in that order:
```
dgw, line 255:
```suggestion
    in the order they are defined:
```

sopel/config/core_section.py
```diff
@@ -302,45 +390,108 @@ class CoreSection(StaticSection):
 
     :default: ``UTC``
 
+    .. highlight:: ini
+
     Used when plugins format times with :func:`sopel.tools.time.format_time`.
+
+    Example with the British time::
```
dgw, line 397:
```suggestion
    For example, to make Sopel fall back on British time::
```

(idk, still kinda messy)

sopel/config/core_section.py
```diff
@@ -302,45 +390,108 @@ class CoreSection(StaticSection):
...
     If set, Sopel will *only* load the plugins named here. All other available
-    plugins will be ignored.
+    plugins will be ignored::
+
+        enable =
+            url
+            xkcd
+            help
+
+    In that case, only ``url``, ``xkcd``, and the ``help`` plugins will be
```
dgw, line 420:
```suggestion
    In that case, only the ``url``, ``xkcd``, and ``help`` plugins will be
```

sopel/config/core_section.py
```diff
@@ -389,57 +564,86 @@ def homedir(self):
...
     """
 
     host_blocks = ListAttribute('host_blocks')
     """A list of hostnames which Sopel should ignore.
 
-    Messages from any user whose connection hostname matches one of these values
-    will be ignored. :ref:`Regular expression syntax <re-syntax>` is supported.
+    Messages from any user whose connection hostname matches one of these
+    values will be ignored. :ref:`Regular expression syntax <re-syntax>` is
+    supported:
```
dgw, line 580: Please don't re-wrap this. It fits perfectly on two lines if the text is allowed to be 80 columns instead of 79.

sopel/config/core_section.py
```diff
@@ -389,57 +564,86 @@ def homedir(self):
...
 
     log_raw = ValidatedAttribute('log_raw', bool, default=False)
     """Whether a log of raw lines as sent and received should be kept.
 
-    See :ref:`Raw Logs`.
+    To enable this logging:
+
+    .. code-block:: ini
+
+        log_raw = yes
```
dgw, line 598:
```suggestion
        log_raw = True
```

I'm partial to using the Pythonic form of a Boolean for these, but you do you. 😹

sopel/config/core_section.py
```diff
@@ -620,18 +856,48 @@ def homedir(self):
...
+    .. highlight:: ini
+
+    **Required**::
+
+        prefix = \\.
 
     It is a regular expression (so the default, ``\\.``, means commands start
     with a period), though using capturing groups will create problems.
+
+    With the default value, user will invoke commands like this::
```
dgw, line 886:
```suggestion
    With the default value, users will invoke commands like this::
```

sopel/config/core_section.py
```diff
@@ -620,18 +856,48 @@ def homedir(self):
...
+
+        <nick> .help
```
dgw, line 888: I'm too lazy today to clone the PR and run Sphinx build, but won't the above `.. highlight:: ini` also try to act on this (and the similar one below with `?`)?

sopel/config/core_section.py
```diff
@@ -620,18 +856,48 @@ def homedir(self):
...
+
+    .. note::
+
+        Remember to change the :attr:`help_prefix` value if you change this
+        one, for example::
+
+            prefix = \\?
+            help_prefix = ?
+
+        In that example, user will invoke commands like this::
```
dgw, line 898:
```suggestion
        In that example, users will invoke commands like this::
```

sopel/config/core_section.py
```diff
@@ -642,11 +908,12 @@ def homedir(self):
     If ``True``, Sopel will send information about the triggered exception to
     the sender of the message that caused the error.
 
-    If ``False``, Sopel will only log the error and will appear to fail silently
-    from the triggering IRC user's perspective.
+    If ``False``, Sopel will only log the error and will appear to fail
+    silently from the triggering IRC user's perspective.
```
dgw, line 912: Another 79 vs. 80 wrapping thing. Sigh. Unnecessary, but at least this one doesn't add any extra lines.

sopel/config/core_section.py
```diff
@@ -712,24 +991,51 @@ def homedir(self):
     """The number of seconds acceptable between pings before timing out.
 
     :default: ``120``
+
+    You can increase the timout (for example up to 200s) like this:
+
+    .. code-block:: ini
+
+        timeout = 200
```
dgw, line 999:
```suggestion
    You can change the timeout like this:

    .. code-block:: ini

        # increase to 200 seconds
        timeout = 200
```

Not necessarily the best alternative, but I don't like the parenthetical here, and I don't like implying that the timeout can be only increased.

sopel/config/core_section.py
```diff
@@ -712,24 +991,51 @@ def homedir(self):
...
     use_ssl = ValidatedAttribute('use_ssl', bool, default=False)
     """Whether to use a SSL/TLS encrypted connection.
 
     :default: ``False``
+
+    Example with SSL on:
+
+    .. code-block:: ini
+
+        use_ssl = yes
```
dgw, line 1012:
```suggestion
        use_ssl = True
```

Same editorial comment as `log_raw`.

sopel/config/core_section.py
```diff
@@ -712,24 +991,51 @@ def homedir(self):
...
     """Whether to require a trusted certificate for encrypted connections.
 
     :default: ``True``
+
+    Example with SSL on:
+
+    .. code-block:: ini
+
+        use_ssl = yes
+        verify_ssl = yes
```
dgw, line 1039:
```suggestion
        use_ssl = True
        verify_ssl = True
```

Same editorial comment as `log_raw`.

sopel/config/core_section.py
Exirel: As discussed on IRC, I don't want to give the impression that it's a Python value (like it would be evaluated as such), when it's not.

I'll also add a small section to explain that you can use more than "yes", but also "y", "on", etc. for any boolean value.

sopel/config/core_section.py
Exirel: Aw. It colors it light blue. Not sure I like it this way.

sopel/config/core_section.py
Exirel: No that's a good idea. I'll add this suggestion and I'll try not to forget about the link!

sopel/config/core_section.py
Exirel: Good point!

sopel/config/core_section.py
dgw: Perhaps this section just needs to use explicit `.. code-block::` blocks with the highlighting set for each. :/

Exirel:
I believe I addressed all the requested changes and suggestions the best I could. I've tweaked the text when they generate orphan (and I made a joke about it in the commit message because I'm feeling extra cheeky today). I also explained more about the Config Parser thing, and what choice, list, and bool options are, and how to set them properly.

I modified the prefix/help_prefix doc a bit, to make it more clear, too.

sopel/config/core_section.py
```diff
@@ -578,16 +578,19 @@ def homedir(self):
...
+    values will be ignored. Both :ref:`regular expression syntax <re-syntax>`
+    and exact match are supported:
 
     .. code-block:: ini
 
         host_blocks =
-            domain\\.com
-            .*\\.domain\\.com
+            exact-match.com
+            .*\\.regex-domain\\.com
```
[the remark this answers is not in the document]
Exirel: There is an ``or`` though, that uses exact match. What would be the correct phrasing for that?

sopel/config/core_section.py
```diff
@@ -872,7 +885,7 @@ def homedir(self):
     """
 
     prefix = ValidatedAttribute('prefix', default='\\.')
-    """The prefix to add to the beginning of commands.
+    """The prefix to add to the beginning of commands as a regular expression.
```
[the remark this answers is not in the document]
Exirel: I really hope so!

sopel/config/core_section.py
```diff
@@ -813,9 +816,19 @@ def homedir(self):
...
     Messages from any user whose nickname matches one of these values will be
-    ignored. :ref:`Regular expression syntax <re-syntax>` is supported.
+    ignored. Both :ref:`regular expression syntax <re-syntax>` and exact match
+    are supported:
+
+    .. code-block:: ini
+
+        nick_blocks =
+            ExactNick
+            _*RegexMatch_*
```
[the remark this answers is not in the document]
dgw: We'll confine discussion to just the one thread: https://github.com/sopel-irc/sopel/pull/1866#discussion_r428909631

sopel/config/core_section.py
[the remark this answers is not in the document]
dgw: It's not a simple matter, is it? Technically, both regex and exact-match _are_ supported. The issue is with managing the user's expectations (with a side of "why does the code try to treat the same value two ways?!", because technical debt).

The possibility for overlap between exact-match and regex syntax is probably greater for nicks than hosts (due to brackets/braces being allowed in nicknames), but it theoretically exists for both.

Maybe it would be enough to mitigate the possibility for confusion by attempting an exact match first, and only trying the regex path if that fails? Then a naïve user accidentally using regex syntax in what they think is an exact match won't be confused that their block rule didn't work.

On the other hand, trying exact matches first has its own potential for accidental false-positives on the nickname side… But users who choose regex-like nicknames deserve to be blocked anyway. 😛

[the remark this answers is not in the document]
Exirel: I see that there is a problem, but I'm not sure what you want me to do about it in this PR.

Since I'm a bit confused, the safe path for me is to revert to a more elusive explanation for now. Something like: "Regular expression are supported (remember to escape characters as needed)."?

I could also add a little warning, something like "We are working on a better block system, see issue [link] for more information and update". I feel like this approach is honest with where the code is at the moment (as you mentioned: technical debt), and I believe honest communication will help the user understand why it isn't that simple.

What do you say?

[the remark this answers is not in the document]
Exirel: Screen Shot 2020-05-22 at 17 05 07

Maybe it could be a "note" instead of a warning, but I wanted you to see how it looks like first.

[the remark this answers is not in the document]
dgw: That screenshot looks good, though I agree about using `.. note::` instead.

I assume this is for both `host_blocks` and `nick_blocks`? Maybe the `.. warning::` would be the bit about regular expression syntax—but that might be too many blocks. 😅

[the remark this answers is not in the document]
Exirel: Done!

sopel/config/core_section.py
```diff
@@ -578,19 +578,25 @@ def homedir(self):
...
-            .*\\.regex-domain\\.com
+            (.+\\.)*domain\\.com
 
     .. seealso::
 
         The :attr:`nick_blocks` list can be used to block users by their nick.
 
+    .. note::
+
+        We are working on a better block system, see `issue #1355`__ for more
```
dgw, line 595:
```suggestion
        We are working on a better block system; see `issue #1355`__ for more
```

sopel/config/core_section.py
```diff
@@ -829,6 +835,13 @@ def homedir(self):
 
         The :attr:`host_blocks` list can be used to block users by their host.
 
+    .. note::
+
+        We are working on a better block system, see `issue #1355`__ for more
```
dgw, line 840:
```suggestion
        We are working on a better block system; see `issue #1355`__ for more
```

Exirel:
@dgw All good now. Even Travis is happy.

docs/source/configuration.rst
```diff
@@ -30,14 +30,46 @@ This file can be generated with ``sopel configure``.
...
+INI file structure
+==================
+
+Sopel uses an INI file structure, and uses `Python's ConfigParser`__ for that
+purpose. Note that the exact behavior of the parser depends on the version of
+Python you are running your instance of Sopel with (indentation and comments
+in particular are handled differently between Python 2.7 and Python 3+).
+
+On top of that, Sopel uses its own configuration class to handle certain type
+of options, such as choice, integer, float, list, and boolean options.
```
dgw, line 49:
```suggestion
On top of that, Sopel uses its own configuration class to handle certain types
of options, such as choice, integer, float, list, and boolean.
```

PR #1871, opened 2020-05-21
url: tweak default_headers and max_bytes for improved URL titles

cottongin:
### Description
What does this do? What problem does it solve? Does it resolve any GitHub issues?
> Adds 'Accept' and 'Accept-Language' headers and spoofs 'User-Agent' while increasing max_bytes so that problematic sites work better for fetching titles

### Checklist

### DIFF
--- a/sopel/modules/url.py
+++ b/sopel/modules/url.py
@@ -17,7 +17,7 @@ import re
 import dns.resolver
 import requests
 
-from sopel import __version__, module, tools
+from sopel import module, tools
 from sopel.config.types import ListAttribute, StaticSection, ValidatedAttribute
 from sopel.tools import web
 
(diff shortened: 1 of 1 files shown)


sopel/modules/url.py
```diff
@@ -27,8 +27,16 @@
...
     from urlparse import urlparse
 
-USER_AGENT = 'Sopel/{} (https://sopel.chat)'.format(__version__)
-default_headers = {'User-Agent': USER_AGENT}
+USER_AGENT = (
+    'Mozilla/5.0 (Windows NT 10.0; Win64; x64) '
+    'AppleWebKit/537.36 (KHTML, like Gecko) '
+    'Chrome/78.0.3904.108 Safari/537.36'
+)
+default_headers = {
```
dgw, line 35: Since you're touching this anyway, I wouldn't say no to fixing the name so it looks properly constant-like (all-caps).

sopel/modules/url.py
```diff
@@ -39,7 +47,7 @@
 # the title. We don't want it too high, or a link to a big file/stream will
 # just keep downloading until there's no more memory. 640k ought to be enough
 # for anybody.
```
dgw, line 49:
```suggestion
# just keep downloading until there's no more memory. 640k ought to be enough
# for anybody, but the modern web begs to differ.
```

It's not 640k any more, so the original comment is no longer accurate 😭

Did you add `* 2` instead of changing the number below so 640k would still "be there"? 😹

sopel/modules/url.py
cottongin: I'm not a huge fan of "constant"-like in Python anyway, and it would require changing other references further on in the plugin. But I'm not opposed to it entirely, I just figured this would be *really* fixed in the future.

sopel/modules/url.py
cottongin: You caught me red-handed lol

sopel/modules/url.py
Exirel: Per the PEP8 (and pylint) style, any module-level attribute should be all-caps.

dgw: Right. The fact that it's in basic `snake_case` is an historical mistake. Not as bad as others in the Sopel codebase, but still a stylistic mistake.

PR #1879, opened 2020-05-31
config: add is_secret attribute to use getpass hidden input

### LINKED ISSUE #819: Hide input of passwords in wizard
This is a 6.0 regression, which I might or might not get done for the actual release.

### LINKED ISSUE #1571: admin: censor settings more appropriately
`admin.py` can be used to display setting values. Some setting values "should not" be displayed (e.g., password, secret keys), however. 

Currently, the plugin censors any setting that `endswith()` "password". Other settings that may not exactly be passwords, but are just as valuable, should similarly be censored. What is the best option to catch settings that should be prevented from being displayed?

Current affected settings (not exhaustive):
- `core.auth_password`
- `github.client_key`
- `github.secret` (this together with `client_key` should _definitely_ not be revealed)

Is `"password" is in` better than `endswith("password")` (similar for "secret", etc...)? If there is a setting like `secret_key` or `password_for_site`, `endswith()` will obviously not censor those. While those examples may not be the best setting names, and can certainly be changed to meet the `endswith()` condition, there is no documentation (I think...) that suggests settings be named in such a way if this behavior is desired.

Also, should `core.owner` be able to view the setting value uncensored since this will be happening in privmsg anyway?

Finally, if a value is censored from a `core.admins` user, should they even be allowed to set it?

_Originally brought up by @dgw in https://github.com/sopel-irc/sopel/issues/1570#issuecomment-486738630_

Exirel:
Fixes #819; closes #1571 

### Description

This PR doesn't try to auto-detect a password/secret attribute, it relies on plugin author to explicitly use a `SecretAttribute`, or to use the `is_secret` parameter. I don't know how it was done before Sopel 6, and the issue #819 doesn't tell either.

When `is_secret` is `True`, the `configure` mehtod uses `getpass.getpass` instead of `input`.

### Checklist

### DIFF
--- a/sopel/config/types.py
+++ b/sopel/config/types.py
@@ -26,6 +26,7 @@ As an example, if one wanted to define the ``[spam]`` section as having an
 
 from __future__ import absolute_import, division, print_function, unicode_literals
 
+import getpass
 import os.path
 import re
 import sys
@@ -83,7 +84,7 @@ class StaticSection(object):
         not the attribute's default.
         """
         clazz = getattr(self.__class__, name)
-        if default is NO_DEFAULT:
+        if default is NO_DEFAULT and not clazz.is_secret:
             try:
                 default = getattr(self, name)
             except AttributeError:
@@ -111,27 +112,60 @@ class BaseValidated(object):
     :param str name: the attribute name to use in the config file
     :param default: the value to be returned if the setting has no value
                     (optional; defaults to :obj:`None`)
-    :type default: str
+    :type default: mixed
+    :param bool is_secret: tell if the attribute is secret/a password
+                           (optional; defaults to ``False``)
 
     ``default`` also can be set to :const:`sopel.config.types.NO_DEFAULT`, if
     the value *must* be configured by the user (i.e. there is no suitable
     default value). Trying to read an empty ``NO_DEFAULT`` value will raise
     :class:`AttributeError`.
     """
-    def __init__(self, name, default=None):
+    def __init__(self, name, default=None, is_secret=False):
         self.name = name
+        """Name of the attribute."""
         self.default = default
+        """Default value for this attribute.
+
+        If not specified, the attribute's default value will be ``None``.
+        """
+        self.is_secret = bool(is_secret)
+        """Tell if the attribute is secret/a password.
+
+        The default value is ``False`` (not secret).
+
+        Sopel's configuration can contain passwords, secret keys, and other
+        private information that must be treated as sensitive data. Such
+        options should be marked as "secret" with this attribute.
+        """
 
     def configure(self, prompt, default, parent, section_name):
-        """
-        With the ``prompt`` and ``default``, parse and return a value from
-        terminal.
+        """Parse and return a value from user's input.
+
+        :param str prompt: text to show the user
+        :param mixed default: default value used if no input given
+        :param parent: usually the parent Config object
+        :type parent: :class:`~sopel.config.Config`
+        :param str section_name: the name of the containing section
+
+        This method displays the ``prompt`` and waits for user's input on the
+        terminal. If no input is provided (i.e. the user just presses "Enter"),
+        the ``default`` value is returned instead.
+
+        If :attr:`.is_secret` is ``True``, the input will be hidden, using the
+        built-in :func:`~getpass.getpass` function.
         """
         if default is not NO_DEFAULT and default is not None:
             prompt = '{} [{}]'.format(prompt, default)
-        value = get_input(prompt + ' ')
+
+        if self.is_secret:
+            value = getpass.getpass(prompt + ' (hidden input) ')
+        else:
+            value = get_input(prompt + ' ')
+
         if not value and default is NO_DEFAULT:
             raise ValueError("You must provide a value for this option.")
+
         value = value or default
         return self.parse(value)
 
@@ -208,9 +242,17 @@ class ValidatedAttribute(BaseValidated):
                       that can be written to the config file safely (optional;
                       defaults to :class:`str`)
     :type serialize: :term:`function`
+    :param bool is_secret: ``True`` when the attribute should be considered
+                           a secret, like a password (default to ``False``)
     """
-    def __init__(self, name, parse=None, serialize=None, default=None):
-        super(ValidatedAttribute, self).__init__(name, default=default)
+    def __init__(self,
+                 name,
+                 parse=None,
+                 serialize=None,
+                 default=None,
+                 is_secret=False):
+        super(ValidatedAttribute, self).__init__(
+            name, default=default, is_secret=is_secret)
         if parse == bool:
             parse = _parse_boolean
             if not serialize or serialize == bool:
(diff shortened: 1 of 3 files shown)


sopel/config/types.py
```diff
@@ -111,27 +112,49 @@ class BaseValidated(object):
...
     the value *must* be configured by the user (i.e. there is no suitable
     default value). Trying to read an empty ``NO_DEFAULT`` value will raise
     :class:`AttributeError`.
     """
-    def __init__(self, name, default=None):
+    def __init__(self, name, default=None, is_secret=False):
         self.name = name
+        """Name of the attribute."""
         self.default = default
+        """Default value for this attribute. (default to ``None``)"""
```
dgw, line 128:
```suggestion
        """Default value for this attribute.
        
        If not specified, the attribute's default value will be ``None``.
        """
```

sopel/config/types.py
```diff
@@ -111,27 +112,49 @@ class BaseValidated(object):
...
+        self.is_secret = bool(is_secret)
+        """Tell if the attribute is secret/a password. (default to ``False``)
```
dgw, line 130:
```suggestion
        """Tell if the attribute is secret/a password.
        
        The default value is ``False`` (not secret).
```

sopel/config/types.py
```diff
@@ -111,27 +112,49 @@ class BaseValidated(object):
...
 
-    def configure(self, prompt, default, parent, section_name):
+        Sopel's configuration can contains password, secret key, private
+        information that must be treated as sensitive data. Such options should
+        be marked as "secret" with this attribute.
```
dgw, line 134:
```suggestion
        Sopel's configuration can contain passwords, secret keys, and other
        private information that must be treated as sensitive data. Such
        options should be marked as "secret" with this attribute.
```

sopel/config/types.py
```diff
@@ -111,27 +112,49 @@ class BaseValidated(object):
...
         """
-        With the ``prompt`` and ``default``, parse and return a value from
-        terminal.
+
+    def configure(self, prompt, default, parent, section_name):
+        """Parse and return a value from user's input.
+
+        This method displays the ``prompt`` and wait for user's input on the
+        terminal. If no input is provided, the ``default`` value is returned
+        instead.
```
dgw, line 142:
```suggestion
        This method displays the ``prompt`` and waits for user's input on the
        terminal. If no input is provided (i.e. the user just presses "Enter"),
        the ``default`` value is returned instead.
```

sopel/config/types.py
```diff
@@ -111,27 +112,49 @@ class BaseValidated(object):
...
+
+        If :attr:`.is_secret` is ``True``, the input will be hidden, using the
+        built-in :func:`getpass.getpass` function.
```
dgw, line 145:
```suggestion
        built-in :func:`~getpass.getpass` function.
```

sopel/config/types.py
```diff
@@ -111,27 +112,49 @@ class BaseValidated(object):
...
-    def configure(self, prompt, default, parent, section_name):
+        Sopel's configuration can contains password, secret key, private
+        information that must be treated as sensitive data. Such options should
+        be marked as "secret" with this attribute.
         """
-        With the ``prompt`` and ``default``, parse and return a value from
-        terminal.
+
+    def configure(self, prompt, default, parent, section_name):
+        """Parse and return a value from user's input.
```
dgw, line 138: This is rough, and I welcome further improvements. Ideally this would appear on all `configure` methods, but I suspect we can just define it once on `BaseValidated`, remove the docstrings on child classes, and rely on inheritance for Sphinx to pull it in?

```suggestion
        """Parse and return a value from user's input.

        :param str prompt: text to show the user
        :param mixed default: default value used if no input given
        :param parent: usually the parent Config object
        :type parent: :class:`~sopel.config.Config`
        :param str section_name: the name of the containing section
```

And yes: You can also say "let's do this in a future PR".

sopel/config/types.py
```diff
@@ -235,16 +266,28 @@ def parse(self, value):
...
-        terminal.
-        """
         if self.parse == _parse_boolean:
             prompt += ' (y/n)'
             default = 'y' if default else 'n'
         return super(ValidatedAttribute, self).configure(prompt, default, parent, section_name)
 
 
+class SecretAttribute(ValidatedAttribute):
+    """A config attribute to store secret/password option.
```
dgw, line 276:
```suggestion
    """A config attribute containing a value which must be kept secret.
```

sopel/config/types.py
```diff
@@ -235,16 +266,28 @@ def parse(self, value):
...
+
+    This attribute is always considered a secret/sensitive data, and otherwise
+    behaves like other any options.
```
dgw, line 279:
```suggestion
    This attribute is always considered to be secret/sensitive data, but
    otherwise behaves like other any option.
```

Exirel:
As mentioned on IRC, I need to change this portion of code:

https://github.com/sopel-irc/sopel/blob/master/sopel/modules/admin.py#L328-L335

Exirel:
Admin plugin modified as expected.

sopel/config/types.py
Exirel: Yup, it works!

Exirel:
So github say that there are changes requested. Well, I've done the changes, @dgw it's ready to check!

sopel/modules/admin.py
```diff
@@ -327,12 +327,22 @@ def set_config(bot, trigger):
...
+
     # Display current value if no value is given
     if not value:
-        if option.endswith("password") or option.endswith("pass"):
+        value = getattr(section, option)
+
+        if descriptor is not None:
+            if getattr(descriptor, 'is_secret', False):
+                # Keep secret option as secret
+                value = "(password censored)"
```
dgw, line 340:
```suggestion
                value = "(secret value censored)"
```

It's not necessarily a password, unlike the "dumb" name-heuristic fallback.

sopel/modules/admin.py
```diff
@@ -327,12 +327,22 @@ def set_config(bot, trigger):
         bot.say(exc.args[1])
         return
 
+    # Get a descriptor class for the option if it's a static section
+    descriptor = getattr(section.__class__, option) if static_sec else None
+
     # Display current value if no value is given
     if not value:
-        if option.endswith("password") or option.endswith("pass"):
+        value = getattr(section, option)
```
dgw, line 335: Maybe it's a micro-optimization, but why assign twice if the option is a secret/password? What was wrong with the previous version's `else` below?

sopel/modules/admin.py
```diff
@@ -327,12 +327,22 @@ def set_config(bot, trigger):
...
+
+        if descriptor is not None:
+            if getattr(descriptor, 'is_secret', False):
+                # Keep secret option as secret
+                value = "(password censored)"
+        elif option.endswith("password") or option.endswith("pass"):
+            # Fallback to guessing if secret, for backward compatiblity
+            # TODO: consider a deprecation warning at some point
```
dgw, line 343: I like the idea of warning about things that look like passwords, but maybe it shouldn't be _here_.

What do you think of emitting it during settings load?

Exirel: Oh yes, that's a good idea.

sopel/modules/admin.py
Exirel: Because it's not a simple `if/else` now, it's a `if (then if/else) / else` (nested if).

At first, `value` is initialized, and then the responsibility of the code that follows this initialization is to override it if necessary, and only to do that. It limits the responsibility of said code to a very small task, reducing its working complexity.

Basically, every if/else following this statement should do one and only on thing: check for & apply override. This is especially useful to reduce the number of `else` case to manage (and test).

sopel/modules/admin.py
Exirel:
```suggestion
            # TODO: consider a deprecation warning when loading settings
```

Exirel:
Done.

PR #1889, opened 2020-06-11
reddit: fix slash-info wrongly calling output funcs as `commanded`

dgw:
### Description
Tin. @kwaaak pointed out on IRC that inline reddit stuff generates a response even if the user/subreddit doesn't exist, when it would make more sense to fail silently. Looks like that was the intent, but the implementation was wrong.

### Checklist


antdude:
Is this related to like r/c? Example:

[12:47pm] Ant> r/c
12:47PM URL> No such subreddit.

dgw:
Yes, this change makes the bot ignore that nonexistent subreddit name instead of outputting the error message you see there.

antdude:
Thank you dgw!

Issue #1821, opened 2020-03-16
irc: trying to join timeout scheduler thread when timeout occurs is a bad idea


Exirel:
### Description

To check for timeout, there is a job manager (`timeout_scheduler`) with a job that runs `_check_timeout`. When this job detect a timeout, it calls `backend.handle_close`, which tries to join the job manager thread.

This is a problem, because in that case, it tries to join itself, and that is not allowed.

A hotfix (see #1820) has been submitted and approved for 7.0.1, and now it's time to come with a proper solution that makes the problem to disapear in 7.1.0.

### Reproduction steps

See logs.

### Expected behavior

1. the bot shouldn't tries to join the timeout scheduler thread within itself
2. a proper cleanup must be performed in any case

### Logs

```
[2020-03-13 15:43:36,014] sopel.irc.backends   ERROR    - Server timeout detected after 147s; closing.
[2020-03-13 15:43:36,014] sopel.tools.jobs     ERROR    - Error while processing job: cannot join current thread
[2020-03-13 15:43:36,015] sopel.irc.backends   ERROR    - Error with the timeout scheduler: cannot join current thread
Traceback (most recent call last):
	File "/usr/lib/python3.8/site-packages/sopel/tools/jobs.py", line 118, in _call
			job.func(self.manager)
	File "/usr/lib/python3.8/site-packages/sopel/irc/backends.py", line 60, in _check_timeout
			backend.handle_close()
	File "/usr/lib/python3.8/site-packages/sopel/irc/backends.py", line 130, in handle_close
			self.timeout_scheduler.join(timeout=15)
	File "/usr/lib/python3.8/threading.py", line 1008, in join
			raise RuntimeError("cannot join current thread")
RuntimeError: cannot join current thread
[2020-03-13 15:43:36,017] sopel.tools.jobs     ERROR    - Error in job scheduler: cannot join current thread
[2020-03-13 15:43:36,017] sopel.irc.backends   ERROR    - Error with the timeout scheduler: cannot join current thread
Traceback (most recent call last):
	File "/usr/lib/python3.8/site-packages/sopel/tools/jobs.py", line 118, in _call
			job.func(self.manager)
	File "/usr/lib/python3.8/site-packages/sopel/irc/backends.py", line 60, in _check_timeout
			backend.handle_close()
	File "/usr/lib/python3.8/site-packages/sopel/irc/backends.py", line 130, in handle_close
			self.timeout_scheduler.join(timeout=15)
[...]

PR #1890, opened 2020-06-11
irc: tweak backend interface

Exirel:
### Description

We used `async_chat.connected` attribute directly in the code, which we shouldn't, so I replaced it with a new method, added to the abstract IRC backend interface.

Also, after further inspection, async_chat's `close` method can be used safely, so now we call it first, then we ask the bot to do whatever is needed with its `on_close` method. Since this method expect the connection to be closed already, it shouldn't be a problem.

The timeout scheduler doesn't need to be joined, since none of its jobs requires to wait that much. I also modified the bot's interface to have a new method (`irc_send`) which will ensure that 1) we don't have any conflict with the `async_chat.send` method, and 2) makes sure to be thread-safe - the abstract class doesn't need to care about that, it's not its job.

For the record, I've run this version of the bot, with these changes, for more than 36h, and I got 0 problem. I connected to Freenode using a secure connection (ssl on), with an authenticated account. So I feel pretty confident about this.

### Checklist

### DIFF
--- a/sopel/irc/backends.py
+++ b/sopel/irc/backends.py
@@ -14,7 +14,7 @@ import logging
 import os
 import socket
 import sys
-from threading import current_thread
+import threading
 
 from sopel.tools.jobs import Job, JobScheduler
 from .abstract_backends import AbstractIRCBackend
@@ -41,7 +41,7 @@ LOGGER = logging.getLogger(__name__)
 
 
 def _send_ping(backend):
-    if not backend.connected:
+    if not backend.is_connected():
         return
     time_passed = (datetime.datetime.utcnow() - backend.last_event_at).seconds
     if time_passed > backend.ping_timeout:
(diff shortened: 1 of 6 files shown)


sopel/irc/backends.py
```diff
@@ -127,18 +142,11 @@ def handle_connect(self):
 
     def handle_close(self):
         """Called when the socket is closed."""
+        LOGGER.debug('Timeout scheduler stop')
```
dgw, line 145:
```suggestion
        LOGGER.debug('Stopping timeout watchdog')
```

Exirel:
> whether to defend his choice

I prefer your version. :+1:

Exirel:
@dgw all good.

Exirel:
Btw it also fixes #1821 and maybe @dgw needs to take note of that for the release notes or something?

dgw:
Not in the style I've been using, not really. It's such a tiny internal change that it'll get lumped in with some other note about refactoring or cleanup. 😁

PR #1794, opened 2019-12-30
Docstring style guide

dgw:
This is a work-in-progress, which I'm adding to as I get time on my holiday trip overseas.

PR #1891, opened 2020-06-16
jobs: prevent parallel execution and plugin specific scheduler

### LINKED ISSUE #1144: feature request: prevent parallel execution of interval
My bot is polling some info via http request every couple of seconds.
I noticed that at least at startup the bot issues multiple polls concurrently - a behavior that I would like to prevent.
This is a simplified version of my approach:

``` python
@sopel.module.interval(5)
def poll_dpm(bot):
    state = botstate()
    if state.POLLING:
        print("polling already ...", flush=True)
        return

    state.POLLING = True
    print("polling ...", flush=True)
```

`state = botstate()` is getting an object which fetches a boolean from the bots memory, bot checks a Threading.Lock which is also stored in the bots memory.

Usually this works, but when the bot is starting, sometimes i still notice that two polls happen in paralell before my lock is set.

Output at startup:

> ...
> Connected.
> Initializing POLLING=False
> polling ...polling ...
> polling already ...

Without the lock this happens a lot more often. Still im facing a mutual exclusion problem.
I believe it could make sense, to natively support enforcing non-paralell execution of scripts.

Exirel:
Closes #1144 

### Description

Initially, it was only about preventing parallel execution of jobs, but then I realized I needed way more than that:

* changed the job object to be registered only once for any set of intervals,
* changed the job object to have similar meta-data than plugin's rules
* added a new scheduler, that can be used in a plugin-specific context (similar to the plugin's rules manager)

Also, I realized that the "catch up" mechanism makes no sense: either the job can be run in time, or it can't, but it won't try to run it more than once because it's late. I can see why it was like that before, but there are several problems with how it was implemented:

* the scheduler is not reliable: it ticks at best every second, at worst every now and then
* a job isn't aware of the catch up mechanism: is it late? is it on time? it can't know, so it can't adapt itself

In that regard, it's best not to over-engineer the solution, and to just take the most optimistic and safe approach: execute only once at a time, then see if it should execute again or a bit later, given the configured intervals.

And here we are, with a lot of changes for something that seem easy at first. The good news is that it was necessary anyway to work on a better documentation spec for Sopel: in the future, we'll be able to list not only rules and commands, but also jobs! But that will be for a future PR.

### Checklist

### DIFF
--- a/sopel/bot.py
+++ b/sopel/bot.py
@@ -22,7 +22,7 @@ from sopel import irc, logger, plugins, tools
 from sopel.db import SopelDB
 import sopel.loader
 from sopel.module import NOLIMIT
-from sopel.plugins import rules as plugin_rules
+from sopel.plugins import jobs as plugin_jobs, rules as plugin_rules
 from sopel.tools import deprecated, Identifier
 import sopel.tools.jobs
 from sopel.trigger import Trigger
@@ -61,6 +61,7 @@ class Sopel(irc.AbstractBot):
         self._running_triggers_lock = threading.Lock()
         self._plugins = {}
         self._rules_manager = plugin_rules.Manager()
+        self._scheduler = plugin_jobs.Scheduler(self)
 
         self._times = {}
         """
@@ -118,12 +119,15 @@ class Sopel(irc.AbstractBot):
         self.shutdown_methods = []
         """List of methods to call on shutdown."""
 
-        self.scheduler = sopel.tools.jobs.JobScheduler(self)
-        """Job Scheduler. See :func:`sopel.module.interval`."""
-
     @property
     def rules(self):
+        """Rules manager."""
         return self._rules_manager
+
+    @property
+    def scheduler(self):
+        """Job Scheduler. See :func:`sopel.module.interval`."""
+        return self._scheduler
 
     @property
     def command_groups(self):
@@ -250,7 +254,7 @@ class Sopel(irc.AbstractBot):
         """
         self.setup_logging()
         self.setup_plugins()
-        self.scheduler.start()
+        self._scheduler.start()
 
     def setup_logging(self):
         """Set up logging based on config options."""
@@ -414,7 +418,7 @@ class Sopel(irc.AbstractBot):
 
         # remove plugin rules, jobs, shutdown functions, and url callbacks
         self._rules_manager.unregister_plugin(name)
-        self.unregister_jobs(jobs)
+        self._scheduler.unregister_plugin(name)
         self.unregister_shutdowns(shutdowns)
         self.unregister_urls(urls)
 
@@ -450,10 +454,14 @@ class Sopel(irc.AbstractBot):
         version='7.1',
         removed_in='8.0')
     def unregister(self, obj):
-        """Unregister a job or a shutdown method.
+        """Unregister a shutdown method.
 
-        :param obj: the job or shutdown method to unregister
+        :param obj: the shutdown method to unregister
         :type obj: :term:`object`
+
+        This method was used to unregister anything (rules, commands, urls,
+        jobs, and shutdown methods), but since everything can be done by other
+        means, there is no use for it anymore.
         """
         callable_name = getattr(obj, "__name__", 'UNKNOWN')
 
@@ -544,20 +552,12 @@ class Sopel(irc.AbstractBot):
 
     def register_jobs(self, jobs):
         for func in jobs:
-            for interval in func.interval:
-                job = sopel.tools.jobs.Job(interval, func)
-                self.scheduler.add_job(job)
-                callable_name = getattr(func, "__name__", 'UNKNOWN')
-                LOGGER.debug(
-                    'Job added "%s", will run every %d seconds',
-                    callable_name,
-                    interval)
+            job = sopel.tools.jobs.Job.from_callable(self.settings, func)
+            self._scheduler.register(job)
 
     def unregister_jobs(self, jobs):
         for job in jobs:
-            callable_name = getattr(job, "__name__", 'UNKNOWN')
-            self.scheduler.remove_callable_job(job)
-            LOGGER.debug('Job callable removed: %s', callable_name)
+            self._scheduler.remove_callable_job(job)
 
     def register_shutdowns(self, shutdowns):
         # Append plugin's shutdown function to the bot's list of functions to
@@ -857,8 +857,8 @@ class Sopel(irc.AbstractBot):
     def on_scheduler_error(self, scheduler, exc):
         """Called when the Job Scheduler fails.
 
-        :param scheduler: the JobScheduler that errored
-        :type scheduler: :class:`sopel.tools.jobs.JobScheduler`
+        :param scheduler: the job scheduler that errored
+        :type scheduler: :class:`sopel.plugins.jobs.Scheduler`
         :param Exception exc: the raised exception
 
         .. seealso::
@@ -870,8 +870,8 @@ class Sopel(irc.AbstractBot):
     def on_job_error(self, scheduler, job, exc):
         """Called when a job from the Job Scheduler fails.
 
-        :param scheduler: the JobScheduler responsible for the errored ``job``
-        :type scheduler: :class:`sopel.tools.jobs.JobScheduler`
+        :param scheduler: the job scheduler responsible for the errored ``job``
+        :type scheduler: :class:`sopel.plugins.jobs.Scheduler`
         :param job: the Job that errored
         :type job: :class:`sopel.tools.jobs.Job`
         :param Exception exc: the raised exception
@@ -939,16 +939,16 @@ class Sopel(irc.AbstractBot):
         LOGGER.info('Shutting down')
         # Stop Job Scheduler
         LOGGER.info('Stopping the Job Scheduler.')
-        self.scheduler.stop()
+        self._scheduler.stop()
 
         try:
-            self.scheduler.join(timeout=15)
+            self._scheduler.join(timeout=15)
         except RuntimeError:
             LOGGER.exception('Unable to stop the Job Scheduler.')
         else:
             LOGGER.info('Job Scheduler stopped.')
 
-        self.scheduler.clear_jobs()
+        self._scheduler.clear_jobs()
 
         # Shutdown plugins
         LOGGER.info(
--- a/sopel/plugins/jobs.py
+++ b/sopel/plugins/jobs.py
@@ -1,0 +1,111 @@
+# coding=utf-8
+"""Sopel's plugin jobs management.
+
+.. versionadded:: 7.1
+
+.. important::
+
+    This is all fresh and new. Its usage and documentation is for Sopel core
+    development and advanced developers. It is subject to rapid changes
+    between versions without much (or any) warning.
+
+    Do **not** build your plugin based on what is here, you do **not** need to.
+
+"""
+# Copyright 2020, Florian Strzelecki <florian.strzelecki@gmail.com>
+#
+# Licensed under the Eiffel Forum License 2.
+from __future__ import absolute_import, division, print_function, unicode_literals
+
+import itertools
+import logging
+
+from sopel import tools
+from sopel.tools import jobs
+
+LOGGER = logging.getLogger(__name__)
+
+
+class Scheduler(jobs.Scheduler):
+    """Plugins's Job Scheduler
+
+    :param manager: bot instance passed to jobs as argument
+    :type manager: :class:`sopel.bot.Sopel`
+
+    Scheduler that stores plugin jobs and behaves like its
+    :class:`parent class <sopel.tools.jobs.Scheduler>`.
+
+    .. versionadded:: 7.1
+
+    .. note::
+
+        This class is a specific implementation of the scheduler, made to store
+        jobs by their plugins and be used by the bot (its ``manager``).
+        It follows a similar interface as the
+        :class:`plugin rules manager <sopel.plugins.rules.Manager>`.
+
+    .. important::
+
+        This is an internal tool used by Sopel to manage its jobs. To register
+        a job, plugin authors should use :func:`sopel.module.interval`.
+
+    """
+    def __init__(self, manager):
+        super(Scheduler, self).__init__(manager)
+        self._jobs = tools.SopelMemoryWithDefault(list)
+
+    def register(self, job):
+        with self._mutex:
+            self._jobs[job.get_plugin_name()].append(job)
+        LOGGER.debug('Job registered: %s', str(job))
+
+    def unregister_plugin(self, plugin_name):
+        """Unregister all the jobs from a plugin.
+
+        :param str plugin_name: the name of the plugin to remove
+        :return: the number of jobs unregistered for this plugin
+        :rtype: int
+
+        All jobs of that plugin will be removed from the scheduler.
+
+        This method is thread safe. However, it won't cancel or stop any
+        currently running jobs.
+        """
+        unregistered_jobs = 0
+        with self._mutex:
+            jobs_count = len(self._jobs[plugin_name])
+            del self._jobs[plugin_name]
+            unregistered_jobs = unregistered_jobs + jobs_count
+
+        LOGGER.debug(
+            '[%s] Successfully unregistered %d jobs',
+            plugin_name,
+            unregistered_jobs)
+
+        return unregistered_jobs
+
+    def clear_jobs(self):
+        with self._mutex:
+            self._jobs = tools.SopelMemoryWithDefault(list)
+
+        LOGGER.debug('Successfully unregistered all jobs')
+
+    def remove_callable_job(self, callable):
+        plugin_name = getattr(callable, 'plugin_name', None)
+        if not self._jobs[plugin_name]:
+            return
+
+        with self._mutex:
+            self._jobs[plugin_name] = [
+                job for job in self._jobs[plugin_name]
+                if job._handler != callable
+            ]
+
+    def _get_ready_jobs(self, now):
+        with self._mutex:
+            jobs = [
+                job for job in itertools.chain(*self._jobs.values())
+                if job.is_ready_to_run(now)
+            ]
+
+        return jobs
--- a/sopel/tools/jobs.py
+++ b/sopel/tools/jobs.py
@@ -1,10 +1,12 @@
 # coding=utf-8
 """Sopel's Job Scheduler: internal tool for job management.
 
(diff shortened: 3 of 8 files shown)


sopel/bot.py
```diff
@@ -394,16 +398,17 @@ def get_plugin_meta(self, name):
...
-        """Unregister a job or a shutdown method.
+        """Unregister a shutdown method.
 
-        :param obj: the job or shutdown method to unregister
+        :param obj: the shutdown method to unregister
         :type obj: :term:`object`
+
+        This method was used to unregister anything (rules, commands, urls,
+        jobs, and shutdown methods), but since everything can be done by other
+        mean, there is no use for it anymore.
```
dgw, line 408:
```suggestion
        means, there is no use for it anymore.
```

Typo fixing reflex.

sopel/bot.py
```diff
@@ -476,20 +481,8 @@ def register_callables(self, callables):
...
-                    callable_name,
-                    interval)
-
-    def unregister_jobs(self, jobs):
-        for job in jobs:
-            callable_name = getattr(job, "__name__", 'UNKNOWN')
-            self.scheduler.remove_callable_job(job)
-            LOGGER.debug('Job callable removed: %s', callable_name)
+            job = sopel.tools.jobs.Job.from_callable(self.settings, func)
+            self._scheduler.register(job)
```
dgw, line 485: Seems to me we should keep either both of these or neither, but not only one.

sopel/plugins/jobs.py
```diff
@@ -0,0 +1,91 @@
...
+import logging
+
+from sopel import tools
+from sopel.tools import jobs
+
+LOGGER = logging.getLogger(__name__)
+
+
+class Scheduler(jobs.JobScheduler):
+    """Scheduler of plugin jobs.
```
dgw, line 30: Should document the params to `__init__()` somewhere.

sopel/plugins/jobs.py
```diff
@@ -0,0 +1,91 @@
...
+        return unregistered_jobs
+
+    def clear_jobs(self):
+        with self._mutex:
+            self._jobs = tools.SopelMemoryWithDefault(list)
+
+        LOGGER.debug('Successfully unregistered all jobs')
+
+    def remove_callable_job(self, callable):
+        """Remove ``callable`` from the job queue."""
```
dgw, line 73:
```suggestion
        """Remove ``callable`` from the job queue.

        :param callable: the callable to remove
        :type callable: :term:`callable`
        """
```

It's kinda silly, but consistency is nice. And our new docstring guidelines (#1794) demand param docs.

sopel/tools/jobs.py
```diff
@@ -39,10 +39,11 @@ def __init__(self, manager):
         self._jobs = []
         self._mutex = threading.Lock()
 
-    def add_job(self, job):
-        """Add a Job to the current job queue."""
+    def register(self, job):
+        """Regsister a Job to the current job queue."""
```
dgw, line 43:
```suggestion
        """Register a Job to the current job queue."""
```

sopel/tools/jobs.py
```diff
@@ -139,19 +142,104 @@ class Job(object):
     execute. Current time is used to decide when the job should be executed
     next so it should only be called right after the function was called.
     """
+    @classmethod
+    def kwargs_from_callable(cls, settings, handler):
```
dgw, line 146: I'm not going to attach a note to every single one, but so many new methods here are lacking docstrings. :/

sopel/tools/jobs.py
```diff
@@ -161,41 +249,43 @@ def is_ready_to_run(self, at_time):
...
         :rtype: boolean
         """
-        return (self.next_time - at_time) <= 0
+        return not self.is_running.is_set() and any(
+            (next_time - at_time) <= 0
+            for next_time in self.next_times.values()
+        )
 
-    def next(self):
+    def next(self, current_time):
```
dgw, line 257: Update docstring with added param.

sopel/tools/jobs.py
```diff
@@ -161,41 +249,43 @@ def is_ready_to_run(self, at_time):
...
 
         return self
 
-    def __str__(self):
-        """Return a string representation of the Job object.
+    def execute(self, manager):
+        """Execute the job's handler and mark itself as running.
 
-        Example result::
+        :param manager: used as argument to the job's handler
```
dgw, line 276: `:type manager:`?

sopel/bot.py
```diff
@@ -394,16 +398,17 @@ def get_plugin_meta(self, name):
         version='7.1',
         removed_in='8.0')
     def unregister(self, obj):
```
dgw, line 400: In hindsight, I'm not sure removing functionality from this method bit by bit was the right approach. We're left with a "deprecated" method that actually won't do what it used to any more, so it's no longer backward compatible.

This is partly on me for not thinking of it before, but since this is kind of a bot internal… should we just delete this now and skip the deprecation cycle? It's either that or restore all functionality to work just like it did in 7.0.4 (and earlier), and maintain it up until 8.0.

sopel/tools/jobs.py
Exirel: Mixed.

sopel/bot.py
Exirel: Hm yeah we should remove it entirely. I didn't do it at first because... I don't know? Maybe I removed them bit by bit in so many different PRs that I didn't think about it too much.

Exirel:
OK, there is more to say about the register/unregister situation in the bot.

1. `bot.register` expects 4 parameters (callables, jobs, urls, shutdown), but `bot.unregister` expect just one callable and deal with it. What is that??? I think we should remove both, and I'd argue that we should not bother with deprecation cycle in this case.
2. `bot.register_*`  methods make sense, because they allow to register a callable, that will be transformed into some sort of object, tied to a plugin.
3. `bot.unregister_*`  methods don't, because what we usually want is to unregister a whole plugin, not just a single function (there should be other means to do that properly, like using rule's label, or command's name)
4. In fact, these `(un)register_*` methods could (should?) be private, because the *right* interface should be about adding/removing a plugin, and not dealing with individual parts. I remember when working on #1479 I had so much trouble making sense of the bot's interface approaching callables, jobs, urls, and so on
5. At some point, shutdown methods shouldn't be registered - the plugin handler can do that for us
6. Also the URL callback registration should be designed like the Rules Manager and the plugin's Scheduler, in a somewhat unified interface

It's a long process. So in this PR, I'll revert my changes on `unregister`, change what I did on `unregister_jobs`, and I'll do another PR for everything else mentioned in this comment, if that sounds good for you @dgw ?

sopel/plugins/jobs.py
Exirel: Sadly, `callable` is not a term of the Python glossary. I'll use `function` instead.

Exirel:
@dgw also, I'll make document `sopel.tools.jobs`. I'll add a big `important` block to tell it's all internal an stuff, but if I don't do that, I'll have trouble documenting `sopel.plugins.jobs`.

sopel/plugins/jobs.py
Exirel: (docstring updated on the base class)

sopel/tools/jobs.py
Exirel: Yup, next commit will be full of them. I didn't want to add docstrings at first, because it's all internal. But heh. Eventually, I prefer *with* docstrings!

sopel/tools/jobs.py
Exirel: Ended with `object`. That will do.

Exirel:
@dgw so, you were totally right in your comments. Because I added *so much* docstrings, and because there was so much to fix/change/revert, I believe a "rebase" couldn't hurt in that specific case. Most of the docstrings have been updated, added, moved, or replaced, so I think you will have to re-review everything *anyway*.

As said earlier, I won't update the `bot.unregister` method again in this PR, I'll do a separate one for that, I think it's better this way. Also, I'm sorry for all the mess. I'm glad you are here doing the review, because it really makes Sopel better. Sometimes, I'm lazy, or I forget something, or I make an honest mistake. And usually, you're here to make sure it's fixed properly, and that's invaluable! :heart: 

So good luck with the review, hope it won't be too difficult this time.

sopel/plugins/jobs.py
```diff
@@ -0,0 +1,111 @@
...
+    :class:`parent class <sopel.tools.jobs.Scheduler>`.
+
+    .. versionadded:: 7.1
+
+    .. note::
+
+        This class is a specific implementation of the scheduler, made to store
+        jobs by their plugins and be used by the bot (its ``manager``).
+        It follows a similar interface as the plugin's
+        :class:`rules manager <sopel.plugins.rules.Manager>`.
```
dgw, line 45:
```suggestion
        It follows a similar interface as the
        :class:`plugin rules manager <sopel.plugins.rules.Manager>`.
```

That short-ass line tho, oh well.

sopel/plugins/jobs.py
```diff
@@ -0,0 +1,111 @@
...
+
+    .. important::
+
+        This is an internal tool used by Sopel to manage its jobs. To register
+        a job, plugin author should uses :func:`sopel.module.interval`.
```
dgw, line 50:
```suggestion
        a job, plugin authors should use :func:`sopel.module.interval`.
```

Let me guess… 🇫🇷? 😁

sopel/tools/jobs.py
```diff
@@ -21,31 +23,74 @@
...
+    job's :meth:`~Job.is_threaded` method).
+
+    It can be started as any other thread::
+
+        # on bot's startup
+        scheduler = jobs.Scheduler(bot)
+        scheduler.start()  # run the thread forever
+
+    Then it runs forever until the :meth:`stop` method is called, usually when
+    the bot shutdowns.
```
dgw, line 44:
```suggestion
    the bot shuts down.
```

sopel/tools/jobs.py
```diff
@@ -21,31 +23,74 @@
...
 
-    It runs forever until the :attr:`stopping` event is set using the
-    :meth:`stop` method.
     """
     def __init__(self, manager):
         threading.Thread.__init__(self)
         self.manager = manager
+        """Job manager, used as argument for jobs."""
         self.stopping = threading.Event()
+        """Stopping flag. See :meth:`stop`"""
```
dgw, line 72:
```suggestion
        """Stopping flag. See :meth:`stop`."""
```

sopel/tools/jobs.py
```diff
@@ -54,20 +99,45 @@ def stop(self):
...
+                if job._handler != callable
             ]
 
     def run(self):
-        """Run forever until :attr:`stopping` event is set."""
+        """Run forever until :meth:`stop` is called.
+
+        This method waits at most a second between each iteration. At each step
+        it retrieves the jobs that are ready for execution, and, well, execute
+        them. See the :meth:`Job.execute` method for more information.
```
dgw, line 131:
```suggestion
        it retrieves the jobs that are ready for execution, and executes them.
        See the :meth:`Job.execute` method for more information.
```

I do appreciate the bit of cheek, but don't think it belongs in this kind of documentation. :/

sopel/tools/jobs.py
```diff
@@ -127,75 +199,284 @@ def _call(self, job):
...
     """Holds information about when a function should be called next.
 
-    :param int interval: number of seconds between calls to ``func``
-    :param func: function to be called
-    :type func: :term:`function`
+    :param intervals: set of intervals; each is a number of seconds between
+                      calls to ``handler``
+    :type intervals: :term:`iterable`
+    :param str plugin: optional job's plugin
+    :param str label: optional job's label
```
dgw, line 206:
```suggestion
    :param str plugin: optional plugin name to which the job belongs
    :param str label: optional label (name) for the job
```

Further suggestions invited! I'm just doing my best with the slightly odd (I assume French-inspired) thing you sometimes do with possessives.

sopel/tools/jobs.py
```diff
@@ -127,75 +199,284 @@ def _call(self, job):
...
+    :param handler: function to be called when the job is ready to execute
+    :type handler: :term:`function`
+    :param str doc: option job's documentation
```
dgw, line 209:
```suggestion
    :param str doc: optional documentation for the job
```

sopel/tools/jobs.py
```diff
@@ -127,75 +199,284 @@ def _call(self, job):
...
+
+    The ``manager`` parameter can be any kind of object; usually it's an
+    instance of :class:`sopel.bot.Sopel`.
+
+    When a job is ready, you can execute it by calling its :meth:`execute`
+    method (providing the appropriate ``manager`` argument)::
+
+        if job.is_ready_to_run(time.time()):
+            job.execute(manager)  # marked as running
+            # "next times" have been updated, the job is not running
```
dgw, line 230:
```suggestion
            # "next times" have been updated; the job is not running
```

Literally the hardest punctuation mark to learn how to use. 😹

sopel/tools/jobs.py
```diff
@@ -127,75 +199,284 @@ def _call(self, job):
...
+            # you can perform post-execute action
 
-    max_catchup = 5
-    """How many seconds the job can get behind.
+        # outside of the with statement, the job is not running anymore
+
+    .. seealso::
+
+        The :class:`sopel.plugins.jobs.Scheduler` class is specifically
+        designed for plugin's jobs, expecting an instance of
```
dgw, line 249:
```suggestion
        designed for plugins' jobs, expecting an instance of
```

sopel/tools/jobs.py
```diff
@@ -127,75 +199,284 @@ def _call(self, job):
...
 
-    This governs how much the scheduling of jobs is allowed to get behind
-    before they are simply thrown out to avoid calling the same function too
-    many times at once.
     """
+    @classmethod
+    def kwargs_from_callable(cls, handler):
+        """Generate the keyword arguments to create a new instance.
+
+        :param callable handler: callable used to generate keyword arguments
```
dgw, line 261: I'm just curious how/if Sphinx renders the type `callable`, since below you split it into a separate `:type:` line with <tt>:term:\`function\`</tt>.

sopel/tools/jobs.py
```diff
@@ -127,75 +199,284 @@ def _call(self, job):
...
+                 intervals,
+                 plugin=None,
+                 label=None,
+                 handler=None,
+                 threaded=True,
+                 doc=None):
+        # scheduling
+        now = time.time()
+        self.intervals = set(intervals)
+        """Set of intervals to execute the job at."""
```
dgw, line 305:
```suggestion
        """Set of intervals at which to execute the job."""
```

"Don't end a sentence with a preposition," they say.

sopel/tools/jobs.py
```diff
@@ -127,75 +199,284 @@ def _call(self, job):
...
+
+        # execution
+        self._handler = handler
+        self._threaded = bool(threaded)
+        self.is_running = threading.Event()
+        """Running flag: it tells if the job is running or not.
+
+        This flag is set and clear automatically by the :meth:`execute` method.
+        It is also set and clear when the job is used with the ``with``
+        statement::
```
dgw, line 325:
```suggestion
        This flag is set and cleared automatically by the :meth:`execute`
        method. It is also set and cleared when the job is used with the
        ``with`` statement::
```

(I'm guessing at text wrapping width; too lazy to open my editor just for this.)

sopel/tools/jobs.py
```diff
@@ -127,75 +199,284 @@ def _call(self, job):
...
+
+            with job:
+                # you do something before executing the job
+                # this ensure that the job is marked as "running"
```
dgw, line 329:
```suggestion
                # this ensures that the job is marked as "running"
```

sopel/tools/jobs.py
```diff
@@ -127,75 +199,284 @@ def _call(self, job):
...
+
+        .. note::
+
+            When set manually or with the ``with`` statement, the
+            :meth:`execute` won't clear this attribute.
```
dgw, line 334:
```suggestion
            :meth:`execute` method won't clear this attribute itself.
```

Still make sense / match what you meant?

sopel/tools/jobs.py
```diff
@@ -127,75 +199,284 @@ def _call(self, job):
...
-        """Return a string representation of the Job object.
+    def execute(self, manager):
+        """Execute the job's handler and mark itself as running.
 
-        Example result::
+        :param object manager: used as argument to the job's handler
+        :return: the return value from the handler's execution
 
-            <Job(2013-06-14 11:01:36.884000, 20s, <function upper at 0x02386BF0>)>
+        This method execute the job's handler, and while doing so, it will
```
dgw, line 455:
```suggestion
        This method executes the job's handler, and while doing so it will
```

sopel/tools/jobs.py
```diff
@@ -127,75 +199,284 @@ def _call(self, job):
...
+        update its running state accordingly:
+
+        1. mark the job as running (set :attr:`is_running`) if not already set
+        2. execute the handler
+        3. if it wasn't already running before, it update the next time the
```
dgw, line 460:
```suggestion
        3. if it wasn't already running before, update the next time the
```

sopel/tools/jobs.py
```diff
@@ -127,75 +199,284 @@ def _call(self, job):
...
+           job can run (:meth:`next`), then mark it as not running anymore
+           (clear :attr:`is_running`)
+
+        Then it returns the handler's return value.
+
+        .. note::
+
+            This method can be safely used with a ``with`` statement, as it
+            set/clear the :attr:`is_running` event only if not previously set
```
dgw, line 469:
```suggestion
            sets/clears the :attr:`is_running` event only if not previously set
```

sopel/plugins/jobs.py
Exirel: No, just me.

sopel/tools/jobs.py
Exirel: I need a (semi) advanced English grammar course. :-/

sopel/tools/jobs.py
Exirel: I think it doesn't. :-/

sopel/tools/jobs.py
Exirel: Actually it's even better.

Exirel:
Change applied and fixed the docstring.

Exirel:
Done as well.

PR #1881, opened 2020-06-01
rules: add find and search rules

Exirel:
Fixes #1757 

Draft PR for now, until #1873 is at least approved.

### Description

Based on #1873 this PR adds two new rules:

- **Find rules**: an anonymous rule that executes its handler for every instance found in the line, allowing to execute a function multiple times for a single trigger object
- **Search rules**: an anonymous rule that works like the original rule, but matches from anywhere in the line, allowing to execute a function exactly once on any word or pattern contained within a line, not just from its beginning

### Checklist

- [x] I have read [CONTRIBUTING.md](https://github.com/sopel-irc/sopel/blob/master/CONTRIBUTING.md)
- [x] I can and do license this contribution under the EFLv2
- [x] No issues are reported by `make qa` (runs `make quality` and `make test`)
- [x] I have tested the functionality of the things this change touches

PR #1892, opened 2020-06-16
reddit: use new `find` rule type to get all slash-refs in line

dgw:
### Description
We're no longer restricted to matching only one `u/` or `r/` reference per line thanks to #1881.

Closes no issue, because I knew I'd remember without one. 😝

### Checklist

### DIFF
--- a/sopel/modules/reddit.py
+++ b/sopel/modules/reddit.py
@@ -20,7 +20,7 @@ import prawcore
 import requests
 
 from sopel.formatting import bold, color, colors
-from sopel.module import commands, example, NOLIMIT, OP, require_chanmsg, rule, url
+from sopel.module import commands, example, find, NOLIMIT, OP, require_chanmsg, url
 from sopel.tools import time
 from sopel.tools.web import USER_AGENT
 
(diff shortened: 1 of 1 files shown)


sopel/modules/reddit.py
```diff
@@ -431,7 +431,7 @@ def get_channel_spoiler_free(bot, trigger):
         bot.say('%s is flagged as spoilers-allowed' % channel)
 
 
-@rule(r'.*(?<!\S)/?(?P<prefix>r|u)/(?P<id>[a-zA-Z0-9-_]+)\b.*')
+@find(r'.*(?<!\S)/?(?P<prefix>r|u)/(?P<id>[a-zA-Z0-9-_]+)\b.*')
```
Exirel, line 434: Do you still need the `.*` to prefix and suffix the pattern? Can't this be simplified to something like `(?P<prefix>r|u)/(?P<id>[a-zA-Z0-9-_]+)`? I didn't dive into the issue so maybe there is a smarter regex to find here.

dgw: Good call, that was one big reason for wanting `find` rules in the first place. 🤦‍♂️

PR #1893, opened 2020-06-17
bot: signal handler now checks if the bot is connected

### LINKED ISSUE #1705: Ctrl-C fails while loading
Maybe we didn't catch all the things in #1478 / #1527, or maybe changes since then regressed? At any rate, on current `master` (d896391375545a9b9fb2e770cd1b61c29ec7c363), pressing Ctrl-C while plugins are loading (before the bot has connected to IRC) prints out the "Got quit signal, shutting down" message—_but doesn't shut down_.

Once the bot has actually connected, Ctrl-C will stop it. It's just not currently possible to interrupt the startup sequence because, e.g., "Crap, I forgot to enable/update <x> plugin/config first!"

Exirel:
Closes #1705 

### Description

(based on #1890)

I moved the signal handler to `sopel.bot.Sopel`, and made sure that it sends QUIT message only if the bot is connected. When the bot is not connected, it raises `KeyboardInterrupt` instead, which is properly handled (hopefully).

### Checklist

### DIFF
--- a/sopel/irc/__init__.py
+++ b/sopel/irc/__init__.py
@@ -179,28 +179,32 @@ class AbstractBot(object):
         try:
             self.backend.run_forever()
         except KeyboardInterrupt:
+            # raised only when the bot is not connected
             LOGGER.warning('Keyboard Interrupt')
-            self.quit('KeyboardInterrupt')
-
-    # Connection Events
+            raise
 
     def on_connect(self):
         """Handle successful establishment of IRC connection."""
+        LOGGER.info('Connected, initiating setup sequence')
+
         # Request list of server capabilities. IRCv3 servers will respond with
         # CAP * LS (which we handle in coretasks). v2 servers will respond with
         # 421 Unknown command, which we'll ignore
+        LOGGER.debug('Sending CAP request')
         self.backend.send_command('CAP', 'LS', '302')
 
         # authenticate account if needed
         if self.settings.core.auth_method == 'server':
+            LOGGER.debug('Sending server auth')
             self.backend.send_pass(self.settings.core.auth_password)
         elif self.settings.core.server_auth_method == 'server':
+            LOGGER.debug('Sending server auth')
             self.backend.send_pass(self.settings.core.server_auth_password)
 
+        LOGGER.debug('Sending nick "%s"', self.nick)
         self.backend.send_nick(self.nick)
+        LOGGER.debug('Sending user "%s" (name: "%s")', self.user, self.name)
         self.backend.send_user(self.user, '+iw', self.nick, self.name)
-
-        LOGGER.info('Connected.')
 
     def on_message(self, message):
         """Handle an incoming IRC message.
--- a/sopel/irc/backends.py
+++ b/sopel/irc/backends.py
@@ -101,7 +101,7 @@ class AsynchatBackend(AbstractIRCBackend, asynchat.async_chat):
 
         :param bytes data: raw line to send
 
-        This uses :meth:`asynchat.async_chat.send` method to send ``data``
+        This uses :meth:`asyncore.dispatcher.send` method to send ``data``
         directly. This method is thread-safe.
         """
         with self.writing_lock:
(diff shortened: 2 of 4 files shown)


sopel/irc/__init__.py
```diff
@@ -179,29 +179,33 @@ def run(self, host, port=6667):
...
+
         # Request list of server capabilities. IRCv3 servers will respond with
         # CAP * LS (which we handle in coretasks). v2 servers will respond with
         # 421 Unknown command, which we'll ignore
+        LOGGER.debug('Sending CAP request')
         self.backend.send_command('CAP', 'LS', '302')
 
         # authenticate account if needed
         if self.settings.core.auth_method == 'server':
+            LOGGER.debug('Sending serveur auth')
```
RhinosF1, line 198: Why serveur auth?

Should it be server?

sopel/irc/__init__.py
```diff
@@ -179,29 +179,33 @@ def run(self, host, port=6667):
...
             self.backend.send_pass(self.settings.core.auth_password)
         elif self.settings.core.server_auth_method == 'server':
+            LOGGER.debug('Sending serveur auth')
```
RhinosF1, line 201: Same as line 198

sopel/irc/__init__.py
Exirel: Pardon my French. :fr: :rofl:

sopel/irc/__init__.py
Exirel:
```suggestion
            LOGGER.debug('Sending server auth')
```

sopel/irc/__init__.py
Exirel:
```suggestion
            LOGGER.debug('Sending server auth')
```

RhinosF1: You would think after studying a language for 7? Years I'd be able to reconigse it!

RhinosF1:
Thanks!

sopel/irc/backends.py
```diff
@@ -101,7 +101,7 @@ def irc_send(self, data):
 
         :param bytes data: raw line to send
 
-        This uses :meth:`asynchat.async_chat.send` method to send ``data``
+        This uses :meth:`asyncore.dispatcher.send` method to send ``data``
```
dgw: This is an interesting thing to catch now. I'd say this should be in #1890, but I merged it earlier… whoops! 😹

sopel/irc/__init__.py
```diff
@@ -179,29 +179,33 @@ def run(self, host, port=6667):
...
             self.backend.send_pass(self.settings.core.server_auth_password)
 
+        LOGGER.debug('Sending nick "%s"', self.nick)
         self.backend.send_nick(self.nick)
+        LOGGER.debug('Sending user "%s" (name: "%s")', self.user, self.name)
         self.backend.send_user(self.user, '+iw', self.nick, self.name)
 
-        LOGGER.info('Connected.')
```
dgw, line 203: I love the extra logger calls everywhere! Not 100% clear on what removing this one does, though. Do we still have a definitive log line that says the bot is done connecting without this?

sopel/irc/backends.py
Exirel: ikr

sopel/irc/__init__.py
Exirel: Yeah, but this was an honest lie: the bot is already connected at the very start of the method, it the setup sequence is just started at the *end* of it: the server need to reply to us about the nick, we still have to join channels, and other stuffs.

So I'd argue that the new logs (I added an INFO log to talk about the init sequence once the bot is connected) is good for *that method*, and that we absolutely need to add way more logs where they belongs (i.e. when we received JOIN acknowledgement for instance).

dgw: Works for me. Guess it's just the rebase/squash left then.

Exirel:
@dgw rebased & squashed!

PR #1894, opened 2020-06-18
rules, emoticons: action commands & commands working together at last

Exirel:
### Description

All I wanted was to do `/me flips table` and have the bot to reply to me with the best emote ever.

However, I forgot just one little detail in my last PR about rules & commands, which is to not set the `intents` attribute in `action_commands`. Everything else was done and ready to work as intended, it was just this dumb line I forgot.

Oh well, what one can do about it?

```
/me flips table
(╯°□°）╯︵ ┻━┻
```

### Checklist

### DIFF
--- a/sopel/modules/emoticons.py
+++ b/sopel/modules/emoticons.py
@@ -8,104 +8,108 @@ https://sopel.chat
 """
 from __future__ import absolute_import, division, print_function, unicode_literals
 
-from sopel.module import commands, example
+from sopel import module
 
 
-@commands('shrug')
-@example('.shrug', r'¯\_(ツ)_/¯')
+@module.commands('shrug')
+@module.action_commands('shrugs')
+@module.example('.shrug', r'¯\_(ツ)_/¯')
 def shrug(bot, trigger):
     bot.say('¯\\_(ツ)_/¯')
 
 
-@commands('happy')
-@example('.happy', 'ᕕ( ᐛ )ᕗ')
+@module.commands('happy')
+@module.example('.happy', 'ᕕ( ᐛ )ᕗ')
 def happy(bot, trigger):
     bot.say('ᕕ( ᐛ )ᕗ')
 
 
-@commands('tableflip', 'tflip')
-@example('.tableflip', '(╯°□°）╯︵ ┻━┻')
-@example('.tflip', '(╯°□°）╯︵ ┻━┻')
+@module.commands('tableflip', 'tflip')
+@module.action_commands('flips table', 'flips a table', 'flips the table')
+@module.example('.tableflip', '(╯°□°）╯︵ ┻━┻')
+@module.example('.tflip', '(╯°□°）╯︵ ┻━┻')
 def tableflip(bot, trigger):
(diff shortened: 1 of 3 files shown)


sopel/modules/emoticons.py
```diff
@@ -8,104 +8,108 @@
...
+@module.example('.happy', 'ᕕ( ᐛ )ᕗ')
 def happy(bot, trigger):
     bot.say('ᕕ( ᐛ )ᕗ')
 
 
-@commands('tableflip', 'tflip')
-@example('.tableflip', '(╯°□°）╯︵ ┻━┻')
-@example('.tflip', '(╯°□°）╯︵ ┻━┻')
+@module.commands('tableflip', 'tflip')
+@module.action_commands('flips table')
```
dgw, line 28: If we're really planning to remove regex support from all command types, gotta do this:

```suggestion
@module.action_commands('flips table', 'flips a table', 'flips the table')
```

If we can keep regex for at least action commands, and maybe nickname commands (side issue):

```suggestion
@module.action_commands('flips (?:(?:a|the) )?table')
```

I didn't think of stuff like this when I just went along with the idea that regex in commands is too confusing. I still agree that for plain old `@commands`, it is, but this case shows how the non-prefixed command types could still benefit from keeping regex support.

This definitely belongs in a separate issue; I'm just taking advantage of this specific case (and the `unflips` one below) to make my argument for me, before anything gets opened. :D

sopel/modules/emoticons.py
```diff
@@ -8,104 +8,108 @@
...
+@module.example('.tableflip', '(╯°□°）╯︵ ┻━┻')
+@module.example('.tflip', '(╯°□°）╯︵ ┻━┻')
 def tableflip(bot, trigger):
     bot.say('(╯°□°）╯︵ ┻━┻')
 
 
-@commands('unflip')
-@example('.unflip', '┬┬ ﻿ノ( ゜-゜ノ)')
+@module.commands('unflip')
+@module.action_commands('unflips')
```
dgw, line 36: See above regex discussion for `flips table`; just presenting the same alternatives here.

```suggestion
@module.action_commands('unflips', 'unflips the table')
```

```suggestion
@module.action_commands('unflips(?: the table)?')
```

sopel/modules/emoticons.py
Exirel: On the other hand, I'm still not convinced that `action_commands` is the right name for it, as they act more like generic rules. So...

sopel/modules/emoticons.py
Exirel: Note that `unflips` already matches `unflips` and `unflips <anything>`, since commands don't check arguments.

If we need to be accurate, this can be done instead:

```suggestion
@module.action_commands('unflips table', 'unflips the table')
```

dgw: Actually, good point. Let's keep only the ones with "table", so "/me unflips" won't do it.

Exirel:
Squashed & rebased.

PR #1901, opened 2020-06-28
doc: cover environment variable support

dgw:
### Description
We had a couple spots in the documentation that mentioned the `SOPEL_CONFIG` var in passing, but no _real_ documentation about it or the ability to override individual settings with envvars.

Now we'll have a whole section (in the CLI chapter) about all supported environment variables.

This is forked off of #1900's WIP branch, so the first commit can be ignored for now. (GitHub may hide it from this PR once #1900 is merged, if its smarts hold up.)

### Checklist

### DIFF
--- a/sopel/config/types.py
+++ b/sopel/config/types.py
@@ -50,9 +50,14 @@ class StaticSection(object):
 
     .. note::
 
-        Subclasses of ``StaticSection`` are named ``PluginnameSection`` by
-        convention—i.e. the plugin's name in strict CamelCase, with the suffix
-        ``Section``. However, any desired class name should work just fine.
+        By convention, subclasses of ``StaticSection`` are named with the
+        plugin's name in CamelCase, plus the suffix ``Section``. For example, a
+        plugin named ``editor`` might name its subclass ``EditorSection``; a
+        ``do_stuff`` plugin might name its subclass ``DoStuffSection`` (its
+        name converted from ``snake_case`` to ``CamelCase``).
+
+        However, this is *only* a convention. Any class name that is legal in
+        Python will work just fine.
 
     """
     def __init__(self, config, section_name, validate=True):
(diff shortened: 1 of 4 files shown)


sopel/config/types.py
```diff
@@ -47,6 +47,13 @@ class StaticSection(object):
 
     This class is intended to be subclassed and customized with added
     attributes containing :class:`BaseValidated`-based objects.
+
+    .. note::
+
+        Subclasses of ``StaticSection`` are named ``PluginnameSection`` by
```
Exirel, line 53: Does that make sense?

```suggestion
        Subclasses of ``StaticSection`` are named ``PluginNameSection`` by
```

dgw: Glad to see you're looking out for the same stuff I do. 😁

I flip-flopped on this one while writing it. On one hand, yes, CamelCase dictates that each word should be capitalized. On the other hand, Sopel encourages all-lowercase plugin names, and usually the name is only one word. Multi-word plugin names tend to end up (or should, at least) in snake_case.

So my thought was that this should illustrate the most common case: a one-word plugin name (like `wikipedia` or `calc`) but in generic form (so `pluginname`). There's probably a clearer way to express that this is a generic form, though, that doesn't violate CamelCase expectations… I'll try to figure one out.

sopel/config/types.py
```diff
@@ -47,6 +47,13 @@ class StaticSection(object):
...
+        convention—i.e. the plugin's name in strict CamelCase, with the suffix
+        ``Section``. However, any desired class name should work just fine.
```
dgw, line 55: Tagging on to @Exirel's apt suggestion on the first line of this block and trying to show an alternative:

```suggestion
        By convention, subclasses of ``StaticSection`` are named with the
        plugin's name in CamelCase, plus the suffix ``Section``. For example, a
        plugin named ``editor`` might name its subclass ``EditorSection``.
        However, any desired class name should work just fine.
```

Perhaps I should also add a note somewhere in `plugin.rst` about the conventional naming scheme for plugins (all lowercase, usually one word, or snake_case if necessary), since I'm already on the subject? Hmm…

dgw:
Totally forgot to come back to this after merging #1900 the other day. 😅

PR #1898, opened 2020-06-27
plugin: rename sopel.module into sopel.plugin

Exirel:
### Description

It partially implements the changes required by #1738 but contains only the very bare minimum changes.
I wanted to make sure that it is backward compatible. As a proof, I'll use the current test-suite.

The next steps will include updating Sopel's tests to use `sopel.plugin`, and to change Sopel's built-in plugins to use `sopel.plugin` aswell.

Note: I had to add `label` to the `__all__` magic constants to make it work properly. I guess I forgot that part back when I added this decorator.

**Edit**: Updated the doc and docstrings, because it wouldn't work otherwise. I didn't try to keep old link to the doc of `sopel.module`. I'm still working on a rewrite of the documentation related to plugins, so this will be probably discussed/taken care of at that point.

### Checklist

- [x] I have read [CONTRIBUTING.md](https://github.com/sopel-irc/sopel/blob/master/CONTRIBUTING.md)
- [x] I can and do license this contribution under the EFLv2
- [x] No issues are reported by `make qa` (runs `make quality` and `make test`)
- [x] I have tested the functionality of the things this change touches

PR #1904, opened 2020-07-09
plugins: handle URL callbacks in new rule system

### LINKED ISSUE #1804: Update URL callback registration mechanism

### LINKED ISSUE #1816: `@url` callables ignore some other plugin decorators
### Description
The `@output_prefix` decorator has no effect when used with `@url`. Output from a `@url`-decorated function is not prefixed with the given string.

### Minimal demo case
```python
@module.url('https?://example.com/')
@module.output_prefix('[Example] ')
def example_url(bot, trigger, match):
    bot.say('I should be prefixed!')
```

### Actual behavior
Triggering the function makes Sopel say, "I should be prefixed!"

### Expected behavior
Triggering the function makes Sopel say, "[Example] I should be prefixed!"

### Environment
- Sopel `.version`: 7.0.0 (also confirmed on 7.0.1-dev @ 85c6a58aa8923922b125e0f65eb0ba5f8c196dd5)
- Sopel installed via: pip (also source)
- Operating system: Ubuntu 18.04 (WSL)
- Python version: 3.6.9
- IRCd `/version`: n/a
- Relevant plugins: n/a

### Additional notes
Decorators confirmed affected:
- `@echo`
- `@output_prefix`

Note that `@url` callables bypass nick/host blocks set with the `.blocks` command because of their unique dispatch logic path, too.

Exirel:
### Description

**TL;DR**:

* URL callbacks are now part of the new rule system
* everything related to `bot.memory['url_callbacks']` is deprecated (but not marked as such in this PR)
* the `match` parameter is obsolete and shouldn't be used anymore - it still work, but not needed anymore
* there is a new `plugin.url_lazy(loader)` that takes a loader and it builds regexes when it needs to
* yeah, I did rewrote the entire world, *again*

The idea is to have `url` and `url_lazy` to be part of the rule system: the bot ask its rule manager to register them, based on what they have. As a result, URL callback are like every other rules: they can take advantage of all the decorators in `sopel.plugin` and its feature such as output prefix and rate limiting.

In effect, that makes the `bot.memory['url_callbacks']` deprecated. This PR takes care of being backward compatible, and we'll have to decide when we want to remove it. (I think Sopel 9 is a good target?)

However, I had to take a difficult decision for the `trigger` and the `match` parameters: now, the `trigger` object is based on the `match` object, making the `match` parameter obsolete. (as they contain the same information) I don't think that's a problem, as the `trigger` object still contains the full IRC line, and it couldn't be used reliably anyway because it would know about the first URL in the IRC message anyway.

Note that doesn't break callable that are supposed to be both a rule/command and a URL callback, because it's not possible: the loader doesn't allow it, and the rule/command status of the callable takes precedence.

### Checklist

### Related issues/PRs

- Replaces #1805
- Closes #1804 (in conjunction with other changes already made to URL handling)
- Closes #1816
- Requires #1906 first (should be rebased after its merged)


dgw:
Well this is exciting.

> * everything related to `bot.memory` is deprecated

I hope this means only the URL-related bits 😂

> * the `match` parameter is obsolete and shouldn't be used anymore

I'm not sure about this. Maybe I'll understand why when I actually have time to review the patch, but at first glance it sounds arbitrary. AFAIK it exists to make handing both URLs and commands/rules with the same callable.

Exirel:
> I hope this means only the URL-related bits joy

Yes! Only for the URL callback related keys in bot.memory.

> AFAIK it exists to make handing both URLs and commands/rules with the same callable.

I'd be on board with that, but the `loader.clean_module` doesn't agree with us: https://github.com/sopel-irc/sopel/blob/65fb76adb6496a3d8c75abd11c352ad02a48e54e/sopel/loader.py#L205-L215 (taken from master as of today). What this if/elif section does is that, if the callable is a generic rule or a command, it won't even try to register it as an URL Callback. So the current code doesn't allow it.

That something that I can probably change, and I'd suggest that in another PR. This one is already *quite* big.

And I reassert that I'm so glad that you took the time to raise this, because I didn't think about it, and it's a great opportunity to improve even further. Thank you. :heart:

dgw:
You rebased this before merge, @Exirel? It's showing commits from #1898 too.

sopel/plugin.py
```diff
@@ -928,12 +932,19 @@ def handle_example_bugs(bot, trigger, match):
...
+        ``match`` parameter, making ``match`` an obsolete parameter, kept
+        for backward compatibility.
+
     .. seealso::
 
         To detect URLs, Sopel uses a matching pattern built from a list of URL
         schemes, configured by
         :attr:`~sopel.config.core_section.CoreSection.auto_url_schemes`.
 
+    .. __: https://docs.python.org/3/library/re.html#match-objects
```
dgw, line 947: Just as a side note, it's possible to link this via intersphinx through `:ref:` instead of hard-coding the URL here.

sopel/modules/bugzilla.py
```diff
@@ -42,34 +39,18 @@ def configure(config):
...
-def shutdown(bot):
-    bot.unregister_url_callback(regex, show_bug)
-
 
-@rule(r'.*https?://(\S+?)'
-      r'(/show_bug.cgi\?\S*?)'
-      r'(id=\d+).*')
+@url(r'https?://(\S+?)'
+     r'(/show_bug.cgi\?\S*?)'
+     r'(id=\d+).*')
```
dgw, line 47: I thought plugins like this were the motivator for creating `plugin.url_lazy`, but you didn't switch to it here. Why not?

sopel/plugins/rules.py
```diff
@@ -343,6 +401,7 @@ class AbstractRule(object):
     def from_callable(cls, settings, handler):
         """Instantiate a rule object from ``settings`` and ``handler``.
 
+        :classmethod:
```
dgw, line 404: Doesn't Sphinx figure this out by itself? (Ditto below, for `kwargs_from_callable`.)

Searching through my (slightly outdated) local HTML docs from the last time I worked on them, there's a _classmethod_ annotation before all instances of these methods in the docs, even inherited ones.

sopel/plugins/rules.py
```diff
@@ -1301,3 +1360,123 @@ def parse(self, text):
...
+        <user> and this URL is https://example.com/other can you get it?
+        <Bot> You triggered an URL callback, with the "/other" path
+
+    Like generic rules, URL callback rules are not triggered by any specific
+    name and they don't have aliases.
+
+    .. note::
+
+        Unlike generic rules and commands, the :func:`~sopel.plugin.url`
+        decorator expect its decorated function to have the bot and the trigger
```
dgw, line 1389:
```suggestion
        decorator expects its decorated function to have the bot and the trigger
```

Found the first nemesi**S** of this PR. You're getting better!

sopel/plugins/rules.py
```diff
@@ -1301,3 +1360,123 @@ def parse(self, text):
...
+        with a third parameter: the ``match`` parameter.
+
+        To use this class with an existing URL callback handler, the
+        :meth:`from_callable` classmethod **must** be used: it will wrap the
+        handler to work as intented. In that case, the ``trigger`` and the
```
dgw, line 1394:
```suggestion
        handler to work as intended. In that case, the ``trigger`` and the
```

sopel/plugins/rules.py
```diff
@@ -1301,3 +1360,123 @@ def parse(self, text):
...
+
+    def match(self, bot, pretrigger):
+        """Match URL in a pretrigger according to the rule.
+
+        :param bot: Sopel instance
+        :type bot: :class:`sopel.bot.Sopel`
+        :param pretrigger: Line to match
+        :type pretrigger: :class:`sopel.trigger.PreTrigger`
+
+        This method looks for URL in the IRC line, and for each it yields
```
dgw, line 1452:
```suggestion
        This method looks for URLs in the IRC line, and for each it yields
```

I _think_ this is what you meant.

sopel/plugins/rules.py
```diff
@@ -1301,3 +1360,123 @@ def parse(self, text):
...
+    def __init__(self,
+                 regexes,
+                 schemes=None,
+                 **kwargs):
+        super(URLCallback, self).__init__(regexes, **kwargs)
+        # prevent mutability of registered schemes
+        self._schemes = tuple(schemes or URL_DEFAULT_SCHEMES)
+
+    def match(self, bot, pretrigger):
+        """Match URL in a pretrigger according to the rule.
```
dgw, line 1445:
```suggestion
        """Match URL(s) in a pretrigger according to the rule.
```

sopel/plugins/rules.py
```diff
@@ -1301,3 +1360,123 @@ def parse(self, text):
...
+        This method looks for URL in the IRC line, and for each it yields
+        `match objects`__ using its regexes.
+
+        .. seealso::
+
+            To detect URLs, this method uses the
+            :attr:`core.auto_url_schemes
+            <sopel.config.core_section.CoreSection.auto_url_schemes>` option.
+
+        .. __: https://docs.python.org/3.6/library/re.html#match-objects
```
dgw, line 1461: Here too, it's possible to use an inline `:ref:` instead of this hard-coded URL. Which is different from the other one I commented on, too. 😁

sopel/trigger.py
```diff
@@ -170,6 +181,11 @@ def __init__(self, own_nick, line):
                 self.tags['intent'] = intent
                 self.args[-1] = message or ''
 
+            # Search URLs if necessary, and after CTCP parsing
+            if ' :' in line:
```
dgw, line 185: Would it not make more sense to search URLs only if the text parameter itself contains `:`?

Also, I don't believe it's strictly required for the final parameter of a command to always have a `:` prefix, e.g. if the value doesn't contain spaces, so it might not always be safe to use that as a delimiter.

sopel/plugin.py
Exirel: Oh? I couldn't find it! I guess I was too tired when I wrote that. (I worked after midnight a couple of time on this so... that might be why)

sopel/modules/bugzilla.py
Exirel: Draft, mistake. :stuck_out_tongue:

sopel/trigger.py
Exirel: Oh... so you could have a PRIVMSG without the ` :` because the text doesn't contain any space? That's weird AF.

sopel/modules/bugzilla.py
Exirel: Fixed, with a bonus.

sopel/plugins/rules.py
Exirel: It does, nice catch.

sopel/trigger.py
Exirel: Welp I'm not going to be smart with that `.urls`, it's fine.

Exirel: Fixed.

Exirel:
@dgw I'll have a look at the Travis error on the `.ip` test tomorrow. Meanwhile, you can have a look, I fixed the PR. I think it's good now.

dgw:
@Exirel The `.ip` "failure" is because the GeoIP database updated to add fields for the test IP that the current regex doesn't expect. Nothing to do with this PR. I thought fetching the DB was mocked out by the VCR patches I did, but apparently it still goes to get the live copy after all. I'll push a commit to fix the regex before bed, so you can keep hacking tomorrow.

dgw:
Welp, turns out _right now_ is right before bed. I'm not sure what to do now, since 19:19 is too early to sleep… 😝

Exirel:
It's ready for the review! I'll edit the description.

@dgw I noted a bug that I should fix directly for the default branch, and not in this PR. But I've added it here so I can actually test live with my working dir.

Exirel:
See also #1906 for the hotfix.

dgw:
Commenting before review, lest I forget: Don't we space-separate names in the commit message? You have e.g. `bot,rules,loader:` but I'd expect `bot, rules, loader:`.

Hopefully I'll finish a review of this tonight or tomorrow.

sopel/plugin.py
```diff
@@ -913,11 +914,14 @@ def url(*url_rules):
         def handle_example_bugs(bot, trigger, match):
             bot.reply('Found bug ID #%s' % match.group(1))
 
-    This should be used rather than the matching in ``trigger``, in order to
-    support e.g. the ``.title`` command.
+    Both ``trigger`` and ``match`` represent the same match on the URL. The
+    ``match`` object is a Python built-in :func:`match object <re.match>`, while
```
dgw, line 918: See how it's intersphinx'd in `Trigger`: https://github.com/sopel-irc/sopel/blob/b022f461e2a8f321c53fe508a78facf8846c3e0d/sopel/trigger.py#L187

and how it appears in [the current live docs](https://sopel.chat/docs/trigger.html#sopel.trigger.Trigger) for `Trigger`'s constructor.

```suggestion
    ``match`` object is a Python built-in :ref:`match object <match-objects>`, while
```

sopel/plugins/rules.py
```diff
@@ -1301,3 +1358,140 @@ def parse(self, text):
...
+        """Match URL(s) in a pretrigger according to the rule.
+
+        :param bot: Sopel instance
+        :type bot: :class:`sopel.bot.Sopel`
+        :param pretrigger: Line to match
+        :type pretrigger: :class:`sopel.trigger.PreTrigger`
+
+        This method looks for :attr:`URLs in the IRC line
+        <sopel.trigger.PreTrigger.urls>`, and for each it yields
+        :func:`match objects <re.match>` using its regexes.
```
dgw, line 1469: See related note in `sopel/plugin.py` for example/etc.

```suggestion
        :ref:`match objects <match-objects>` using its regexes.
```

sopel/trigger.py
```diff
@@ -315,6 +332,14 @@ class Trigger(unicode):
     example, when setting ``mode -m`` on the channel ``#example``, args would
     be ``('#example', '-m')``
     """
+    urls = property(lambda self: self._pretrigger.urls)
+    """A tuple containing all URLs found in the text message.
+
+    :type: tuple
+
+    These are the URLs found in the text message of the IRC line, usually after
+    the ` :` section.
```
dgw, line 341:
```suggestion
    These are the URLs found in the last parameter of the IRC line.
```

Since you decided not to be smart, and just always run `search_urls()` on `self.args[-1]`, there's no need to even mention <tt>&nbsp;:</tt> at all.

The last parameter also isn't necessarily always the text message component, e.g. for a kick/ban with no message.

sopel/trigger.py
dgw: Servers will _probably_ normalize what gets forwarded to other clients, but it's not guaranteed.

sopel/modules/bugzilla.py
```diff
@@ -42,34 +40,35 @@ def configure(config):
...
-      r'(/show_bug.cgi\?\S*?)'
-      r'(id=\d+).*')
+@plugin.url_lazy(_bugzilla_loader)
 def show_bug(bot, trigger, match=None):
     """Show information about a Bugzilla bug."""
-    match = match or trigger
-    domain = match.group(1)
+    domain = trigger.group(1)
     if domain not in bot.config.bugzilla.domains:
         return
```
dgw, line 70: Why do we still need this if the whole thing's using URL callbacks now? It'll never match unless the domain was in the config list when `_bugzilla_loader()` was called by the rule setup.

sopel/bot.py
```diff
@@ -573,22 +572,21 @@ def unregister_shutdowns(self, shutdowns):
...
+                    self.settings, func)
+                self._rules_manager.register_url_callback(rule)
+
+            if url_lazy_loader:
+                try:
+                    rule = plugin_rules.URLCallback.from_callable_lazy(
+                        self.settings, func)
+                    self._rules_manager.register_url_callback(rule)
+                except plugins.exceptions.PluginError as err:
+                    LOGGER.error('Can not register URL callback: %s', err)
```
dgw, line 589: Did I miss this before? I feel like maybe. Whoops.

```suggestion
                    LOGGER.error('Cannot register URL callback: %s', err)
```

sopel/plugin.py
```diff
@@ -928,6 +932,12 @@ def handle_example_bugs(bot, trigger, match):
 
         More than one pattern can be provided as positional argument at once.
 
+    .. versionchanged:: 7.1
+
+        The ``trigger`` parameter now represents the same match as the
+        ``match`` parameter, making ``match`` an obsolete parameter, kept
+        for backward compatibility.
```
dgw, line 939: Question is how long we should keep `match` around in the calls. Since it (_should be_) defined as a kwarg in plugin code (`match=None`), not passing it shouldn't cause fatal errors. Might make the code fail to work as expected, though, if it's written to ignore `trigger` and use `match` only.

Can we say it'll be removed in 9.0? That should be long enough.

sopel/plugin.py
```diff
@@ -946,6 +956,54 @@ def actual_decorator(function):
...
+            bot.say('URL found: %s' % trigger.group(0))
+
+    .. versionadded:: 7.1
+
+    .. important::
+
+        There are few differences with the :func:`url` decorator:
+
+        * the decorated callable does **not** have a ``match`` parameter
+        * the decorated callable accepts **one and only one** ``loader``
```
dgw, line 998: I doubt it'd be too difficult to accept multiple `loader`s for parity with the other decorators.

We still have #1619 open for 8.0, and let's not make more work for that one. 😁

sopel/plugins/exceptions.py
```diff
@@ -16,3 +16,12 @@ def __init__(self, name):
         message = 'Plugin "%s" not registered' % name
         self.plugin_name = name
         super(PluginNotRegistered, self).__init__(message)
+
+
+class ImproperlyConfigured(PluginError):
```
dgw, line 21: I'm confused as to why we need to add this when `sopel.config.ConfigurationError` [already exists](https://github.com/sopel-irc/sopel/blob/a5a6e2dcf3ad18e582ed063a41670106c6fc2be5/sopel/config/__init__.py#L78)? Since the error is ultimately due to a missing value in the config, that seems like a perfectly appropriate exception to use.

sopel/plugins/rules.py
```diff
@@ -1301,3 +1358,140 @@ def parse(self, text):
...
+        Unlike generic rules and commands, the :func:`~sopel.plugin.url`
+        decorator expects its decorated function to have the bot and the
+        trigger with a third parameter: the ``match`` parameter.
+
+        To use this class with an existing URL callback handler, the
+        :meth:`from_callable` classmethod **must** be used: it will wrap the
+        handler to work as intended. In that case, the ``trigger`` and the
+        ``match`` arguments will be the same when the rule executes.
+
+        This behavior makes the ``match`` parameter obsolete.
```
dgw, line 1395: If we elect to mark `match` as deprecated and plan to remove it in 9.0 (or whenever), a note should be added here too.

sopel/trigger.py
```diff
@@ -80,6 +82,16 @@ class PreTrigger(object):
...
 
+    .. py:attribute:: urls
+        :type: tuple
+
+        The :class:`tuple` of URLs found in the IRC line's text, for
+        ``PRIVMSG`` and ``NOTICE`` only. For other messages, this will be an
+        empty ``tuple``.
+
+        These URLs are found in the text message of the IRC line, usually after
+        the ` :` section.
```
dgw, line 93: This is no longer accurate. It always searches the last parameter now, regardless of whether it's the "text" parameter.

```suggestion
        The :class:`tuple` of URLs found in the IRC line's final parameter.
        This will be an empty ``tuple`` if there are no URLs found.
```

Actually, now that I've made this suggestion… why is it documented both here as a `py:attribute` _and_ below in the `property` docstring? Do we need both, or can we drop down to documenting it in just one place?

sopel/plugin.py
Exirel: I don't want to, for 2 reasons:

* as a function, a dev can compose it however they want: calling another function, generating a specific list, etc. whatever they fancy
* I don't want to deal with corner cases, like "oh, one of the 3 loaders raised an exception, do we ignore it and keep the other? do we crash even if only one failed?" or something like "which one of the 10 loaders failed to deliver?" or "if there is the same pattern twice from 2 different loaders, what do we do?" or "if there are two patterns that are different but match the same thing, how should the decorator handle them?"

Lazy loading is a powerful feature, and as such, can be quite dangerous. I prefer to limit it in a way that the dev is ultimately both in control of and responsible for the outcome.

sopel/plugins/exceptions.py
Exirel: Hm. I didn't consider this. I wanted to catch `PluginError`, more than anything else, and I still wanted something specific to both plugins & settings, which `ConfigurationError` isn't.

I thought that `ConfigurationError` couldn't carry the semantic required by the plugin. For instance, the config can tell you if the list is empty when it shouldn't, but it can't tell you that you have twice the same value, or that 2 values are mutually exclusive - this is more or less up to the plugin.

So, I don't know. I need to think about it.

sopel/trigger.py
Exirel: One of them is for the `PreTrigger` class, the other for the `Trigger` class, which inherit from `str`, not `PreTrigger`.

dgw: Bah, of course. Then just worry about the wording matching the implementation. 😸

sopel/plugin.py
dgw: About corner cases, doesn't the same potential exist for e.g. one of several `rule` patterns to be invalid regex and fail to compile? Actually, all of those issues in the second point have counterparts in the existing multi-use/multi-param decorators (except for the identical-pattern one, maybe). Even a single loader function could generate multiple patterns that match the same thing by accident (or design), so why is that considered a corner case? Why not just copy the behavior of the other decorators—if something doesn't load properly, the plugin doesn't register. You're done, no need to handle corner cases.

I see where you're coming from on this, but it would be weird for the plugin API to explicitly prevent shooting yourself in the foot _only here_ when it's allowed everywhere else. Helping the user (plugin dev) avoid shooting their own foot sounds appealing, but there's only so much safety possible without somehow embedding an AGI into Sopel's rule manager. 😁

Exirel: Indeed, we can't prevent all the bullet-to-foot possibilities. After all, we are humans, and so are the users. So for me, it's about the trade off we can offer: how far can you go with one bullet? One foot? Both feet? OK, enough of this analogy. 😛 

I consider that the trade-off is as follow:

* option 1 (`url`): you get as many patterns as you want, but they are static (not limit but less power). I like it because you can statically check that the patterns are the right ones.
* option 2 (`url_lazy`): you can define the regexes when the software is running, but you can't define more than one loader (more power but strict limit). I like it because you have to test one and only one function. You don't have to run cross-validation tests or anything. You still have to make sure that all the possible states of your function are accounted for, but at least you have only one function to test.

I guess I think with unit tests in mind first, and I try to get the appropriate trade-off between having power and being error prone.

dgw: This is just to explain where I'm coming from. I threw it together, and it's not necessarily the _most_ complete argument, but I wanted to make sure this doesn't get forgotten (since it's been 4 days since we last chatted about it!)

On the other hand, I can see a plugin developer wanting to reuse certain pieces of regex-generating functionality. It's not even a toy example, but the current implementation (only one loader, period) forces:

```py
def _lazy_spam(settings):
    # figure out spam rules

def _lazy_eggs(settings):
    # figure out eggs rules

def _loader_func_eggspam(settings):
    for pattern in itertools.chain(_lazy_spam(settings), _lazy_eggs(settings)):
        yield pattern

@plugin.url_lazy(_lazy_spam)
def spam_only(bot, trigger):
    # whatever this does

@plugin.url_lazy(_loader_func_eggspam)
def spam_and_eggs(bot, trigger):
    # whatever happens here
```

With multiple loaders/decorators allowed for `url_lazy`, it's certainly cleaner to read:

```py
def _lazy_spam(settings):
    # figure out spam rules

def _lazy_eggs(settings):
    # figure out eggs rules

@plugin.url_lazy(_lazy_spam)
def spam_only(bot, trigger):
    # whatever this does

@plugin.url_lazy(_lazy_spam, _lazy_eggs)
def spam_and_eggs(bot, trigger):
    # whatever happens here

# Alternatively, we might want to leave additional positional args to `url_lazy`
# available for future enhancements, like passing extra args along to the loader
@plugin.url_lazy(_lazy_spam)
@plugin.url_lazy(_lazy_eggs)
def spam_and_eggs(bot, trigger):
    # same thing
```

On the _other_ other hand, I can see where implementing and documenting it as "one and only one loader function" _now_ doesn't preclude a later change to allow multiple—but allowing multiple from the start ties our hands later due to back-compat goals. I'm not saying that multi-loader support is a blocker for accepting this patch or anything, but I'm sure if I actually sat down and thought for more than a few minutes I could devise a real use-case for sharing loader logic like this. Whether a first-class function (like I did) or a lambda, it's pretty ugly either way if that's the only means of doing it.

sopel/plugin.py
Exirel: Yes. Note that I've seen `match` argument as named argument instead of keyword argument: nothing in the doc specify that anyway.

sopel/trigger.py
Exirel: Reading the code, it sounds like it matches the implementation: the detection of URLs is done within this `if` block:

```
if self.event == 'PRIVMSG' or self.event == 'NOTICE':
    # [...]
    self.urls = tuple(web.search_urls(self.args[-1], schemes=url_schemes))
```

And in a PRIVMSG/NOTICE, it's usually (not always, then, I guess), after the `:`. I'd like to keep the information about PRIVMSG/NOTICE, what do you suggest?

sopel/plugin.py
Exirel: Maybe all we need is more tools then? Have a look at this one:

https://github.com/sopel-irc/sopel/pull/1904/commits/e55eb1c4766a5bb6a49b4dc14a2bd0b47ad9eb37

sopel/modules/bugzilla.py
Exirel: Lol at me.

sopel/plugin.py
Exirel: Note that the same logic can be used with variations:

* make a chain loaders that check for uniqueness
* have a `rule_lazy` decorator that will behave like `url_lazy`... and the chain loaders function can be reused *without modification*!

sopel/trigger.py
dgw: Docs for the `text` attribute already talk about it, so you could just refer there?

sopel/plugin.py
dgw: Hmm, this could be a good approach in theory. I'd still prefer if such `_lazy` decorators just accepted multiple arguments directly, though.

Also: This `chain_loaders` function belongs somewhere else, probably in `tools`. `plugin` should contain decorators only.

sopel/trigger.py
Exirel: Oh yes absolutely, good idea!

sopel/plugin.py
Exirel: Yes `tools` sound like a better idea indeed!

I don't think it's good only in theory, I find it more practical to tell developers, by design, that they need to be mindful about how to merge the result of more than one loader. It also begs the question why someone would need to merge loaders, that are probably already used by another rule(s).

It feels like an edge case: who would like to have 2 rules that trigger on the same line, in the same plugin, for the same regex? Take the bugzilla, it's a good example for URL callbacks: you only need one, because the variation is about the domain name. For URLs, I believe this is probably the main if not the only reason to have a lazy loader. Domain name, base path variation, etc. that's probably it.

A use case I can think of is something like that:

```
def _regex_base(settings):
    return re.escape(settings.gitlab.my_gitlab_domain)

def _loader_pulls(settings):
    base = _regex_base(settings)
    return [base + pull_pattern]

def _loader_issue(settings):
    base = _regex_base(settings)
    return [base + issue_pattern]

@plugin.url_lazy(_loader_egg)
def pull_url(bot, trigger):
    # something something pull request

@plugin.url_lazy(_loader_spam)
def issue_url(bot, trigger):
    # something something issue
```

In that case, URL callback shares the domain name, which is known from a configuration value, but everything else depend on the type of URL callback you need.

I understand the other decorators accepts more than one pattern/name at a time, but I don't think they can be compared with their lazy versions.

Exirel: After further consideration and many conversations (with many people, including - yes - myself), I decided to follow @dgw's opinion. My conclusion to all of these conversations was: it's an edge case, if someone complain, there is nothing they can't fix with some more code.

sopel/plugins/exceptions.py
Exirel: @dgw keep or drop? I still like it, but I decided to document the `url_lazy` decorator with the `PluginError` instead.

Exirel:
@dgw some new stuffs in here! I request a new review! :heart:

Exirel:
>     * 1 for Module is imported with 'import' and 'import from'

Oh yes, that's because I have to import manually the right function depending on the Python's version: in 2.7, the function to inspect a function's parameters is not the same as the one in 3.3+. So this alarm can be ignored, and will be removed automatically when we drop support for Python 2.7.

sopel/tools/__init__.py
```diff
@@ -769,3 +769,41 @@ def get_raising_file_and_line(tb=None):
...
+        def _lazy_spam(settings):
+            # figure out spam rules
+
+        def _lazy_eggs(settings):
+            # figure out eggs rules
+
+        @plugin.url_lazy(
+            plugin.chain_loaders(_lazy_spam, _lazy_eggs))
+        def spam_and_eggs(bot, trigger):
+            # whatever happens here
```
dgw, line 795:
```suggestion
    This function takes any number of lazy loaders as arguments and merges them
    together into one. It's primarily a helper for lazy rule decorators such as
    :func:`sopel.plugin.url_lazy`.
```

We don't need that old example any more now that `@plugin.url_lazy` uses this function itself, automatically.

sopel/plugin.py
```diff
@@ -988,60 +987,25 @@ def loader(settings):
         def my_url_handler(bot, trigger):
             bot.say('URL found: %s' % trigger.group(0))
 
-    .. versionadded:: 7.1
+    Unlike the :func:`url` decorator, the handler does not have a ``match``
+    parameter.
```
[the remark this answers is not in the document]
Exirel: Oh you're right, I should remove that.

sopel/plugins/exceptions.py
dgw: @Exirel What if keep, but under a different name like `PluginConfigurationError` or `PluginSettingsError`?

Exirel: `PluginSettingsError` it is.

Exirel:
@dgw ready to squash whenever you want.

dgw:
@Exirel I looked at "changes since [my] last review" and I'm happy. Squash away, at least locally, so you can move on to that documentation thing you wanted to tackle (tomorrow? IIRC)

Exirel:
OK on this one, I think we are good. I could potentially work on deprecating the memory/url_callback thing, but that can be left to another PR.

PR #1910, opened 2020-07-28
find: switch to deque for recent line storage

dgw:
### Description
Besides giving us roughly O(1) performance for appends instead of O(n), we also get automatic trimming of old entries.

Since I was already switching things up, I also eliminated the need to reverse the list (now deque) of lines before iterating through it when executing the find-and-replace operation. Lines are now stored in reverse chronological order.

### Checklist

### Notes
I'm letting Travis do the QA checks, and will update that checkbox later.

@cottongin first suggested the idea, and expressed interest in helping test this change. I did do basic testing myself (including inspecting the data in memory with the ol' ipython plugin).

Probably should revisit a couple of my own plugins that also store recent messages like this and convert them over, since I just lazily copied find's logic.

### DIFF
--- a/sopel/modules/find.py
+++ b/sopel/modules/find.py
@@ -6,13 +6,15 @@ using the sed notation (s///) commonly found in vi/vim.
 
 Copyright 2011, Michael Yanovich, yanovich.net
 Copyright 2013, Elsie Powell, embolalia.com
-Includes contributions from: dgw, Matt Meinwald, and Morgan Goose
+Copyright 2020, dgw, technobabbl.es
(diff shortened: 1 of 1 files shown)


cottongin:
> @cottongin expressed interest in helping test this change. I did do basic testing myself (including inspecting the data in memory with the ol' ipython plugin).

In my 18 hours or so of use I've yet to see any glaring issues.  I like this a lot, though that could be because it was my idea. 😝

dgw:
> I like this a lot, though that could be because it was my idea. 😝

I edited the OP. Thought I'd credited you for the idea already, but my sleepy self forgot to do that.

PR #1911, opened 2020-07-28
instagram: fix for being blocked

cottongin:
### Description
This fixes Sopel parsing instagram links when blocked using the traditional method.

### Checklist

### DIFF
--- a/sopel/modules/instagram.py
+++ b/sopel/modules/instagram.py
@@ -39,11 +39,24 @@ def instaparse(bot, trigger, match):
     try:
         bot.say(parse_insta_json(json))
     except ParseError:
-        LOGGER.exception(
-            "Unable to find Instagram's payload for URL %s", instagram_url)
-        bot.say(
-            "Unable to parse this Instagram URL. "
-            "It's probably a temporary error; try again in a bit.")
+        try:
+            json = get_oembed_json(instagram_url)
+            bot.say(parse_oembed_json(json))
+        except ParseError:
+            LOGGER.exception(
+                "Unable to find Instagram's payload for URL %s", instagram_url)
+            bot.say(
+                "Unable to parse this Instagram URL. "
+                "It's probably a temporary error; try again in a bit.")
+        else:
+            LOGGER.info(
+                "Used fallback to oEmbed metadata; "
+                "Sopel's IP may be blocked by Instagram.")
+
+
+def get_oembed_json(url):
+    response = get("https://api.instagram.com/oembed/?url={}".format(url))
+    return response.json()
 
 
 def get_insta_json(url):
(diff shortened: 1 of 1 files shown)


sopel/modules/instagram.py
```diff
@@ -139,3 +148,12 @@ def parse_insta_json(json):
 
     # Build the message
     return ' | '.join(parts)
+
+
+def parse_oembed_json(json):
+    if not json:
+        raise ParseError("No valid JSON returned")
+    return "[insta] Post by {} | {}".format(
```
dgw, line 156: How about noting that this is fallback/limited parsing somehow?

```suggestion
    return "[insta|fallback] Post by {} | {}".format(
```

sopel/modules/instagram.py
```diff
@@ -39,11 +39,20 @@ def instaparse(bot, trigger, match):
     try:
         bot.say(parse_insta_json(json))
     except ParseError:
-        LOGGER.exception(
-            "Unable to find Instagram's payload for URL %s", instagram_url)
-        bot.say(
-            "Unable to parse this Instagram URL. "
-            "It's probably a temporary error; try again in a bit.")
+        try:
```
dgw, line 42: Also log info is always helpful. Maybe `debug` instead of `info`, idk.

```suggestion
        try:
            LOGGER.info(
                "Falling back to OEmbed metadata; "
                "Sopel's IP may be blocked by Instagram.")
```

dgw: I'm moving this log output somewhere else.

sopel/modules/instagram.py
```diff
@@ -39,11 +39,20 @@ def instaparse(bot, trigger, match):
...
+            json = get_oembed_json(instagram_url)
+            bot.say(parse_oembed_json(json))
+        except ParseError:
+            LOGGER.exception(
+                "Unable to find Instagram's payload for URL %s", instagram_url)
+            bot.say(
+                "Unable to parse this Instagram URL. "
+                "It's probably a temporary error; try again in a bit.")
```
dgw, line 50:
```suggestion
                "It's probably a temporary error; try again in a bit.")
        else:
            LOGGER.info(
                "Used fallback to oEmbed metadata; "
                "Sopel's IP may be blocked by Instagram.")
```

sopel/modules/instagram.py
cottongin: As discussed - at length - on IRC, we couldn't really settle on a format here that makes complete sense without confusing _someone_, and the existing parsing method logic already supports a `[insta] Post by user | Title` format if the other parts aren't found in the data. Thus I think sticking with my original output here is for the "best".

Perhaps this can be re-visited later.

sopel/modules/instagram.py
cottongin: This I like and will implement, commit and rebase, and force push.

dgw:
Nice. I'll defer to @kwaaak for final approval (subject to "no, _you_ do it!" 😜)

dgw:
Amended the merge commit to match @cottongin's retitling. 😝

Issue #827, opened 2015-06-28
wikipedia: Use db rather than config to determine default language per channel


embolalia:

dgw:
This is easy to do, and we should do it sooner rather than later so we can deprecate the config language value per channel during 7.x and remove it in 8.0 without the deprecation and removal being too close together. Next time I'm in the mood to work on something fairly simple, hopefully I'll remember this.

Exirel:
With #1916 merged all that is left is to wait for Sopel 8 to remove this setting.

PR #1916, opened 2020-08-03
wikipedia: implement language preferences in DB for channels & nicks

dgw:
### Description
If the DB does not contain any relevant settings for the calling nick or the channel in which the command was run, falls back to the existing `lang_per_channel` config value (if set), then `default_lang`.

Note that the `lang_per_channel` setting is considered deprecated after this change. It has to stay defined in the section class (or it wouldn't be loaded), but it's been removed from everywhere possible and documented as deprecated in the code.

### Checklist

### Notes
Upon merging this PR, update #827 to request removal of fallback to deprecated `lang_per_channel` setting for 8.0 milestone.

### DIFF
--- a/sopel/modules/wikipedia.py
+++ b/sopel/modules/wikipedia.py
@@ -13,7 +13,7 @@ import re
 from requests import get
 
 from sopel.config.types import StaticSection, ValidatedAttribute
-from sopel.module import commands, example, NOLIMIT, output_prefix, url
+from sopel.module import commands, example, NOLIMIT, OP, output_prefix, require_chanmsg, url
 from sopel.tools.web import quote, unquote
 
 try:  # TODO: Remove fallback when dropping py2
@@ -88,7 +88,10 @@ class WikipediaSection(StaticSection):
     default_lang = ValidatedAttribute('default_lang', default='en')
     """The default language to find articles from (same as Wikipedia language subdomain)."""
     lang_per_channel = ValidatedAttribute('lang_per_channel')
-    """List of ``#channel:langcode`` pairs to define Wikipedia language per channel."""
+    """List of ``#channel:langcode`` pairs to define Wikipedia language per channel.
+
+    Deprecated: Will be removed in Sopel 8. Use ``.wpclang`` to manage per-channel language settings.
+    """
 
 
 def setup(bot):
@@ -100,13 +103,32 @@ def configure(config):
     | name | example | purpose |
     | ---- | ------- | ------- |
     | default\\_lang | en | The default language to find articles from (same as Wikipedia language subdomain) |
-    | lang\\_per\\_channel | #YourPants:en,#TusPantalones:es | List of #channel:langcode pairs to define Wikipedia language per channel |
     """
     config.define_section('wikipedia', WikipediaSection)
     config.wikipedia.configure_setting(
         'default_lang',
         "Enter the default language to find articles from."
     )
+
+
+def choose_lang(bot, trigger):
+    """Determine what language to use for queries based on sender/context."""
+    user_lang = bot.db.get_nick_value(trigger.nick, 'wikipedia_lang')
+    if user_lang:
+        return user_lang
+
+    if not trigger.sender.is_nick():
+        channel_lang = bot.db.get_channel_value(trigger.sender, 'wikipedia_lang')
+        if channel_lang:
+            return channel_lang
+
+    if bot.config.wikipedia.lang_per_channel:
+        customlang = re.search('(' + trigger.sender + r'):(\w+)',
+                               bot.config.wikipedia.lang_per_channel)
+        if customlang is not None:
+            return customlang.group(2)
+
+    return bot.config.wikipedia.default_lang
 
 
 def mw_search(server, query, num):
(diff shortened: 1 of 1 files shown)


sopel/modules/wikipedia.py
```diff
@@ -109,6 +111,26 @@ def configure(config):
     )
 
 
+def choose_lang(bot, trigger):
+    """Determine what language to use for queries based on sender/context."""
+    user_lang = bot.db.get_nick_value(trigger.nick, 'wikipedia_lang')
+    channel_lang = None if trigger.sender.is_nick() else bot.db.get_channel_value(trigger.sender, 'wikipedia_lang')
```
Exirel, line 117: Instead of using `if/elif/else` structure, I'd use a `if then return` structure:

```
if user_lang:
    return user_lang

if not trigger.sender.is_nick():
    channel_lang = bot.db.get_channel_value(trigger.sender, 'wikipedia_lang')
    if channel_lang is not None:
        return channel_lang

if bot.config.wikipedia.lang_per_channel:
    # [...]
    return lang

return bot.config.wikipedia.default_lang
```

sopel/modules/wikipedia.py
```diff
@@ -88,7 +88,10 @@ class WikipediaSection(StaticSection):
     default_lang = ValidatedAttribute('default_lang', default='en')
     """The default language to find articles from (same as Wikipedia language subdomain)."""
     lang_per_channel = ValidatedAttribute('lang_per_channel')
-    """List of ``#channel:langcode`` pairs to define Wikipedia language per channel."""
+    """List of ``#channel:langcode`` pairs to define Wikipedia language per channel.
```
Exirel, line 91: How was that supposed to work with a ValidatedAttribute before that? Like a list attribute (separated by new-lines)?

I don't think it needs to be modified, but it's... weird.

sopel/modules/wikipedia.py
```diff
@@ -262,3 +275,48 @@ def wikipedia(bot, trigger):
...
+        bot.reply(
+            "Your current Wikipedia language is: {}"
+            .format(
+                bot.db.get_nick_value(
+                    trigger.nick, 'wikipedia_lang',
+                    bot.config.wikipedia.default_lang)
+            )
+        )
+    else:
+        bot.db.set_nick_value(trigger.nick, 'wikipedia_lang', trigger.group(3))
```
Exirel, line 293: Not even a check on the validity of the lang here?

sopel/modules/wikipedia.py
dgw: mfw I originally started writing a multi-return structure and second-guessed myself that the reviewer (almost always you) would prefer a single-return. 😹

If you really feel strongly about it, I'll go back to what I was going to do in the first place.

sopel/modules/wikipedia.py
dgw: Definitely not going to change how it works at the same time as deprecating it. 😅

Anyway, you can see in the deleted example under `configure()` that it was a plain old `ValidatedAttribute` that kinda-sorta looked like a `ListAttribute` but "parsed" by the regex search I moved into `choose_lang()` instead of by splitting on `,`.

Yes it was a weird way of doing things. Yes that's one of the reasons I'm quite happy to deprecate that mess. 😉

sopel/modules/wikipedia.py
dgw: Considered it.
Partially implemented it.
Realized that there's no nice way of making a regex strict enough to be useful, and I don't want to hard-code a list of valid language codes. (1. There are a lot of them. 2. Wikimedia might add more at any time.)
Deleted it. 😹

About all we could do is verify that the lang contains only a-z and hyphen, which is not particularly helpful. The user will figure out very soon if they set it wrong anyway, and it's easily changed again.

sopel/modules/wikipedia.py
Exirel: I usually prefer a multi-return structure when the only thing I need is to return a value, without modification shared between cases.

For instance, I'd do a `if/elif/else` if the result of this block had to be modified by something in all cases. Otherwise, multi-return has my preference.

Given that you wanted to go with that in the first place, I think it's sane to revert to it. Unless you want to have - how did we say? - room for improvement later? 😄

sopel/modules/wikipedia.py
Exirel: Yeah, we could use https://pypi.org/project/pycountry/ at some point, but I think we can leave that for the future.

sopel/modules/wikipedia.py
dgw: I always like leaving room for improvement, but couldn't think of what might need to be shoehorned in later. So this is restructured.

sopel/modules/wikipedia.py
dgw: Note for the future: `pycountry` language lookup won't recognize `simple`, `zh-min-nan`, `roa-tara`, or any of several other [perfectly valid](https://meta.wikimedia.org/wiki/List_of_Wikipedias) Wikipedia subdomains.

PR #1918, opened 2020-08-04
trigger: add a plain attribute to PreTrigger and Trigger

### LINKED ISSUE #1768: `trigger` should have an easy way to get plain-text (no formatting)
Note that this is a distinct feature from #462, which requests the ability to match rules on the plain-text version of each line.

My notes from back when I wrote this idea down proposed `trigger.plain` as the home for this data. We could even parallel `trigger.group` with `trigger.plain.group`… Bikeshedding welcome, I guess. This isn't going into the next release (7.0), so there's time to consider and discuss.

Exirel:
### Description

- `sopel.formatting.plain` to strip known control codes from a text
- `sopel.trigger.PreTrigger.plain` is `args`'s last item without its control code
- `sopel.trigger.Trigger.plain` exposes `sopel.trigger.PreTrigger.plain`
- Closes #1768

### Checklist

### DIFF
--- a/sopel/formatting.py
+++ b/sopel/formatting.py
@@ -95,10 +95,10 @@ CONTROL_NON_PRINTING = [
     # '\x1d', CONTROL_ITALIC
     # '\x1e', CONTROL_STRIKETHROUGH
     # '\x1f', CONTROL_UNDERLINE
+    '\x7f',
 ]
 
 # Regex to detect Control Pattern
-
 COLOR_PATTERN = re.escape(CONTROL_COLOR) + r'((\d{1,2},\d{2})|\d{2})?'
 HEX_COLOR_PATTERN = '%s(%s)?' % (
     re.escape(CONTROL_HEX_COLOR),
--- a/sopel/formatting.py
+++ b/sopel/formatting.py
@@ -8,6 +8,7 @@
 # Licensed under the Eiffel Forum License 2.
 from __future__ import absolute_import, division, print_function, unicode_literals
 
+import re
 import string
 import sys
 
@@ -32,6 +33,7 @@ __all__ = [
     'strikethrough',
     'monospace',
     'reverse',
+    'plain',
     # utility class
     'colors',
 ]
@@ -59,6 +61,31 @@ CONTROL_MONOSPACE = '\x11'
 """The control code to start or end monospace formatting."""
 CONTROL_REVERSE = '\x16'
 """The control code to start or end reverse-color formatting."""
+
+# Regex to detect Control Pattern
+
+COLOR_PATTERN = re.escape(CONTROL_COLOR) + r'((\d{1,2},\d{2})|\d{2})?'
+HEX_COLOR_PATTERN = '%s(%s)?' % (
+    re.escape(CONTROL_HEX_COLOR),
+    '|'.join([
+        '(' + ','.join([r'[a-fA-F0-9]{6}', r'[a-fA-F0-9]{6}']) + ')',
+        r'[a-fA-F0-9]{6}'
+    ])
+)
+
+PLAIN_PATTERN = '|'.join([
+    # strip color code
+    '(' + COLOR_PATTERN + ')',
+    '(' + HEX_COLOR_PATTERN + ')',
+    '(' + re.escape(CONTROL_BOLD) + ')',
+    '(' + re.escape(CONTROL_ITALIC) + ')',
+    '(' + re.escape(CONTROL_UNDERLINE) + ')',
+    '(' + re.escape(CONTROL_STRIKETHROUGH) + ')',
+    '(' + re.escape(CONTROL_MONOSPACE) + ')',
+    '(' + re.escape(CONTROL_REVERSE) + ')',
+    '(' + re.escape(CONTROL_NORMAL) + ')',
+])
+PLAIN_REGEX = re.compile(PLAIN_PATTERN)
 
 
 # TODO when we can move to 3.3+ completely, make this an Enum.
@@ -247,3 +274,12 @@ def reverse(text):
         that understand it (e.g. mIRC) can be unpredictable. Use it carefully.
     """
     return ''.join([CONTROL_REVERSE, text, CONTROL_REVERSE])
+
+
+def plain(text):
+    """Return the text without any IRC formatting.
+
+    :param str text: text with potential IRC formatting control code(s)
+    :rtype: str
+    """
+    return PLAIN_REGEX.sub('', text)
--- a/sopel/formatting.py
+++ b/sopel/formatting.py
@@ -65,9 +65,9 @@ CONTROL_REVERSE = '\x16'
 CONTROL_NON_PRINTING = [
     '\x00',
     '\x01',
-    # '\x02', CONTROL_BOLD
-    # '\x03', CONTROL_COLOR
-    # '\x04', CONTROL_HEX_COLOR
+    '\x02',  # CONTROL_BOLD
+    '\x03',  # CONTROL_COLOR
+    '\x04',  # CONTROL_HEX_COLOR
     '\x05',
     '\x06',
     '\x07',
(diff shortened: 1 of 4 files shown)


sopel/formatting.py
```diff
@@ -247,3 +274,12 @@ def reverse(text):
         that understand it (e.g. mIRC) can be unpredictable. Use it carefully.
     """
     return ''.join([CONTROL_REVERSE, text, CONTROL_REVERSE])
+
+
+def plain(text):
+    """Return the text without any IRC formatting.
+
+    :param str text: text with potential IRC formatting control code
```
dgw, line 282:
```suggestion
    :param str text: text with potential IRC formatting control code(s)
```

sopel/trigger.py
```diff
@@ -80,6 +80,10 @@ class PreTrigger(object):
         For lines that do *not* contain ``:``, :attr:`text` will be the last
         argument in :attr:`args` instead.
 
+    .. py:attribute:: plain
+
+        The last argument of the IRC command without control code.
```
dgw, line 85:
```suggestion
        The last argument of the IRC command with control codes stripped.
```

sopel/trigger.py
```diff
@@ -175,6 +180,10 @@ def __init__(self, own_nick, line):
             # Account is the second arg `...JOIN #Sopel account :realname`
             self.tags['account'] = self.args[1]
 
+        # get plain text message
+        if self.args:
+            self.plain = formatting.plain(self.args[-1])
```
dgw, line 185: Why not use `self.text`? It's already initialized to either the last arg, or the part of the line after <tt>'&nbsp;:'</tt> if present.

```suggestion
        # get plain text message
        if self.text:
            self.plain = formatting.plain(self.text)
```

sopel/trigger.py
```diff
@@ -315,6 +324,13 @@ class Trigger(unicode):
     example, when setting ``mode -m`` on the channel ``#example``, args would
     be ``('#example', '-m')``
     """
+    plain = property(lambda self: self._pretrigger.plain)
+    """The text without control code.
+
+    :type: str
+
+    This is the text of the trigger object without control code.
```
dgw, line 332:
```suggestion
    """The text without formatting control codes.

    :type: str

    This is the text of the trigger object without formatting control codes.
```

sopel/trigger.py
Exirel: Because `self.text`, in the case of a CTCP, will contain `\x01ACTION text\x01` which is not what I thought we wanted.

Exirel:
Good nitpicking! I replied about `self.text` vs `self.args[-1]`. Should be good now.

sopel/trigger.py
dgw: Hmm, arguably `self.text` shouldn't contain the CTCP bytes either, but that's a separate conversation.

dgw:
@Exirel Good to squash if we aren't changing anything else. Thoughts on stripping known ranges of control characters instead of only the ones Sopel knows about via its formatting constants?

Exirel:
A range like from 01 to 99 or something like that? I don't know/I don't have a strong opinion. What does the IRC protocol say about control code? What should a IRC client do when there is a control code it doesn't know about?

dgw:
Per RFC 1459, the trailing parameter (message text) may contain any sequence of octets except NUL, CR, or LF. A few more characters are excluded from use in middle parameters or the command target, but generally anything goes.

In short, the protocol says nothing about control codes at all. The formatting characters are all _de facto_ standards initiated by client developers, mostly originating from mIRC and then adopted by others, and even stripping out the common ones Sopel knows about could theoretically lead to removing something that wasn't inserted as formatting by a client that doesn't know about that.

It's purely a judgement call—how useful is it to remove all non-printing characters from the `plain` trigger, vs. how likely is it that doing so would break something else? (I'd say, pretty useful and not likely, respectively.)

Exirel:
Oh I see. So maybe I can remove all of these: https://en.wikipedia.org/wiki/C0_and_C1_control_codes#Basic_ASCII_control_codes

Exirel:
Done. @dgw I took the list from Wikipedia and rolled with it. Should be all good now!

sopel/formatting.py
```diff
@@ -60,6 +62,67 @@
...
+    '\x1b',
+    '\x1c',
+    # '\x1d', CONTROL_ITALIC
+    # '\x1e', CONTROL_STRIKETHROUGH
+    # '\x1f', CONTROL_UNDERLINE
+]
+
+# Regex to detect Control Pattern
+
+COLOR_PATTERN = re.escape(CONTROL_COLOR) + r'((\d{1,2},\d{2})|\d{2})?'
```
dgw, line 102: Probably clearer without the extra blank
```suggestion
# Regex to detect Control Pattern
COLOR_PATTERN = re.escape(CONTROL_COLOR) + r'((\d{1,2},\d{2})|\d{2})?'
```

sopel/formatting.py
```diff
@@ -247,3 +310,12 @@ def reverse(text):
...
     return ''.join([CONTROL_REVERSE, text, CONTROL_REVERSE])
+
+
+def plain(text):
+    """Return the text without any IRC formatting.
+
+    :param str text: text with potential IRC formatting control code(s)
+    :rtype: str
+    """
+    return PLAIN_REGEX.sub('', text)
```
dgw, line 321: Since I was thinking of making at least part of this PR myself, I looked at [ZNC's implementation](https://github.com/znc/znc/blob/700fdab3fb61061dd1b6ae46a191e040cc434c8d/src/ZNCString.cpp#L1393-L1440) a while back. Could we maybe use an iterative approach instead of regex? Since this is going to be called on every message Sopel receives, it's worth thinking about perf.

My simplistic mockup (it's past midnight, after all 😛) of a "dumb" `plain()` function using iteration did pretty well against the regex-based implementation you included here:

```
>>> text_text = "This is a string with \x02control\x02 \x1dcodes\x1d in it."
>>> from timeit import timeit
>>> from sopel.formatting import plain
>>> def plain_dumb(text):
...     out = []
...     for c in text:
...         if ord(c) < 0x20:
...             continue
...         out.append(c)
...     return ''.join(out)
... 
>>> assert plain(test_text) == plain_dumb(test_text)
>>> timeit(stmt='plain(test_text)', globals=globals())
7.396932500007097
>>> timeit(stmt='plain_dumb(test_text)', globals=globals())
3.493459499964956
```

Obviously my `plain_dumb()` function needs a little additional state-machine logic to account for formatting codes that take parameters, but even with that I bet it'd still be a large net perf gain vs. using regex.

(Commenting here because the infra above for this implementation is really long, but the actual function is just one line of "real code". 😝)

sopel/formatting.py
```diff
@@ -60,6 +62,67 @@
...
+    # '\x16', CONTROL_REVERSE
+    '\x17',
+    '\x18',
+    '\x19',
+    '\x1a',
+    '\x1b',
+    '\x1c',
+    # '\x1d', CONTROL_ITALIC
+    # '\x1e', CONTROL_STRIKETHROUGH
+    # '\x1f', CONTROL_UNDERLINE
```
dgw, line 97: Whatever the final implementation, it (IMO) should include character code `7F` in what gets stripped.

```suggestion
    # '\x1f', CONTROL_UNDERLINE
    '\x7f',
```

sopel/formatting.py
Exirel: **TL;DR**: Will try. It's way funnier to do the iterative approach, the regex is my safety net, because at Sopel's scale, the difference is nothing.

Long version: you got a fair point, of a function ~2.1 times faster than the regex version. However, since `timeit` runs the code about a million times, if we scale down to what we can expect with an IRC server - say, about 1000 messages per minutes at best? - we can argue that the overhead of the `plain` function is close to 7.4ms per minute with regex, 3.5ms per minute with iterative approach. So if I can't pull the iterative approach, we are fine. I'll definitively try it, because it's way more fun, but I'm not worried, and I hope you are not either.

Since all the tests are written independently from the implementation, I can safely try. So: hurray! \o/

sopel/formatting.py
Exirel: Will do!

Exirel:
I rebased on master, and didn't change anything from the commits, and I didn't squashed them. So @dgw you shouldn't have any trouble in the future review.

Now, let's work on that iterative approach.

sopel/formatting.py
Exirel: After a bit of work I came up with e35457574f6ebca43e4709b70940d26a4923cba5 which is a trade-off: I couldn't get a proper strip function with the (hex) color control codes, it was all messy and eventually more time consuming (yeah, perf!).

So here it is, if there is a (hex) color control code, it uses a regex. Otherwise, it doesn't. Works fine, tests are all green, so I'm pretty happy about it.

You can test the perf, it's way better, but not as good as a ~3.49s for 1,000,000: more like 4.5 without colors, 5.5 with colors. Not bad at all!

Exirel:
Hey that's weird, I thought I left a message here to say that I figured a way to do it (mostly) with an iterative approach rather than the regex one. Apparently, I did not.

So, @dgw, you can review the last commit. I benched it and it's faster than the regex only version. Basically, as long as a message doesn't contain control code for colors, this will use the basic iterative approach (which, in our case, is just a good all for-in loop). However for color control code, I couldn't find a better version, performance-wise, than the regex one.

IIRC I took the regex-only version from 7ms per 1,000 messages, to roughly 3 to 5ms per 1,000 messages with the mixed approach. So, better. Review welcome!

PR #1920, opened 2020-08-07
tools.compile_rule: better whitespace handling for `$nick` placeholder

### LINKED ISSUE #1885: rule: $nick in regex pattern is confusing
### The problem

When using `sopel.module.rule`, you can have the variable `$nick` in the regex pattern:

```
r'$nickhello' # will match "Bot: hello"
r'$nick hello' # won't match "Bot: hello", but "Bot:  hello" (two spaces)
```

The `$nick` variable is super confusing as it should probably accept `\s*` instead of `\s+` so a pattern like `r'$nick hello'` would be way easier to read and be less prone to error.

### The solution

I don't know yet. Is `\s*` an acceptable change? Do we need to wait until Sopel 8 for that?

### Notes

More documentation about these variables is required. Also this comes from a review by @dgw on #1873 where he points out that my comment should be more visible, and not just in a corner of the test suite.

dgw:
### Description
If the placeholder is followed by a space, replace it with a pattern that will consume extra whitespace but doesn't require any.

If the placeholder is NOT followed by a space, use the original pattern that requires at least one space to match.

Meant to resolve #1885.

### Checklist

### DIFF
--- a/sopel/tools/__init__.py
+++ b/sopel/tools/__init__.py
@@ -214,9 +214,11 @@ def compile_rule(nick, pattern, alias_nicks):
         nick = re.escape(nick)
 
     pattern = pattern.replace('$nickname', nick)
-    pattern = pattern.replace('$nick', r'{}[,:]\s+'.format(nick))
+    pattern = pattern.replace('$nick ', r'{}[,:]\s*'.format(nick))  # @rule('$nick hi')
+    pattern = pattern.replace('$nick', r'{}[,:]\s+'.format(nick))  # @rule('$nickhi')
     flags = re.IGNORECASE
     if '\n' in pattern:
+        # See https://docs.python.org/3/library/re.html#re.VERBOSE
         flags |= re.VERBOSE
     return re.compile(pattern, flags)
 


sopel/tools/__init__.py
```diff
@@ -214,9 +214,10 @@ def compile_rule(nick, pattern, alias_nicks):
         nick = re.escape(nick)
 
     pattern = pattern.replace('$nickname', nick)
-    pattern = pattern.replace('$nick', r'{}[,:]\s+'.format(nick))
+    pattern = pattern.replace('$nick ', r'{}[,:]\s*'.format(nick))  # @rule('$nick hi')
+    pattern = pattern.replace('$nick', r'{}[,:]\s+'.format(nick))  # @rule('$nickhi')
     flags = re.IGNORECASE
-    if '\n' in pattern:
+    if '\n' in pattern:  # TODO: Why do we have this? IRC lines can't contain \n
```
Exirel, line 220:
```suggestion
    if '\n' in pattern:
        # handle verbose regular expression
        # see https://docs.python.org/3/library/re.html#re.VERBOSE
```

dgw: I did something like this, but without applying the suggestion. Good link to have in there!

PR #1924, opened 2020-08-13
plugin: rename plural decorators to be consistent

dgw:
### Description
* `plugin.action_commands` is now `plugin.action_command`
* `plugin.commands` is now `plugin.command`
* `plugin.nickname_commands` is now `plugin.nickname_command`

The legacy imports in `sopel.module` have the old names preserved by using the `as` keyword.

Should be the last thing we need to do for #1619 (except make sure to call out the renames when writing the changelog for 7.1). I did a manual find-and-replace in all files to rename all Sphinx references to these decorators, but the small number of changes it turned up made me suspicious and I'd love a second set of eyes to double-check those. Yes, even though I also went through and reviewed all ~600 matches of "commands" in the source code & doc files.

### Checklist

### DIFF
--- a/docs/source/plugin.rst
+++ b/docs/source/plugin.rst
@@ -56,7 +56,7 @@ However, sometimes multiple words are needed for clarity or disambiguation;
 
         from sopel import plugin
 
-        @plugin.commands('hello')
+        @plugin.command('hello')
         def say_hello(bot, trigger):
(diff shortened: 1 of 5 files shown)


dgw:
(Today I learned that it's possible to minimize browser windows to the system tray by accident. So that was fun.)

PR #1927, opened 2020-08-22
jobs: fix job 'running state'

Exirel:
### Description

For some reason it was broken? I don't understand what happened here. I'll probably have to investigate further to understand the "why did I do that?" later. That will do at the moment, so no worries.

### Checklist

### DIFF
--- a/sopel/tools/jobs.py
+++ b/sopel/tools/jobs.py
@@ -185,12 +185,10 @@ class Scheduler(threading.Thread):
             self._call(job)
 
     def _call(self, job):
-        """Wrapper for collecting errors from plugins."""
+        """Wrap the job's execution to handle its state and errors."""
         try:
-            job.execute(self.manager)
-        except KeyboardInterrupt:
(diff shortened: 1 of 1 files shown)


dgw:
@Exirel I came back to properly review this but the docstring fix isn't done yet. :S

Guess the real question is: Is a threaded job _always_ considered `is_running` even if it's, um, idle? Not sure I understand the intended behavior or use of the flag, then. 😕

Exirel:
@dgw the goal is to prevent such job to run multiple time in parallel. Like, you have a job and you want it in a thread because it takes, like, 30s. It's fine, since it' should run only every 5min, but then you have a very weird case where it takes more than 5 mins, and "oh no" you don't want it to run again.

I think I wanted to be "too smart" with the `with job: job.execute()` thing, and it didn't come out as intended initially.

In short: sloppy job on my part. Will fix the docstring asap. Soon. This week.

PR #1930, opened 2020-08-28
core: change/recover in-use nick; fix obsolete USER command syntax

### LINKED ISSUE #300: Quit after timeout from server

Exirel:
### Description

Basic implementation to fix #300 (no configuration, no alternative, nothing):

1. on connection, if we receive a "nick already used" event, we change the current nick, adding an extra ` _` to its end
2. on auth with a nickserv account, if `bot.nick` is not `bot.settings.core.nick`, we try to use `NickServ GHOST`
3. when `bot.settings.core.nick` QUIT, we try to get back to that nick

Maybe this should be optional, maybe we could have a list of nick aliases to use for that purpose. But I wanted to show what is possible with the current state of the code. It tested it by using a copy of a config file, and checking that the second instance ghost the first one and get its nick back.

From another user point of view, here are the logs:

```
15:25:05 → Exibot joined (~Exibot@hostmask)
15:27:33 → Exibot_ joined (~Exibot@hostmask)
15:27:33 ⇐ Exibot quit (~Exibot@hostmask) Disconnected by services
15:27:34 Exibot_ → Exibot
```

After some discussions, I also fixed an improper use of the `USER` command, and made some minor code-style changes.

I'll do an Exirel and say: everything else is good for another PR!

### Checklist

### DIFF
--- a/sopel/coretasks.py
+++ b/sopel/coretasks.py
@@ -98,8 +98,17 @@ def auth_after_register(bot):
     else:
         return
 
+    # nickserv-based auth method needs to check for current nick
     if auth_method == 'nickserv':
-        bot.say('IDENTIFY %s' % auth_password, auth_target or 'NickServ')
+        if bot.nick != bot.settings.core.nick:
+            LOGGER.warning('Sending nickserv GHOST command.')
+            bot.say(
+                'GHOST %s %s' % (bot.settings.core.nick, auth_password),
+                auth_target or 'NickServ')
+        else:
+            bot.say('IDENTIFY %s' % auth_password, auth_target or 'NickServ')
+
+    # other methods use account instead of nick
     elif auth_method == 'authserv':
         bot.write(('AUTHSERV auth', auth_username + ' ' + auth_password))
     elif auth_method == 'Q':
@@ -120,6 +129,21 @@ def _execute_perform(bot):
         command = command.replace('$nickname', bot.config.core.nick)
         LOGGER.debug(command)
         bot.write((command,))
+
+
+@plugin.event(events.ERR_NICKNAMEINUSE)
+@plugin.thread(False)
+@plugin.priority('high')
+@plugin.unblockable
+def on_nickname_in_use(bot, trigger):
+    LOGGER.error(
+        'Nickname already in use! '
+        '(Nick: %s; Sender: %s; Args: %r)',
+        trigger.nick,
+        trigger.sender,
+        trigger.args,
+    )
+    bot.change_current_nick(bot.nick + '_')
 
 
 @module.require_privmsg("This command only works as a private message.")
@@ -597,6 +621,11 @@ def track_quit(bot, trigger):
     for channel in bot.channels.values():
         channel.clear_user(trigger.nick)
     bot.users.pop(trigger.nick, None)
+
+    if trigger.nick == bot.settings.core.nick and trigger.nick != bot.nick:
+        # old nick is now available, let's change nick again
+        bot.change_current_nick(bot.settings.core.nick)
+        auth_after_register(bot)
 
 
 @module.event('CAP')
--- a/sopel/irc/__init__.py
+++ b/sopel/irc/__init__.py
@@ -46,7 +46,6 @@ except ImportError:
     has_ssl = False
 
 from sopel import tools
-from sopel.tools import events
 from sopel.trigger import PreTrigger
 from .backends import AsynchatBackend, SSLAsynchatBackend
 from .isupport import ISupport
@@ -294,6 +293,15 @@ class AbstractBot(object):
 
         self.last_error_timestamp = datetime.now()
         self.error_count = self.error_count + 1
+
+    def change_current_nick(self, new_nick):
+        """Change the current nick without configuration modification.
+
+        :param str new_nick: new nick to be used by the bot
+        """
+        self._nick = tools.Identifier(new_nick)
+        LOGGER.debug('Sending nick "%s"', self.nick)
+        self.backend.send_nick(self.nick)
 
     def on_close(self):
         """Call shutdown methods."""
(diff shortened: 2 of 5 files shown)


Exirel:
Note that it should fix #300 which is the current oldest issue we have for now.

sopel/coretasks.py
```diff
@@ -98,8 +98,17 @@ def auth_after_register(bot):
     else:
         return
 
+    # nickserv-based auth method need to check for current nick
```
dgw, line 101:
```suggestion
    # nickserv-based auth method needs to check for current nick
```

sopel/coretasks.py
```diff
@@ -122,6 +131,21 @@ def _execute_perform(bot):
...
 
 
+@plugin.event(events.ERR_NICKNAMEINUSE)
+@plugin.thread(False)
+@plugin.priority('high')
+@plugin.unblockable
+def on_nickname_in_use(bot, trigger):
+    LOGGER.error(
+        'Nickname already in use! '
+        '(Nick: %s ; Sender: %s ; Args: %r)',
```
dgw, line 141: If there's no technical reason to have these spaces, I think they look weird in log output.

```suggestion
        '(Nick: %s; Sender: %s; Args: %r)',
```

sopel/irc/__init__.py
```diff
@@ -301,6 +295,17 @@ def on_error(self):
         self.last_error_timestamp = datetime.now()
         self.error_count = self.error_count + 1
 
+    def change_current_nick(self, new_nick):
+        """Change the current nick without configuration modification.
+
+        :param str new_nick: New nick to be used by the bot.
```
dgw, line 301: This doesn't follow our normal style for `:param:` docs.

```suggestion
        :param str new_nick: new nick to be used by the bot
```

sopel/irc/abstract_backends.py
```diff
@@ -26,6 +26,17 @@ def is_connected(self):
         """
         raise NotImplementedError
 
+    def on_irc_error(self, pretrigger):
+        """Action to perform when the server send an error event.
```
dgw, line 30:
```suggestion
        """Action to perform when the server sends an error event.
```

sopel/irc/abstract_backends.py
```diff
@@ -26,6 +26,17 @@ def is_connected(self):
...
+
+        :param pretrigger: Pretrigger object with the error event.
```
dgw, line 32: Ditto the other `:param:` note.

```suggestion
        :param pretrigger: PreTrigger object with the error event
```

sopel/irc/__init__.py
```diff
@@ -301,6 +295,17 @@ def on_error(self):
...
+        """
+        self._nick = tools.Identifier(new_nick)
+        LOGGER.debug('Sending nick "%s"', self.nick)
+        self.backend.send_nick(self.nick)
+        LOGGER.debug('Sending user "%s" (name: "%s")', self.user, self.name)
+        self.backend.send_user(self.user, '+iw', self.nick, self.name)
```
dgw, line 307: This is going to cause `ERR_ALREADYREGISTRED` when called later to regain the bot's configured nick after the ghost is killed or times out. No side effects, but it would be Better Behaved™ if Sopel sent this only when necessary.

Thanks to the handler for `RPL_WELCOME`, we already have a `bot.connection_registered` flag. But there's also no indication (to me) that a failed `NICK` command blocks the `USER` sent later in `on_connect()` here: https://github.com/sopel-irc/sopel/blob/f7bf3da022e75475c44ecbdad23edbbbe677a6fd/sopel/irc/__init__.py#L204-L207

That will still send the `USER` command even if `NICK` triggers `ERR_NICKNAMEINUSE` and fires off the handler in coretasks that comes back to call this function.

Based on all of that, I'm fairly confident that this suggestion will work just fine, but definitely test it if you apply it. 😝

```suggestion
```

sopel/irc/__init__.py
```diff
@@ -301,6 +295,17 @@ def on_error(self):
         self.last_error_timestamp = datetime.now()
         self.error_count = self.error_count + 1
 
+    def change_current_nick(self, new_nick):
+        """Change the current nick without configuration modification.
+
+        :param str new_nick: New nick to be used by the bot.
+        """
+        self._nick = tools.Identifier(new_nick)
```
dgw, line 303: Updating the bot's internal state before even sending the command makes me nervous. What if the new nick is in use, or is invalid, and the server rejects it? I'd rather update `self._nick` later, upon receiving the `NICK` event confirming the change back from the network.

But I also have to express frustration that during initial registration, the IRC protocol gives only silence—the absence of an error—as indication that the chosen NICK was accepted. This leap of faith seems unavoidable. 🙁

dgw (changes requested):
Found a few nemese**S** to slay. Despite a bit of doom and gloom in a couple of my line notes, this seems like a good approach. I'm just a bit mad at shortcomings of the IRC protocol and/or Sopel's threaded/async architecture. 😅

One other thing I couldn't touch in line notes because of GitHub's arbitrary restrictions:

```
./sopel/irc/__init__.py:49:1: F401 'sopel.tools.events' imported but unused
```

sopel/irc/__init__.py
Exirel: As discussed on IRC:

1. yes, it'll work
2. Modern IRC should be `USER <username> 0 * :<real name>` and we are sending `USER <username> +iw <nick> :<real name>` on connect so let's change that too.

dgw (approved):
I'll probably forget to include the USER fix in release notes, but that's what updating PR titles is for. 🤪

PR #1932, opened 2020-09-04
docs: add space before attribute lists

dgw:
### Description
Main motivation: Class/Module members/attributes were too close together in production docs. Finally got motivated to fix it.

### Checklist

### Notes
Might want to amend the commit/PR title if that spacing is the only tweak. Thinking I might find a few more before this is actually ready to go, but who knows? I might not.

### DIFF
--- a/docs/source/_static/custom.css
+++ b/docs/source/_static/custom.css
@@ -1,0 +1,3 @@
+.section dl.py:not(:first-of-type) {
+    margin-top: 1.5em;
+}


docs/source/_static/custom.css
```diff
@@ -0,0 +1,3 @@
+.section dl.py:not(:first-of-type) {
```
Exirel, line 1: Achieve the same effect without the `:not(...)`  stuff that might not be supported everywhere.

```suggestion
.section dl.py + dl.py {
```

docs/source/_static/custom.css
```diff
@@ -0,0 +1,3 @@
...
+    margin-top: 1.5em;
```
Exirel, line 2: This suggestion can be ignored, I just like it more with even more space:

```suggestion
    margin-top: 2em;
```

Exirel (approved):
:+1: (with nitpicking)

docs/source/_static/custom.css
dgw: If you're worried about the `:not()` selector, I don't think it's worth worrying about with 95% support and working in every modern browser under the sun: https://caniuse.com/#feat=mdn-css_selectors_not

The `:first-of-type` selector is almost as broadly supported: https://caniuse.com/#feat=mdn-css_selectors_first-of-type

Only depressingly old browsers will have trouble with this QoL style.

That said, if you have another suggestion for avoiding extra space between the dl and headings, I'm all eyes. This was just the first solution I devised. It might have other edge cases I didn't consider.

docs/source/_static/custom.css
dgw: I tried 2em first 😂

docs/source/_static/custom.css
Exirel: I checked how `:first-of-type` works exactly, because I wasn't sure, and it's slightly different from `x + y`:

```
<div>
    <h2></h2>
    <dl></dl> <!-- :first-of-type -->
    <dl></dl> <!-- :not(:first-of-type) ; dl + dl--> 
    <h2></h2>
    <dl></dl><!-- :not(:first-of-type) -->
    <dl></dl><!-- :not(:first-of-type) ; dl + dl -->
</div>
```

In our case, I think Sphinx never include other elements than `dl` at the same level, so it's technically the same in our context. However, in the general case, they don't achieve the same thing.

It's up to you. I prefer the `x + y` approach because I works better for me and how I represent things.

dgw: Looking back at the above metrics, plus https://caniuse.com/mdn-css_selectors_adjacent_sibling: I finally choose to stick with how I originally wrote the selector. Real-world stats say it has slightly _higher_ support than `+` (I was surprised too), and it's safer in case we end up doing something later that makes Sphinx give us an unexpected `dl` at the same level as something else.

dgw:
I'd rather ship this than wait to collect more tweaks. Marked ready and reworded the commit message.

docs/source/_static/custom.css
Exirel: It's a trick of the light.

Exirel (approved):

PR #1933, opened 2020-09-04
config: small refactoring and unit tests for types

Exirel:
### Description

All I wanted was to add a new type of attribute (something like a `FlagAttribute` to simplify how boolean attributes are handled in config). But then, I discovered an interface I really don't like.

So instead I wrote some unit-test, refactored some bit. I would like to go further, but I need some approval first.
Also, all these tests are still good and shouldn't go to waste or kept in a forgotten local branch.

### Checklist

### DIFF
--- a/sopel/config/types.py
+++ b/sopel/config/types.py
@@ -95,19 +95,20 @@ class StaticSection(object):
         passed, the current value of the setting will be ignored, even if it is
         not the attribute's default.
         """
-        clazz = getattr(self.__class__, name)
-        if default is NO_DEFAULT and not clazz.is_secret:
+        attribute = getattr(self.__class__, name)
+        if default is NO_DEFAULT and not attribute.is_secret:
             try:
                 default = getattr(self, name)
             except AttributeError:
                 pass
             except ValueError:
                 print('The configured value for this option was invalid.')
-                if clazz.default is not NO_DEFAULT:
-                    default = clazz.default
+                if attribute.default is not NO_DEFAULT:
+                    default = attribute.default
         while True:
             try:
-                value = clazz.configure(prompt, default, self._parent, self._section_name)
+                value = attribute.configure(
+                    prompt, default, self._parent, self._section_name)
             except ValueError as exc:
                 print(exc)
             else:
@@ -194,7 +195,9 @@ class BaseValidated(object):
             raise ValueError("You must provide a value for this option.")
 
         value = value or default
-        return self.parse(value)
+        section = getattr(parent, section_name)
+
+        return self._parse(value, parent, section)
 
     def serialize(self, value, *args, **kwargs):
         """Take some object, and return the string to be saved to the file.
@@ -212,25 +215,32 @@ class BaseValidated(object):
     def __get__(self, instance, owner=None):
         if instance is None:
             # If instance is None, we're getting from a section class, not an
-            # instance of a session class. It makes the wizard code simpler
+            # instance of a section class. It makes the wizard code simpler
             # (and is really just more intuitive) to return the descriptor
             # instance here.
             return self
 
+        value = None
         env_name = 'SOPEL_%s_%s' % (instance._section_name.upper(), self.name.upper())
         if env_name in os.environ:
             value = os.environ.get(env_name)
         elif instance._parser.has_option(instance._section_name, self.name):
             value = instance._parser.get(instance._section_name, self.name)
-        else:
-            if self.default is not NO_DEFAULT:
-                return self.default
-            raise AttributeError(
-                "Missing required value for {}.{}".format(
-                    instance._section_name, self.name
-                )
+
+        settings = instance._parent
+        section = getattr(settings, instance._section_name)
+        return self._parse(value, settings, section)
+
+    def _parse(self, value, settings, section):
+        if value is not None:
+            return self.parse(value)
+        if self.default is not NO_DEFAULT:
+            return self.default
+        raise AttributeError(
+            "Missing required value for {}.{}".format(
+                section._section_name, self.name
             )
-        return self.parse(value)
+        )
 
     def __set__(self, instance, value):
         if value is None:
@@ -238,8 +248,14 @@ class BaseValidated(object):
                 raise ValueError('Cannot unset an option with a required value.')
             instance._parser.remove_option(instance._section_name, self.name)
             return
-        value = self.serialize(value)
+
+        settings = instance._parent
+        section = getattr(settings, instance._section_name)
+        value = self._serialize(value, settings, section)
         instance._parser.set(instance._section_name, self.name, value)
+
+    def _serialize(self, value, settings, section):
+        return self.serialize(value)
 
     def __delete__(self, instance):
         instance._parser.remove_option(instance._section_name, self.name)
@@ -503,7 +519,10 @@ class ListAttribute(BaseValidated):
             else:
                 values.append(value)
             value = get_input(each_prompt + ' ')
-        return self.parse(self.serialize(values))
+
+        section = getattr(parent, section_name)
+        values = self._serialize(values, parent, section)
+        return self._parse(values, parent, section)
 
 
 class ChoiceAttribute(BaseValidated):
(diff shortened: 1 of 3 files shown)


test/config/test_config_types.py
```diff
@@ -0,0 +1,364 @@
...
+def test_list_parse_legacy_coma_strip():
+    option = types.ListAttribute('foo', strip=False)
+    assert option.parse("""value 1, # value 2   ,   value 3""") == [
+        'value 1',
+        ' # value 2   ',
+        '   value 3',
+    ]
+
+
+def test_list_parse_new_lines_legacy_coma():
```
dgw, line 249:
```suggestion
def test_list_parse_new_lines_legacy_comma():
```

test/config/test_config_types.py
```diff
@@ -0,0 +1,364 @@
...
+
+
+def test_list_serialize_value_error():
+    option = types.ListAttribute('foo')
+
+    with pytest.raises(ValueError):
+        option.serialize('value 1')
+
+    with pytest.raises(ValueError):
+        option.serialize(('1', '2', '3'))  # tuple are not allowed
```
dgw, line 293:
```suggestion
        option.serialize(('1', '2', '3'))  # tuples are not allowed
```

Or

```suggestion
        option.serialize(('1', '2', '3'))  # tuple is not allowed
```

sopel/config/types.py
```diff
@@ -194,7 +195,9 @@ def configure(self, prompt, default, parent, section_name):
             raise ValueError("You must provide a value for this option.")
 
         value = value or default
-        return self.parse(value)
+
+        section = getattr(parent, section_name)
+        return self._parse(value, parent, section)
```
dgw, line 200: Seems weird to have assignment, blank line, assignment, use both assignments in return. To me it makes more sense to have assignment, assignment, blank line, use assigned values.

```suggestion
        section = getattr(parent, section_name)

        return self._parse(value, parent, section)
```

(Nitpick mode: Active. 😝)

sopel/config/types.py
```diff
@@ -216,31 +219,43 @@ def __get__(self, instance, owner=None):
             # (and is really just more intuitive) to return the descriptor
```
dgw, line 219: Can't touch the line above this (GitHub staff, please read and implement isaacs/github#284)…

I think "session class" is a typo. Could be wrong, but it _sounds_ incorrect. Doesn't this just mean "from a section class, not a section instance"? It's confusingly worded.

sopel/config/types.py
```diff
@@ -216,31 +219,43 @@ def __get__(self, instance, owner=None):
...
             # instance here.
             return self
-
+        value = None
```
dgw, line 222: Let's leave the blank line above this.

```suggestion

        value = None
```

sopel/config/types.py
```diff
@@ -216,31 +219,43 @@ def __get__(self, instance, owner=None):
...
+
+        settings = instance._parent
+        section = getattr(settings, instance._section_name)
+        return self._parse(value, settings, section)
+
+    def _parse(self, value, settings, section):
+        if value is not None:
+            return self.parse(value)
+        if self.default is not NO_DEFAULT:
+            return self.default
```
dgw, line 237: No suggestion or anything here, just wondering if we ever talked about if and/or how `None` should be allowed as a setting value in its own right. If not, we probably should at some point. I guess we'd need a new empty class like `UnsetOption` to differentiate then.

dgw (changes requested):
I've gone through most of this, but could use a bit of background on what you were trying to accomplish with the `FilenameAttribute` changes. Without context, I just see removing a bunch of stuff that doesn't _appear_ to be replaced with new code doing the same thing. (Looking at magic `__set__`/`__get__` methods and env-var handling, specifically.)

Surely there's a connection between these changes and the ones above in `BaseValidated` etc., that I'm just not seeing. 🙁 And if that's not the code-review equivalent of "it's not you, it's me", I don't know what is. 😅

Sidebar: Now that `FilenameAttribute` creating the file/directory it's set to has been shoved in my face, I'm not sure I like that behavior. I'll add that to my list of things to think about, and write up an issue if I still don't like it after investigating.

test/config/test_config_types.py
```diff
@@ -0,0 +1,364 @@
...
+def test_validated_parse_int():
+    option = types.ValidatedAttribute('foo', parse=int)
+    assert option.parse('0') == 0
+    assert option.parse('1') == 1
+    assert option.parse('7814') == 7814
+    assert option.parse('-1') == -1
+    assert option.parse(1) == 1
+
+    with pytest.raises(ValueError):
+        assert option.parse('785.56')
```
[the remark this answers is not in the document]
Exirel: Wrong copy/paste.

sopel/config/types.py
Exirel: It's the underlying implementation (Python's ConfigParser) that's leaking up the abstraction layer: basically, setting an option to "None" means "unset".

sopel/config/types.py
Exirel: I think I was too tired when I wrote all of these, it's not the first comment where I think I should have known better.

sopel/config/types.py
Exirel: Yup it's a typo, I'll fix that.

Exirel: Fixed that for you.

Exirel:
> I've gone through most of this, but could use a bit of background on what you were trying to accomplish with the `FilenameAttribute` changes.

Before:

* `BaseValidated.parse(value, *args, **kwargs)`: public, official, `sopel.config.types`'s interface
* `BaseValidated.__get__`: I'm NOT going to use anything but `parse(value)`
* `ValidatedAttribute.parse(value)`: I laught at your public interface and do something that is NOT compatible but will force you to deal with my dumb new interface FOREVER
* `FilenameAttribute.parse(value, settings, section)`: lol I'm drunk I don't know what I'm doing (also I'm going to create file & dir in your back)
* same for `serialize`

After:

* `BaseValidated.parse(value, *args, **kwargs)`: same but we now know that, in reality, it's more `parse(value)` because of other classes
* `BaseValidated._parse(value, settings, section)`: the hack that allows to mix together the initial interface, the shitty version from `ValidatedAttribute`, and the drunk one from `FilenameAttribute`
* `FilenameAttribute.parse(value)`: now you're following the (still shitty) interface set by other classes, now it's unified!
* same for `_serialize` and `serialize`

If I could, I'd wreak havoc on `BaseValidated.parse/serialize` directly, this way I'd not have to care about creating an intermediate layer to make it works. It's your call.

> Sidebar: Now that `FilenameAttribute` creating the file/directory it's set to has been shoved in my face, I'm not sure I like that behavior. I'll add that to my list of things to think about, and write up an issue if I still don't like it after investigating.

Yeah I hate that too, but there are plugins (at least one is built-in IIRC) that expects this behaviour. It annoys me.

This whole config annoys me.

dgw:
> If I could, I'd wreak havoc on `BaseValidated.parse/serialize` directly, this way I'd not have to care about creating an intermediate layer to make it works. It's your call.

Seems like something that could be changed for… what would you say to 9.0? Though I guess, as usual, it's a trade-off between maintaining backward compatibility with older code that implements custom config types (probably rare) and making the current code more straightforward. Yeah, what else is new. 😁

But honestly, just because implementing custom config setting types seems like such an obscure thing to do, I'm not sure it's worth the effort of making sure warnings about a Grand Interface Change track across documentation and changelogs for multiple Sopel releases. I need more of a cost/benefit analysis than exists right now; "wreak havoc" doesn't tell me much except hint that it'd be a breaking change. 😝 Abstraction layers aren't inherently bad if there's a purpose for them. We do use SQLAlchemy, after all…

> > Sidebar: Now that `FilenameAttribute` creating the file/directory it's set to has been shoved in my face, I'm not sure I like that behavior. I'll add that to my list of things to think about, and write up an issue if I still don't like it after investigating.
> 
> Yeah I hate that too, but there are plugins (at least one is built-in IIRC) that expects this behaviour. It annoys me.

Maybe for 8.0 we can plan an extra kwarg for `FilenameAttribute` that can switch off auto-creating the given path? Obviously all existing code is built for the current behavior, so that would have to remain the default (for a while), but new code that doesn't want it could opt out. Shall I make a new issue for that? 😺

Exirel:
> just because implementing custom config setting types seems like such an obscure thing to do

Basically, all I need is to add 2 mandatory parameters to `parse`  and `serialize`. That would break any subclasses that override these methods without the proper signature, because they would be called with "unknown arguments".

Except, if they do it right, they should have this signature at the moment:

    def parse(self, value, *args, **kwargs):

So in theory, I *could* add the 2 parameters I want. However, since Sopel's own subclasses *don't* follow this interface, I'm not sure what other people have done. Did they follow the initial interface? Did they follow what Sopel does?

That's what I call "wreak havoc" on that code: making sure the code gets its shit together.

Exirel:
Oh, here is the main issue I have right now with what Sopel does:

```
if isinstance(descriptor, types.FilenameAttribute):
    value = descriptor.parse(value, bot.config, descriptor)
else:
    value = descriptor.parse(value)
```

So here, I've to use my new `_parse(value, settings, section)` method, because one may assume that `parse` takes only one argument.

Exirel:
@dgw any news on that one?

sopel/config/types.py
```diff
@@ -95,19 +95,20 @@ def configure_setting(self, name, prompt, default=NO_DEFAULT):
...
                 print('The configured value for this option was invalid.')
-                if clazz.default is not NO_DEFAULT:
-                    default = clazz.default
+                if attribute.default is not NO_DEFAULT:
+                    default = attribute.default
         while True:
             try:
-                value = clazz.configure(prompt, default, self._parent, self._section_name)
+                value = attribute.configure(
+                    prompt, default, self._parent, self._section_name)
```
dgw, line 111: This was not a particularly long line, but I won't block merging on a style change like this. Eventually we _do_ want to have a sensible line length, just probably a bit more than 80 col.

sopel/config/types.py
```diff
@@ -639,16 +640,5 @@ def parse(self, value, main_config, this_section):
...
-        """Used to validate ``value`` when it is changed at runtime.
-
-        :param main_config: the config object which contains this attribute
-        :type main_config: :class:`~sopel.config.Config`
-        :param this_section: the config section which contains this attribute
-        :type this_section: :class:`~StaticSection`
-        :return: the ``value``, if it is valid
-        :rtype: str
-        :raise ValueError: if the ``value`` is not valid
-        """
```
dgw, line 652: Hmm, other `serialize` methods do have a docstring describing their purpose relative to that class. In this case we don't want the one from `BaseValidated` to get inherited, because it doesn't document args and says "must be implemented in subclasses" (which doesn't make sense, because it _is_ implemented and this _is_ a subclass).

dgw (commented):
Even after switching my diff view from Unified to Split, I'm still only about 90% sure I understand all the changes. But there's a hefty new test suite to verify behavior, so while you have a chance to address my one actionable nitpick I'll go run the tests on current `master` and see if anything that shouldn't be broken, is.

sopel/config/types.py
Exirel: Fixed it.

Exirel:
And done.

dgw (approved):
I don't know how I keep forgetting about this PR. Let's go, though.

PR #1936, opened 2020-09-11
loader, plugin: load decorated callables only

### LINKED ISSUE #859: Module members with attributes named interval, rule, etc. confused for decorated functions
Modules with 'interval' config option don't work, they throw this error:

TypeError: 'ValidatedAttribute' object is not iterable (file "/home/julen/projects/sopel/sopel/bot.py", line 177, in register)

If i change the config option name to something else they work properle

Tryed with this to discard any other problem:

interval.py
import sopel
from sopel.config.types import ValidatedAttribute, StaticSection

class IntervalSection(StaticSection):
    interval = ValidatedAttribute('interval', default='5')

def configure(config):
    config.define_section('test', IntervalSection)
    config.test.configure_setting(
        'interval',
        'interval to work'
    )

def setup(bot=None):
    if not bot:
        return
    bot.config.define_section('test', IntervalSection)

Exirel:
### Description

Fix #859 and that's it.

### Checklist

### DIFF
--- a/sopel/loader.py
+++ b/sopel/loader.py
@@ -234,8 +234,11 @@ def clean_module(module, config):
     urls = []
     for obj in itervalues(vars(module)):
         if callable(obj):
+            is_sopel_callable = getattr(obj, '_sopel_callable', False) is True
             if getattr(obj, '__name__', None) == 'shutdown':
                 shutdowns.append(obj)
+            elif not is_sopel_callable:
+                continue
             elif is_triggerable(obj):
                 clean_callable(obj, config)
                 callables.append(obj)
--- a/test/test_loader.py
+++ b/test/test_loader.py
@@ -63,6 +63,25 @@ def shutdown():
 def ignored():
     pass
 
+
+@sopel.module.rate(10)
+def ignored_rate():
+    pass
+
+
+class Ignored:
+    def __init__(self):
+        self.rule = [r'.*']
+
+    def __call__(self, bot, trigger):
+        pass
+
+ignored_obj = Ignored()
+
+def ignored_trickster():
+    pass
+
+ignored_trickster._sopel_callable = True
 """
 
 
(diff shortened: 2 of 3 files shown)


sopel/loader.py
```diff
@@ -234,15 +234,16 @@ def clean_module(module, config):
...
-            elif is_triggerable(obj):
+            elif is_sopel_callable and is_triggerable(obj):
                 clean_callable(obj, config)
                 callables.append(obj)
-            elif hasattr(obj, 'interval'):
+            elif is_sopel_callable and hasattr(obj, 'interval'):
                 clean_callable(obj, config)
                 jobs.append(obj)
-            elif is_url_callback(obj):
+            elif is_sopel_callable and is_url_callback(obj):
```
dgw, line 246: Was there a reason you didn't go with a well-placed `continue` instead of adding `is_sopel_callable and` to every `if`?

```python
    for obj in itervalues(vars(module)):
        if callable(obj):
            is_sopel_callable = getattr(obj, '_sopel_callable', False) is True
            if getattr(obj, '__name__', None == 'shutdown':
                shutdowns.append(obj)
            elif not is_sopel_callable:
                continue  # skip it
            elif is_triggerable(obj):
                # et cetera; I'm not gonna retype everything xD
```

(GitHub won't let me make an easily-applied suggestion because the range includes deleted lines. Sigh.)

I get that `setup` (irrelevant here) and `shutdown` won't be decorated with something that adds the `_sopel_callable` attribute, but a `continue` after checking for the special cases should be fine and avoids re-evaluating the same condition over and over.

dgw (commented):
There's one question in line notes. Besides that, I'm also curious if it would be useful to add the `_sopel_callable` attribute to the relevant tests in `test/test_plugin.py`—maybe even some `assert not hasattr()`s for decorators that aren't supposed to add it? Just a thought. 😸

sopel/loader.py
Exirel: Yeah you're right. Why I did that is beyond me.

Exirel:
> maybe even some `assert not hasattr()`s for decorators that aren't supposed to add it?

My initial approach was to test the behaviour of the loader function, more than anything else. After all I decided to add a "protected" attribute (starting with `_`), as to make it not public.

So yeah, I'm still not sure. The behaviour is tested, and I added some corner cases to make sure everything is handled properly. So I feel like it's OK.

test/test_loader.py
```diff
@@ -194,10 +204,23 @@ def test_clean_module(testplugin, tmpconfig):
...
+    assert test_mod.ignored_rate not in jobs
+    assert test_mod.ignored_rate not in shutdowns
+    assert test_mod.ignored_rate not in urls
+    # object with a triggerable attribute are ignored by default
+    assert loader.is_triggerable(test_mod.ignored_obj)
     assert test_mod.ignored_obj not in callables
     assert test_mod.ignored_obj not in jobs
     assert test_mod.ignored_obj not in shutdowns
     assert test_mod.ignored_obj not in urls
+    # trickster function is ignored: it's still not a propre plugin callable
```
dgw, line 218:
```suggestion
    # trickster function is ignored: it's still not a proper plugin callable
```

dgw (approved):
Just one typo to fix while you're squashing. :shipit:

Exirel:
Done.

PR #1938, opened 2020-09-12
tools: new `SopelIdentifierMemory` class

### LINKED ISSUE #1341: Inconsistent SopelMemory behavior with Identifier keys & lookups by string
Looking up a `unicode`/`str` key that is not all-lowercase in a `SopelMemory` (dict) that uses `Identifier`s as keys works inconsistently. For example, `'Sopel' == Identifier('Sopel')`, but as far as `SopelMemory.__contains__()` is concerned it's not equal.

The cause is, `unicode`/`str` objects are case sensitive when hashed, whereas `Identifier` always returns `self._lowered.__hash__()`. Among other things, this affects the `ip` module looking up hostnames by nick (as string) from `bot.users` without casting to `Identifier` first (functionality introduced in #1033).

I can think of two solutions. One is just always casting to `Identifier` when indexing dictionaries by nickname (or channel name), but that's not as batteries-included as I'd like. The other solution, which I'm toying with, is creating a subclass of `SopelMemory` that's designed for using `Identifier`s as keys, and always normalizes everything. Call it an `IdentifierMemory` or `Idict` (Sopel already has a `Ddict`), for now. (Other ideas are welcome, as always.)

I'd like to get this resolved for 6.6.0, ideally. It's already been "broken" for some time, just patched around in my (and presumably others') modules by always manually casting string inputs to `Identifier` when needed. I don't think those manual casts should be necessary.

----

**Thoughts on what direction to take**

`SopelMemory` could be re-implemented as an extension of `collections.MutableMapping`, since extending `dict` [might not be the best idea](https://stackoverflow.com/a/3387975/5991). As Jochen Ritzel says there, what we really need is the _interface_ of a `dict`. What's under the hood doesn't matter that much. Abstract base classes are available from Python 2.6 and up, so this option is safe on the compatibility front. The main concern with it is performance—the `dict` built-in type is generally implemented in lower-level code (for example, C in CPython implementations), and defining a custom class in Python would mean slower operations on the object.

I'm going to also investigate overriding more of `dict` to get the desired behavior, of course. It would be ideal to preserve as much performance as possible. Just haven't looked deep enough into that path yet; focused on too many other things.

Exirel:
### Description

In an attempt to fix #1341, I decided to create a new class that specifically deals with `tools.Identifier` as its keys: nothing more, nothing less, just `Identifier`.

My idea is that we *want* `SopelMemory` to consider `"nick"` and `Identifier("nick")` as two separated keys, and what we need is a specific **type** of memory, one that is specifically designed to store and retrieve values from `Identifier` key. So that's what I did here.

### Checklist

### DIFF
--- a/sopel/tools/__init__.py
+++ b/sopel/tools/__init__.py
@@ -751,6 +751,46 @@ class SopelMemoryWithDefault(defaultdict):
         return self.__contains__(key)
 
 
+class SopelIdentifierMemory(SopelMemory):
+    """Special Sopel memory that stores ``Identifier`` as key.
+
+    This is a convenient subclass of :class:`SopelMemory` that always casts its
+    keys as instances of :class:`Identifier`::
+
+        >>> from sopel import tools
+        >>> memory = tools.SopelIdentifierMemory()
+        >>> memory['Exirel'] = 'king'
+        >>> list(memory.items())
+        [(Identifier('Exirel'), 'king')]
+        >>> tools.Identifier('exirel') in memory
+        True
+        >>> 'exirel' in memory
+        True
+
+    As seen in the example above, it is possible to perform various operations
+    with both ``Identifier`` and :class:`str` objects, taking advantage of the
+    case-insensitive behavior of ``Identifier``.
+
+    .. note::
+
+        Internally, it will try to do ``key = tools.Identifier(key)``, which
+        will raise an exception if it cannot instantiate the key properly::
+
+            >>> memory[1] = 'error'
+            AttributeError: 'int' object has no attribute 'lower'
+
+    .. versionadded:: 7.1
+    """
+    def __getitem__(self, key):
+        return super(SopelIdentifierMemory, self).__getitem__(Identifier(key))
+
+    def __contains__(self, key):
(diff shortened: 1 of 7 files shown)


sopel/tools/__init__.py
```diff
@@ -751,6 +751,46 @@ def contains(self, key):
...
+        True
+
+    As seen in the example above, it is possible to perform various operations
+    with both ``Identifier`` and :class:`str` objects, taking advantage of the
+    case-insensitive behavior of ``Identifier``.
+
+    .. note::
+
+        Internally, it will try to do ``key = tools.Identifier(key)``, which
+        will raise an exception if it cannot instanciate the key properly::
```
dgw, line 777: Let's go with the "common" spelling.

```suggestion
        will raise an exception if it cannot instantiate the key properly::
```

dgw (changes requested):
Those tests you added in `tools` have disabused me of any remaining notion that you're a humble man. 😛 Just one doc nitpick and a rebase. If you want to fix d451590's commit message so it capitalizes the new class name correctly, that's great too.

dgw (approved):
This is a nice, simple addition. Honestly can't wait to make use of it in my own plugins, too.

Issue #1648, opened 2019-06-27
tld: Expand data sources


dgw:
The `tld` plugin is not as useful as it once was. Queries for TLDs that definitely exist often fail, and successful queries return incomplete data much of the time.

Relying on Wikipedia is probably not our best long-term option…

https://github.com/sopel-irc/sopel/blob/b105fe4aaa6c1cd258337e60a4f17c1a0751ecb5/sopel/modules/tld.py#L23

At one point I was toying with the idea of adding [ICANNWiki](https://icannwiki.org/) as a source, or replacing Wikipedia entirely. I lost interest in that specific idea when I realized that ICANNWiki actually lacks _more_ details for many of its listed TLDs than Wikipedia usually does—but I still think we should try to improve the output of `.tld` commands.

Hence, this feature issue. It's not yet tied to a version milestone, and I would love to hear suggestions for other data sources that we might use for the `.tld` command besides our one Wikipedia page listing a bunch of TLDs (but not all of them).

kwaaak:
The module relies a lot on editorial content on Wikipedia.

We could use [this IANA text file](https://data.iana.org/TLD/tlds-alpha-by-domain.txt) for searching and at least give a positive result if the domain exists.
If there is more information available on Wikipedia (or other sources), display that as well.

There is no native python whois module, we could use [IANA Whois](https://www.iana.org/whois?q=net) for basic information like creation date and status.

TigerP:
You could also try using [RDAP](https://www.icann.org/rdap), which should be available for all gTLD's from august 26th 2019. But probably not for ccTLD's.

dgw:
I'm putting the finishing touches on a rewrite of this plugin. Whether it's enough to consider this issue closed is undecided yet, but it is a massive improvement nonetheless.

Sneak peek output:

```irc
<SopelTest> [tld] Name: .aq | Entity: Antarctica | IDN: ? | DNSSEC: No | SLD: Yes | IPv6: ? | Notes: Defined by
            the Antarctic Treaty as everything south of latitude 60°S. AQ domain names are available to government […]
```

dgw:
Considering this "fixed" by #1939. With IANA's list for checking existence (thanks to @kwaaak's above suggestion) and better (dynamic!) table parsing off of Wikipedia for the details, this feature should be secure Enough™ until and unless a proper API for this kind of data can be found.

PR #1939, opened 2020-09-17
tld: comprehensive rewrite

dgw:
### Description
In short, this fixes Sopel being unable to find known-valid TLDs. But it also makes things _better_ than before. No more ugly, unmaintainable regex!

* Pulls from IANA for verifying that a TLD is recognized
* If a TLD is not in the IANA list, it stops and says so
* If the given TLD exists in the IANA list, return at least a confirmation of existence
* If the TLD exists in the Wikipedia article we've been using for years, also include more details from there
* Wikipedia table data is now transformed into an in-memory dictionary using `HTMLParser`

Both the IANA and Wikipedia data sources are cached locally, persisted/restored to/from the bot's database on shutdown/startup, and periodically updated by a plugin job (minimum age 7 days).

I think about the only thing I didn't do was add example-tests. But I haven't decided _against_ doing so. I've just not gotten to it.

### Checklist

### See also
This might be sufficient to consider #1648 as "done". I've been unable to find any viable alternative data source that actually provides the information contained in Wikipedia's tables—even the basics like whether a TLD allows second-level registrations or supports DNSSEC/IDN/etc.

### DIFF
--- a/sopel/modules/tld.py
+++ b/sopel/modules/tld.py
@@ -17,7 +17,7 @@ import sys
 
 import requests
 
-from sopel import plugin, tools
+from sopel import formatting, plugin, tools
 
 if sys.version_info.major >= 3:
     unicode = str
@@ -88,6 +88,7 @@ class WikipediaTLDListParser(HTMLParser):
         self.current_cell = ''
         self.rows = []
         self.tables = []
+        self.finished = False
 
     def handle_starttag(self, tag, attrs):
         if tag == 'td' or tag == 'th':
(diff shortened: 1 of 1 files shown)


dgw:
Update: I was gathering examples of the new output and found [one table](https://en.wikipedia.org/wiki/List_of_Internet_top-level_domains#Internationalized_country_code_top-level_domains) that gets parsed wrong. Will need to investigate that one. A quick look at the HTML just now didn't reveal any obvious structural difference, and I unfortunately can't spend any more time on this until Friday evening (9/18 in central US, for our non-American project members) at the earliest.

Still feels good to have _mostly_ done, even if a gremlin or two remain lurking.

Exirel (approved):
AFAICT it's working fine.

So yeah, sure, one table is messing with us. It's still far better than the current version.

dgw:
I'll just merge this as-is and follow up with another PR to fix that table later (or an issue, if I don't get to it soon). Like @Exirel said, "it's working fine", and I agree that one table misbehaving a bit shouldn't block the _vast_ improvement that is the rest of this PR.

PR #1940, opened 2020-09-18
isup: fix error-series HTTP status codes; get away from assuming TLD

dgw:
### Description
Unbelievably, `isup` has been sometimes (often?) returning "looks fine" for sites that send back error status codes, e.g. HTTP 4xx, 5xx. Those should be considered "down".

I also threw in a couple other little tweaks, the important one being that a bare name like `google` will no longer assume `google.com` as the input. (There's one last TODO left for myself before this PR is actually _ready_ for review, about properly handling full URLs that contain a `.` but refer to a LAN name.)

### Checklist

### DIFF
--- a/sopel/modules/isup.py
+++ b/sopel/modules/isup.py
@@ -57,7 +57,9 @@ def handle_isup(bot, trigger, secure=True):
     """
     try:
         site = get_site_url(trigger.group(2))
-        response = requests.head(site, verify=secure).headers
+        response = requests.head(site, verify=secure)
+        response.raise_for_status()
+        response = response.headers
     except ValueError as error:
         bot.reply(str(error))
     except requests.exceptions.SSLError:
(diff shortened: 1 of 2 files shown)


Exirel:
> e.g. HTTP 4xx, 5xx. Those should be considered "down".

I understand what you say with that quoted "down", and I'd say we need to inform the end-user about the difference between being unreachable, falling into a server error (5xx), and just a URL that cannot be accessed from here (4xx). For instance, 401/403 means the site requires authentication/authorization, not that it is down. The same way, a 404 may mean that you are trying an URL that doesn't return anything, and yet, the site itself isn't down, you can still access it.

So, maybe we should change the message for these cases instead.

dgw:
Hmm… That makes a lot of sense, though I was (unconsciously?) associating our `.isup` command with `isup.me`/`downforeveryoneorjustme.com`—which just calls unreachable, timeout, HTTP error, whatever… all "down". It doesn't get specific. Maybe I should "pull an @Exirel" and say let's get more specific in a separate PR? 😜

Exirel:
> let's get more specific in a separate PR? 😜

Yes!

dgw:
Took care of my last TODO. Ready for a real look. 😸

Exirel (approved):
As discussed on IRC, I'd have done that differently, and that doesn't change the fact that this implementation works perfectly. So let's 🚢 it!

PR #1944, opened 2020-09-22
plugins: command-regexp functions are now private

### LINKED ISSUE #1887: tools: command-regexp functions shouldn't be public
The major rule-system rework in #1873 is a great excuse to take the `get_*_regexp()` and `get_*_pattern()` functions from `sopel.tools` and either shove them into someplace like `sopel.plugins.rules.utils` or encapsulate the pattern generation into the different rule object classes.

@Exirel let's come back to this after #1873 is finished.

**Full to-deprecate list (for 7.1):**
- [x] `compile_rule()` [#2027]
- [x] `get_command_regexp()` [#1944]
- [x] `get_command_pattern()` [#1944]
- [x] `get_nickname_command_regexp()` [#1944]
- [x] `get_nickname_command_pattern()` [#1944]
- [x] `get_action_command_regexp()` [#1944]
- [x] `get_action_command_pattern()` [#1944]

**Full to-delete list (for 8.0):**
- [x] `compile_rule()` [#2141]
- [x] `get_command_regexp()` [#2141]
- [x] `get_command_pattern()` [#2141]
- [x] `get_nickname_command_regexp()` [#2141]
- [x] `get_nickname_command_pattern()` [#2141]
- [x] `get_action_command_regexp()` [#2141]
- [x] `get_action_command_pattern()` [#2141]

Exirel:
### Description

Fix #1887 

* moved everything into `sopel.plugins.rules`, directly into each rule class
* removed/deprecated a bunch of stuff from `sopel.test_tools` (this entire module should probably be deprecated and removed at some point anyway)
* reworked imports in `sopel.test_tools`
* changed how the `example` decorator runs a test (this should be transformed further, probably in a future PR, for 7.1 or 8.x)
* removed obsolete tests and fixtures
* made sure the doc is still OK

### Checklist

### DIFF
--- a/sopel/plugins/rules.py
+++ b/sopel/plugins/rules.py
@@ -56,6 +56,7 @@ __all__ = [
     'URLCallback',
 ]
 
+
 LOGGER = logging.getLogger(__name__)
 
 IGNORE_RATE_LIMIT = 1  # equivalent to sopel.plugin.NOLIMIT
@@ -1017,6 +1018,24 @@ class Command(NamedRuleMixin, Rule):
         <Bot> You just invoked the command 'dummy' (as 'dummy-alias')
 
     """
+    # This regexp matches equivalently and produces the same
+    # groups 1 and 2 as the old regexp: r'^%s(%s)(?: +(.*))?$'
+    # The only differences should be handling all whitespace
+    # like spaces and the addition of groups 3-6.
+    PATTERN_TEMPLATE = r"""
+        (?:{prefix})({command}) # Command as group 1.
+        (?:\s+              # Whitespace to end command.
+        (                   # Rest of the line as group 2.
+        (?:(\S+))?          # Parameters 1-4 as groups 3-6.
+        (?:\s+(\S+))?
+        (?:\s+(\S+))?
+        (?:\s+(\S+))?
+        .*                  # Accept anything after the parameters.
+                            # Leave it up to the plugin to parse the line.
+        ))?                 # Group 2 must be None, if there are no parameters.
+        $                   # EoL, so there are no partial matches.
+    """
+
     @classmethod
     def from_callable(cls, settings, handler):
         prefix = settings.core.prefix
@@ -1093,13 +1112,18 @@ class Command(NamedRuleMixin, Rule):
         * the rule's name (escaped for regex if needed),
         * all of its aliases (escaped for regex if needed),
 
-        and everything is then given to :func:`sopel.tools.get_command_regexp`,
-        which creates a compiled regex to return.
+        to create a compiled regex to return.
         """
         name = [self.escape_name(self._name)]
         aliases = [self.escape_name(alias) for alias in self._aliases]
         pattern = r'|'.join(name + aliases)
-        return tools.get_command_regexp(self._prefix, pattern)
+
+        # Escape all whitespace with a single backslash.
+        # This ensures that regexp in the prefix is treated as it was before
+        # the actual regexp was changed to use the verbose syntax.
+        prefix = re.sub(r"(\s)", r"\\\1", self._prefix)
+        pattern = self.PATTERN_TEMPLATE.format(prefix=prefix, command=pattern)
+        return re.compile(pattern, re.IGNORECASE | re.VERBOSE)
 
 
 class NickCommand(NamedRuleMixin, Rule):
@@ -1124,6 +1148,22 @@ class NickCommand(NamedRuleMixin, Rule):
 
     Apart from that, it behaves exactly like a :class:`generic rule <Rule>`.
     """
+    PATTERN_TEMPLATE = r"""
+        ^
+        $nickname[:,]? # Nickname.
+        \s+({command}) # Command as group 1.
+        (?:\s+         # Whitespace to end command.
+        (              # Rest of the line as group 2.
+        (?:(\S+))?     # Parameters 1-4 as groups 3-6.
+        (?:\s+(\S+))?
+        (?:\s+(\S+))?
+        (?:\s+(\S+))?
+        .*             # Accept anything after the parameters. Leave it up to
+                       # the plugin to parse the line.
+        ))?            # Group 1 must be None, if there are no parameters.
+        $              # EoL, so there are no partial matches.
+    """
+
     @classmethod
     def from_callable(cls, settings, handler):
         nick = settings.core.nick
@@ -1196,15 +1236,15 @@ class NickCommand(NamedRuleMixin, Rule):
         * the rule's name (escaped for regex),
         * all of its aliases (escaped for regex),
 
-        and everything is then given to
-        :func:`sopel.tools.get_nickname_command_regexp`, which creates a
-        compiled regex to return.
+        to create a compiled regex to return.
         """
         name = [self.escape_name(self._name)]
         aliases = [self.escape_name(alias) for alias in self._aliases]
         pattern = r'|'.join(name + aliases)
-        return tools.get_nickname_command_regexp(
-            self._nick, pattern, self._nick_aliases)
+        return tools.compile_rule(
+            self._nick,
+            self.PATTERN_TEMPLATE.format(command=pattern),
+            self._nick_aliases)
 
 
 class ActionCommand(NamedRuleMixin, Rule):
@@ -1226,6 +1266,21 @@ class ActionCommand(NamedRuleMixin, Rule):
     Apart from that, it behaves exactly like a :class:`generic rule <Rule>`.
     """
     INTENT_REGEX = re.compile(r'ACTION', re.IGNORECASE)
+    PATTERN_TEMPLATE = r"""
+        ({command})         # Command as group 1.
+        (?:\s+              # Whitespace to end command.
+        (                   # Rest of the line as group 2.
+        (?:(\S+))?          # Parameters 1-4 as groups 3-6.
+        (?:\s+(\S+))?
+        (?:\s+(\S+))?
+        (?:\s+(\S+))?
+        .*                  # Accept anything after the parameters.
+                            # Leave it up to the plugin to parse
+                            # the line.
+        ))?                 # Group 2 must be None, if there are no
+                            # parameters.
+        $                   # EoL, so there are no partial matches.
+    """
 
     @classmethod
     def from_callable(cls, settings, handler):
@@ -1267,14 +1322,13 @@ class ActionCommand(NamedRuleMixin, Rule):
         * the rule's name (escaped for regex if needed),
         * all of its aliases (escaped for regex if needed),
 
-        and everything is then given to
-        :func:`sopel.tools.get_action_command_regexp`, which creates a compiled
-        regex to return.
+        to create a compiled regex to return.
         """
         name = [self.escape_name(self._name)]
         aliases = [self.escape_name(alias) for alias in self._aliases]
         pattern = r'|'.join(name + aliases)
-        return tools.get_action_command_regexp(pattern)
+        pattern = self.PATTERN_TEMPLATE.format(command=pattern)
+        return re.compile(pattern, re.IGNORECASE | re.VERBOSE)
 
     def match_intent(self, intent):
         """Tell if ``intent`` is an ``ACTION``.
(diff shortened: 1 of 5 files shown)


sopel/plugins/rules.py
```diff
@@ -1196,15 +1241,15 @@ def get_rule_regex(self):
...
         """
         name = [self.escape_name(self._name)]
         aliases = [self.escape_name(alias) for alias in self._aliases]
         pattern = r'|'.join(name + aliases)
-        return tools.get_nickname_command_regexp(
-            self._nick, pattern, self._nick_aliases)
+        return tools.compile_rule(
+            self._nick,
+            self.PATTERN_TEMPLATE.format(command=pattern),
+            self._nick_aliases)
```
dgw, line 1252: I must have forgotten to mention it in that issue you were working from, but the new rule system is a perfect excuse to deprecate `tools.compile_rule()` too, and move its behavior into the `Rule` type.

dgw (changes requested):
I have to ask: Are all of these changes _really_ related to the PR/commit title, `plugins: command-regexp functions are now private`? This is a rare occasion when I really think you're doing too much in one commit, unless all of the refactoring and such is truly necessary to do all at once. I'd say at least some of it probably isn't, and should be a separate commit or PR.

Two things about the overall approach here, though:

1. The existing functions in `tools` should be marked deprecated until removal in 8.0, instead of yanked out in a point release with no warning. But nothing says we have to keep their implementation; what would you say to building the implementation inside `Rule` and its subclasses such that those `tools` functions would just call a classmethod or act as an accessor for the appropriate class constant?
2. Can we put `tools.compile_rule` into `plugins.rules` too? Same thing about deprecation vs. outright removal in 7.1, but it has no reason to exist outside the rule system _especially_ if the rule system is its primary user.

sopel/tools/__init__.py
```diff
@@ -223,133 +223,6 @@ def compile_rule(nick, pattern, alias_nicks):
...
-        (?:\s+(\S+))?
-        .*                  # Accept anything after the parameters.
-                            # Leave it up to the plugin to parse
-                            # the line.
-        ))?                 # Group 2 must be None, if there are no
-                            # parameters.
-        $                   # EoL, so there are no partial matches.
-        """.format(command=command)
-
-
```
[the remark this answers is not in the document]
Exirel: OK I'll rework that.

Exirel:
> This is a rare occasion when I really think you're doing too much in one commit

I totally got carried away with that one, with a very bad case of tunnel vision. I'll see if I can separate that properly.

sopel/plugins/rules.py
Exirel: Hm. It requires a bit more work. I shall do that in another PR :tm:

sopel/tools/__init__.py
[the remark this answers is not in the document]
Exirel: I get these functions back with a deprecation warning!

Exirel:
> I'll see if I can separate that properly.

Just did! I had to rebase everything, so I took that opportunity to fix the merge conflict as well.

sopel/test_tools.py
```diff
@@ -263,6 +260,7 @@ def insert_into_module(func, module_name, base_name, prefix):
     setattr(module, func.__name__, func)
 
 
+@tools.deprecated('py.test run tests from @example directly', '7.1', '8.0')
```
dgw, line 263: I think this is what you meant (but it's already been the case for a _long_ time, so we probably should've deprecated this sooner 😅)

```suggestion
@tools.deprecated('pytest now runs @plugin.example tests directly', '7.1', '8.0')
```

I want to mention `@module.example` too somehow, but can't think of a way that doesn't make the deprecation message extra long and cluttered. :/

sopel/plugins/rules.py
```diff
@@ -56,6 +57,11 @@
     'URLCallback',
 ]
 
+if sys.version_info.major >= 3:
+    # TODO: remove when dropping Python 2.7
+    unicode = str
+
+
```
dgw, line 64: Why is this ported over to here? Nothing references the name `unicode` that I can see.

sopel/plugins/rules.py
dgw: I'll probably make an issue for this, because with no plans to release any 7.2 branch, our last chance to deprecate anything else we want to remove in 8.0 (like `tools.compile_rule`) is 7.1.

test/test_tools.py
```diff
@@ -5,8 +5,6 @@
 from datetime import timedelta
 import re
 
-import pytest
```
dgw, line 8: Won't we get a drop in coverage by removing these tests while the tested functions still exist? Coveralls reported -0.1% for the current patch. I think deleting these tests is premature.

I also think it's important to make sure the equivalent behavior is tested in `plugins.rules` and its various classes. Obviously `<RuleTypeClass>.get_rule_regex()` must be called in the course of setting up the tests in `test/plugins/test_plugins_rules.py`, so its behavior in each class is validated—but there's e.g. minimal validation of command rules' match groups (AFAICT, only groups 0 & 1).

sopel/tools/__init__.py
```diff
@@ -223,6 +223,7 @@ def compile_rule(nick, pattern, alias_nicks):
     return re.compile(pattern, flags)
 
 
+@deprecated('regexp functions are now private', '7.1', '8.0')
```
dgw, line 226: I'm gonna wordsmith a bit in this file. Thought about just making the suggestion once, but it's probably easier on you if I suggest it everywhere so you can just batch-apply the change if you like it. 😁

```suggestion
@deprecated('command regexp tools are now private', '7.1', '8.0')
```

sopel/tools/__init__.py
```diff
@@ -240,6 +241,7 @@ def get_command_regexp(prefix, command):
     return re.compile(pattern, re.IGNORECASE | re.VERBOSE)
 
 
+@deprecated('regexp functions are now private', '7.1', '8.0')
```
dgw, line 244:
```suggestion
@deprecated('command regexp tools are now private', '7.1', '8.0')
```

sopel/tools/__init__.py
```diff
@@ -269,6 +271,7 @@ def get_command_pattern(prefix, command):
         """.format(prefix=prefix, command=command)
 
 
+@deprecated('regexp functions are now private', '7.1', '8.0')
```
dgw, line 274:
```suggestion
@deprecated('command regexp tools are now private', '7.1', '8.0')
```

sopel/tools/__init__.py
```diff
@@ -287,6 +290,7 @@ def get_nickname_command_regexp(nick, command, alias_nicks):
     return compile_rule(nick, get_nickname_command_pattern(command), alias_nicks)
 
 
+@deprecated('regexp functions are now private', '7.1', '8.0')
```
dgw, line 293:
```suggestion
@deprecated('command regexp tools are now private', '7.1', '8.0')
```

sopel/tools/__init__.py
```diff
@@ -311,6 +315,7 @@ def get_nickname_command_pattern(command):
         """.format(command=command)
 
 
+@deprecated('regexp functions are now private', '7.1', '8.0')
```
dgw, line 318:
```suggestion
@deprecated('command regexp tools are now private', '7.1', '8.0')
```

sopel/tools/__init__.py
```diff
@@ -322,6 +327,7 @@ def get_action_command_regexp(command):
     return re.compile(pattern, re.IGNORECASE | re.VERBOSE)
 
 
+@deprecated('regexp functions are now private', '7.1', '8.0')
```
dgw, line 330:
```suggestion
@deprecated('command regexp tools are now private', '7.1', '8.0')
```

dgw (commented):
Overall not a whole lot to say; mostly repeated suggestions for wording that's used in a half-dozen spots.

I complained about removing tests for the deprecated `tools` functions, but there was no place to leave a line note about making them just call the corresponding `plugins.rules` implementations until removal (what I suggested before).

sopel/plugins/rules.py
Exirel: Oh weird. I think an initial version used that. I shall remove that!

test/test_tools.py
Exirel: I consider this:

* deprecated functions **must not** be modified once they are deprecated
* these functions have been tested before deprecation without issues

So I was a bit quick to delete these tests, I admit.

However, about the actual tests: https://github.com/sopel-irc/sopel/blob/b16cfcd8e1197e4e85727f00f56c69f01c890aac/test/plugins/test_plugins_rules.py#L1421-L1501

So it partially cover this, because only the commands are, not the nickname/action commands.

dgw: Honestly, though, these "deprecated functions" aren't actually deprecated until this PR gets merged, so you are still allowed to change them. 😝 But my main concern is behavior drift, making sure the deprecated functions _and_ the new stuff in `plugins.rules` are doing the same things. Plus, uncovered lines make me sad. (_You_ did this to me!)

It's possible that I'm being overly paranoid, but heh. Your job is being a test-writing wizard, and my job is fretting about backward compatibility to ridiculous extremes. 😁

Hmm, there _are_ a few places where match groups other than 0 & 1 get checked for action & nick commands, but they're asserted to be `None`. We don't have _positive_ tests for groups in those command types to check that they get set correctly, only tests that verify they _aren't_ set when not expected.

The `rules` test stuff is enough to be a "separate PR" kind of thing now, so I'll leave it alone. But I'll reserve the right to take a stab at turning the deprecated functions into wrappers around the new stuff.

dgw: Update: Pretty simple to turn the pattern functions into wrappers, actually. I should do the same for the regexp ones.

dgw:
@Exirel Pushed a couple commits to address my own feedback about test coverage and stuff. Instead of -0.1% coverage, the last PR build reported only -0.004%. 🏆

When it's time to straighten out the PR history, I suggest:
* "fixup" the appropriate commit with the partial revert (currently 3176636)
* either leave e4cd829 standalone or merge it with 1f4df30 (along with the other amendment, 452818c)

Exirel:
I'm really not fan of your solution here: when something is deprecated, it's for a reason. This should not be modified, and there is only one way to ensure that: don't change the code. Now, if we want to change the code in, say, the regex we use for good, we'll have to take care of *deprecated functions we don't want to change*.

dgw:
These _deprecated functions we don't want to change_ are meant to return the regex or pattern for the type of command they specify. The regex isn't going to change between now and 8.0, at which point these will be gone.

But let's say we _do_ want to change something. We'd have made that change in _these_ functions before; now we make it in `plugins.rules` instead. I don't see a difference. If someone's code is going to break because the pattern/regex returned by this tool function changed, it's going to break regardless of whether said pattern/regex came from `tools/__init__.py` or `plugins/rules.py`. 🤷‍♂️

Exirel:
@dgw rebased, keeping your changes. Should be good now.

sopel/plugins/rules.py
```diff
@@ -1267,14 +1322,13 @@ def get_rule_regex(self):
...
-        :func:`sopel.tools.get_action_command_regexp`, which creates a compiled
-        regex to return.
+        to create a compiled regex to return.
         """
         name = [self.escape_name(self._name)]
         aliases = [self.escape_name(alias) for alias in self._aliases]
         pattern = r'|'.join(name + aliases)
-        return tools.get_action_command_regexp(pattern)
+        pattern = self.PATTERN_TEMPLATE.format(command=pattern)
+        return re.compile(pattern, re.IGNORECASE | re.VERBOSE)
```
dgw, line 1331: Reusing the name `pattern` like this is a little bit weird, but I won't block on it.

dgw (approved):
Extra tests for the regex args, too, it's magnificent!

sopel/plugins/rules.py
Exirel: I was tempted to rename that as ` base_pattern` or ` command_pattern`, but heh, I'm fine with just `pattern`.

Exirel:
Rebase because I want to work on the next step asap.

PR #1948, opened 2020-10-01
tools: make `time` and `web` always accessible by importing `tools`

dgw:
### Description
It's really silly that anyone who wants to use the `time` submodule has to do `import sopel.tools.time` even if they've already done `from sopel import tools`. This makes that unnecessary.

The intention here is to make `from sopel import tools` the only import needed to use the `tools`, `tools.events`, `tools.time`, and `tools.web` modules, without any extra nonsense.

### Checklist


Exirel (approved):
👍 smart

PR #1949, opened 2020-10-01
tld: add .tldcache command

dgw:
### Description
Allows viewing last update time for both caches, forcing the cached data to update, and clearing the cached data from both bot memory and the bot's database.

Perhaps a *bit* messier than it could be, but I'm tired of resetting my test bot's database values for tonight.

Not mentioned in commit log: This change means anyone running from source will have to manually delete the 'tld' plugin's stored plugin values from the database, as the timestamp format has changed.

### Checklist

### DIFF
--- a/sopel/modules/tld.py
+++ b/sopel/modules/tld.py
@@ -15,6 +15,7 @@ import logging
 import re
 import sys
 
+import pytz
 import requests
 
 from sopel import formatting, plugin, tools
@@ -29,7 +30,8 @@ else:
 LOGGER = logging.getLogger(__name__)
 
 
-DATE_FORMAT = '%Y-%m-%d %H:%M:%S'
+DATE_FORMAT = '%Y-%m-%d %H:%M:%S %z'
+DEFAULT_CACHE_TIME = '2000-01-01 00:00:00 +0000'
 IANA_LIST_URI = 'https://data.iana.org/TLD/tlds-alpha-by-domain.txt'
 WIKI_PAGE_URI = 'https://en.wikipedia.org/wiki/List_of_Internet_top-level_domains'
 r_tld = re.compile(r'^\.(\S+)')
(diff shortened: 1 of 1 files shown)


dgw:
Yes, #1948 being accepted _would_ mean that I should change the imports in here.

dgw:
That import thing that both LGTM and Travis flagged is exactly why I also opened #1948. Ignoring until that's either merged or dismissed with cause.

Exirel (approved):

dgw:
I probably won't see until tonight, but that force-push _should_ pass CI now that #1948 got merged.

dgw:
Looks like I already have something to fix when I get home: py2.7 doesn't have `datetime.timezone`. Fun.

sopel/modules/tld.py
```diff
@@ -9,7 +9,7 @@
 """
 from __future__ import absolute_import, division, print_function, unicode_literals
 
-from datetime import datetime
+from datetime import datetime, timezone
```
Exirel, line 12:
```suggestion
from datetime import datetime
```

Exirel: HA I can't do the other suggestion, which is to do that:

```
import pytz
```

Right before the `import requests` down below.

sopel/modules/tld.py
```diff
@@ -204,13 +205,13 @@ def _update_tld_data(bot, which):
         LOGGER.error("Asked to update unknown cache type '%s'.", which)
         return
 
-    now = datetime.now()
+    now = datetime.now(tz=timezone.utc)
```
Exirel, line 208:
```suggestion
    now = datetime.now(tz=pytz.utc)
```

Exirel (commented):
Apart from that missing `import pytz` that should do it.

sopel/modules/tld.py
dgw: Gotta love GitHub's artificial restrictions on PR reviews. 🙄

dgw:
@Exirel Believe it or not, I had already come up with the exact same changes on my own by the time your suggestions hit my email inbox. (And even if not, I'm also too lazy to apply suggestions and rebase when I could just `git commit --amend --no-edit` 😏)

Exirel (approved):

PR #1950, opened 2020-10-01
isup: more specific errors

dgw:
### Description
Follow-up to #1940, wherein @Exirel requested more specific error messages than just "seems down".

### Checklist

### DIFF
--- a/sopel/modules/isup.py
+++ b/sopel/modules/isup.py
@@ -62,7 +62,6 @@ def handle_isup(bot, trigger, secure=True):
         site = get_site_url(trigger.group(2))
         response = requests.head(site, verify=secure, timeout=(10.0, 5.0))
         response.raise_for_status()
-        response = response.headers
     except ValueError as error:
         bot.reply(str(error))
     except requests.exceptions.SSLError:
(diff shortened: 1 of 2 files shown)


sopel/modules/isup.py
```diff
@@ -60,22 +60,36 @@ def handle_isup(bot, trigger, secure=True):
...
+    except requests.ConnectionError:
+        bot.say(
+            '{} looks down to me (connection error).'
+            .format(site))
     else:
         if response:
             bot.say(site + ' looks fine to me.')
-        else:
-            bot.say(site + ' is down from here.')
+        else:  # TODO: Is it even possible to get here any more?
```
Exirel, line 91: Nope you can remove that: either there is an exception, or it works, nothing in-between.

Exirel (approved):
One small nitpick you can fix, but it's good to go.

sopel/modules/isup.py
```diff
@@ -60,22 +60,36 @@ def handle_isup(bot, trigger, secure=True):
     """
     try:
         site = get_site_url(trigger.group(2))
```
Exirel, line 62: If you can check that argument, that would do it for #1001 !

Exirel (changes requested):
Actually, a small nitpick to fix!

sopel/modules/isup.py
dgw: Will refactor it out.

sopel/modules/isup.py
dgw: Doesn't this snippet of `get_site_url()` already do that?

https://github.com/sopel-irc/sopel/blob/50ace55f2b6309b59792aeda332e12ed79472a16/sopel/modules/isup.py#L32-L34

I guess it could be checked _in_ `handle_isup()` instead, but that validation function is already covered by unit tests (including empty `site` arg, though `None` wasn't tested; I just added that).

Exirel: Great!

Exirel (approved):
👍 🎉

PR #1961, opened 2020-10-14
admin: add .chanlist command to show what channels Sopel is in

dgw:
### Description
Tin.

This was requested on our IRC channel. Instead of just telling the user how to write their own command to do it, I realized that such a feature probably *does* belong in the admin plugin that ships with Sopel.

### Checklist
  - Note: I'm relying on tests and other code I've patched recently to prove that `max_messages` works as intended

### DIFF
--- a/sopel/modules/admin.py
+++ b/sopel/modules/admin.py
@@ -154,6 +154,19 @@ def temporary_part(bot, trigger):
 
 
 @plugin.require_privmsg
+@plugin.require_admin
+@plugin.command('chanlist', 'channels')
+@plugin.priority('low')
+def channel_list(bot, trigger):
+    """Show channels Sopel is in."""
+    channels = ', '.join(sorted(bot.channels.keys()))
+
+    # conservative assumption about how much room we have in the line to make
+    # sure `max_messages` won't actually truncate anything
+    bot.say(channels, max_messages=1 + len(channels) // 400)
+
+
+@plugin.require_privmsg
 @plugin.require_owner
 @plugin.command('restart')
 @plugin.priority('low')


sopel/modules/admin.py
```diff
@@ -153,6 +153,21 @@ def temporary_part(bot, trigger):
...
 
+@plugin.require_privmsg
+@plugin.require_admin
+@plugin.command('chanlist', 'channels')
+@plugin.priority('low')
+def channel_list(bot, trigger):
+    """Show channels Sopel is in."""
+    channels = list(bot.channels.keys())
+    channels.sort()
+    channels = ', '.join(channels)
```
Exirel, line 164: Same result, less steps:

```suggestion
    channels = ', '.join(sorted(bot.channels.keys()))
```

As a side note: if you remember, I wasn't bothered by re-assigning twice a value to a variable (it was `pattern`). However, there is something I didn't mention back then: it doesn't bother me when the variable stay of the same type. This is not the case here: first it's a list, then it's a string. That isn't a very good practice, and I tend to avoid that.

But heh it doesn't matter here, as you can do everything in one go.

Exirel (changes requested):
Oh yeah good call with the PR. I saw the feature request on IRC and it's a good idea. Just a nitpick to have a slightly more elegant code.

sopel/modules/admin.py
dgw: I'm really amused right now, because I started off with a one-liner and then split it into steps for clarity before making the PR. 😹

Exirel (approved):
I like the `max_messages` bit.

Issue #1886, opened 2020-06-06
STATUSMSG not handled in `Identifier.is_nick()`


RhinosF1:
### Description
Freenode adds a @ to the channel name in a +mz channel or +z channel where the user is quieted.

This means #channel becomes @#channel and the it's impossible to use trigger.sender to kick a user, for example, from the channel based on what they said.

### Reproduction steps
1. Create a +mz channel on Freenode and set log_raw to True
2. op your bot have a user who can only speak to ops say something.
3. You will see the channel is recorded as @#channel

### Expected behavior
The @ is dropped from trigger.sender (but maybe something like trigger.was_moderated is added to indicate it)

(PS: I can attempt it tonight)

### Logs
```
2020-06-05 22:47:01,739 << :test-lta!uid348912@gateway/web/irccloud.com/x-zbfyoyydyhcwvgmu PRIVMSG @#miraheze-overflow :sing and dance with me
```

### Environment
- Sopel `.version`: 7.0.1 & 7.0.4 tested
- Sopel installed via: pip (both)
- Python version: 3.5.3 (both)
- Operating system: Debian Stretch (9.7) for .1 & Debian 4.9.144-3 (2019-02-02) x86_64 for .4
- IRCd `/version`: Freenode (https://freenode.net/news/seven-118)

dgw:
Going to leave these here for whoever first decides to work on implementing this (could be me, could be someone else): https://modern.ircdocs.horse/#statusmsg-parameter & https://modern.ircdocs.horse/#privmsg-message

PR #1963, opened 2020-10-17
core: use CHANTYPES for `Identifier.is_nick()` if advertised

dgw:
### Description
Very simple: If the server's ISUPPORT info includes CHANTYPES, update the list of prefixes in `tools`. Does so through a private helper function to prevent breaking anything in case a bad value gets through.

Admittedly, this approach is a bit overengineered, and maybe also runs the update in a suboptimal location. This is why we have code review. x) **Edit:** Leave it to @Exirel, who says he'll have better ideas Soon™.

### Checklist
  - Well it's just replacing one tuple with another, so…

### DIFF
--- a/sopel/coretasks.py
+++ b/sopel/coretasks.py
@@ -24,7 +24,7 @@ import re
 import sys
 import time
 
-from sopel import loader, module, plugin
+from sopel import loader, module, plugin, tools
 from sopel.config import ConfigurationError
 from sopel.irc import isupport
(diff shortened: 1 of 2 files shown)


dgw:
Per a rather lengthy IRC discussion, this opens a whole can of worms that we'd best leave for 8.0.

dgw:
Rejecting my own patch in favor of further discussion at #1886 and/or #2115. We need an approach that works reasonably simply **and** allows for both CHANTYPES & STATUSMSG.

Issue #1738, opened 2019-11-09
Master plan to fix the "plugin"/"module" confusion


dgw:
Since long before this project was called "Sopel", the `.py` files one would drop in to add functionality have been referred to as "modules". The code uses it, the docs use it—that term is everywhere. Thing is, Python already has something called "modules", and that makes stuff confusing.

Going forward, we've more-or-less decided to start calling our stuff "plugins" instead. The rewritten machinery to manage them is in `sopel.plugins` (see #1479). We (those of us who work on the project on a regular basis) are pretty consistent about referring to them as "plugins" on IRC and in new documentation. But there are still a _lot_ of references to "modules" everywhere.

Working on this will probably last the entire life-cycle of Sopel 7. It's not trivial to make such a wide-reaching change in how stuff gets named (or documented), so the implementation plan below is intentionally slow-paced.

Here are the things that [we know] need to be done, sorted by when they need to happen:

  - **Before 7.0 is released**
    - [ ] Fix [the per-channel configuration stuff](https://github.com/sopel-irc/sopel/blob/f50e094f6b4afa2c47ce360ba13259f9c19f068f/sopel/bot.py#L509-L529) (in `bot.py`) to call them "plugins"
      - There's documentation both here and in the website repo that will need updating to go along with this, as we'll want to rename the new config options etc. before people start using them in a stable release
    - [ ] Update the _new_ CLI structure
      - We currently have `sopel configure --modules`, for example, which replaces the old `sopel --configure-modules` option, but the new CLI should call them "plugins"—especially since we also have a new `sopel-plugins` management tool to match
      - Obviously the old CLI options should still map to their new counterparts; we're in a transition period specifically to avoid yanking those out with no warning
  - **Through the 7.x life-cycle**
    - [ ] Add `~/.sopel/plugins` support, working just like `~/.sopel/modules` does in 7.0
      - Transitional period, again, because it's always nice to do that if possible
    - [ ] Audit documentation and code comments for references to "modules" that actually refer to "plugins" and update them
      - This probably should be split into a list of sub-tasks, but I'll leave that for another time
    - [ ] Initiate rename of `sopel.module` (all the decorator magic) to `sopel.plugin`
      - Transitional period might last all the way until 9.0, not sure yet
  - **For 8.0**
    - [ ] Remove support for loading plugins from `~/.sopel/modules` by default
      - Obviously it can still be added back with the `core.extra` setting, but renaming the folder is even easier
[...]

PR #1964, opened 2020-10-18
Audit docs (mostly) for incorrect use of the term "module"

dgw:
### Description
This is the final step for 7.1 in #1738. Making these commits was literally just me going through about 1,400 find-in-files results for the string "module". Fortunately I was able to skip a lot with negative regex lookarounds (e.g. decorators, object attributes).

I mixed in a few little tweaks to other stuff, because without them it would have been a _very_ small PR indeed.

In summary:

* coretasks: audit for outdated uses of "module" vs. "plugin" (6fa4bb1)
* irc: audit for outdated use of "module" vs. "plugin" (650e672)
  * Some of this actually changes an argument name, something we might not want to do right now (and in fact the "module" moniker goes all the way down into the `CapReq` type from `irc.utils`).
  * More importantly, it seems that nothing actually *uses* the `cap_req()` function provided here. Something to address in a future patch, perhaps.
* irc.utils: fix docs for `CapReq` param `prefix` (f8ed9e5)
  * The empty string is also allowed; it means "request this if no other CapReq forbids it".
  * Here I did _not_ change the parameter name `module` or anything else, yet. I'm still not entirely sure how this and `cap_req()` fit together, and we'll probably end up dropping most or all of the above commit.
* logger: decorate `get_logger()` with deprecation info (e02a286)
  * I think this was deprecated before we added that `warning_in` parameter for `tools.deprecated()` in #1872, but now we won't have to remember to decorate this function later.
* plugins: audit docs for outdated use of "module" vs. "plugin" (d5854f8)
* plugins.handlers: small fixes (all uses of "module" are correct) (850a178)
* irc.utils: rename "module" param & attribute to "plugin" (33b094ba)
  * Kept `CapReq.module` as a `@tools.deprecated` `@property`. Looks like nothing was accessing it, but let's still keep it around until 8.0 just in case.

### Checklist
  - Note: There are minimal code changes (none if the single code-touching commit to `irc` is rejected)

### Notes
Yes, I'm aware that I threw some unrelated stuff in here, not that long after admonishing @Exirel for doing the same in #1944. I don't regret it. That stuff needs fixing, too. 😜

### DIFF
--- a/sopel/coretasks.py
+++ b/sopel/coretasks.py
@@ -1,7 +1,7 @@
 # coding=utf-8
 """Tasks that allow the bot to run, but aren't user-facing functionality
 
-This is written as a module to make it easier to extend to support more
+This is written as a plugin to make it easier to extend to support more
 responses to standard IRC codes without having to shove them all into the
 dispatch function in bot.py and making it easier to maintain.
 """
@@ -37,7 +37,7 @@ if sys.version_info.major >= 3:
 LOGGER = logging.getLogger(__name__)
 
 batched_caps = {}
-who_reqs = {}  # Keeps track of reqs coming from this module, rather than others
+who_reqs = {}  # Keeps track of reqs coming from this plugin, rather than others
 
 
 def setup(bot):
--- a/sopel/irc/__init__.py
+++ b/sopel/irc/__init__.py
@@ -334,11 +334,11 @@ class AbstractBot(object):
         logger = logging.getLogger('sopel.raw')
         logger.info('\t'.join([prefix, line.strip()]))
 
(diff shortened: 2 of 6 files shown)


Exirel (approved):
👍

dgw:
@Exirel You still OK with this, including the new commit I just added (33b094ba)? The "deprecated property" solution was super obvious but I was so distracted by other stuff when putting together the rest of this PR that I didn't think of it. 😅

Exirel:
> You still OK with this, including the new commit I just added 

Yes, all good to me!

PR #1965, opened 2020-10-19
choose: safely handle formatted options

### LINKED ISSUE #1877: choose: IRC formatting can make weird output
### Description
Input with certain formatting characters in certain locations can mess up the `choose` plugin's output:

```
<user>  .ch Sherlock, BBC's \x1DSherlock\x1D, \x1FTorchwood
<sopel> user: Your options: Sherlock, BBC's \x1DSherlock, Torchwood. My choice: Sherlock
```

Or, to recreate the IRC rendering using Markdown formatting:

> &lt;user&gt; &nbsp;.ch Sherlock, BBC's _Sherlock_, ~~Torchwood~~
> &lt;sopel&gt; user: Your options: Sherlock, BBC's _Sherlock, Torchwood. My choice: Sherlock_

Some IRC formatting characters get stripped here, but not all; Python considers _some_ control characters as "whitespace" for the purposes of `str.strip()`: https://github.com/sopel-irc/sopel/blob/d7bc97a94521fc4c7a4656fa8c0158e1833f72f4/sopel/modules/choose.py#L41
Combine that with formatting in the middle of a choice (obviously ignored by `strip()`) and we can get pretty messy output indeed.

For the example I gave above, the problem comes from Python removing the second `\x1D` (end italic) from the end of choice 2 and the `\x1F` (begin strikethrough) from the beginning of choice 3, meaning that the first `\x1D` (italic) is never canceled and gets applied through the end of the output line.

The obvious workaround is to tell users "don't use formatting for choices". That's probably not helpful.

The obvious _code_ workaround is to strip _all_ formatting from choices, forcing them all to be plain text. We most certainly can do that, but it would be better to find a solution that allows users to format choices however they want without breaking the plugin's output.

Perhaps the simplest solution is to apply a formatting reset (`\x0F`) before the comma, turning the first line here into `'\x0F, '.join(`: https://github.com/sopel-irc/sopel/blob/d7bc97a94521fc4c7a4656fa8c0158e1833f72f4/sopel/modules/choose.py#L45-L48
But it would be more elegant to only include a reset character when needed, and place it inside the quotes if the option contains a comma.

dgw:
### Description
Briefly, this resolves #1877. No longer will unterminated formatting, or leading/trailing formatting characters, cause formatting issues in the `.choose` command's output.

The custom "strip"/"cleanup" function that makes it all possible comes replete with <del>16</del> <del><ins>20</ins></del> <ins>23</ins> passing cases and 8 expect-fail cases for future enhancements—though I'm sure there are weird cases I haven't covered with the existing test params. Suggestions welcome, and/or we can just see what breaks once this hits the wild.

### Checklist

### DIFF
--- a/sopel/modules/choose.py
+++ b/sopel/modules/choose.py
@@ -12,8 +12,51 @@ https://sopel.chat
 from __future__ import absolute_import, division, print_function, unicode_literals
 
 import random
+import unicodedata
 
-from sopel import plugin
+from sopel import formatting, plugin
+
+
+def _format_safe(text):
+    """Remove excess whitespace and terminate IRC formatting.
+
+    :param str text: text to clean of whitespace
+    :rtype: str
+
+    Our own take on ``str.strip()`` that skips stripping off IRC formatting
+    and makes sure any formatting codes are closed if necessary.
+    """
+    start = end = 0
+
+    pos = 0
+    while pos < len(text):
+        is_whitespace = unicodedata.category(text[pos]) == 'Zs'
+        is_non_printing = (
+            text[pos] in formatting.CONTROL_NON_PRINTING and
+            text[pos] not in formatting.CONTROL_FORMATTING
+        )
+        if not is_whitespace and not is_non_printing:
+            start = pos
+            break
+        pos += 1
+
+    pos = len(text) - 1
+    while pos >= 0:
+        is_whitespace = unicodedata.category(text[pos]) == 'Zs'
+        is_non_printing = (
+            text[pos] in formatting.CONTROL_NON_PRINTING and
+            text[pos] not in formatting.CONTROL_FORMATTING
+        )
+        if not is_whitespace and not is_non_printing:
+            end = pos + 1
+            break
+        pos -= 1
+
+    safe = text[start:end]
+    if any(c in safe for c in formatting.CONTROL_FORMATTING):
+        safe += formatting.CONTROL_NORMAL
+
+    return safe
 
 
 @plugin.command('choose', 'choice', 'ch')
@@ -40,7 +83,7 @@ def choose(bot, trigger):
         choices = trigger.group(2).split(delim)
         if len(choices) > 1:
             break
-    choices = [choice.strip() for choice in choices]
+    choices = [_format_safe(choice) for choice in choices]
     pick = random.choice(choices)
 
     # Always use a comma in the output
--- a/sopel/modules/choose.py
+++ b/sopel/modules/choose.py
@@ -28,6 +28,7 @@ def _format_safe(text):
     """
     start = end = 0
 
+    # strip left
     pos = 0
     while pos < len(text):
         is_whitespace = unicodedata.category(text[pos]) == 'Zs'
@@ -40,6 +41,7 @@ def _format_safe(text):
             break
         pos += 1
 
+    # strip right
     pos = len(text) - 1
     while pos >= 0:
         is_whitespace = unicodedata.category(text[pos]) == 'Zs'
@@ -52,8 +54,10 @@ def _format_safe(text):
             break
         pos -= 1
 
+    # build the final string
     safe = text[start:end]
     if any(c in safe for c in formatting.CONTROL_FORMATTING):
+        # if it includes IRC formatting, append reset character just in case
         safe += formatting.CONTROL_NORMAL
 
     return safe
--- a/sopel/modules/choose.py
+++ b/sopel/modules/choose.py
@@ -16,16 +16,29 @@ import unicodedata
 
 from sopel import formatting, plugin
 
+# Remove when dropping py2 support
+try:
+    str = unicode
+except NameError:
+    pass
+
 
 def _format_safe(text):
     """Remove excess whitespace and terminate IRC formatting.
 
     :param str text: text to clean of whitespace
     :rtype: str
+    :raises TypeError: if the passed ``text`` is not a string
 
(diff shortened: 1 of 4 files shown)


dgw:
Wow, I really whiffed that. Was so focused on making sure my _new_ tests behaved that I neglected the existing ones.

dgw:
Should be all fixed, or at least the full `make test` suite passes locally. Got distracted by sopel-irc/sopel-github#77 first.

test/modules/test_modules_isup.py
```diff
@@ -1,5 +1,5 @@
 # coding=utf-8
-"""Tests for Sopel's ``remind`` plugin"""
+"""Tests for Sopel's ``isup`` plugin"""
```
Exirel, line 2: lol

sopel/modules/choose.py
```diff
@@ -12,8 +12,51 @@
...
+    while pos >= 0:
+        is_whitespace = unicodedata.category(text[pos]) == 'Zs'
+        is_non_printing = (
+            text[pos] in formatting.CONTROL_NON_PRINTING and
+            text[pos] not in formatting.CONTROL_FORMATTING
+        )
+        if not is_whitespace and not is_non_printing:
+            end = pos + 1
+            break
+        pos -= 1
```
Exirel, line 53: For some reason I don't like `while` loop to iterate over a string, mostly because 1. you need to be sure of your exit condition and 2. you can already iterate over a string like any other iterable.

So I started playing with `enumerate`, to get both the current character and its position, played a bit with `reversed` and text's length to do that from the end of the string, and _voilà_, no more `while` loop, no more `pos` to remember, which means no risk of an off by one error, or a missing exit condition.

```suggestion
    # strip left
    for i, c in enumerate(text):
        is_whitespace = unicodedata.category(c) == 'Zs'
        is_non_printing = (
            c in formatting.CONTROL_NON_PRINTING and
            c not in formatting.CONTROL_FORMATTING
        )
        if not is_whitespace and not is_non_printing:
            start = i
            break

    # strip right
    for i, c in enumerate(reversed(text)):
        is_whitespace = unicodedata.category(c) == 'Zs'
        is_non_printing = (
            c in formatting.CONTROL_NON_PRINTING and
            c not in formatting.CONTROL_FORMATTING
        )
        if not is_whitespace and not is_non_printing:
            end = len(text) - i
            break
```

I checked with `timeit` with different string, and there is not a lot of difference: sometimes it's slightly faster, sometimes it's slightly slower.

sopel/modules/choose.py
```diff
@@ -12,8 +12,51 @@
...
+def _format_safe(text):
+    """Remove excess whitespace and terminate IRC formatting.
+
+    :param str text: text to clean of whitespace
+    :rtype: str
+
+    Our own take on ``str.strip()`` that skips stripping off IRC formatting
+    and makes sure any formatting codes are closed if necessary.
+    """
+    start = end = 0
```
Exirel, line 29:
```suggestion
    if not text:
        return ''

    start = end = 0
```

This is a minor optimization, but heh, worth a shot.

Exirel (approved):
As far as I'm concerned, it works and it does the job, so good for me.

Well OK yes I suggested some changes, but they are pretty minor.

sopel/modules/choose.py
dgw: I like this idea, but maybe not the exact implementation. If `text` is a falsey non-string value, it should fail—e.g. this would return `''` for `_format_safe(None)`, a case that should raise `TypeError` or `ValueError`. Since I already have a couple more test tweaks in mind, I can do something about this at the same time.

sopel/modules/choose.py
dgw: I'm going to steal the comments from this, at least.

For longer strings the `enumerate()` approach was consistently ~10-15% slower than `while` loops in my own `timeit` experiments from before this suggestion existed—significant enough to push me toward manually keeping track of `pos`. And even ignoring execution time, wrapping a generator inside another generator in the right-stripping code using `enumerate(reversed())` feels like unnecessary overhead.

Admittedly, there _was_ an off-by-one error in this code while I was first developing it, but that's why I wrote tests. Needing to be sure of the exit condition isn't a bad thing.

dgw (commented):
In which @dgw declines to lazily apply @Exirel's suggestions. 😁

sopel/modules/choose.py
Exirel: I'm, like: yeah, OK. But really I'm like: yeah, whatever.

Also, the very sassy me would say: this wouldn't happen with a proper static type system. Which is true, but not helpful at all.

sopel/modules/choose.py
Exirel: I'm not surprised by the off-by-one error. It's a classic.

Also I don't think this function will be modified anytime soon, I just hope no one will ever want to do it because `while` loops are evil.

sopel/modules/choose.py
dgw: I think you'll like the commit message on 0c763cf then. 😏

dgw:
I'm going to stop messing with this now. 23 passing + 8 xfail test cases is enough for this one silly little function. 😛

dgw:
> I'm going to stop messing with this now.

Yeah, that didn't happen. But I'm _really_ done now. CI passed on all versions this time!

Exirel (approved):

PR #1967, opened 2020-10-20
plugin: keep old-school plugin devs happy w/plural command decorators

dgw:
### Description
Enough people expressed dismay that our upcoming migration to `sopel.plugin` (from `sopel.module`) would get rid of the plural `commands` decorator family. So we can have both, and plugin devs may choose depending on how they decorate stuff.

```py
@plugin.command('justone')
def one_name(bot, trigger):
    pass

@plugin.commands('multiple', 'names')
def two_names(bot, trigger):
    pass
```

I gave the aliases very short docstrings that just link back to the main entry, so Sphinx wouldn't duplicate the whole shebang.
Locally built preview:

image

@Exirel, is adding tests for the aliases worthwhile? I don't feel like bothering. 😝

### Checklist

### DIFF
--- a/sopel/module.py
+++ b/sopel/module.py
@@ -13,15 +13,15 @@ from __future__ import absolute_import, division, print_function, unicode_litera
 # Therefore, don't add anything to this import list. Ever.
 from sopel.plugin import (  # noqa
     ADMIN,
-    action_command as action_commands,
-    command as commands,
+    action_commands,
+    commands,
     echo,
     event,
     example,
     HALFOP,
     intent,
     interval,
-    nickname_command as nickname_commands,
+    nickname_commands,
     NOLIMIT,
     OP,
     OPER,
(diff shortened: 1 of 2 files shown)


Exirel (changes requested):
You need to update `sopel.module`'s imports as well.

dgw:
@Exirel Wait, so now you're _in favor_ of touching deprecated stuffs? 😼

Exirel:
@dgw your PR is about un-deprecating `commands`, tho.

dgw:
@Exirel `plugin.commands` wasn't deprecated; it never existed (except briefly in an unreleased state, between #1898 and #1924, for about a month). It's just, after the little row we had over touching deprecated code in #1944, I was surprised that you _wanted_ me to do so here. It makes sense. Importing with `as` is silly if not needed. Just didn't expect it.

Exirel (approved):

PR #1969, opened 2020-10-23
cli, setup: hard-code Python 2 EOL message

dgw:
### Description
Updates to Python 2.x stopped months ago. Any future release is certain to be installed and run after support stopped, therefore there's no need for wasting CPU cycles on date comparisons.

### Checklist

### DIFF
--- a/sopel/cli/run.py
+++ b/sopel/cli/run.py
@@ -30,11 +30,9 @@ if sys.version_info < (2, 7) or (
     tools.stderr('Error: Sopel requires Python 2.7+ or 3.3+.')
     sys.exit(1)
 if sys.version_info.major == 2:
-    now = time.time()
-    state = 'has reached end of life'
-    if now >= 1588291200:  # 2020-05-01 00:00:00 UTC
-        state += ' and will receive no further updates'
-    tools.stderr('Warning: Python 2.x %s. Sopel 8.0 will drop support for it.' % state)
+    tools.stderr(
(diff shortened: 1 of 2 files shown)


Exirel (approved):

sopel/cli/run.py
```diff
@@ -30,11 +30,9 @@
...
     sys.exit(1)
 if sys.version_info.major == 2:
-    now = time.time()
-    state = 'has reached end of life'
-    if now >= 1588291200:  # 2020-05-01 00:00:00 UTC
-        state += ' and will receive no further updates'
-    tools.stderr('Warning: Python 2.x %s. Sopel 8.0 will drop support for it.' % state)
+    tools.stderr(
+        'Warning: Python 2.x has reached end of life and will receive '
+        'no further updates. Sopel 8.0 will drop support for it.')
```
half-duplex, line 35: brace style nitpick
```python
        'no further updates. Sopel 8.0 will drop support for it.'
    )
```
I also prefer double quotes >.>

half-duplex (changes requested):

sopel/cli/run.py
dgw: NAK for brace style here; it's quite clear this way with only one arg. For calls with multiple args, I'd definitely do it your way.

As for double quotes… meh. I just kept what was already here, and I don't feel like putting this through another two-hour Travis queue just for that.

PR #1970, opened 2020-10-25
tld: Use Wiki API

half-duplex:
### Description
Use the Wikipedia API to fetch only [article content](https://en.wikipedia.org/wiki/Special:ApiSandbox#action=parse&format=json&page=List_of_Internet_top-level_domains&prop=text&utf8=1&formatversion=2) instead of scraping the *entire* page.
I'd like to use a better parser, but we probably don't want to add dependencies to the larger sopel repo just for this.
`make qa` currently fails in the same way it does on `master`.

### Checklist

### DIFF
--- a/sopel/modules/tld.py
+++ b/sopel/modules/tld.py
@@ -31,7 +31,7 @@ LOGGER = logging.getLogger(__name__)
 
 DATE_FORMAT = '%Y-%m-%d %H:%M:%S'
 IANA_LIST_URI = 'https://data.iana.org/TLD/tlds-alpha-by-domain.txt'
-WIKI_PAGE_URI = 'https://en.wikipedia.org/wiki/List_of_Internet_top-level_domains'
+WIKI_PAGE_NAME = 'List_of_Internet_top-level_domains'
 r_tld = re.compile(r'^\.(\S+)')
 r_idn = re.compile(r'^(xn--[A-Za-z0-9]+)')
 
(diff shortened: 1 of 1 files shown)


half-duplex:
Ok, I dunno why qa fails on my box, but it passes in CI :shrug:

dgw:
> `make qa` currently fails in the same way it does on `master`.

😕

> Ok, I dunno why qa fails on my box, but it passes in CI 🤷

Good; I was about to say `master` doesn't fail anything so WTF 😅

sopel/modules/tld.py
```diff
@@ -234,8 +234,20 @@ def _update_tld_data(bot, which):
...
+                    "action": "parse",
+                    "format": "json",
+                    "prop": "text",
+                    "utf8": 1,
+                    "formatversion": 2,
+                    "page": WIKI_PAGE_NAME,
+                },
+            ).json()
+            tld_data = tld_response["parse"]["text"]
+        except Exception:
```
dgw, line 250: Don't really like that change. Any exception that isn't from Requests is very unexpected and should fail loudly instead of just logging a warning.

I guess in addition to `RequestException`s, you could also catch ~~`JSONDecodeError`~~ (not available in py2) `ValueError` just in case decoding the JSON fails for some reason? But any other kind of error really shouldn't get swept under the `except Exception` rug. You'll find a number of `# TODO: Be specific` comments throughout Sopel's code because we're trying to eliminate this pattern wherever possible.

dgw (changes requested):
tl;dr: `except Exception` bad

sopel/modules/tld.py
half-duplex: I can't think of anything but syntax errors, requests errors, JSONDecodeError, or KeyError. What would cause ValueError?
Still, fair.

dgw: In older Pythons, `json` throws `ValueError` if decoding the JSON document fails. In newer Pythons, it throws `JSONDecodeError`—a subclass of `ValueError`. So, catching `ValueError` is valid to catch failed JSON decoding across all Python versions Sopel supports in 7.x.

sopel/modules/tld.py
```diff
@@ -234,8 +234,21 @@ def _update_tld_data(bot, which):
...
+                    "action": "parse",
+                    "format": "json",
+                    "prop": "text",
+                    "utf8": 1,
+                    "formatversion": 2,
+                    "page": WIKI_PAGE_NAME,
+                },
+            ).json()
+            tld_data = tld_response["parse"]["text"]
+        # py3.3 needs ValueError instead of more specific json.decoder.JSONDecodeError
```
dgw, line 250: It's all py<3.5, not only 3.3

half-duplex: frick

dgw (approved):
Man, I got distracted by the whole Travis CI build queue thing… This is a meaningful reduction in the amount of HTML the bot needs to transfer and parse. Why not?

Exirel (approved):

PR #1972, opened 2020-10-27
ci: test on Python 3.9

dgw:
It took a few weeks for Travis CI to add Python 3.9 stable to the available versions, but it finally arrived.

This CI configuration has been [validated](https://travis-ci.org/github/sopel-irc/sopel/builds/739190149) on a test branch.

Letting this jump the usual merge order might make sense, after which it should be possible for me to just "kick" older PRs through Travis' web UI and make them rerun with the added job. Could be annoying if they continue under-allocating build workers to the .org platform, but migrating onto .com is currently a last-resort thing due to the migration function _still_ being labeled "Beta" (some things like build logs do _not_ transfer).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,6 +19,8 @@ matrix:
       python: 3.7
     - name: "Python 3.8 on Bionic"
       python: 3.8
+    - name: "Python 3.9 on Bionic"
+      python: 3.9
 git:
   submodules: false
 branches:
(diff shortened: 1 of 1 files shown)


Exirel (approved):

dgw:
Merging. Will likely trigger manual rebuilds as necessary on older PRs before merging, just to make sure they don't break on 3.9 (unlikely, but better off safe).

PR #1973, opened 2020-10-30
config: warn if unexpected settings exist

### LINKED ISSUE #1570: config: Warn if unexpected settings exist
When loading the config file, Sopel should warn the user if it finds a setting name that isn't defined in the `StaticSection` for that section. (Obviously, this logic will have to ignore the per-channel config sections added in #1235—or handle those, too, since they also have clearly defined properties.)

Such a warning _should not_ prevent starting the bot or halt/block loading the module to whom the setting belongs; it should be purely informational.

This would very much help users avoid typos after manually editing their config files, and aid in debugging issues like the one @HumorBaby and I just spent literally _hours_ on—which turned out to be the result of a misnamed setting in the `.cfg` file Sopel was using. (The fact that the incorrect name made more sense than the correct one is a bug for another repo.)

Exirel:
### Description

Fix #1570

To implement this, I added new methods and changed the bot's `setup()` method:

* `sopel.config.Config.get_defined_sections()` retrieves all defined sections, so it's easier to know which sections of the config file are defined in order to check them for unexpected settings
* `sopel.bot.Sopel.setup()` now calls a new method called `post_setup()`, to start the scheduler and perform post-setup checks

### Checklist

### DIFF
--- a/sopel/config/__init__.py
+++ b/sopel/config/__init__.py
@@ -159,6 +159,30 @@ class Config(object):
         else:
             return os.path.dirname(self.filename)
 
+    def get_defined_sections(self):
+        """Retrieve all defined static sections of this configuration.
+
+        :return: all instances of :class:`~sopel.config.types.StaticSection`
+                 defined for this configuration file
+        :rtype: list
+
+        When a plugin defines a section (using :meth:`define_section`), it
+        associates a :class:`~sopel.config.types.StaticSection` for the section.
+        This method allows to retrieve these instances of ``StaticSection``,
+        and only these.
+
+        .. versionadded:: 7.1
+        """
+        sections = (
+            (name, getattr(self, name))
+            for name in self.parser.sections()
+        )
+        return [
+            (name, section)
+            for name, section in sections
+            if isinstance(section, types.StaticSection)
+        ]
+
     def save(self):
         """Write all changes to the config file.
 
--- a/sopel/bot.py
+++ b/sopel/bot.py
@@ -254,7 +254,7 @@ class Sopel(irc.AbstractBot):
         """
         self.setup_logging()
         self.setup_plugins()
-        self._scheduler.start()
+        self.post_setup()
 
     def setup_logging(self):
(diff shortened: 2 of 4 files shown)


sopel/config/__init__.py
```diff
@@ -159,6 +159,31 @@ def homedir(self):
         else:
             return os.path.dirname(self.filename)
 
+    def get_defined_sections(self):
+        """Retrieve all defined static section of this configuration.
```
dgw, line 163:
```suggestion
        """Retrieve all defined static sections of this configuration.
```

sopel/config/__init__.py
```diff
@@ -159,6 +159,31 @@ def homedir(self):
...
+
+        :return: a list of all instance of
+                 :class:`~sopel.config.types.StaticSection` defined for this
+                 configuration file.
+        :rtype: list
```
dgw, line 168: nemesi**S** and elimination of redundant info

```suggestion
        :return: all instances of :class:`~sopel.config.types.StaticSection`
                 defined for this configuration file
        :rtype: list
```

The only downside to this change is the widow it introduces in our current doc styles (note "file" on its own line):

image

sopel/config/__init__.py
```diff
@@ -159,6 +159,31 @@ def homedir(self):
...
+
+        When a plugin defines a section (using :meth:`define_section`), it
+        associate a :class:`~sopel.config.types.StaticSection` for the section.
```
dgw, line 171:
```suggestion
        associates a :class:`~sopel.config.types.StaticSection` for the section.
```

dgw (changes requested):
This looks pretty straightforward. 🚀 My suggested changes are all in documentation.

dgw:
@Exirel Go ahead and squash, then we'll let the final history finish checks before I approve it… at the rate Travis is running jobs, maybe tomorrow. 😝

dgw (approved):
🥳

PR #1975, opened 2020-10-31
plugin: replace intent by ctcp

Exirel:
### Description

This is part of #1683: it replaces `sopel.module.intent` by `sopel.plugin.ctcp`. This does not change how "intents" are managed internally by Sopel, which should be done later, possibly in Sopel 8 or maybe Sopel 9.

Also, because CTCP is mostly used for ``ACTION``, this is now the default: you can use `ctcp`, `ctcp()`, or `ctcp('ACTION')` to achieve the same result.

I updated the doc and tests accordingly.

### Checklist

### DIFF
--- a/sopel/module.py
+++ b/sopel/module.py
@@ -19,7 +19,7 @@ from sopel.plugin import (  # noqa
     event,
     example,
     HALFOP,
-    intent,
+    ctcp as _future_ctcp,
     interval,
     nickname_commands,
     NOLIMIT,
@@ -41,3 +41,21 @@ from sopel.plugin import (  # noqa
     url,
     VOICE,
 )
+
+
+def intent(*intent_list):
+    """Decorate a callable to trigger on intent messages.
+
+    :param str intent_list: one or more intent(s) on which to trigger (really,
+                            the only useful value is ``ACTION``)
+
+    .. versionadded:: 5.2.0
+
+    .. deprecated:: 7.1
+
+    .. important::
+
+        This will be removed in Sopel 9, as the IRCv3 intent specification is
+        long dead. You can use :func:`@ctcp <sopel.plugin.ctcp>` instead.
+    """
+    return _future_ctcp(*intent_list)
--- a/sopel/plugin.py
+++ b/sopel/plugin.py
@@ -21,11 +21,11 @@ __all__ = [
     'action_commands',
     'command',
     'commands',
+    'ctcp',
     'echo',
     'event',
     'example',
     'find',
-    'intent',
     'interval',
     'label',
     'nickname_command',
@@ -539,10 +539,10 @@ def action_command(*command_list):
         command names!
 
         If you need to use a regex pattern, please use the :func:`rule`
-        decorator instead, with the :func:`intent` decorator::
+        decorator instead, with the :func:`ctcp` decorator::
 
             @rule(r'hello!?')
-            @intent('ACTION')
+            @ctcp('ACTION')
                 # Would trigger on "/me hello!" and "/me hello"
 
     """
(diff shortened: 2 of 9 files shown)


sopel/plugin.py
```diff
@@ -615,25 +615,42 @@ def add_attribute(function):
...
 
-def intent(*intent_list):
-    """Decorate a callable to trigger on a message with any of the given intents.
+def ctcp(function=None, *command_list):
+    """Decorate a callable to trigger on CTCP commands (mostly, ``ACTION``).
 
-    :param str intent_list: one or more intent(s) on which to trigger (really,
-                            the only useful value is ``action``)
+    :param str ctcp_command: one or more CTCP command(s) on which to trigger
+                             (really, the only useful value is ``ACTION``)
```
dgw, line 622: This is not true of CTCPs.

```suggestion
```

If you want to hint at what plugin devs might use the decorator for, I suggest:

```suggestion
                             (the most commonly useful value is ``ACTION``)
```

sopel/plugin.py
```diff
@@ -615,25 +615,42 @@ def add_attribute(function):
...
 
 
-def intent(*intent_list):
-    """Decorate a callable to trigger on a message with any of the given intents.
+def ctcp(function=None, *command_list):
+    """Decorate a callable to trigger on CTCP commands (mostly, ``ACTION``).
 
-    :param str intent_list: one or more intent(s) on which to trigger (really,
-                            the only useful value is ``action``)
+    :param str ctcp_command: one or more CTCP command(s) on which to trigger
```
dgw, line 621: The param's name is `command_list` though? This is how we've done it for other variable-length decorators like `event` above.

```suggestion
    :param str command_list: one or more CTCP command(s) on which to trigger
```

sopel/plugin.py
```diff
@@ -615,25 +615,42 @@ def add_attribute(function):
...
+        This is now ``ctcp`` instead of ``intent``, and it can be called
+        without argument, assuming ``ACTION`` in that case.
 
     .. note::
 
-        This system will be replaced and marked deprecated in or before Sopel
-        8.0, then removed in Sopel 9.0, as the IRCv3 spec for intents is dead.
+        This used to be ``@intent``, for a long dead feature in the IRCv3 spec.
+        It is now replaced by ``@ctcp``, which can be used without arguments.
+        In that case, Sopel will assume to trigger on ``ACTION``.
```
dgw, line 633:
```suggestion
        In that case, Sopel will assume it should trigger on ``ACTION``.
```

sopel/plugin.py
```diff
@@ -615,25 +615,42 @@ def add_attribute(function):
...
-    :param str intent_list: one or more intent(s) on which to trigger (really,
-                            the only useful value is ``action``)
+    :param str ctcp_command: one or more CTCP command(s) on which to trigger
+                             (really, the only useful value is ``ACTION``)
 
-    .. versionadded:: 5.2.0
+    .. versionadded:: 7.1
+
+        This is now ``ctcp`` instead of ``intent``, and it can be called
+        without argument, assuming ``ACTION`` in that case.
```
dgw, line 627:
```suggestion
        without argument, in which case it will assume ``ACTION``.
```

dgw (changes requested):
Aside from the docstring stuff I made suggestions for, there are a couple other places that we should think about sooner, rather than later:

* `Trigger` objects parse CTCP into an `intent` tag, something that should be replaced by a `ctcp` attribute perhaps. (For 7.1, it's totally fine to make a `@property` that just returns the value of `self.tags['intent']`; what's important is providing the future-compatible API ASAP.)
* `Rule`'s interface and subclasses define a `match_intent()` method that needs replacing. Since those objects are documented explicitly as "internal" (and quite new anyway), we needn't worry about providing a deprecation warning/timeline.
* And of course, since `@module.intent` simply wraps the `@plugin.ctcp` implementation, there's no need to keep `intents` as the attribute name used for storing the values for Sopel's loader. This, too, is purely internal; I'm just mentioning all this stuff for completeness.

Wherever we land on the above, anything _not_ changed in this PR should get an explicit mention in #1683 for tracking purposes.

sopel/plugin.py
Exirel: My mistake!

dgw (approved):
Go on and squash when ready. I'll take my notes about what else still needs to change and port them into #1683.

Exirel:
Ready.

Exirel:
@dgw I figured out I could just add `trigger.ctcp` property right now, so I did just that.

sopel/trigger.py
```diff
@@ -286,6 +286,22 @@ class Trigger(unicode):
     ``PRIVMSG``. Other event types like ``NOTICE``, ``NICK``, ``TOPIC``,
     ``KICK``, etc. must be requested using :func:`.plugin.event`.
     """
+    ctcp = property(lambda self: self.tags.get('intent', None))
+    """The CTCP command (if any).
+
+    :type: str
+
+    Common CTCP commands are ``ACTION``, ``VERSION``, and ``TIME``. Other
+    commands include ``USERINFO``, ``PING``, and they are used for ``DCC``.
```
dgw, line 295:
```suggestion
    commands include ``USERINFO``, ``PING``, and various ``DCC`` operations.
```

sopel/trigger.py
```diff
@@ -286,6 +286,22 @@ class Trigger(unicode):
...
+
+    .. versionadded:: 7.1
+
+    .. note::
+
+        This attribute must be used instead of the ``intent`` tag that will be
+        removed in a future version of Sopel, as it represents a feature in
+        IRCv3 spec that is long dead.
```
dgw, line 303:
```suggestion
    .. important::

        Use this attribute instead of the ``intent`` tag in :attr:`tag <tags>`.
        Message intents never made it past the IRCv3 draft stage, and Sopel will
        drop support for them in a future release.
```

I'm suggesting approximately the fifth version of this docstring. It took a ton of self-restraint to skip mentioning "Sopel should have already removed these ages ago". 😂

sopel/module.py
```diff
@@ -41,3 +41,21 @@
...
+                            the only useful value is ``ACTION``)
+
+    .. versionadded:: 5.2.0
+
+    .. deprecated:: 7.1
+
+    .. seealso::
+
+        This will be removed in Sopel 9, as the intent feature in IRCv3 spec is
+        long dead. You can use :func:`@ctcp <sopel.plugin.ctcp>` instead.
```
dgw, line 59: This should perhaps also be `.. important::`, and adapt some of the wording changes I applied to the `Trigger.ctcp` property's docstring if you like. I've taken a stab at that below.

```suggestion
    .. important::

        This will be removed in Sopel 9, as the IRCv3 intent specification is
        long dead. You can use :func:`@ctcp <sopel.plugin.ctcp>` instead.
```

dgw (changes requested):
Based on added changes, I have some docstring suggestions (as usual), and a request that `sopel/modules/seen.py` be updated to use the new `trigger.ctcp` property.

Exirel:
> a request that `sopel/modules/seen.py` be updated to use the new `trigger.ctcp` property.

Done!

dgw (approved):
Oops, I forgot about this one. 😅 Go ahead and squash when ready!

Exirel:
Rebased to fix a merge conflict.

dgw (commented):
I missed a spot last time. :|

dgw (approved):
I'm ready when Travis is.

**Edit:** LGTM is once again the slow one, thanks to holiday lull in Travis builds.

PR #1976, opened 2020-10-31
core: fix invalid usage of r-string for SASL auth token

### LINKED ISSUE #1974: Invalid format of `AUTHENTICATE` message
## Description

Sopel uses a literal `\0` instead of the null byte in AUTHENTICATE payloads.

### Reproduction steps

1. use this config:

```
auth_username = jilles
auth_password = sesame
auth_method = sasl
```

Sopel will then send this to the server:

```
AUTHENTICATE amlsbGVzXDBqaWxsZXNcMHNlc2FtZQ==
```

### Expected behavior

It should send this:

```
AUTHENTICATE amlsbGVzAGppbGxlcwBzZXNhbWU=
```

base64-decoded, they can be represented as Python literals like this, respectively: `"jilles\\0jilles\\0sesame"` and `"jilles\0jilles\0sesame"`.

### Environment
- Sopel `.version`: master
- Sopel installed via: `setup.py install`

###

This is because of this change: https://github.com/sopel-irc/sopel/pull/1928/files#diff-a3d2e4c9269312d687f6dbab17168611fd71629ef36b099985b9ba4bd293c853R838

this is equivalent to replacing `"\0"` with `"\\0"`.

I'm guessing this was to fix some sort of lint warning, but it also changes the behavior. Instead, you should probably use a `bytes` object.

Exirel:
### Description

Fix #1974 

@dgw if this can be merged in priority, as this bug prevent from using SASL in dev (I don't use SASL so I didn't notice).

### Checklist

### DIFF
--- a/sopel/coretasks.py
+++ b/sopel/coretasks.py
@@ -835,7 +835,7 @@ def auth_proceed(bot, trigger):
     else:
         return
     sasl_username = sasl_username or bot.nick
-    sasl_token = r'\0'.join((sasl_username, sasl_username, sasl_password))
+    sasl_token = '\x00'.join((sasl_username, sasl_username, sasl_password))
     send_authenticate(bot, sasl_token)
 
 


sopel/coretasks.py
```diff
@@ -835,7 +835,7 @@ def auth_proceed(bot, trigger):
     else:
         return
     sasl_username = sasl_username or bot.nick
-    sasl_token = r'\0'.join((sasl_username, sasl_username, sasl_password))
+    sasl_token = '\0'.join((sasl_username, sasl_username, sasl_password))
```
dgw, line 838: Why not do it in a non-deprecated way?

```suggestion
    sasl_token = '\x00'.join((sasl_username, sasl_username, sasl_password))
```

dgw (changes requested):

sopel/coretasks.py
Exirel: TIL.

dgw (approved):
CI ran really fast on this… We should do all our stuff on the weekends! 🤣

PR #1981, opened 2020-11-01
adminchannel: allow non-op .topic in -t channels

half-duplex:
### Description
Allow `.topic` for all users in channels without `+t` set.
Keeps topic mask related things op-only.

Depends on #1980. Improved with #1982. Fixes #1842.

### Checklist
- [x] I have read [CONTRIBUTING.md](https://github.com/sopel-irc/sopel/blob/master/CONTRIBUTING.md)
- [x] I can and do license this contribution under the EFLv2
- [x] No issues are reported by `make qa` (runs `make quality` and `make test`)
- [x] I have tested the functionality of the things this change touches

PR #1982, opened 2020-11-01
plugin: new `require_bot_privilege` decorator

Exirel:
### Description

This wasn't requested, and probably not needed that much. However, when reading the code for #1981 I realized that we repeat a lot of code, so I figured out I could add these:

* `bot.has_channel_privilege(channel, level)` will tell you if the **bot** itself has the right privilege level,
* `plugin.require_bot_privilege` is a new decorator that uses this new method,

which is probably useful only for channel administration. Probably.

Anyway, it has tests, and it gave me some ideas for the future (in particular, how we could manage more complex privilege check). So it's good!

### Second round description

OK, after second thought, I realized I could improve the `sopel.tools.target.Channel` class, by adding privilege related methods, and use one of these methods into the bot.

However, I found out that there is no test for `Channel` (or `User`), and that's not good, so I kind of hijacked this PR to add these tests, while trying to stay on topic.

### Checklist

### DIFF
--- a/docs/source/plugin/anatomy.rst
+++ b/docs/source/plugin/anatomy.rst
@@ -80,6 +80,18 @@ better to limit who can trigger them. There are decorators for that:
   trigger the rule
 * :func:`sopel.plugin.require_owner`: only the bot's owner can trigger the rule
 
+Sometimes it's not the channel privilege level of the user who triggers a
+command that matters, but the **bot's** privilege level. For that, there are
+two options:
+
+* :func:`sopel.plugin.require_bot_privilege`: this decorator is similar to
+  the ``require_privilege`` decorator, but it checks the bot's privilege level
+  instead of the user's; works only for channel messages, not private messages;
+  and you probably want to use it with the ``require_chanmsg`` decorator.
+* :meth:`bot.has_channel_privilege() <sopel.bot.Sopel.has_channel_privilege>`
+  is a method that can be used to check the bot's privilege level in a channel,
+  which can be used in any callable.
+
 .. __: https://ircv3.net/specs/extensions/account-tag-3.2
 
 Rate limiting
--- a/sopel/bot.py
+++ b/sopel/bot.py
@@ -205,7 +205,7 @@ class Sopel(irc.AbstractBot):
 
         :param str channel: a channel the bot is in
         :param int privilege: privilege level to check
-        :raise RuntimeError: when the channel is unknown
+        :raise ValueError: when the channel is unknown
 
         This method checks the bot's privilege level in a channel, i.e. if it
         has this level or higher privileges::
@@ -216,9 +216,10 @@ class Sopel(irc.AbstractBot):
 
         The ``channel`` argument can be either a :class:`str` or a
         :class:`sopel.tools.Identifier`, as long as Sopel joined said channel.
+        If the channel is unknown, a :exc:`ValueError` will be raised.
         """
         if channel not in self.channels:
-            raise RuntimeError('Unknown channel %s' % channel)
+            raise ValueError('Unknown channel %s' % channel)
 
         return self.channels[channel].has_privilege(self.nick, privilege)
 
--- a/sopel/bot.py
+++ b/sopel/bot.py
@@ -200,6 +200,28 @@ class Sopel(irc.AbstractBot):
 
         return self.users.get(self.nick).hostmask
 
+    def has_channel_privilege(self, channel, privilege):
+        """Tell if the bot has a ``privilege`` level or above in a ``channel``.
+
+        :param str channel: a channel the bot is in
+        :param int privilege: privilege level to check
+        :raise RuntimeError: when the channel is unknown
+
+        This method checks the bot's privilege level in a channel, i.e. if it
+        has this level or higher privileges::
+
+            >>> bot.channels['#chan'].privileges[bot.nick] = plugin.OP
+            >>> bot.has_channel_privilege('#chan', plugin.VOICE)
+            True
+
+        The ``channel`` argument can be either a :class:`str` or a
+        :class:`sopel.tools.Identifier`, as long as Sopel joined said channel.
+        """
+        if channel not in self.channels:
+            raise RuntimeError('Unknown channel %s' % channel)
+
+        return self.channels[channel].privileges[self.nick] >= privilege
+
     # signal handlers
 
     def set_signal_handlers(self):
--- a/sopel/bot.py
+++ b/sopel/bot.py
@@ -220,7 +220,7 @@ class Sopel(irc.AbstractBot):
         if channel not in self.channels:
             raise RuntimeError('Unknown channel %s' % channel)
 
-        return self.channels[channel].privileges[self.nick] >= privilege
+        return self.channels[channel].has_privilege(self.nick, privilege)
 
     # signal handlers
 
--- a/sopel/modules/adminchannel.py
+++ b/sopel/modules/adminchannel.py
@@ -28,16 +28,14 @@ def default_mask(trigger):
 
 
 @plugin.require_chanmsg
-@plugin.require_privilege(plugin.OP, ERROR_MESSAGE_NO_PRIV)
+@plugin.require_privilege(plugin.OP, ERROR_MESSAGE_NO_PRIV, reply=True)
+@plugin.require_bot_privilege(plugin.OP, ERROR_MESSAGE_NOT_OP, reply=True)
 @plugin.command('op')
 def op(bot, trigger):
     """
     Command to op users in a room. If no nick is given,
     Sopel will op the nick who sent the command
     """
-    if bot.channels[trigger.sender].privileges[bot.nick] < plugin.OP:
-        bot.reply(ERROR_MESSAGE_NOT_OP)
-        return
     nick = trigger.group(2)
     channel = trigger.sender
     if not nick:
@@ -46,16 +44,14 @@ def op(bot, trigger):
 
 
 @plugin.require_chanmsg
-@plugin.require_privilege(plugin.OP, ERROR_MESSAGE_NO_PRIV)
+@plugin.require_privilege(plugin.OP, ERROR_MESSAGE_NO_PRIV, reply=True)
+@plugin.require_bot_privilege(plugin.OP, ERROR_MESSAGE_NOT_OP, reply=True)
 @plugin.command('deop')
 def deop(bot, trigger):
     """
     Command to deop users in a room. If no nick is given,
     Sopel will deop the nick who sent the command
     """
-    if bot.channels[trigger.sender].privileges[bot.nick] < plugin.OP:
-        bot.reply(ERROR_MESSAGE_NOT_OP)
-        return
     nick = trigger.group(2)
     channel = trigger.sender
     if not nick:
@@ -64,16 +60,14 @@ def deop(bot, trigger):
 
 
 @plugin.require_chanmsg
-@plugin.require_privilege(plugin.OP, ERROR_MESSAGE_NO_PRIV)
+@plugin.require_privilege(plugin.OP, ERROR_MESSAGE_NO_PRIV, reply=True)
+@plugin.require_bot_privilege(plugin.HALFOP, ERROR_MESSAGE_NOT_OP, reply=True)
 @plugin.command('voice')
 def voice(bot, trigger):
     """
     Command to voice users in a room. If no nick is given,
     Sopel will voice the nick who sent the command
     """
-    if bot.channels[trigger.sender].privileges[bot.nick] < plugin.HALFOP:
-        bot.reply(ERROR_MESSAGE_NOT_OP)
-        return
     nick = trigger.group(2)
     channel = trigger.sender
     if not nick:
@@ -82,16 +76,14 @@ def voice(bot, trigger):
 
 
 @plugin.require_chanmsg
-@plugin.require_privilege(plugin.OP, ERROR_MESSAGE_NO_PRIV)
+@plugin.require_privilege(plugin.OP, ERROR_MESSAGE_NO_PRIV, reply=True)
+@plugin.require_bot_privilege(plugin.HALFOP, ERROR_MESSAGE_NOT_OP, reply=True)
 @plugin.command('devoice')
 def devoice(bot, trigger):
     """
     Command to devoice users in a room. If no nick is given,
     Sopel will devoice the nick who sent the command
     """
-    if bot.channels[trigger.sender].privileges[bot.nick] < plugin.HALFOP:
-        bot.reply(ERROR_MESSAGE_NOT_OP)
-        return
     nick = trigger.group(2)
     channel = trigger.sender
     if not nick:
@@ -100,14 +92,12 @@ def devoice(bot, trigger):
 
 
 @plugin.require_chanmsg
-@plugin.require_privilege(plugin.OP, ERROR_MESSAGE_NO_PRIV)
+@plugin.require_privilege(plugin.OP, ERROR_MESSAGE_NO_PRIV, reply=True)
+@plugin.require_bot_privilege(plugin.HALFOP, ERROR_MESSAGE_NOT_OP, reply=True)
 @plugin.command('kick')
 @plugin.priority('high')
 def kick(bot, trigger):
     """Kick a user from the channel."""
-    if bot.channels[trigger.sender].privileges[bot.nick] < plugin.HALFOP:
-        bot.reply(ERROR_MESSAGE_NOT_OP)
-        return
     text = trigger.group().split()
     argc = len(text)
     if argc < 2:
@@ -153,7 +143,8 @@ def configureHostMask(mask):
 
 
 @plugin.require_chanmsg
-@plugin.require_privilege(plugin.OP, ERROR_MESSAGE_NO_PRIV)
+@plugin.require_privilege(plugin.OP, ERROR_MESSAGE_NO_PRIV, reply=True)
+@plugin.require_bot_privilege(plugin.HALFOP, ERROR_MESSAGE_NOT_OP, reply=True)
 @plugin.command('ban')
 @plugin.priority('high')
 def ban(bot, trigger):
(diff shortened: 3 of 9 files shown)


docs/source/plugin/anatomy.rst
```diff
@@ -80,6 +80,17 @@ better to limit who can trigger them. There are decorators for that:
   trigger the rule
 * :func:`sopel.plugin.require_owner`: only the bot's owner can trigger the rule
 
+Sometimes, it's not **who** triggers a command that matter, but the bot's
```
half-duplex, line 83: Since this is contrasting priv check types not identities, I would say "Sometimes it's not the privileges of the user who triggers a command that matter," or switch the emphasis to "triggers". With the latter or as currently written, s/matter/matters/.

half-duplex (commented):
Seems useful. Lgtm besides nitpicks.

sopel/modules/adminchannel.py
```diff
@@ -154,17 +144,15 @@ def configureHostMask(mask):
 
 
 @plugin.require_chanmsg
-@plugin.require_privilege(plugin.OP, ERROR_MESSAGE_NO_PRIV)
+@plugin.require_privilege(plugin.OP, ERROR_MESSAGE_NO_PRIV, reply=True)
+@plugin.require_bot_privilege(plugin.HALFOP, ERROR_MESSAGE_NOT_OP, reply=True)
```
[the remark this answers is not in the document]
Exirel: I've honestly no idea.

docs/source/plugin/anatomy.rst
Exirel: Done!

Exirel:
Re-requesting reviews because I added a new commit with some tests on `sopel.tools.target.Channel` with new methods.

docs/source/plugin/anatomy.rst
```diff
@@ -80,6 +80,18 @@ better to limit who can trigger them. There are decorators for that:
   trigger the rule
 * :func:`sopel.plugin.require_owner`: only the bot's owner can trigger the rule
 
+Sometimes it's not the privileges of the user who triggers a command that
+matters, but the **bot's privilege level** in a channel. For that, there are
+two options:
+
+* :func:`sopel.plugin.require_bot_privilege`: this decorator is similar to
+  the ``require_privilege`` decorator, but it checks the bot's privilege level
+  instead of the user's; works only for channel messages, not private messages,
```
dgw, line 89:
```suggestion
  instead of the user's; works only for channel messages, not private messages;
```

docs/source/plugin/anatomy.rst
```diff
@@ -80,6 +80,18 @@ better to limit who can trigger them. There are decorators for that:
   trigger the rule
 * :func:`sopel.plugin.require_owner`: only the bot's owner can trigger the rule
 
+Sometimes it's not the privileges of the user who triggers a command that
+matters, but the **bot's privilege level** in a channel. For that, there are
+two options:
```
dgw, line 85:
```suggestion
Sometimes it's not the channel privilege level of the user who triggers a
command that matters, but the **bot's** privilege level. For that, there are
two options:
```

I'm just spitballing here, mostly to avoid the confusing "does-it-need-an-S-or-not" issue with the old phrasing. 😁

sopel/bot.py
```diff
@@ -200,6 +200,28 @@ def hostmask(self):
 
         return self.users.get(self.nick).hostmask
 
+    def has_channel_privilege(self, channel, privilege):
+        """Tell if the bot has a ``privilege`` level or above in a ``channel``.
+
+        :param str channel: a channel the bot is in
+        :param int privilege: privilege level to check
+        :raise RuntimeError: when the channel is unknown
```
dgw, line 208: I'm not sold on this exception type. Passing in a channel without knowing if the bot is in it first is a caller error. Wouldn't `ValueError` make sense?

sopel/plugin.py
```diff
@@ -896,6 +897,44 @@ def guarded(bot, trigger, *args, **kwargs):
     return actual_decorator
 
 
+def require_bot_privilege(level, message=None, reply=False):
+    """Decorate a function to require the bot a privilege level or above.
```
dgw, line 901:
```suggestion
    """Decorate a function to require a minimum channel privilege for the bot.
```

I even wrote this in my editor to make sure it would fit in 80 cols. 😛

sopel/plugin.py
```diff
@@ -896,6 +897,44 @@ def guarded(bot, trigger, *args, **kwargs):
...
+
+    :param int level: required privilege level to use this command
+    :param str message: optional message said to insufficiently privileged user
```
dgw, line 904: I wanted to correct "insufficiently privileged user" because I know exactly where that was copied-and-pasted from 😝 and I just kept going after that.

```suggestion
    :param int level: minimum channel privilege the bot needs for this function
    :param str message: optional message said if the bot's channel privilege
                        level is insufficient
```

sopel/tools/target.py
```diff
@@ -107,9 +108,193 @@ def add_user(self, user, privs=0):
...
+        return self.privileges.get(Identifier(nick), 0) >= privilege
+
+    def is_oper(self, nick):
+        """Tell if a user has the OP privilege level.
+
+        :param str nick: a user's nick in this channel
+        :rtype: bool
+
+        Unlike :meth:`has_privilege`, this method checks if the user has been
+        explicitely granted the OPER privilege level::
```
dgw, line 148:
```suggestion
        explicitly granted the OPER privilege level::
```

sopel/tools/target.py
```diff
@@ -107,9 +108,193 @@ def add_user(self, user, privs=0):
...
+
+            There are other methods to check the exact privilege level of a
+            user, such as :meth:`is_oper`, :meth:`is_owner`, :meth:`is_admin`,
+            :meth:`is_op`, :meth:`is_halfop`, and :meth:`is_voiced`.
+
+        """
+        return self.privileges.get(Identifier(nick), 0) >= privilege
+
+    def is_oper(self, nick):
+        """Tell if a user has the OP privilege level.
```
dgw, line 142:
```suggestion
        """Tell if a user has the OPER (operator) privilege level.
```

sopel/tools/target.py
```diff
@@ -107,9 +108,193 @@ def add_user(self, user, privs=0):
...
+        return self.privileges.get(Identifier(nick), 0) & plugin.OPER
+
+    def is_owner(self, nick):
+        """Tell if a user has the OWNER privilege level.
+
+        :param str nick: a user's nick in this channel
+        :rtype: bool
+
+        Unlike :meth:`has_privilege`, this method checks if the user has been
+        explicitely granted the OWNER privilege level::
```
dgw, line 174:
```suggestion
        explicitly granted the OWNER privilege level::
```

sopel/tools/target.py
```diff
@@ -107,9 +108,193 @@ def add_user(self, user, privs=0):
...
+        return self.privileges.get(Identifier(nick), 0) & plugin.OWNER
+
+    def is_admin(self, nick):
+        """Tell if a user has the ADMIN privilege level.
+
+        :param str nick: a user's nick in this channel
+        :rtype: bool
+
+        Unlike :meth:`has_privilege`, this method checks if the user has been
+        explicitely granted the ADMIN privilege level::
```
dgw, line 200:
```suggestion
        explicitly granted the ADMIN privilege level::
```

sopel/tools/target.py
```diff
@@ -107,9 +108,193 @@ def add_user(self, user, privs=0):
...
+        return self.privileges.get(Identifier(nick), 0) & plugin.ADMIN
+
+    def is_op(self, nick):
+        """Tell if a user has the OP privilege level.
+
+        :param str nick: a user's nick in this channel
+        :rtype: bool
+
+        Unlike :meth:`has_privilege`, this method checks if the user has been
+        explicitely granted the OP privilege level::
```
dgw, line 226:
```suggestion
        explicitly granted the OP privilege level::
```

sopel/tools/target.py
```diff
@@ -107,9 +108,193 @@ def add_user(self, user, privs=0):
...
+        return self.privileges.get(Identifier(nick), 0) & plugin.HALFOP
+
+    def is_voiced(self, nick):
+        """Tell if a user has the VOICE privilege level.
+
+        :param str nick: a user's nick in this channel
+        :rtype: bool
+
+        Unlike :meth:`has_privilege`, this method checks if the user has been
+        explicitely granted the VOICE privilege level::
```
dgw, line 278:
```suggestion
        explicitly granted the VOICE privilege level::
```

dgw (changes requested):
I did the editing thing again, with a bit of bonus commenting on actual code.

sopel/bot.py
Exirel: It's kind of a generic `Exception` tbh:

* `ValueError`: when the value is not in range.
* `TypeError`: when the value is not of the right type
* `RuntimeError`: generic error when something doesn't work at runtime because the dev handle all the cases
* `AssertionError`: a pre or post conditions is invalid

We could have something more accurate, but I'm not sure anyone will bother using it. I'm reluctant to use `assert` (that would raise `AssertionError`, which can be done manually) because a) `assert` can be deactivated; and b) it's not common in the bot's code to do so.

I found few places where it's done, and they confused me (I think it was in Channel and User?), because they are that rare.

sopel/tools/target.py
half-duplex: Still reads "OP" in latest patch

half-duplex:
`is_oper` is misleading - several levels deeper, what it's actually checking for is +y/+Y in the channel, which when present (and not remapped) does indicate oper, but when absent or unsupported may make is_oper() returning False confusing. `is_owner` as a Channel method is also more easily confused with `Trigger.owner`/`plugin.require_owner`, though that's not a hill I'll die on.

More generally, I'm not sold on the need for `is_*()` helpers. I can't think of any mainstream case that would necessitate ease of checking for exactly a certain privilege level. While it is documented, I can definitely see someone thinking e.g. `is_voice()` would tell them if a user could speak when they really want `has_privilege(..., plugin.VOICE)`.
My inclination would be to keep the API smaller and less confusing, and let the few people who need to check for a specific presence and absence check `Channel.privileges` themselves.

`is_*()` also further enshrines those priv modes in code when it's theoretically possible to have more/different ones. If we're going to drop any pretense of supporting others, that's fine, but we should acknowledge what we're doing.

This started as a tangential thought, but... Ideally we would track +y/+Y separately, so a bot could avoid harassing an oper joining on official business... Maybe we should track the actual user priv modes, and add only a `has_mode("mal", "+h")` type thing to allow for generic specific mode querying? Might require `WHOX` or something to work properly though.

sopel/bot.py
dgw: > * `ValueError`: when the value is not in range

This applies. The expected range of values is "a channel out of the set that the bot knows about", and passing a channel that the bot does _not_ know about is out of range.

About assertions, we… probably should remove those from the `User` and `Channel` objects. They're left over from the initial implementation of user tracking (ea7f14459d3babf1640de504903ee67cd3152878) and I can't imagine anyone getting mad about changing them to proper `raise`s with real exceptions.

dgw:
As far as `is_oper()` goes, I had the same thought but wound up distracted by docstring cleanup. As usual.

Ignoring the specific cases that could be confusing, though: I see @half-duplex's point about enshrining the privilege modes Sopel "supports". My question is then, how could we rework Sopel's channel privilege tracking and requirement enforcement to be more dynamic? If Sopel assigns bitwise values to modes from the server's `PREFIX` parameter dynamically, it would necessarily compromise plugins' ability to require certain privileges.

Example: On Network A with only RFC-defined modes, `+v` and `+o`, the maximum (bitwise) user privilege mask would be `11`; but Network B with `+v`, `+h`, `+o`, and `+a` would have up to `1111`. More importantly, if Sopel assigns bit positions to modes in the order given in `PREFIX`, `+o` from network A (`10`) wouldn't be equivalent to the `+o` from Network B (`0100`). How could plugins reasonably require "mode `+o` or higher" then?

I'm not saying that it's impossible, but it _would_ require a thoughtful rewrite of all decorators etc. dealing with privileges. That's obviously out of scope for this PR, though it should inform our decision about whether to introduce the `is_*()` family of privilege-checking functions at all.

And to _stop_ ignoring the potentially confusing cases: I say we explicitly **should not** have `is_oper()` as implemented. Out of all the de facto non-RFC modes that Sopel supports, those are among the weirdest and least widely supported. See data below—which also calls into question whether reworking Sopel's privilege tracking to dynamically read the available user privilege modes from the server's `PREFIX` is even worthwhile in this context.

<details>
<summary>IRC networks' advertised PREFIX modes</summary>
<p>I (well, my Python console) downloaded all 477 network pages currently linked from <tt>netsplit.de</tt>'s network list and analyzed the <tt>PREFIX</tt> tokens, where available.</p>
<p>Of the 317 network pages that matched my <tt>PREFIX</tt> regex, here are how many advertised each mode letter:</p>
<pre>
v: 317
o: 317
h: 272
a: 206
q: 205
Y: 47
V: 10
u: 2
Z: 2
y: 2
d: 1
f: 1
</pre>
<p>Outside of the six buckets Sopel hard-codes (v, h, o, a, q, y/Y), the most common is advertised on only 10 networks out of over 300 surveyed. Maybe we should just keep doing what we're doing.</p>
</details>

sopel/bot.py
Exirel: OK will do that!

Exirel:
> And to _stop_ ignoring the potentially confusing cases: I say we explicitly **should not** have `is_oper()` as implemented.

I honestly didn't know about OPER privilege level: it was here, in the code, as a constant, so I used it, but that's pretty much the full extent of my knowledge about it. It sound sane to remove `is_oper` in any case.

Now, Sopel has all these constants for years now, and they assume a system where each level is a different power of 2. It's already the assumption made by the code, both when registering user's privileges in channel and when checking for a user's privilege (with decorators such as `require_privilege`).

I've made, in this PR, 2 decisions:

* decorators use the lax `has_privilege`, which requires a minimum level, so there is no behavior change in any case
* `is_*` methods (apart from OPER) are shortcut for something that developers are already doing anyway, so any change that affect this assumption will break things, and I intend these methods to be a way to 1. add deprecation warning if needed and 2. change the underlying way to register privileges without changing the interface

To support these decisions, I've documented how to protect a callable for the bot's privilege level, but there is no specific documentation about how to deal with the `Channel` object itself. The autodoc part is usually overlooked and more for advanced users, and as seen on IRC, it's usually the doc people don't read anyway.

Now I've a question: do we really care about exotic mode/privilege levels? I mean, IRC is quite old, the only 2 privileges that really matters are VOICE and OP, everything else is mostly luck based, and it's not like Sopel offer some kind of enterprise level support. I think this topic suffer a bit from overthinking the problems.

sopel/bot.py
Exirel: Fixed that.

Exirel:
I modified the `RuntimeError` to `ValueError`, and rebased to be more up to date, given there wasn't more comment about the code itself and more a conversation around the privileges.

Is there a final statement about that? Should I just remove the methods around non-OP/VOICED privilege levels? Should I keep them?

sopel/tools/target.py
```diff
@@ -107,9 +108,193 @@ def add_user(self, user, privs=0):
...
+        This method checks the user's privilege level in this channel, i.e. if
+        it has this level or higher privileges::
+
+            >>> channel.add_user(some_user, plugin.OP)
+            >>> channel.has_privilege(some_user.nick, plugin.VOICE)
+            True
+
+        The ``nick`` argument can be either a :class:`str` or a
+        :class:`sopel.tools.Identifier`. If the user is not in this channel,
+        it will be considered has not having any privilege.
```
dgw, line 130:
```suggestion
        it will be considered as not having any privilege.
```

Must have missed this one before. Oops.

sopel/tools/target.py
```diff
@@ -107,9 +108,193 @@ def add_user(self, user, privs=0):
...
+
+        .. seealso::
+
+            There are other methods to check the exact privilege level of a
+            user, such as :meth:`is_oper`, :meth:`is_owner`, :meth:`is_admin`,
+            :meth:`is_op`, :meth:`is_halfop`, and :meth:`is_voiced`.
+
+        """
```
dgw, line 138: This goes with the idea below (see line note in `is_oper()`).

```suggestion
        .. important::

            Not all IRC networks support all privilege levels. If you intend
            for your plugin to run on any network, it is safest to rely only
            on the presence of standard modes: ``+v`` (voice) and ``+o`` (op).

        """
```

sopel/tools/target.py
```diff
@@ -107,9 +108,193 @@ def add_user(self, user, privs=0):
...
+
+        Note that you can always have more than one privilege level::
+
+            >>> channel.add_user(some_user, plugin.OPER | plugin.VOICE)
+            >>> channel.is_oper(some_user.nick)
+            True
+            >>> channel.is_voiced(some_user.nick)
+            True
+
+        """
```
dgw, line 164: This is an idea. If completely applied, this same note would also appear in the docs for `is_owner()`, `is_admin()`, `is_halfop()`, and all four relevant constants in `sopel.plugin` (`OPER`, `OWNER`, `ADMIN`, `HALFOP`).

```suggestion
        .. important::

            Not all IRC networks support this privilege mode. If you are
            writing a plugin for public distribution, ensure your code behaves
            sensibly if only ``+v`` (voice) and ``+o`` (op) modes exist.

        """
```

sopel/tools/target.py
```diff
@@ -107,9 +108,193 @@ def add_user(self, user, privs=0):
...
+        return self.privileges.get(Identifier(nick), 0) & plugin.OP
+
+    def is_halfop(self, nick):
+        """Tell if a user has the HALFOP privilege level.
+
+        :param str nick: a user's nick in this channel
+        :rtype: bool
+
+        Unlike :meth:`has_privilege`, this method checks if the user has been
+        explicitely granted the HALFOP privilege level::
```
dgw, line 252: I missed another typo 😩

```suggestion
        explicitly granted the HALFOP privilege level::
```

dgw (commented):
By way of answering @Exirel's question about whether to keep the OPER-related stuff around, I've gone through most of this again and annotated a few typos I missed, plus put forward an idea for warnings in the documentation so plugin developers unfamiliar with how fragmented IRC mode support can be are aware.

Exirel:
Thanks for the review, I added the `important` to the other methods and constants.

dgw (approved):
Ready for history cleanup at your leisure 🚀

PR #1985, opened 2020-11-08
tld: fix parsing stored cache timestamps that are invalid or on py2

dgw:
### Description
Instead of the plugin failing to load, it now merely assumes the cached TLD information is outdated. The cache will repopulate as normal when:

* any user calls `.tld` for the first time
* an admin calls `.tldcache update`
* the `update_caches()` job runs (first expected ~1 hour after startup)

### Checklist

### DIFF
--- a/sopel/modules/tld.py
+++ b/sopel/modules/tld.py
@@ -30,10 +30,18 @@ else:
 LOGGER = logging.getLogger(__name__)
 
 
-DATE_FORMAT = '%Y-%m-%d %H:%M:%S %z'
-DEFAULT_CACHE_TIME = '2000-01-01 00:00:00 +0000'
+def _strptime_as_utc(time_string):
+    return datetime.strptime(
+        time_string, DATE_FORMAT).replace(tzinfo=pytz.utc)
+
+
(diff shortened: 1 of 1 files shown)


sopel/modules/tld.py
```diff
@@ -32,8 +32,12 @@
 
 DATE_FORMAT = '%Y-%m-%d %H:%M:%S %z'
```
Exirel, line 33: Bad news, the `%z` is not valid for Python 2.7. I think it was parsed properly since Python 3.2 only.

Exirel (changes requested):
I don't really know how to fix that properly. I should have seen that before, sorry. :disappointed:

sopel/modules/tld.py
dgw: I checked in [the py2 docs](https://docs.python.org/2/library/datetime.html#strftime-strptime-behavior) and `%z` is listed? Not sure what you mean "not valid".

dgw:
New commit should fix the `%z` issue. Guess I'll have to squash later, or merge a commit that failed CI (🤮).

Exirel (approved):

PR #1988, opened 2020-11-12
currency: add thousands separator, dynamic precision, and optional HTTPS for Fixer

Duckle29:
### Description
The current code only adds more precision when converting bitcoin. Other crypto currencies are also valued such that they return very small numbers.
This change finds the required amount of digits to display at least 2 non-zero decimals.

It also adds thousands separators to the reply.

### Checklist

One issue was reported by make qa, but in help.py (ambiguous variable name l)

### DIFF
--- a/sopel/modules/currency.py
+++ b/sopel/modules/currency.py
@@ -17,12 +17,13 @@ import requests
 
 from sopel import plugin
 from sopel.config import types
+from sopel.tools import web
 
 
 PLUGIN_OUTPUT_PREFIX = '[currency] '
 FIAT_PROVIDERS = {
     'exchangerate.host': 'https://api.exchangerate.host/latest?base=EUR',
-    'fixer.io': 'http://data.fixer.io/api/latest?base=EUR&access_key={}',
+    'fixer.io': '//data.fixer.io/api/latest?base=EUR&access_key={}',
     'ratesapi.io': 'https://api.ratesapi.io/api/latest?base=EUR',
 }
 CRYPTO_URL = 'https://api.coingecko.com/api/v3/exchange_rates'
@@ -47,6 +48,8 @@ class CurrencySection(types.StaticSection):
     """Which data provider to use (some of which require no API key)"""
     fixer_io_key = types.ValidatedAttribute('fixer_io_key', default=None)
     """API key for Fixer.io (widest currency support)"""
+    fixer_use_ssl = types.BooleanAttribute('fixer_use_ssl', default=False)
+    """Whether to use SSL (HTTPS) for Fixer API"""
     auto_convert = types.BooleanAttribute('auto_convert', default=False)
     """Whether to convert currencies without an explicit command"""
 
@@ -58,6 +61,7 @@ def configure(config):
     | auto\\_convert | False | Whether to convert currencies without an explicit command |
     | fiat\\_provider | ratesapi.io | Which data provider to use (some of which require no API key) |
     | fixer\\_io\\_key | 0123456789abcdef0123456789abcdef | API key for Fixer.io (widest currency support) |
+    | fixer\\_use\\_ssl | False | Whether to use SSL (HTTPS) for Fixer API (requires paid API access) |
     """
     config.define_section('currency', CurrencySection, validate=False)
     config.currency.configure_setting(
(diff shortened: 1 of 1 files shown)


sopel/modules/currency.py
```diff
@@ -79,10 +79,17 @@ def build_reply(amount, base, target, out_string):
     rate = float(rate_raw)
     result = float(rate * amount)
 
-    if target == 'BTC':
-        return out_string + ' {:.5f} {},'.format(result, target)
+    digits = 0
+    while 1 / 10**digits > result:
+        digits += 1
+        if digits >= 8:
+            break
```
Exirel, line 86: Since it's a while loop, the exit condition can be directly inside the while expression:

```suggestion
    # up to 10 (8+2) digits precision when result is less than 1
    # as smaller results need more precision
    while digits < 8 and 1 / 10**digits > result:
        digits += 1
```

And a bit of inline comment to help future readers to understand what is going on here.

Exirel (changes requested):
The code is a bit confusing at first, so it needs at least a few comments to make it easier for future reader to understand what is going on.

Furthermore, our commit convention is to use this format:

```
plugin-name: reason
```

Not `[Plugin Name] reason`.

sopel/modules/currency.py
```diff
@@ -79,10 +79,17 @@ def build_reply(amount, base, target, out_string):
     rate = float(rate_raw)
     result = float(rate * amount)
 
-    if target == 'BTC':
-        return out_string + ' {:.5f} {},'.format(result, target)
+    digits = 0
+    # up to 10 (8+2) digits precision when result is less than 1
+    # as smaller results need more precision
+    while digits < 8 and 1 / 10**digits > result:
+        digits += 1
```
dgw, line 86: What if instead of `while`, something like:

```suggestion
    # smaller results (usually cryptocurrencies) need more precision
    whole, decimal = repr(result).split('.')
    digits = min(1, len(whole))
    digits += len(decimal) - len(decimal.lstrip('0'))
```

Not sure how it would handle roundoff errors, though. Some (e.g. the classic 0.30000000000000004) would be fine, but surely it would give weird results if Python is capable of e.g. 0.1 turning into 0.0999.

Just a thought. @Exirel mentioned the `while` loop isn't his favorite, so I did my best to come up with _an idea_ even if we end up not using it.

Duckle29: While I agree the un-commented while loop can be confusing, I think with a bit of an explanation of the intuitive way it works could help that. Not sure if I like that there's more comments than code tho :P 
```suggestion
    
    # up to 10 (8+2) digits precision when result is less than 1
    # as smaller results (usually cryptocurrencies) need more precision
    # Move the decimal point to the right by dividing by 10
    # Stop when we have enough digits to show something in result, or we've hit our limit
    digits = 0
    while digits < 8 and 1 / 10**digits > result:
        digits += 1
```
Also, it seems repr returns engineering notation on very small numbers
```
>>> repr(0.000001)
'1e-06'
```

dgw: Sure would be nice to find a constant-time algorithm, though, even if the O(n) one is limited to 8 iterations. In this case it's the principle, not the actual performance impact, that has _me_ interested.

> Also, it seems repr returns engineering notation on very small numbers

Annoying. I guess [one](https://stackoverflow.com/a/38847691/5991) of [these](https://stackoverflow.com/a/38983595/5991) would be needed to stringify the number as expected, but at that point it's probably more efficient to do it another way.

dgw:
Merging #1989 should fix the spurious CI failures. Keep an eye on [this build](https://travis-ci.org/github/sopel-irc/sopel/builds/743555648)—if it passes, feel free to squash/rebase away.

dgw:
d59d59e04807d706be637b5b79b54dcbb77de3fb is pretty ridiculous. What in the world is Fixer doing with HTTPS that it costs too much to provide for all users?!

Duckle29:
image I guess it's to make this look nice and stair-case shaped. It's really dumb. I didn't have issues with this before, but API complained today :/

Exirel (approved):
Let's hope this doesn't ~brake~ break too often...

dgw:
Brake? Even if that's a typo, I don't see what would break.

Exirel:
The Fixer API, or any other currency API. It feels like it's going downhill like the weather APIs. 😞

dgw:
Ah, Fixer is just getting the sh*t monetized out of it. There are several other Fixer-compatible sites out there; the two we merged a few days ago are just a random selection. Still thinking about that true "custom" option, to provide one last-ditch escape hatch for 7.1.x clients with a free-entry URL config value.

PR #1989, opened 2020-11-13
requirements, url: adapt to urllib3 exception type change

dgw:
### Description
urllib3 1.26 changed its `UnicodeError` to a custom exception type, breaking some of our `url` plugin's tests (plus the real-world behavior they check, of course). Now `urllib3.exceptions.LocationParseError` is raised (a subclass of `urllib3.exceptions.LocationValueError`) instead.

The solution is threefold:

1. Distrust urllib3's point releases; pin it to 1.26.x or lower now
2. Catch the custom exception type's superclass now
3. Stop catching `UnicodeError` and drop our support for urllib3<1.26 in Sopel 8 (or switch to urllib3 2.0+ only, if it's released by then)

### Checklist

### Notes
Probably will backport this to 7.0.x and cut a new release. Don't want to touch that branch any more, but I think it's going to be impossible to get a working new Sopel install on v7.0.6 because of this change.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -11,7 +11,11 @@ geoip2>=4.0,<5.0; python_version >= '3.6'
 maxminddb<2.0; python_version < '3.6'
 ipaddress<2.0; python_version < '3.3'
 requests>=2.0.0,<3.0.0
+# transitive dependency of requests
+# 2.0 will drop EOL Python 2.7 & 3.5, just like Sopel 8 plans to
+urllib3<1.27; python_version != '3.3' and python_version != '3.4'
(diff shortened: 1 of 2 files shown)


Exirel (approved):
Yeah, the lock is fine for now.

dgw:
Older Python/pip versions didn't like an unrestricted "double" dependency, so I fixed that. No other changes.

dgw:
Needed another pin for Python 3.4, because pip apparently can't tell that 1.24.x satisfies both <1.27 (from us) _and_ <1.25 (from requests), and it just installed 1.25.x anyway.

PR #1991, opened 2020-11-18
add 'h' alias to help plugin command

xnaas:


dgw:
I'm not inclined to add an extra alias to this one when it's due to be replaced by the external plugin (`sopel-help`) after the next Sopel release.

Separately, I'd rather leave 'h' for some other plugin. I don't really like the existing 'commands' alias either…

PR #1993, opened 2020-11-25
find: use slash or pipe in sed-like command

### LINKED ISSUE #1188: [find] Upgrade s/ubstitution/editing/ to accept any delimiter, like sed
Sometimes it's just tricky to use `/` as the delimiter when correcting one's own mistake, like in a URL. The `find.py` module could use regex backreferences to allow the use of any character (except `\` and newlines) as a delimiter, to make life a bit easier.

Exirel:
### Description

Attempted fix for #1188

So it was kind of messy because working with backreference to negate them is (not) surprisingly hard to do. I got [some help from StackOverflow](https://stackoverflow.com/questions/8055727/negating-a-backreference-in-regular-expressions) but I had to let go of the "you can escape the separator to use it inside the substitution/editing text ".

But heh, on the other hand, as one of the authors of the previous regex, if you want both / and | as separator AND having both in a substitution message:

> you're shit out of luck because this is the fucking regex of death as it is.

PS: it was on @dgw 's task list, but I did it anyway out of curiosity for regex backreference.

### Checklist

### DIFF
--- a/sopel/modules/find.py
+++ b/sopel/modules/find.py
@@ -50,10 +50,11 @@ def collectlines(bot, trigger):
     # Update in-memory list of the user's lines in the channel
     line_list = bot.memory['find_lines'][trigger.sender][trigger.nick]
     line = trigger.group()
-    if line.startswith("s/"):  # Don't remember substitutions
+    if line.startswith('s/') or line.startswith('s|'):
+        # Don't remember substitutions
         return
     # store messages in reverse order (most recent first)
-    elif line.startswith("\x01ACTION"):  # For /me messages
+    elif line.startswith('\x01ACTION'):  # For /me messages
         line = line[:-1]
         line_list.appendleft(line)
     else:
@@ -113,20 +114,39 @@ def kick_cleanup(bot, trigger):
 
 # Match nick, s/find/replace/flags. Flags and nick are optional, nick can be
 # followed by comma or colon, anything after the first space after the third
-# slash is ignored, you can escape slashes with backslashes, and if you want to
-# search for an actual backslash followed by an actual slash, you're shit out of
-# luck because this is the fucking regex of death as it is.
+# slash is ignored, and you can use either a slash or a pipe.
+# If you want to search for an actual slash AND a pipe in the same message,
+# you can escape your separator, in old and/or new.
 @plugin.rule(r"""(?:
-            (\S+)           # Catch a nick in group 1
-          [:,]\s+)?         # Followed by colon/comma and whitespace, if given
-          s/                # The literal s/
-          (                 # Group 2 is the thing to find
-            (?:\\/ | [^/])+ # One or more non-slashes or escaped slashes
-          )/(               # Group 3 is what to replace with
-            (?:\\/ | [^/])* # One or more non-slashes or escaped slashes
-          )
-          (?:/(\S+))?       # Optional slash, followed by group 4 (flags)
-          """)
+             (?P<nick>\S+)    # Catch a nick in group 1
+             [:,]\s+)?        # Followed by optional colon/comma and whitespace
+             s(?P<sep>/)      # The literal s and a separator / as group 2
+             (?P<old>         # Group 3 is the thing to find
+               (?:\\/|[^/])+  # One or more non-slashes or escaped slashes
+             )
+             /                # The separator again
+             (?P<new>         # Group 4 is what to replace with
+               (?:\\/|[^/])*  # One or more non-slashes or escaped slashes
+             )
+             (?:/             # Optional separator followed by group 5 (flags)
+                (?P<flags>\S+)
+             )?
+            """)
+@plugin.rule(r"""(?:
+             (?P<nick>\S+)    # Catch a nick in group 1
+             [:,]\s+)?        # Followed by optional colon/comma and whitespace
+             s(?P<sep>\|)     # The literal s and a separator | as group 2
+             (?P<old>         # Group 3 is the thing to find
+               (?:\\\||[^|])+  # One or more non-pipe or escaped pipe
+             )
+             \|               # The separator again
+             (?P<new>         # Group 4 is what to replace with
+               (?:\\\||[^|])* # One or more non-pipe or escaped pipe
+             )
+             (?:|             # Optional separator followed by group 5 (flags)
+                (?P<flags>\S+)
+             )?
+            """)
 @plugin.priority('high')
 def findandreplace(bot, trigger):
     # Don't bother in PM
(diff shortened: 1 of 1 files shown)


sopel/modules/find.py
```diff
@@ -50,10 +50,11 @@ def collectlines(bot, trigger):
     # Update in-memory list of the user's lines in the channel
     line_list = bot.memory['find_lines'][trigger.sender][trigger.nick]
     line = trigger.group()
-    if line.startswith("s/"):  # Don't remember substitutions
+    if line.startswith('s/') or line.startswith('s|'):
+        # Don't remember substitutions
         return
     # store messages in reverse order (most recent first)
-    elif line.startswith("\x01ACTION"):  # For /me messages
+    elif line.startswith('\x01ACTION'):  # For /me messages
```
dgw, line 57: Unrelated style change :|
Exi I thought you were above that 😹

sopel/modules/find.py
```diff
@@ -113,20 +114,22 @@ def kick_cleanup(bot, trigger):
...
-          )/(               # Group 3 is what to replace with
-            (?:\\/ | [^/])* # One or more non-slashes or escaped slashes
-          )
-          (?:/(\S+))?       # Optional slash, followed by group 4 (flags)
-          """)
+             (\S+)          # Catch a nick in group 1
+             [:,]\s+)?      # Followed by colon/comma and whitespace, if given
+             s([/|])        # The literal s and a separator as group 2 (/ or |)
+             (              # Group 3 is the thing to find
+               (?:(?!\2).)+ # One of more that isn't the separator from group 2
```
dgw, line 125:
```suggestion
               (?:(?!\2).)+ # One or more that isn't the separator from group 2
```

sopel/modules/find.py
```diff
@@ -113,20 +114,22 @@ def kick_cleanup(bot, trigger):
...
+             )
+             \2             # The separator, from group 2.
+             (              # Group 4 is what to replace with
+               (?:(?!\2).)* # One of more that isn't the separator from group 2
```
dgw, line 129:
```suggestion
               (?:(?!\2).)* # One or more that isn't the separator from group 2
```

sopel/modules/find.py
```diff
@@ -113,20 +114,22 @@ def kick_cleanup(bot, trigger):
...
+             )
+             (?:\2(\S+))?   # Optional slash, followed by group 5 (flags)
```
dgw, line 131:
```suggestion
             (?:\2(\S+))?   # Optional separator from group 2, followed by group 5 (flags)
```

dgw (commented):
I unlinked #1188 because that requests support for using _any_ character as delimiter; this PR adds only one alternative, and removes the escaping feature.

This might get skipped for 7.1, not sure. I'd like to experiment with lookarounds for handling escapes, and/or libraries like [sedparse](https://pypi.org/project/sedparse/) that could be used after this plugin gets extracted to its own package in the run-up to Sopel 8 (since we're conservative about dependencies primarily so no one has to install the whole universe even if they disable the whole core set).

sopel/modules/find.py
Exirel: Probably a minor change I didn't think through that much.

Exirel:
> This might get skipped for 7.1, not sure. I'd like to experiment with lookarounds for handling escapes

Works for me, either way I gave it a shot!

dgw:
I'll hold on an approving review for now, but everything is addressed. I just have to do that lookaround research/experiment(s) and/or decide if this is worth breaking the escaped-delimiter feature in a point release.

dgw:
@Exirel Were you thinking of simplifying this to the easier solution of two rules? Though I still want to investigate lookarounds.

Exirel:
@dgw updated, see if you like it better. Escaping is back!

sopel/modules/find.py
```diff
@@ -113,38 +114,58 @@ def kick_cleanup(bot, trigger):
...
+             (?:/             # Optional separator followed by group 5 (flags)
+                (?P<flags>\S+)
+             )?
+            """)
+@plugin.rule(r"""(?:
+             (?P<nick>\S+)    # Catch a nick in group 1
+             [:,]\s+)?        # Followed by optional colon/comma and whitespace
+             s(?P<sep>\|)     # The literal s and a separator | as group 2
+             (?P<old>         # Group 3 is the thing to find
+               (?:\\/|[^/])+  # One or more non-slashes or escaped slashes
```
dgw, line 140: I don't think this is right. This is the `|` version of the rule.

sopel/modules/find.py
```diff
@@ -113,38 +114,58 @@ def kick_cleanup(bot, trigger):
...
-          """)
+             (?P<nick>\S+)    # Catch a nick in group 1
+             [:,]\s+)?        # Followed by optional colon/comma and whitespace
+             s(?P<sep>/)      # The literal s and a separator / as group 2
+             (?P<old>         # Group 3 is the thing to find
+               (?:\\/|[^/])+  # One or more non-slashes or escaped slashes
+             )
+             /                # The separator again
+             (?P<new>         # Group 4 is what to replace with
+               (?:\\/|[^/])*  # One or more that isn't the separator
```
dgw, line 129:
```suggestion
               (?:\\/|[^/])*  # One or more non-slashes or escaped slashes
```

Match above.

dgw (changes requested):
I actually like this approach a lot. It just looks like there might be a little mistake in one of the rules?

Exirel:
Should be better now.

sopel/modules/find.py
```diff
@@ -113,38 +114,58 @@ def kick_cleanup(bot, trigger):
...
+             (?:/             # Optional separator followed by group 5 (flags)
+                (?P<flags>\S+)
+             )?
+            """)
+@plugin.rule(r"""(?:
+             (?P<nick>\S+)    # Catch a nick in group 1
+             [:,]\s+)?        # Followed by optional colon/comma and whitespace
+             s(?P<sep>\|)     # The literal s and a separator | as group 2
+             (?P<old>         # Group 3 is the thing to find
+               (?:\\\||[^|])+  # One or more non-slashes or escaped slashes
```
dgw, line 140: It's not a slash, though…?

```suggestion
               (?:\\\||[^|])+  # One or more non-pipe or escaped pipe
```

sopel/modules/find.py
```diff
@@ -113,38 +114,58 @@ def kick_cleanup(bot, trigger):
...
+             )
+             \|               # The separator again
+             (?P<new>         # Group 4 is what to replace with
+               (?:\\\||[^|])* # One or more that isn't the separator
```
dgw, line 144:
```suggestion
               (?:\\\||[^|])* # One or more non-pipe or escaped pipe
```

dgw (commented):
Here we can see why this was deemed "the fucking regex of death" at one point. Pattern's so long you get lost in it! 😂

dgw (approved):
Yep, this looks ready to squash out the fixups 🚀

PR #1995, opened 2020-11-30
modules: use `bot.say()`'s `trailing` parameter

dgw:
### Description
This updates every core plugin I could find with an existing handler for "too long" output (`instagram`, `reddit`, `tld`, `wiktionary`, & `wikipedia`) to use `bot.say()`'s new `trailing` parameter from #1958.

Two things that _might_ be worth amending, if others think my thought process or assumptions were wrong:

* `instagram`: I rearranged the output to put the caption (most likely to be over-length) at the end. A subsequent commit to this branch (to `wikipedia`) showed a neat hack that would allow leaving the caption where it was, in the middle of the line, but it seems to make the most sense at the end anyway.
* `reddit`: I did _not_ change submission output to use `trailing`, because:
  * The title is the very first thing in the output
  * The submission's link might also be very long, and we can't truncate those
  * If the link is long enough, it plus the rest of the metadata could be longer than allowed anyway, even without the title
  * Basically, there were too many edge cases I could see, so I didn't touch the current post/submission behavior

### Checklist

### DIFF
--- a/sopel/modules/wikipedia.py
+++ b/sopel/modules/wikipedia.py
@@ -193,7 +193,7 @@ def say_section(bot, trigger, server, query, section):
         return
 
     msg = '{} - {} | "{}"'.format(page_name, section.replace('_', ' '), snippet)
-    bot.say(msg)
+    bot.say(msg, trailing=' […]"')
 
 
 def mw_section(server, query, section):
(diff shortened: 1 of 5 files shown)


sopel/modules/wikipedia.py
```diff
@@ -193,7 +200,7 @@ def say_section(bot, trigger, server, query, section):
         return
 
     msg = '{} - {} | "{}"'.format(page_name, section.replace('_', ' '), snippet)
-    bot.say(msg)
+    bot.say(msg, trailing=' [...]"')
```
Exirel, line 203: Why not this bad boy here too?

```suggestion
    bot.say(msg, trailing=' […]"')
```

Exirel (approved):

sopel/modules/wikipedia.py
dgw: Oh, I was matching what it appended before ([`wikipedia.py` L236](https://github.com/sopel-irc/sopel/pull/1995/files#diff-c5ee1fef33221c1015f3a3313648113e1be501376eeea2354c013db3926958c2L236)), but there's no real reason to do so and I already changed the ellipsis style in this PR's `wiktionary.py` patch… I'm just silly? 😁

Exirel: That's why I already approved the PR, I knew you would fix it or have a good reason. ❤️

Issue #1496, opened 2019-03-09
NAMESX support & better privilege tracking


dgw:
As a follow-up to #1482, I think it would be beneficial for Sopel to detect NAMESX support on the server, and activate it if available.

Servers with WHOX and NAMESX available would be ideal, but we should also update privilege tracking at the same time to make sure it does not remove mask bits added by (for example) a NAMES reply if a later WHO response includes only the highest mode prefix. Right now, if only one of WHO and NAMES includes multiple mode prefixes, it opens up Sopel to race conditions that result in different stored privilege information depending on the order in which replies were received and processed.

We should probably start parsing the 005 `RPL_ISUPPORT` numeric (see #1082), which would give us access to the server flags for NAMESX, WHOX, NICKLEN (see #1305), and a bunch of other useful stuff. For example, I found that irc-framework [handles NAMESX](https://github.com/kiwiirc/irc-framework/blob/1ba628acc6589fa9d0c23405fbbbb240c9f68e55/src/commands/handlers/registration.js#L70-L74) in a pretty nice way.

`coretasks` currently decides whether WHOX is available based on a set of enabled capabilities. This is OK, and it probably works—but enabling features based on ISUPPORT flags would be an ideal addition.

Tagging as high-priority and targeting for Sopel 7 because I think getting privilege-tracking correct is a really foundational piece of how well Sopel functions.

dgw:
I think this is mitigated Enough™ by periodic WHO updates (see #1664) that we can leave it for 7.1.

dgw:
ISUPPORT parameters are now available thanks to #1758. That was the minimum requirement for starting on this, so we're set for beginning early 7.1.

Exirel:
So, I've read a bit about `NAMESX`, and it appears that it has been [deprecated for IRCv3](https://defs.ircdocs.horse/defs/isupport.html), and we should use the [`multi-prefix`](https://ircv3.net/specs/extensions/multi-prefix-3.1.html) capability instead.

For that to work, as I understand it:

1. request `multi-prefix` when available
2. make sure to store all privileges for a user instead of the higher one only
3. make sure to parse properly `RPL_NAMREPLY` event with more than one privilege

For `WHOX`, I haven't think about it yet.

dgw:
[...]

PR #1997, opened 2020-12-10
coretasks: NAMESX when multi-prefix is not available

Exirel:
### Description

As per the discussion in #1496 this makes Sopel send a `PROTOCTL NAMESX`  command when:

1. we receive a RPL_ISUPPORT message from the server
2. multi-prefix is not an available capability
3. NAMESX is now activated

### Checklist

### DIFF
--- a/sopel/coretasks.py
+++ b/sopel/coretasks.py
@@ -237,6 +237,10 @@ def startup(bot, trigger):
 @module.rule('are supported by this server')
 def handle_isupport(bot, trigger):
     """Handle ``RPL_ISUPPORT`` events."""
+    # remember if NAMESX is known to be supported, before parsing RPL_ISUPPORT
+    namesx_support = 'NAMESX' in bot.isupport
+
+    # parse ISUPPORT message from server
     parameters = {}
     for arg in trigger.args:
         try:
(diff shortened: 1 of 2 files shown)


sopel/coretasks.py
```diff
@@ -248,6 +252,14 @@ def handle_isupport(bot, trigger):
 
     bot._isupport = bot._isupport.apply(**parameters)
 
+    # does NAMESX has been updated?
```
dgw, line 255:
```suggestion
    # was NAMESX support status updated?
```

sopel/coretasks.py
```diff
@@ -248,6 +252,14 @@ def handle_isupport(bot, trigger):
...
+    if not namesx_support and 'NAMESX' in bot.isupport:
+        # yes it does!
```
dgw, line 257:
```suggestion
        # yes it was!
```

dgw (changes requested):
The actual implementation of this is so stupidly simple that I can't imagine why I didn't think of this approach when I dug into NAMESX stuff a month ago. Of _course_ it works to check if the bot already knows about NAMESX being advertised. Big brain time!

Just editing the comments like I do. One of them reminded me of a very rude meme phrase starting, "Do you are have" and ending with an insult, so I had to fix. x)

dgw (approved):
So I just thought of a different way to write the comments added to `handle_isupport()` but it's not enough of an improvement to bother.

Are you planning to work on the WHOX part of #1496? If you are, I should merge this next (like later tonight when Travis' queue is shorter) so you can build on it.

Exirel:
> Are you planning to work on the WHOX part of #1496? If you are, I should merge this next (like later tonight when Travis' queue is shorter) so you can build on it.

Not really. I don't get how WHOX is supposed to work (yet). Maybe another time!

dgw:
OK, one of us will look at WHOX in the coming weeks. I tweaked how Sopel detects WHOX availability in #1998, for a start. Need more research before going any further, but not sure when I'll get the time to do that. Probably _next_ (not this coming) weekend, at the earliest, and I have doubts about that timeline. It's gonna be a busy month. :-/

Issue #1996, opened 2020-12-03
instagram: oembed endpoint deprecated


r4f4:
### Description
When sopel tries to get info on an instagram link it fails with the message:
```
Unexpected error (Expecting value: line 1 column 1 (char 0)) from <user> at 2020-12-03 16:08:50.365477. Message was: [...]
```
That happens because, instead of a json, sopel gets the following message:

```
This endpoint is deprecated. Please read https://developers.facebook.com/docs/instagram/oembed  to learn how to migrate your endpoint
```

So not only has the endpoint changed, but it also requires an access token now.

### Reproduction steps
1. Have the instagram plugin enabled
2. Connect to server, join a channel, ...?
3. Send the bot a message containing an instagram url
4. The bot displays an error message

### Expected behavior
Sopel is able to display instagram metadata

### Logs
```
[2020-12-03 16:08:50,365] sopel.bot            ERROR    - Unexpected error (Expecting value: line 1 column 1 (char 0)) from <user> at 2020-12-03 16:08:50.365477. Message was: https://www.instagram.com/p/CIV90SGp5uL/
Traceback (most recent call last):
  File "/home/sopel/sopel/bot.py", line 643, in call_rule
    rule.execute(sopel, trigger)
  File "/home/sopel/sopel/plugins/rules.py", line 921, in execute
    exit_code = self._handler(bot, trigger)
  File "/home/sopel/sopel/plugins/rules.py", line 1482, in execute_handler
    return handler(bot, trigger, match=trigger)
  File "/home/sopel/sopel/modules/instagram.py", line 51, in instaparse
    json = get_oembed_json(instagram_url)
  File "/home/sopel/sopel/modules/instagram.py", line 67, in get_oembed_json
    return response.json()
  File "/usr/lib/python3.6/site-packages/requests/models.py", line 850, in json
    return complexjson.loads(self.text, **kwargs)
  File "/usr/lib64/python3.6/json/__init__.py", line 354, in loads
    return _default_decoder.decode(s)
  File "/usr/lib64/python3.6/json/decoder.py", line 339, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
  File "/usr/lib64/python3.6/json/decoder.py", line 357, in raw_decode
    raise JSONDecodeError("Expecting value", s, err.value) from None
```

### Environment
- Sopel `.version`: 82e3939a471e5f4d2909a8fb3cfbdad6d14388c9
- Sopel installed via: source
[...]

PR #1999, opened 2020-12-12
instagram: remove "deprecated" (actually removed) oEmbed fallback

dgw:
### Description
As reported in #1996, Facebook "deprecated" the oEmbed API endpoint. However, instead of a true deprecation where the endpoint simply starts returning a warning (or even just is marked as deprecated in docs), FB essentially *removed* the endpoint. It no longer returns valid JSON at all, let alone the expected result. Sopel receives plain text stating that the endpoint is "deprecated", with a link for more details.

The replacement API endpoint requires an API key, and this seems like a good time to consider removing the Instagram plugin to its own package. I will prepare that change separately, mostly so this fix can be easily cherry-picked into 7.0.x for maintenance, but also because it could be enough to wait for removing the plugin entirely until Sopel 8.0.

I'm not personally signed up for Instagram's developer program, so it wouldn't be fair if I alleged that FB never sent a warning before this endpoint stopped working. They might have. I'm just taking issue with the use of the term "deprecated" for something that is *nonfunctional* rather than merely unmaintained or scheduled for removal.

### Checklist
- [x] I have tested the functionality of the things this change touches (note: minimally)

### Notes
Requesting review from @kwaaak specifically, as the original plugin author.
@cottongin I'd love a review from you, too, since the fallback logic I'm gutting is from your PR #1911.

I intend to backport this tweak onto 7.0.x, for release as soon as this is merged, since the "Unexpected error" messages described in #1996 are pretty ugly.

### DIFF
--- a/sopel/modules/instagram.py
+++ b/sopel/modules/instagram.py
@@ -47,24 +47,11 @@ def instaparse(bot, trigger, match):
     try:
         bot.say(parse_insta_json(json))
     except ParseError:
-        try:
-            json = get_oembed_json(instagram_url)
-            bot.say(parse_oembed_json(json))
-        except ParseError:
-            LOGGER.exception(
-                "Unable to find Instagram's payload for URL %s", instagram_url)
-            bot.say(
-                "Unable to parse this Instagram URL. "
-                "It's probably a temporary error; try again in a bit.")
-        else:
-            LOGGER.info(
-                "Used fallback to oEmbed metadata; "
-                "Sopel's IP may be blocked by Instagram.")
-
-
-def get_oembed_json(url):
-    response = get("https://api.instagram.com/oembed/?url={}".format(url))
-    return response.json()
+        LOGGER.exception(
+            "Unable to find Instagram's payload for URL %s", instagram_url)
+        bot.say(
+            "Unable to fetch info about this Instagram URL. "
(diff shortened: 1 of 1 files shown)


kwaaak:
Does this still work for anyone? Maybe if the bot runs on an innocent residential IP?
Requests from anything resembling a server are flat-out denied.
I feel like the `"Instagram is probably blocking my IP address. :(` error generates a false hope of this ever going to work again.
Fail silently? Remove completely and get an ugly generic <title> display?

dgw:
I couldn't actually test from my home IP until today, but no dice here either. My local Sopel instance spits out the same `JSONDecodeError` as my server-hosted one.

Maybe we should just close this, merge #2000, and rebuild the plugin using the key-based API access in its own package? That was always going to be the future, but it looks like this stopgap won't help anything despite what I hoped.

kwaaak:
There is a [Python library on Github/PyPI](https://github.com/realsirjoe/instagram-scraper) which could be the nuts and bolts of a standalone plugin. No developer account needed, works with username and password.
I have not tested it myself. It might not be healthy to use your account with it, but it is a more realistic way than getting access to the proper API.

dgw:
Closing in favor of #2000 and a new standalone plugin package.

PR #2003, opened 2020-12-14
tests: deprecate test_tools; replace with tests.pytest_plugin

Exirel:
### Description

Since the last refactoring, the only purpose left for `sopel.test_tools`  was to provide utility functions for `sopel.plugin.example`. Since this feature always requires pytest, I figured that it would be better to nuke `sopel.test_tools` in Sopel 8 and to move everything still relevant to that plugin.

A second round of refactoring in Sopel 8 might be done at some point to rework how pytest is used to discover example test, but that's for another time and certainly not something to do in this PR.

### Checklist

### DIFF
--- a/sopel/test_tools.py
+++ b/sopel/test_tools.py
@@ -1,12 +1,17 @@
 # coding=utf-8
-"""This module has classes and functions that can help in writing tests.
+"""This module provided tools that helped to write tests.
 
-.. note::
+.. deprecated:: 7.1
 
-   This module formerly contained mock classes for bot, bot wrapper, and config
-   objects. Those are deprecated, and will be removed in Sopel 8.0. New code
-   should use the new :mod:`.mocks`, :mod:`.factories`, and
-   :mod:`.pytest_plugin` added in Sopel 7.0.
+    This module will be **removed in Sopel 8**.
+
+    It formerly contained mock classes for the bot, its wrapper, and its config
+    object. As the module is deprecated, so are they, and they will be removed
+    as well.
+
+    New code should use the :mod:`pytest plugin <sopel.tests.pytest_plugin>`
+    for Sopel; or should take advantage of the :mod:`~sopel.tests.mocks` and
+    :mod:`~sopel.tests.factories` modules, both added in Sopel 7.0.
 
 """
 # Copyright 2013, Ari Koivula, <ari@koivu.la>
@@ -24,7 +29,7 @@ try:
 except ImportError:
     import configparser as ConfigParser
 
-from sopel import bot, config, loader, plugins, tools, trigger
+from sopel import bot, config, tools
 
 
 __all__ = [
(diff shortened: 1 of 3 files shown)


sopel/test_tools.py
```diff
@@ -1,12 +1,11 @@
...
-.. note::
+.. deprecated:: 7.1
 
-   This module formerly contained mock classes for bot, bot wrapper, and config
-   objects. Those are deprecated, and will be removed in Sopel 8.0. New code
-   should use the new :mod:`.mocks`, :mod:`.factories`, and
-   :mod:`.pytest_plugin` added in Sopel 7.0.
+    This module is replaced by the
+    :mod:`pytest plugin <sopel.tests.pytest_plugin>` for Sopel; it will be
+    **removed in Sopel 8**.
```
dgw, line 8: This note doesn't fully cover the old mock classes, which are still here and also will be removed when this module gets deleted. Should mention the `mocks` & `factories` modules too.

sopel/tests/pytest_plugin.py
```diff
@@ -5,11 +5,163 @@
...
+            )
+    return disable_setup
+
+
+def get_example_test(tested_func, msg, results, privmsg, admin,
+                     owner, repeat, use_regexp, ignore=[]):
+    """Get a function that calls ``tested_func`` with fake wrapper and trigger.
+
+    :param callable tested_func: a Sopel callable that accepts a
+        :class:`~.bot.SopelWrapper` and a :class:`~.trigger.Trigger`
```
dgw, line 52: Oof, can we re-wrap this ugly mess? 😅

sopel/tests/pytest_plugin.py
```diff
@@ -5,11 +5,163 @@
...
+    to an existing module to be used with pytest.
+    """
+    func.__module__ = module_name
+    module = sys.modules[module_name]
+    # Make sure the func method does not overwrite anything.
+    for i in range(1000):
+        func.__name__ = str("%s_%s_%s" % (prefix, base_name, i))
+        if not hasattr(module, func.__name__):
+            break
+    setattr(module, func.__name__, func)
```
dgw, line 162: I know this bug is from the original code that you just cut-and-pasted here, but it really annoys me that there's no exception raised if some insane plugin already has 1000 conflicting objects and the code will simply overwrite the last one (`prefix_basename_999`) if it hasn't found a free slot yet.

Obviously this isn't a practical issue ~~except for plugins _you_ write 😝~~ but still.

sopel/tests/pytest_plugin.py
```diff
@@ -5,11 +5,163 @@
...
+
+    When using ``@example`` for a plugin callable with an expected output,
+    pytest will be used to run it as a test. In order to work, this fixture
+    must be added to the plugin to set up the plugin before running the test.
+    """
+    @pytest.fixture(autouse=True)
+    def disable_setup(request, monkeypatch):
+        setup = getattr(request.module, "setup", None)
+        isfixture = hasattr(setup, "_pytestfixturefunction")
+        if setup is not None and not isfixture and py.builtin.callable(setup):
```
dgw, line 37: Point of interest: `py` is in maintenance mode. Would it be too much trouble to replace its single use here with something supported, or ideally something that needs only stdlib?

I guess `pytest` itself depends on `py`, but that's not guaranteed to remain the case. If Sopel's own code is going to import `py`, it should be in Sopel's own requirements list.

dgw (commented):
I'm not exactly going in order, but this is the PR that caught my attention and then I got distracted by plumbing some of the weird shit that's left in the example-test code. See line notes. 😁

sopel/tests/pytest_plugin.py
Exirel: I suggest that we do that in another PR. I really don't want to change too much in one PR, given that here I'm just moving code around without changing any behaviour.

If you don't mind.

sopel/test_tools.py
Exirel: Yeah OK I'll update that.

sopel/tests/pytest_plugin.py
Exirel: Wait. I just realized... we are using `py` just for... `callable`??? What the...

OK nevermind, I'll remove that. It's stupid.

Exirel:
I did some force-push because really its just a few lines. Damn, I hate `py`.

sopel/test_tools.py
```diff
@@ -1,12 +1,18 @@
...
+
+    It contains mock classes for bot, bot wrapper, and config.
 
    This module formerly contained mock classes for bot, bot wrapper, and config
    objects. Those are deprecated, and will be removed in Sopel 8.0. New code
-   should use the new :mod:`.mocks`, :mod:`.factories`, and
-   :mod:`.pytest_plugin` added in Sopel 7.0.
+   should use the pytest plugin directly; or should take advantage of the
+   :mod:`.mocks` and :mod:`.factories` modules, both added in Sopel 7.0.
```
dgw, line 15: Is this as you intended it? The mixture of 3- and 4-space indents tells me perhaps not, plus the fact that it repeats itself a bit ("It contains mock classes for ..." followed by "This module formerly contained mock classes for ...").

Also, will Sphinx handle `.mocks` and `.factories` correctly from this context? They're inside `sopel.tests`.

dgw (commented):
Hooray for fixing what I mentioned before! 🥳 Though I'm not sure what happened with the module docstring for `sopel.test_tools`. I swear it wasn't like this before.

sopel/test_tools.py
Exirel: It's close to what I intended, but not exactly that. Indeed, there is a repeat. Weird. I'll see to fix that, but the general idea is to say:

1. use the pytest plugin directly
2. or use sopel.tests' mocks and factories

For the `:mod:.mocks`, I don't know how, but yes, it works! My guess is that `sopel.test_tools` is documented within the `sopel.tests` namespace, and so Sphinx uses that as its reference point. It's weird, and I'll change that because it's way too confusing.

dgw:
Go ahead and do any history cleanup you were planning. I'd guess 3f10b010e7cd55d89848306ad464d69cebc37fb1 will get fixup'd into fe5e29c0f10b23b8e8e2aeabd0e540268b8cf602?

Exirel:
Already did.

dgw:
Did you? Without 3f10b010e7cd55d89848306ad464d69cebc37fb1, that first commit introduces mixed indentation (3 & 4 spaces) into the module docstring for `test_tools`.

Exirel:
Pretty sure it's fixed... did I do something wrong? I don't see what you are talking about.

I remember the problem, but I think I did fixed and rebased properly.

dgw:
I can even see it on mobile. First commit:



Fix commit, replacing mixed indentation and redundant verbiage with a consistent and concise docstring:

Exirel:
Oh, OK. Weird.

Exirel:
OK, now it should be good!

dgw (approved):
I was putting off re-reviewing this because of the line count, but I forgot that it was only waiting for history cleanup. 😅

Issue #1765, opened 2019-11-25
Additional flake8 plugins we'd like to use


dgw:
I have a list of flake8 plugins I'd like to add after we release 7.0:

- [ ] `flake8-future-import` for easier enforcement of the required `__future__` imports
- [ ] `flake8-docstrings` to enforce blank lines (or not) around docstrings, and flag things that aren't documented but should be
- [ ] `flake8-import-order` to check import sorting (kind of redundant with the existing `isort` project in #1752, but has the benefit of being integrated with our current checks; in-progress by me, @dgw, in a personal branch named `flake8-import-order`)

There are probably more plugins that would be useful, which anyone may suggest here. I'll modify this checklist as needed.

Exirel:
I tend to completely ignore the "unused argument" warning in my code: sometimes, you need that argument in your interface, but not every implementation will need it, or even maybe most of them won't. It's OK, it's part of the life of an interface and its implementations.

I think there are better way to improve the code quality than checking that, and adding `# noqa: U100` all over the place is not really my cup of tea.

My suggestion: we are all done for 7.1 as far as I'm concerned.

dgw:
I still have a WIP branch for flake8-docstrings, and I don't intend to ship 7.1 without finishing it.

Unused args, fine, that decision can wait until after 7.1 and we check how many `# noqa`s we'd need.

half-duplex:
aaaaaaaaaaaaa
You should comment on issues with draft stuff you've worked on, cause I just spent most of the evening on it [again](https://github.com/half-duplex/sopel/tree/flake8-docstrings).
Though yours looks like a challenging rebase anyway by now

dgw:
I _did_ self-assign this issue… But if you're considerably further along than mine (and I'm sure you are), don't worry about it.

Exirel:
I'm looking at your branch @half-duplex and I really don't like this syntax:

```
"""
Synopsis of multi-line docstring.

Long description.
"""
```

I don't see the need for that change.

Also, I don't see the point of adding short docstring just for the sake of *having* a docstring.

Exirel:
Also, if you want to flag stuffs without docstring: use *pylint*.

half-duplex:
> I'm looking at your branch @half-duplex and I really don't like this syntax

I wasn't going to break `"""Synopsis` lines, but dgw's branch added that warning.

> Also, I don't see the point of adding short docstring just for the sake of _having_ a docstring.

[...]

PR #2016, opened 2021-01-15
[POC] all: fix deepsource checks

RhinosF1:
### Description
Cleans up a bunch of possible style issues per https://deepsource.io/gh/RhinosF1/sopel/issues/?category=recommended

This is a proof of concept but if you like the feedback then I can go further or it could be configured on your end too.

### Checklist


Exirel:
Almost a year now, and I don't think we gave it a proper review. Probably because, as you may have guest, I'm not a huge fan of adding yet another tools with yet another opinions on code style.

I think it's fair to say we can close this. What do you say @dgw ?

sopel/config/__init__.py
```diff
@@ -156,8 +156,7 @@ def homedir(self):
             configured = self.parser.get('core', 'homedir')
         if configured:
             return configured
-        else:
-            return os.path.dirname(self.filename)
+        return os.path.dirname(self.filename)
```
dgw, line 159: But this kind of thing is also good.

sopel/config/__init__.py
```diff
@@ -276,7 +275,7 @@ def define_section(self, name, cls_, validate=True):
         current_name = str(current.__class__)
         new_name = str(cls_)
         if (current is not None and not isinstance(current, self.ConfigSection) and
-                not current_name == new_name):
+                current_name != new_name):
```
dgw, line 278: This is good, too.

dgw (commented):
> I'm not a huge fan of adding yet another tools with yet another opinions on code style.

> What do you say @dgw ?

I say what's below in the line notes. **tl;dr:** This tool does have _some_ useful rules, but I can only see documentation on ignoring issues in specific lines. One reason I think flake8 has been our preference for code style is that it lets you globally decide that you don't care about a certain style rule, and in many cases there are multiple rules to choose from (e.g. putting binary operators at the beginning or end of a line).

If there's really no way to customize the style rules applied by DeepSource, then the way forward is to look for flake8 plugins that enforce the DeepSource rules I/we _do_ like and add those (to the wishlist in #1765, probably, so they can be implemented at an opportune moment).

PR #2026, opened 2021-02-06
tools.time: improve seconds_to_human

### LINKED ISSUE #1824: tools.time feature: "Raw" mode for seconds_to_human()

Exirel:
### Description

Fix #1824 

Expose intermediate functions: one to split a time in seconds into a tuple of (years, months, days, hours, minutes, seconds) and the other that take that information to return a tuple of 2-value tuple with (value, unit).

Also, while I was working on making `second_to_human` more reusable, I added a `granularity` parameter, because why not.

### Checklist

### DIFF
--- a/sopel/tools/time.py
+++ b/sopel/tools/time.py
@@ -7,6 +7,15 @@ import datetime
 import pytz
 
 
+# various time unit in seconds, approximation for months and years
+SECONDS = 1
+MINUTES = 60 * SECONDS
+HOURS = 60 * MINUTES
+DAYS = 24 * HOURS
+MONTHS = int(30.5 * DAYS)
+YEARS = 365 * DAYS
+
+
 def validate_timezone(zone):
     """Return an IETF timezone from the given IETF zone or common abbreviation.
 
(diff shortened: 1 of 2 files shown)


dgw (changes requested):
Just some nemesi**S** fighting in here. I'd like for the `granularity` parameter to round the value up if it's close to the next unit (e.g. 1h20m = "1 hour" but 1h40m = "2 hours"), but I don't even need to wait for you to say "let's do that in a future PR" because I already think that. x) I'll put that idea on the list for 8.0.

dgw (approved):
Imagine forgetting to re-review a PR for two weeks. 🤯

Exirel:
Squashed & rebased.

PR #2028, opened 2021-02-06
url: urlexclude/allow commands; auto-title ignores URLs in ANY other command

### LINKED ISSUE #1870: Allow changing the URL blacklist from IRC
### The problem
You have to restart the bot and modify its config file to change the blacklist.

### The solution
We could make the URL blacklist available from IRC and editable.

https://github.com/Pix1234/ZppixBot-Source/blob/master/modules/urls.py does this and could be inspiration

Edit: in particular, this bit here https://github.com/Pix1234/ZppixBot-Source/blob/master/modules/urls.py#L242-L289

Exirel:
### Description

Close #1870

And because nothing is simple: I also added aliases for `urlpban` and `urlpunban` to ban/unban patterns instead of URL directly.

### Checklist

### DIFF
--- a/sopel/modules/url.py
+++ b/sopel/modules/url.py
@@ -153,6 +153,96 @@ def shutdown(bot):
             pass
 
 
+@plugin.command('urlexclude', 'urlpexclude', 'urlban', 'urlpban')
+@plugin.example('.urlpexclude example\\.com/\\w+', user_help=True)
+@plugin.example('.urlexclude example.com/path', user_help=True)
+@plugin.output_prefix('[url] ')
+def url_ban(bot, trigger):
+    """Exclude a URL from auto title.
+
+    Use ``urlpexclude`` to exclude a pattern instead of a URL.
+    """
+    url = trigger.group(2)
+
+    if not url:
+        bot.reply('This command requires a URL to exclude.')
+        return
+
+    if trigger.group(1) in ['urlpexclude', 'urlpban']:
+        # validate regex pattern
+        try:
+            re.compile(url)
+        except re.error as err:
+            bot.reply('Invalid regex pattern: %s' % err)
+            return
+    else:
+        # escape the URL to ensure a valid pattern
+        url = re.escape(url)
+
+    patterns = bot.settings.url.exclude
+
+    if url in patterns:
+        bot.reply('This URL is already excluded from auto title.')
+        return
+
+    # update settings
+    patterns.append(url)
+    bot.settings.url.exclude = patterns  # set the config option
+    bot.settings.save()
+    LOGGER.info('%s excluded the URL pattern "%s"', trigger.nick, url)
+
+    # re-compile
+    bot.memory['url_exclude'] = [re.compile(s) for s in patterns]
+
+    # tell the user
+    bot.reply('This URL is now excluded from auto title.')
+
+
+@plugin.command('urlallow', 'urlpallow', 'urlunban', 'urlpunban')
+@plugin.example('.urlpallow example\\.com/\\w+', user_help=True)
+@plugin.example('.urlallow example.com/path', user_help=True)
+@plugin.output_prefix('[url] ')
+def url_unban(bot, trigger):
+    """Allow a URL for auto title.
+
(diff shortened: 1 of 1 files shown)


cottongin:
I like this, my only suggestion is what do you think about renaming the commands to use a different word than "ban"?

At least to me using "ban" implies banning someone from using the plugin. I know it also includes "url" in the command name, so perhaps it is less confusing to others, but the plugin is also named `url.py`... (maybe a better suggestion would be to rename the plugin `title.py` or something)

Exirel:
If you have something better than `ban`, I'll take it. `urlexclude` maybe?

cottongin:
Yes sure, exclude makes more sense to me

sopel/modules/url.py
```diff
@@ -153,6 +153,96 @@ def shutdown(bot):
...
+        try:
+            re.compile(url)
+        except re.error as err:
+            bot.reply('Invalid regex pattern: %s' % err)
+            return
+    else:
+        # escape the URL to ensure a valid pattern
+        url = re.escape(url)
+
+    patterns = bot.settings.url.exclude or []
```
dgw, line 182: No need to fall back on an empty list; `ListAttribute` already presents as empty list if not set.

```suggestion
    patterns = bot.settings.url.exclude
```

sopel/modules/url.py
```diff
@@ -153,6 +153,96 @@ def shutdown(bot):
...
+
+    if url in patterns:
+        bot.reply('This URL is already banned from auto title.')
+        return
+
+    # update settings
+    patterns.append(url)
+    bot.settings.url.exclude = patterns
```
dgw, line 190: Not necessary if `patterns` is a simple assignment to `bot.config.url.exclude`; it'll be the same object reference.

```suggestion
```

sopel/modules/url.py
```diff
@@ -153,6 +153,96 @@ def shutdown(bot):
...
+        try:
+            re.compile(url)
+        except re.error as err:
+            bot.reply('Invalid regex pattern: %s' % err)
+            return
+    else:
+        # escape the URL to ensure a valid pattern
+        url = re.escape(url)
+
+    patterns = bot.settings.url.exclude or []
```
dgw, line 227: See comments/suggestions in `url_ban()` above.

```suggestion
    patterns = bot.settings.url.exclude
```

sopel/modules/url.py
```diff
@@ -153,6 +153,96 @@ def shutdown(bot):
...
+
+    if url not in patterns:
+        bot.reply('This URL is not banned from auto title.')
+        return
+
+    # update settings
+    patterns.remove(url)
+    bot.settings.url.exclude = patterns
```
dgw, line 235: See `url_ban()` comments.

```suggestion
```

dgw (changes requested):
"urlpunban" lol 😹

I couldn't take the time just now to think about the `# re-compile` steps, but surely there's a way to manage `bot.memory['url_exclude']` without running a whole list comprehension every time?

Exirel:
> I couldn't take the time just now to think about the `# re-compile` steps, but surely there's a way to manage `bot.memory['url_exclude']` without running a whole list comprehension every time?

I think there is, but I couldn't bother to do something complicated without the proper unit-test to back me up. I think the tradeoff is good for now. I know I tend to be a bit lazy with tests when it comes to plugin, because I never know if I really want to invest in them right now, or wait for when they are all external plugins. After all, I did a 100% test coverage on sopel-help, and I'm doing that for remind too...

dgw (dismissed):
Squash at will, sir. 🖖

Exirel:
OK, I squashed and rebased after some further fix and changes, making it worthy of a second full review, I think.

I've settled for `urlexclude` and `urlallow` in hope of making @cottongin happy. We discussed this with @dgw but neither of us had an opinion about it. We decided to keep the `url(pun)ban` because it's funny.

dgw (approved):
Retitled to reflect the extra fix Exi sneakily added, so I remember to mention it in the changelog.

PR #2029, opened 2021-02-25
url: restore error output when not all `.title` URLs could be fetched

### LINKED ISSUE #1946: Error messages were removed from `.title` command
In #1510, the error feedback Sopel used to output if it couldn't find a title for one or more URL given to the `.title` command got removed. This is kind of regression-y, and was shipped already in 7.0.x before someone mentioned it. Here's how it looked: https://github.com/sopel-irc/sopel/blob/806f296fdb74f197397bcdd6eb5464cd07a638cc/sopel/modules/url.py#L132-L137 Might not be practical to do exactly the same thing now due to use of generators, but something similar would be enough.

dgw:
### Description
Tin. Goes slightly beyond the previous error messages with a special case for failing to fetch all URLs in a list of 2+. Fixes #1946.

### Checklist
- [x] No issues are reported by `make quality` (I didn't touch anything covered by tests)

### DIFF
--- a/sopel/modules/url.py
+++ b/sopel/modules/url.py
@@ -174,16 +174,30 @@ def title_command(bot, trigger):
         else:
             urls = [bot.memory['last_seen_url'][trigger.sender]]
     else:
-        urls = web.search_urls(
-            trigger,
-            exclusion_char=bot.config.url.exclusion_char)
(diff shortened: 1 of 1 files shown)


Exirel (approved):
Neat.

PR #2033, opened 2021-03-01
bot: emit warning when memory['url_callbacks'] is used manually in a plugin's setup

Exirel:
### Description

Tin. Replace #1809 
I had to mark callbacks to prevent raising the same warning more than once.

I accept the trade-off: this won't cover 100% of cases and I'm fine with that. It'll cover the most common one, when it was common to declare URL callbacks in the `setup` hook because `@url` didn't work well.

### Checklist

### DIFF
--- a/sopel/bot.py
+++ b/sopel/bot.py
@@ -331,6 +331,8 @@ class Sopel(irc.AbstractBot):
                 try:
                     if plugin.has_setup():
                         plugin.setup(self)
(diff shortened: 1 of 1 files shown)


half-duplex (approved):
Looks sound, unless we want to deprecate `register_url_callback()` too as discussed in IRC.

dgw (approved):
I'm going along with @half-duplex's nitpick, which I'll just apply manually.

PR #2044, opened 2021-03-25
config.types, test: add `BooleanAttribute` setting type

### LINKED ISSUE #2019: Sopel should have a BooleanAttribute config type
### The problem
It's really unintuitive to have `ValidatedAttribute` with `bool` as the `parse` parameter, because all that happens is the object silently uses something completely different to parse _and serialize_ its values.

### The solution
Let's add a `BooleanAttribute` type in ~~8.0~~ 7.1 that uses `ConfigParser.getboolean()` instead of bespoke parse/serialize functions.

### Notes
Bonus points if a `ValidatedAttribute` with `parse == bool` gets internally transformed into a `BooleanAttribute` so we can remove the private helpers `_parse_boolean` and `_serialize_boolean` immediately.

Of course the special handling of `parse == bool` will get taken out eventually, but that `if` statement has to stick around anyway so it can emit a deprecation warning until 9.0.

dgw:
### Description
Was going to move the helper functions used in `ValidatedAttribute` when `parse == bool`, but it would've required some ridiculous polymorphism. Better to just leave the old stuff—which will also make it easier to mark as deprecated.

Maybe we need to add tests that cover the `default` parameter to some or all of these setting types? I'll chew on that.

When complete, this will implement and close #2019.

### Checklist

### DIFF
--- a/sopel/config/types.py
+++ b/sopel/config/types.py
@@ -30,8 +30,9 @@ import getpass
 import os.path
 import re
 import sys
+import traceback
 
-from sopel.tools import get_input
+from sopel.tools import get_input, stderr
 
 if sys.version_info.major >= 3:
     unicode = str
@@ -299,10 +300,24 @@ class ValidatedAttribute(BaseValidated):
                  is_secret=False):
         super(ValidatedAttribute, self).__init__(
             name, default=default, is_secret=is_secret)
+
         if parse == bool:
             parse = _parse_boolean
             if not serialize or serialize == bool:
                 serialize = _serialize_boolean
+
+            # deprecation warning
+            # can't use tools.deprecated in this case, unfortunately
+            # we need this to be conditional
+            msg = (
+                'Deprecated since 7.1, '
+                'will be removed in 9.0: '
+                'Use BooleanAttribute instead of ValidatedAttribute with parse=bool')
+            stderr(msg)
+            # Only display the last stack frame
+            trace = traceback.extract_stack()
+            stderr(traceback.format_list(trace[:-1])[-1][:-1])
+
         self.parse = parse or self.parse
         self.serialize = serialize or self.serialize
 
--- a/sopel/config/types.py
+++ b/sopel/config/types.py
@@ -30,9 +30,8 @@ import getpass
 import os.path
 import re
 import sys
-import traceback
 
-from sopel.tools import get_input, stderr
+from sopel.tools import deprecated, get_input
 
 if sys.version_info.major >= 3:
     unicode = str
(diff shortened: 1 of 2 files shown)


dgw:
Simply could not think of a better approach to emitting deprecation warnings for `ValidatedAttribute(..., parse=bool)`. Obviously decorating `ValidatedAttribute.__init__()` as `@deprecated` would cover a LOT more cases than just the Boolean ones. Decorating the helpers `_parse_boolean()` and `_serialize_boolean()` caused tons of spam in the terminal.

Exirel:
> Decorating the helpers `_parse_boolean()` and `_serialize_boolean()` caused tons of spam in the terminal. man_shrugging

Indeed. If you want to reduce the duplicated code, you can:

1. put everything from the `if bool` block into a function/a method
2. put the deprecation warning on that function/method
3. call it in the `if bool` and nowhere else

The content of said function would be something like:

```python
@deprecate(...)
def _get_deprecated_boolean_handlers(parser=None, serializer=None):
    # ... logic from that if bool block
    return the_boolean_parser, the_boolean_serializer
```

Problem solved.

dgw:
> Problem solved.

Not really. I mean technically, it deduplicates the code, but the warning printed is this:

```
Deprecated since 7.1, will be removed in 9.0: Use BooleanAttribute instead of ValidatedAttribute with parse=bool
  File "/mnt/c/Users/dgw/github/sopel/sopel/config/types.py", line 317, in __init__
    parse, serialize = _deprecated_special_bool_handling(serialize)
```

That doesn't tell the user or plugin developer where the outdated usage lies.

If `tools.deprecated()` had a kwarg to specify which stack frame to show, that might help, but it's way out of scope for this PR. That said, I already put together the bare bones of such a new kwarg; I'd just need to document it to make it PR-ready.

Exirel:
We can always add the field's name in the deprecation message, that should help a bit.

dgw:
I'm not at all sure how you'd do that from a decorator on a helper function called from the object constructor, but if I can learn something from you again, I'm all eyes.

Exirel:
I'd do that within the constructor:

```
def __init__(self, name, ...):
    @deprecate("some message with the name %s" % name)
    def _get_deprecated_boolean_handlers(parser=None, serializer=None):
        # ... logic from that if bool block
        return the_boolean_parser, the_boolean_serializer

    if bool:
        parser, serialiazer = _get_deprecated_boolean_handlers()
```

Not perfect, but it works and it's temporary.

dgw:
I guess defining stuff inside other defs is not unheard of in Sopel's code, e.g. the old signal handlers. But even that approach would benefit from the PR I'm about to open against `tools.deprecated`.

And don't worry, I _did_ see your muttering on IRC about using `warnings.warn()`. That's something `tools.deprecated` could do instead of directly printing to stderr itself. I've also mused recently that that decorator should output to the logs along with everything else. Lots to think about for 8.0!

Exirel:
> even that approach would benefit from the PR

Yup, and if that PR can be merged first, go for it. I think it would be the better option here.

dgw:
Pushed using the new kwarg to `tools.deprecated`. If this doesn't pass CI I'll do a rebase later.

Might want to clean up history later anyway, tbf.

dgw:
I cleaned up the history. @Exirel presumably will be the one to review this, as it uses his suggested approach.

Exirel (approved):
👍

PR #2046, opened 2021-03-27
tools: add `stack_frame` kwarg to `deprecated` decorator

dgw:
### Description
Allows specifying which stack frame to show in the traceback, if the default (the most recent) isn't helpful. Sometimes the parent caller or a more distant ancestor is more informative to the end user.

### Checklist

### Notes
Inspired by challenges encountered when working on #2044. This should be merged first if approved.

I seriously considered making `stack_frame` just a plain integer, and not requiring the negative sign, just because 99% (maybe even 100%) of all use cases are going to want a frame relative to the _end_ of the list generated by `traceback.extract_stack()`. But it's more intuitive to make the argument the same value as the list index used inside the decorator…

### DIFF
--- a/sopel/tools/__init__.py
+++ b/sopel/tools/__init__.py
@@ -65,6 +65,8 @@ def deprecated(
                            will be removed
     :param str warning_in: optional version number when the decorated function
                            should start emitting a warning when called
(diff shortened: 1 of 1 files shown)


Exirel (approved):
I love it.

PR #2048, opened 2021-03-28
irc.isupport, announce: fix unexpected error (get) from .announce

### LINKED ISSUE #2040: Unexpected error (get) from .announce
### Description
.announce fatals with unexpected error (get)

### Reproduction steps
1. Upgrade to latest master
2. Connect to server, join a few channels
3. Use .announce
4. Check your logs

### Expected behavior
No fatals, it works 

### Logs
```
[2021-03-14 16:02:42,710] sopel.bot            ERROR    - Unexpected error (get) from RhinosF1 at 2021-03-14 16:02:42.710332. Message was: .announce this is a test --rf1
  File "/srv/sopelbots/devvenv/lib/python3.7/site-packages/sopel/bot.py", line 698, in call_rule
    rule.execute(sopel, trigger)
  File "/srv/sopelbots/devvenv/lib/python3.7/site-packages/sopel/plugins/rules.py", line 996, in execute
  File "/srv/sopelbots/devvenv/lib/python3.7/site-packages/sopel/plugin.py", line 914, in guarded
    return function(bot, trigger, *args, **kwargs)
  File "/srv/sopelbots/devvenv/lib/python3.7/site-packages/sopel/modules/announce.py", line 37, in announce
    size = bot.isupport.get('TARGMAX', {}).get('PRIVMSG', 1)
  File "/srv/sopelbots/devvenv/lib/python3.7/site-packages/sopel/irc/isupport.py", line 192, in __getattr__
    raise AttributeError(name)
AttributeError: get
```

### Environment
- Sopel `.version`: latest master
- Sopel installed via: build our own wheel using build & then pip install
- Python version: 3.7.3
- Operating system: Debian Buster
- IRCd `/version`: ircd-seven-1.1.9(20191023-4b033a7b9221,charybdis-3.4-dev). egan.freenode.net
- Relevant plugins: announce

### Notes
Anything else we should know?

Exirel:
### Description

Fix #2040

### Checklist

### DIFF
--- a/sopel/modules/announce.py
+++ b/sopel/modules/announce.py
@@ -34,8 +34,14 @@ def announce(bot, trigger):
         bot.reply('Announce what? I need a message to say.')
         return
 
-    size = bot.isupport.get('TARGMAX', {}).get('PRIVMSG', 1)
+    size = 1
+    try:
+        size = bot.isupport.TARGMAX.get('PRIVMSG', size)
+    except AttributeError:
+        pass
+
     channels = _chunks(bot.channels.keys(), size)
     for cgroup in channels:
         bot.say(trigger.group(2), ','.join(cgroup))
+
     bot.reply('Announce complete.')
--- a/sopel/irc/isupport.py
+++ b/sopel/irc/isupport.py
@@ -123,7 +123,7 @@ ISUPPORT_PARSERS = {
     'SAFELIST': _no_value,
     'SILENCE': _optional(int),
     'STATUSMSG': _optional(tuple),
-    'TARGMAX': _optional(_map_items(int)),
+    'TARGMAX': _optional(_map_items(int), default=tuple()),
     'TOPICLEN': int,
     'USERLEN': int,
 }
@@ -371,10 +371,15 @@ class ISupport(object):
                 'PRIVMSG': 3,
                 'WHOIS': 1,
             }
+            >>> isupport['TARGMAX']  # internal representation
+            (('JOIN', None), ('PRIVMSG', 3), ('WHOIS', 1))
 
         This attribute is not available if the server does not provide the
         right information, and accessing it will raise an
         :exc:`AttributeError`.
+
+        The internal representation of ``TARGMAX`` is a tuple of 2-value
+        tuples as seen above.
 
         .. seealso::
 
@@ -384,9 +389,5 @@ class ISupport(object):
         if 'TARGMAX' not in self:
             raise AttributeError('TARGMAX')
 
-        targmax = self['TARGMAX']
-
-        if targmax is None:
-            return {}
-
+        # always return a dict if None or empty tuple
         return dict(self['TARGMAX'] or [])
--- a/sopel/irc/isupport.py
+++ b/sopel/irc/isupport.py
@@ -204,6 +204,16 @@ class ISupport(object):
         elif name not in self.__dict__:
             raise AttributeError('Unknown attribute')
 
+    def get(self, name, default=None):
+        """Retrieve value for the feature ``name``.
+
+        :param str name: feature to retrieve
+        :param default: default value if the feature is not advertised
+                        (defaults to ``None``)
+        :return: the value for that feature, if advertised, or ``default``
+        """
+        return self[name] if name in self else default
+
     def apply(self, **kwargs):
         """Build a new instance of :class:`ISupport`.
 
--- a/test/irc/test_irc_isupport.py
+++ b/test/irc/test_irc_isupport.py
@@ -508,6 +508,13 @@ def test_parse_parameter_targmax():
     assert value == (('PRIVMSG', 3),)
 
 
+def test_parse_parameter_targmax_optional():
+    key, value = isupport.parse_parameter('TARGMAX=')
+
+    assert key == 'TARGMAX'
+    assert value == tuple()
+
+
 def test_parse_parameter_targmax_many():
     key, value = isupport.parse_parameter('TARGMAX=PRIVMSG:3,WHOIS:1')
 
--- a/test/irc/test_irc_isupport.py
+++ b/test/irc/test_irc_isupport.py
@@ -76,6 +76,22 @@ def test_isupport_removed_parameter():
     assert '-AWAYLEN' not in instance
 
 
+def test_isupport_get():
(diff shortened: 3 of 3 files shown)


sopel/modules/announce.py
```diff
@@ -34,7 +34,7 @@ def announce(bot, trigger):
         bot.reply('Announce what? I need a message to say.')
         return
 
-    size = bot.isupport.get('TARGMAX', {}).get('PRIVMSG', 1)
+    size = dict(bot.isupport.get('TARGMAX') or []).get('PRIVMSG', 1)
```
dgw, line 37: If `TARGMAX` isn't advertised, you're going to run `[].get('PRIVMSG', 1)`?

And why not use the `default` kwarg you were so careful to add on `ISupport.get()`?

dgw (commented):
I have a confuse.

sopel/modules/announce.py
Exirel: In its raw form, `TARMAX` is parse as a list of items, that's why I transform it into a `dict( get targmax )` before accessing its `PRIVMSG` key (`dict(...).get(...)`).

And I don't like relying on default, but in that case I think I was too cautious, it could be this instead:

```suggestion
    size = dict(bot.isupport.get('TARGMAX', [])).get('PRIVMSG', 1)
```

And I need to change [this line](https://github.com/sopel-irc/sopel/blob/96c55aff852bf40bca56de49b2bc30378bf1c819/sopel/irc/isupport.py#L126) so the default value for TARGMAX is actually a tuple instead of `None`.

dgw: > In its raw form, `TARGMAX` is parsed as a tuple of items, that's why I transform it into a `dict( get targmax )` before accessing its `PRIVMSG` key (`dict(...).get(...)`).

I still have a confuse, because [Sopel's unstable docs say](https://unstable.docs.sopel.chat/irc.html#sopel.irc.isupport.ISupport.TARGMAX) it's already a dict:

> _property_ **TARGMAX**
>
> Expose `TARGMAX` as a dict, if advertised by the server.
>
> This exposes information about the maximum number of arguments for each command:
>
> ```repl
> >>> isupport.TARGMAX
> {
>     'JOIN': None,
>     'PRIVMSG': 3,
>     'WHOIS': 1,
> }
> ```
>
> This attribute is not available if the server does not provide the right information, and accessing it will raise an `AttributeError`.

_Note: Stable docs also say this. I just quoted the unstable ones to make sure the language was as up-to-date as possible._

Now I notice that accessing stuff via property/`bot.isupport.TARGMAX` gives you _different data_ (a dict) vs. treating `bot.isupport` as a dict-ish object via `__getitem__`/`bot.isupport['TARGMAX']` (returns the internal data structure, a tuple of tuples)… and I'm mad that I didn't notice before because this is _not good_ API design IMO. Especially because the difference isn't documented, but it's just unintuitive anyway.

For _this_ PR, the solution is likely something like this:

```py
size = 1

try:
    size = bot.isupport.TARGMAX.get('PRIVMSG', 1)
except AttributeError:
    # just don't try to group targets
    pass
```

For later, the `ISupport` stuff needs revisiting. It's _probably_ not important enough for 7.1, and any changes to how the `__getitem__` access works would be "breaking" and therefore not shippable until 8.0 anyway. But I'm going to make damn sure it's on our radar. I can think of no good reason to expose the internal data structure of any `ISupport` property if it's any more complicated than a scalar.

> And I need to change [this line](https://github.com/sopel-irc/sopel/blob/96c55aff852bf40bca56de49b2bc30378bf1c819/sopel/irc/isupport.py#L126) so the default value for TARGMAX is actually a tuple instead of `None`.

Might want to change [this line](https://github.com/sopel-irc/sopel/blob/96c55aff852bf40bca56de49b2bc30378bf1c819/sopel/irc/isupport.py#L382) too, so it'll use an empty tuple instead of an empty list? That's nitpicking, yes.

Exirel: > because this is not good API design IMO

I tend to disagree with your opinion on that. By using the modern.ircdocs.horse I followed their documentation: when an item is a list, then it is parsed as a tuple (which is just an immutable list).

Although there is no documentation that states that `isupport['TARGMAX']` is a tuple, there is no documentation either that says it's a dict. There is an example given in the ISupport's docstring:

    >>> isupport = ISupport(chanlimit=(('&', None), ('#', 70)))
    >>> isupport['CHANLIMIT']
    (('&', None) ('#', 70))
    >>> isupport.CHANLIMIT  # some parameters are also properties
    {
        '&': None,
        '#': 70,
    }

And it clearly shows the difference between internal keys, and properties. It's even more clear when you look at the `CHANMODES` properties, which only return the 4 first item in the tuple, because everything outside of A, B, C, and D chan modes are unspecified.

I admit, there is a lot to document here. I think the best option is to document the `ISUPPORT_PARSERS` mapping, which contains how each value is parsed.

> so it'll use an empty tuple instead of an empty list?

No, it's just a shortcut to create an empty dict if necessary. You really tend to forget that there is a `dict(...)` surrounding stuff the way I forget about S everywhere. :grin:

Exirel:
Design consideration aside (that we should totally start to address in 8.0), I think @dgw is right that the documentation is not good enough. In an attempt to improve the situation with a small increment, I pushed a new commit with this:

image

It's not enough, I think a note should be added about the differences between internal representation and attributes, and we should probably give an explanation of how the isupport parser system works—even though it's all pretty internal stuff in here, and nothing will replace a good knowledge of IRC spec.

dgw (approved):
I've come to terms with how this interface works for 7.x, plus both of us are probably ready to move on until we start on 8.0. The documentation side might be more helpful if it placed more emphasis on interacting with `bot.isupport` using `getattr()`? But that's for a separate PR, and likely 8.0.

sopel/modules/announce.py
dgw: > No, it's just a shortcut to create an empty dict if necessary. You really tend to forget that there is a `dict(...)` surrounding stuff the way I forget about S everywhere. 😁

Well we all have to have _some_ quirk. 😁

Let's leave things as they are and talk about documentation improvements for 8.0, sound good? If we come up with justifiable API changes during that process, so be it.

PR #2052, opened 2021-04-03
find_updates: return in ALL the appropriate failure cases

dgw:
### Description
Even if there haven't been enough check failures to trigger a warning to Sopel's owner, it still needs to skip trying to print the `info` about whatever new release is available because that variable doesn't exist.

Missed in #1779 / 951fccaccf5f4bdb70058a326543c1d82354c453, oops.

### Checklist
- [x] I have read [CONTRIBUTING.md](https://github.com/sopel-irc/sopel/blob/master/CONTRIBUTING.md)
- [x] I can and do license this contribution under the EFLv2
- [x] No issues are reported by `make qa` (runs `make quality` and `make test`)
- [ ] I have tested the functionality of the things this change touches (see Notes)

### Notes
Relying on reporter (@RhinosF1) to see if the bug is still reproducible with this patch.

PR #2056, opened 2021-04-30
find_updates, test: create test to verify error case behavior

dgw:
### Description
Validates the behavior fixed in #2052.

### Checklist
  - On Python 3.8, _and_ (accidentally) 3.3. The latter's the version I was worried about breaking, anyway.

### Notes
Haven't milestoned this yet, as I'm not sure if I want to add a new dependency (even only for dev) on 7.1.x this late. Looking for second opinions before I decide!

### DIFF
--- a/test/modules/test_modules_find_updates.py
+++ b/test/modules/test_modules_find_updates.py
@@ -1,0 +1,48 @@
+# coding=utf-8
+"""Tests for Sopel's ``find_updates`` plugin"""
+from __future__ import absolute_import, division, print_function, unicode_literals
+
+import pytest
+import requests.exceptions
+
+from sopel.modules import find_updates
+
+
+TMP_CONFIG = """
+[core]
+owner = Admin
+nick = Sopel
+enable =
+    find_updates
+host = chat.freenode.net
+"""
+
+
+@pytest.fixture
+def mockbot(configfactory, botfactory):
+    tmpconfig = configfactory('default.ini', TMP_CONFIG)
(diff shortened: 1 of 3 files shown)


dgw:
Guess what happens when you forget to force-push? Oh yeah, CI failures. x)

Exirel:
Look at you, writing tests! I'm so proud of you. ❤️

test/modules/test_modules_find_updates.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+TMP_CONFIG = """
+    [core]
+    owner = Admin
+    nick = Sopel
+    enable =
+        find_updates
+    host = chat.freenode.net
+    """
```
Exirel, line 18:
```suggestion
[core]
owner = Admin
nick = Sopel
enable =
    find_updates
host = chat.freenode.net
"""
```

And then py2.7 will stop complaining.

dgw: Aw hell, lol. I was so worried about py3.3 breaking that I didn't test 2.7 😹

dgw:
@Exirel now that py2.7 is happy (😤), any thoughts about holding this for 8.0 vs. dropping it into 7.1? The hard version pin should prevent anything from breaking too much during maintenance of 7.1.x, plus it's only a dev-dep.

Exirel:
Let's go with 7.1, it's a dev thing.

Exirel (approved):

PR #2058, opened 2021-05-03
tools: Convert `deprecated` decorator to use logging output

dgw:
### Description
No more `stderr()` calls!

One notable exception: If a `core` section setting uses some deprecated behavior, that warning will not be logged in the usual way because logging isn't set up yet. The bot _has_ to load its core settings before it can get logging up, since that's where the logging itself is configured. Core plugins' `StaticSection`s get loaded later, after logging is available, so they work as expected.

We core devs will just have to pay attention to the console output during testing in the future if any more config stuff used by the `core` section gets deprecated. It's not something that end-users (or even plugin devs) will need to worry about.

### Checklist
  - Before: Lots of deprecation warnings from core plugins about using `BooleanAttribute` instead of `ValidatedAttribute` with `parse=bool`, printed directly to the console.
  - After: Same warnings, but they all go through the logging formatter instead and appear in the log file. (Well, _almost_ all; see note above).

### Notes
I worked on this quite intently for a day or two last month, but a new work project began consuming the majority of my thoughts and I forgot to even mention this, let alone propose it as a pull request. I remembered this upon reviewing my list of local branches after cleaning up a bunch of merged work.

My next PR will probably be to replace the deprecated behavior that spawned all these warnings, because I didn't do so in #2044. This patch is best tested first, since these are the _only_ deprecation warnings present in Sopel's own code at this point.

### DIFF
--- a/sopel/tools/__init__.py
+++ b/sopel/tools/__init__.py
@@ -17,6 +17,7 @@ from __future__ import absolute_import, division, print_function, unicode_litera
 import codecs
 from collections import defaultdict
 import functools
+import inspect
 import logging
 import os
 import re
@@ -168,10 +169,31 @@ def deprecated(
     def deprecated_func(*args, **kwargs):
         if not (warning_in and
                 parse_version(warning_in) >= parse_version(__version__)):
-            stderr(text)
-            # Only display the last frame
+            original_frame = inspect.stack()[-stack_frame]
+            mod = inspect.getmodule(original_frame[0])
+            module_name = None
+            if mod:
+                module_name = mod.__name__
+            if module_name:
+                if 'sopel.' in module_name:
+                    # core, or core plugin
+                    logger = logging.getLogger(module_name)
+                else:
+                    # probably a plugin; use Sopel's public API for getting the
+                    # logger for a plugin
+                    logger = get_logger(module_name.replace('sopel_modules.', ''))
+            else:
+                # don't know the module/plugin name, but we want to make sure
+                # the log line is still output, so fudge it
+                logger = logging.getLogger('sopel.<unknown>')
+
+            # Format only the desired stack frame
             trace = traceback.extract_stack()
-            stderr(traceback.format_list(trace[:-1])[stack_frame][:-1])
+            trace_frame = traceback.format_list(trace[:-1])[stack_frame][:-1]
+
+            # Warn the user
+            logger.warning(text + "\n" + trace_frame)
+
         return func(*args, **kwargs)
 
     return deprecated_func
--- a/sopel/tools/__init__.py
+++ b/sopel/tools/__init__.py
@@ -175,17 +175,21 @@ def deprecated(
             if mod:
                 module_name = mod.__name__
             if module_name:
-                if 'sopel.' in module_name:
+                if module_name.startswith('sopel.'):
                     # core, or core plugin
                     logger = logging.getLogger(module_name)
                 else:
                     # probably a plugin; use Sopel's public API for getting the
                     # logger for a plugin
-                    logger = get_logger(module_name.replace('sopel_modules.', ''))
+                    if module_name.startswith('sopel_modules.'):
+                        # namespace package plugins have a prefix, obviously
+                        # TODO: use str.removeprefix() when we drop Python<3.9
+                        module_name = module_name.replace('sopel_modules.', '', 1)
+                    logger = get_logger(module_name)
             else:
                 # don't know the module/plugin name, but we want to make sure
-                # the log line is still output, so fudge it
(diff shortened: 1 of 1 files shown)


sopel/tools/__init__.py
```diff
@@ -168,10 +169,31 @@ def decorator(func):
...
+                    # core, or core plugin
+                    logger = logging.getLogger(module_name)
+                else:
+                    # probably a plugin; use Sopel's public API for getting the
+                    # logger for a plugin
+                    logger = get_logger(module_name.replace('sopel_modules.', ''))
+            else:
+                # don't know the module/plugin name, but we want to make sure
+                # the log line is still output, so fudge it
+                logger = logging.getLogger('sopel.<unknown>')
```
Exirel, line 188: Why would you fake it? You can just use the regular logger:

```suggestion
                logger = logging.getLogger(__name__)
```

It would be fine by me.

Exirel (commented):
I don't know if I can approve or not. I'm a bit confused, and not really sold on the idea of using Logging instead of just stderr.

I mean, what are the point of `deprecated`? If it's to deprecate part of the code, as mentioned, it shouldn't even be a simple string, but a true `warning.warn`. If it's to deprecate a Sopel's owner feature, then the `@deprecated` decorator is not really well designed for that.

In either case, I'm not convinced by the usage of logging here—that plus the weird "won't work before logging is setup" part.

dgw:
It's not that it does _nothing_ before logging is set up; the deprecation warning will appear _only_ in console output, because the file logger and other stuff doesn't exist yet. Those messages look exactly the same as before (but aren't on stderr any more). After logging is set up, they come out looking like every other log message with the package name and all the other details.

The point of _this_ is making deprecation warnings appear in the log file. Currently they don't, and the only way to see them is looking at Sopel's console output (on stderr). Maybe `warnings.warn` is better than just sending a log message, but I'd have to learn how to write that approach first. 😁

sopel/tools/__init__.py
dgw: Wouldn't this make the source module appear to be `sopel.tools` or something like that? The point of fudging it is to make the log entry stand out, because this should never be unknown.

Exirel:
> deprecation warning will appear _only_ in console output

Yeah, but who can fix those warnings, if it's a *developer* problem? Not the bot's owner, they can't do that, so having that warning won't help them anyway. Only the plugin author needs that warning, and only them can act on that. Why would a developer use the log file when developing when they can just, you know... read the console output. It's too weird for me!

dgw:
> Not the bot's owner

It seems quite common for bot owners to have custom plugins, actually. At the very least, exposing the deprecated usage to the bot owner _with info about which module_ makes it easy for bot owners to report the problem to plugin maintainers if it's not custom code.

Let's face it: Not all Sopel plugins have really active developers who hang on our every commit looking for stuff they'll need to update! 😂

Exirel:
> Let's face it: Not all Sopel plugins have really active developers who hang on our every commit looking for stuff they'll need to update! 😂


Sadly, I can't argue with that. :(

sopel/tools/__init__.py
Exirel: The traceback + WARNING level should be enough in my opinion.

Exirel:
OK, enough chit chat: you made your point, and I think it's a reasonable approach. If I'm not happy in Sopel 8.0, then it's up to me to find a better solution. And if I can't find one, then we are all good.

So, LGTM!

sopel/tools/__init__.py
dgw: New commit should take care of this lingering concern. 😸

Exirel (approved):

sopel/tools/__init__.py
```diff
@@ -168,10 +169,31 @@ def decorator(func):
...
                 parse_version(warning_in) >= parse_version(__version__)):
-            stderr(text)
-            # Only display the last frame
+            original_frame = inspect.stack()[-stack_frame]
+            mod = inspect.getmodule(original_frame[0])
+            module_name = None
+            if mod:
+                module_name = mod.__name__
+            if module_name:
+                if 'sopel.' in module_name:
```
Exirel, line 178: I didn't realized that first but I think this should be better:

```suggestion
                if module_name.startswith('sopel.'):
```

dgw: Oh yes, this is fair. It's probably what I intended, lol. Will apply this _and_ tweak the `.replace()` below so it only removes `sopel_modules.` from the package name if that's how it starts.

Exirel: All good!

Exirel:
Looks like it's ready to squash&go!

PR #2060, opened 2021-05-05
currency: API switch & backup option

dgw:
### Description
exchangeratesapi.io was bought and started requiring an API key. 🙄

Rather than just replace it with another Fixer-compatible service, I decided to build out the logic for choosing a provider because I found _two_ good replacements (partly thanks to @xnaas). Setting `fixer_io_key` still uses Fixer regardless of the chosen `fiat_provider`, for backward compatibility reasons, but also logs a warning at each rate update.

### Checklist

### Notes
Every time I touch the VCR.py cassettes, I worry that something about the format will be incompatible with one of the older Python versions and break CI. We'll see. If something does break, I'll try re-recording with py3.3 (which I can run locally, somehow).

### DIFF
--- a/sopel/modules/currency.py
+++ b/sopel/modules/currency.py
@@ -20,7 +20,7 @@ from sopel.config import types
 
 
 PLUGIN_OUTPUT_PREFIX = '[currency] '
-FIAT_URL = 'https://api.exchangeratesapi.io/latest?base=EUR'
+FIAT_URL = 'https://api.ratesapi.io/api/latest?base=EUR'
 FIXER_URL = 'https://data.fixer.io/api/latest?base=EUR&access_key={}'
 CRYPTO_URL = 'https://api.coingecko.com/api/v3/exchange_rates'
 EXCHANGE_REGEX = re.compile(r'''
--- a/sopel/modules/currency.py
+++ b/sopel/modules/currency.py
@@ -20,8 +20,11 @@ from sopel.config import types
 
 
 PLUGIN_OUTPUT_PREFIX = '[currency] '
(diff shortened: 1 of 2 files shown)


sopel/modules/currency.py
```diff
@@ -176,20 +196,32 @@ def update_rates(bot):
...
-        response = requests.get(FIXER_URL.format(bot.config.currency.fixer_io_key))
+        LOGGER.debug('Updating fiat rates from Fixer.io')
+        if bot.config.currency.fiat_provider != 'fixer.io':
+            # Warning about future behavior change
+            LOGGER.warning(
+                'fiat_provider is set to %s, but Fixer.io API key is present; '
+                'using Fixer anyway. In Sopel 8, fiat_provider will take precedence.',
+                bot.config.currency.fiat_provider)
+
+        response = requests.get(FIAT_PROVIDERS['fixer.io'].format(bot.config.currency.fixer_io_key))
```
Exirel, line 218: I hope FixerIO keys are always url-safe, otherwise this might fail.

Exirel (commented):
Ha, the good old VCR and old Python version.

sopel/modules/currency.py
dgw: I just did exactly the same thing as what was there before.
It used to be `response = requests.get(FIXER_URL.format(bot.config.currency.fixer_io_key))`, same potential problem. Nobody has complained yet 🤷‍♂️😅

Exirel: Ha! Fair enough.

Exirel (approved):

PR #2062, opened 2021-05-07
sopel: drop support for Python < 3.6

### LINKED ISSUE #1801: Python2 Deprecation
- [x] `setup.py` manual py2 version checks
- [x] `setup.cfg` version restrictions
- [x] `sopel/cli/run.py` py2 version check "sopel 8.0"
- [x] remove unnecessary `__future__`s
- [x] add available `__future__`s
- [x] remove `str = unicode`s
- [x] remove `checkstyle.sh` py3_unicode check
- [x] audit `sys.version_info` uses
- [x] replace `tools.iteritems(x)` uses with `x.items()`
- [x] search for "py2"/"python2" comments

Exirel:
### Description

Fix #1801 and hopefully nothing else so @dgw can have a good time removing deprecated stuffs himself. :grin: 

### Checklist

### DIFF
--- a/checkstyle.sh
+++ b/checkstyle.sh
@@ -1,30 +1,10 @@
 #!/bin/sh
 
-find_source_files() {
-    find . -name '*.py' -size +0 -print | grep -ve './docs' -e 'env' -e './contrib' -e './conftest.py'
-}
-files=$(find_source_files)
 # For now, go through all the checking stages and only die at the end
 exit_code=0
 
 if ! flake8; then
     echo "ERROR: flake8 does not pass."
-    exit_code=1
-fi
-
-# Find files which use the unicode type but (heuristically) don't make it py3
-# safe
-fail_py3_unicode=false
-for file in $(find_source_files); do
-    if grep -qle 'unicode(' -e 'class .*(unicode)' $file; then
-        if ! grep -ql 'unicode = str' $file; then
-            echo "Suspicious 'unicode' use: $file"
-            fail_py3_unicode=true
-        fi
-    fi
-done
-if $fail_py3_unicode; then
-    echo "ERROR: Above files use unicode() but do not make it safe for Python 3."
     exit_code=1
 fi
 
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -1,31 +1,13 @@
-coveralls<2.0; python_version < '3.5'
-coveralls>=2.0; python_version >= '3.5'
-flake8<3.6.0; python_version == '3.3'
-flake8>=3.7.0,<3.8.0; python_version != '3.3'
+coveralls>=2.0
+flake8>=3.7.0,<3.8.0
 flake8-coding
 flake8-future-import<0.4.6
-flake8-import-order; python_version > '3.3'
-flake8-import-order<=1.18.1; python_version <= '3.3'
-# transitive dependency of Sphinx
-# added f-strings in 2.x, but it's not worth doing a whole suite of version markers
-MarkupSafe<2.0
-pytest<3.3; python_version == '3.3'
-pytest>=4.6,<4.7; python_version != '3.3'
-pytest-vcr==1.0.2; python_version != '3.3'
-pytest-vcr==0.3.0; python_version == '3.3'
-PyYAML<5.1; python_version == '3.3'
-PyYAML<5.3; python_version == '3.4'
+flake8-import-order
+pytest>=4.6,<4.7
+pytest-vcr==1.0.2
 requests-mock==1.9.1
-setuptools<40.0; python_version == '3.3'
-# use Sphinx 3.x until dev begins on Sopel 8 and we drop the dead snakes
-sphinx<4.0
-# autoprogram extension added type annotations in 0.1.6
-# such annotations require Python 3.5
-sphinxcontrib-autoprogram<0.1.6; python_version < '3.5'
+sphinx
 # further constrain autoprogram version because the new (in 2021) maintainer
 # already demonstrated a willingness to make major changes in patch versions
-sphinxcontrib-autoprogram<=0.1.7; python_version >= '3.5'
-vcrpy==2.1.1; python_version == '2.7'
-vcrpy<1.12.0; python_version == '3.3'
-vcrpy<2.1.0; python_version == '3.4'
-vcrpy<3.0.0; python_version >= '3.5'
+sphinxcontrib-autoprogram<=0.1.7
+vcrpy<3.0.0
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -1,7 +1,7 @@
 coveralls>=2.0
-flake8>=3.7.0,<3.8.0
+flake8
 flake8-coding
-flake8-future-import<0.4.6
+flake8-future-import
 flake8-import-order
 pytest>=4.6,<4.7
 pytest-vcr==1.0.2
--- a/pytest_run.py
+++ b/pytest_run.py
@@ -11,7 +11,7 @@ Licensed under the Eiffel Forum License 2.
 
 https://sopel.chat
 """
-from __future__ import absolute_import, division, print_function, unicode_literals
+from __future__ import generator_stop
 
 if __name__ == "__main__":
     import sys
--- a/pytest_run.py
+++ b/pytest_run.py
@@ -1,5 +1,4 @@
 #!/usr/bin/env python
-# coding=utf-8
 """This is a script for running pytest from the command line.
 
 This script exists so that the project directory gets added to sys.path, which
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,23 +1,7 @@
-xmltodict<0.12.0; python_version == '3.3'
-xmltodict==0.12; python_version != '3.3'
+xmltodict==0.12
 pytz
 praw>=4.0.0,<6.0.0
-# transitive dependency of praw; v0.18 introduced f-string syntax
-update-checker<0.18; python_version < '3.6'
-geoip2<3.0; python_version <= '3.5' and python_version != '2.7'
-geoip2>=3.0,<4.0; python_version == '2.7'
-geoip2>=4.0,<5.0; python_version >= '3.6'
-# transitive dependency of geoip2; v2 dropped py2.7 & py3 < 3.6
-maxminddb<2.0; python_version < '3.6'
-ipaddress<2.0; python_version < '3.3'
+geoip2>=4.0,<5.0
 requests>=2.0.0,<3.0.0
-# transitive dependency of requests
-# 2.0 will drop EOL Python 2.7 & 3.5, just like Sopel 8 plans to
-urllib3<1.27; python_version != '3.3' and python_version != '3.4'
-urllib3<1.23; python_version == '3.3'
-urllib3<1.25; python_version == '3.4'
-dnspython<2.0; python_version == '2.7'
-dnspython<1.16.0; python_version == '3.3'
-dnspython<3.0; python_version >= '3.4'
-sqlalchemy<1.3; python_version == '3.3'
-sqlalchemy<1.4; python_version != '3.3'
+dnspython<3.0
+sqlalchemy<1.4
--- a/setup.cfg
+++ b/setup.cfg
@@ -16,17 +16,14 @@ classifiers =
     License :: Eiffel Forum License (EFL)
     License :: OSI Approved :: Eiffel Forum License
     Operating System :: POSIX :: Linux
-    Programming Language :: Python :: 2.7
-    Programming Language :: Python :: 3.3
-    Programming Language :: Python :: 3.4
-    Programming Language :: Python :: 3.5
     Programming Language :: Python :: 3.6
     Programming Language :: Python :: 3.7
     Programming Language :: Python :: 3.8
+    Programming Language :: Python :: 3.9
     Topic :: Communications :: Chat :: Internet Relay Chat
 
 [options]
-python_requires = >=2.7, !=3.0.*, !=3.1.*, !=3.2.*, <4
+python_requires = >=3.6, <4
 packages = find:
 zip_safe = false
 
--- a/setup.cfg
+++ b/setup.cfg
@@ -61,4 +61,4 @@ exclude =
     env/*,
     contrib/*,
     conftest.py
-accept-encodings = utf-8
+no-accept-encodings = True
--- a/setup.cfg
+++ b/setup.cfg
@@ -53,14 +53,10 @@ ignore =
     # These are forbidding certain __future__ imports. The future-import plugin
     # has errors both for having and not having them; we want to have these until
     # Sopel no longer supports Python versions that require them.
-    FI50,FI51,FI53,FI54,FI55,
+    FI55,
     # These would require future imports that are not needed any more on Sopel's
-    # oldest supported Python version (2.7).
-    FI12,FI16,FI17,
-    # We ignore the error for missing generator_stop because it's only available
-    # in Python 3.5+ (switch this to FI55 in the above list when Sopel drops
-    # support for Python older than 3.5)
-    FI15,
+    # oldest supported Python version (3.6).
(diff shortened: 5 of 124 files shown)


dgw:
This will have to drop CI for the older Pythons too. Not going to worry about it until 7.1 is done.

LGTM's new alerts are interesting. Annoying, but interesting. I thought we were importing `division` from the future to avoid such things.

Exirel:
> This will have to drop CI for the older Pythons too. Not going to worry about it until 7.1 is done.

There are many other things to do, yes. In this PR, I focused only on removing/adapting the code, but I tried not to touch anything that would be used by the CI or the githook stuff.

> I thought we were importing `division` from the future to avoid such things.

Which is not necessary with Python 3 as the [true division is now by default](https://docs.python.org/fr/3.6/library/__future__.html). I'm pretty sure it's because LGTM is configured with Py2.7.

Also, It's funny to see that the CI fails on Py2.7 to 3.5, and is fine for 3.6+—just as planned!

dgw:
> I tried not to touch anything that would be used by the CI or the githook stuff.

> Also, It's funny to see that the CI fails on Py2.7 to 3.5, and is fine for 3.6+—just as planned!

Leaving the CI stuff, at least, means this PR can never pass checks. Passing CI is required to merge. Dropping support for a Python version _should_ include removing tests for that version, because we shouldn't waste resources testing something unsupported.

And speaking of unsupported things, the githooks are `contrib` stuff, so it's fine not to worry about those. Doesn't look like @HumorBaby included anything Python-version-specific in those, though.

> I'm pretty sure it's because LGTM is configured with Py2.7.

It's probably because we still have Python 2.7 in CI and/or requirements files. LGTM assumes Python 3 if it doesn't see any versions specified; see https://lgtm.com/help/lgtm/python-extraction#python_setup-step

Exirel:
Yup yup, I agree with that. Which makes me realized that I should have opened that as a DRAFT and not a ready PR, my mistake!

My reasoning about CI stuff was to give you the code first, and see what needed to be done exactly for the CI—something I know too little about at the moment—and to do that with you. So yeah, absolutely, the CI need to be modified.

My guess is that you're probably too busy to look at that now, and I'm sure it's better to wait after 7.1 is released than right now.

dgw:
Dropped back to draft status. You're correct, I'd rather spend my time on 7.1 stuff now, and tackle this after it's released. 😸

checkstyle.sh
```diff
@@ -12,20 +8,4 @@ if ! flake8; then
     exit_code=1
 fi
 
-# Find files which use the unicode type but (heuristically) don't make it py3
-# safe
```
dgw, line 16: What would you say to making this into a check for using `unicode` _at all_? It should all be `str` now in py3-only code.

pytest_run.py
```diff
@@ -11,7 +11,7 @@
 
 https://sopel.chat
 """
-from __future__ import absolute_import, division, print_function, unicode_literals
+from __future__ import generator_stop
```
dgw, line 14: I am not going to mark every one of these, but what's the chance of subtle behavior changes from adding this? (Yes, I realize it's "mandatory" on three of our four currently-supported Python versions, 3.7+.)

setup.cfg
```diff
@@ -16,17 +16,13 @@ classifiers =
     License :: Eiffel Forum License (EFL)
     License :: OSI Approved :: Eiffel Forum License
     Operating System :: POSIX :: Linux
-    Programming Language :: Python :: 2.7
-    Programming Language :: Python :: 3.3
-    Programming Language :: Python :: 3.4
-    Programming Language :: Python :: 3.5
     Programming Language :: Python :: 3.6
     Programming Language :: Python :: 3.7
     Programming Language :: Python :: 3.8
```
dgw, line 21:
```suggestion
    Programming Language :: Python :: 3.8
    Programming Language :: Python :: 3.9
```

Might as well.

sopel/__init__.py
```diff
@@ -35,12 +35,11 @@
...
 loc = locale.getlocale()
-if sys.version_info.major > 2:
-    if not loc[1] or 'UTF-8' not in loc[1]:
-        print('WARNING!!! You are running with a non-UTF8 locale environment '
-              'variables (e.g. LC_ALL is set to "C"), which makes Python 3 do '
-              'stupid things. If you get strange errors, please set it to '
-              'something like "en_US.UTF-8".', file=sys.stderr)
+if not loc[1] or 'UTF-8' not in loc[1]:
+    print('WARNING!!! You are running with a non-UTF8 locale environment '
+          'variables (e.g. LC_ALL is set to "C"), which makes Python 3 do '
```
dgw, line 40:
```suggestion
          'variable (e.g. LC_ALL is set to "C"), which makes Python 3 do '
```

So this typo has been here for _quite_ a while.

sopel/tools/__init__.py
```diff
@@ -31,16 +31,13 @@
...
-    iteritems = dict.items
-    itervalues = dict.values
-    iterkeys = dict.keys
-else:
-    iteritems = dict.iteritems
-    itervalues = dict.itervalues
-    iterkeys = dict.iterkeys
+
+# Kept for backward compatibility
+# TODO: consider removing that
```
dgw, line 36: For the record, I would be quite happy to deprecate these, as they have no place in Sopel's `tools` module.

It could be done in a separate PR though, and if you agree just lmk and I'll add it to a task list somewhere.

setup.py
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env python
 # coding=utf-8
```
dgw, line 2: As a side note, since UTF-8 is the default in Python 3, we don't need these encoding cookies any more.

sopel/config/__init__.py
```diff
@@ -48,19 +48,16 @@ class SpamSection(config.types.StaticSection):
...
 from sopel import tools
 from . import core_section, types
 
-if sys.version_info.major < 3:
-    import ConfigParser
-else:
-    basestring = str
-    import configparser as ConfigParser
+
+basestring = str
```
dgw, line 60: Why keep this?

sopel/modules/ip.py
```diff
@@ -8,31 +8,23 @@
...
-from __future__ import absolute_import, division, print_function, unicode_literals
+from __future__ import generator_stop
 
 import logging
 import os
 import socket
 import tarfile
+# urlretrieve has been put under urllib.request in Python 3.
+# It's also deprecated so this should probably be replaced with
+# urllib2.
```
dgw, line 19:
```suggestion
```

This comment is somewhat mistaken. `urllib.request.urlretrieve()` is part of the Legacy Interface, and "might become" deprecated in the future but still is not marked so [even in Python 3.10](https://docs.python.org/3.10/library/urllib.request.html#urllib.request.urlretrieve) (so far). There's also no such thing as a `urllib2` module in Python 3.9 or 3.10 (I looked).

Might as well remove the misleading stuff while we're already in here cleaning up.

sopel/test_tools.py
```diff
@@ -17,11 +17,10 @@
...
 # Licensed under the Eiffel Forum License 2.
-from __future__ import absolute_import, division, print_function, unicode_literals
+from __future__ import generator_stop
 
 import os
 import re
-import sys
 import tempfile
 
 try:
```
dgw, line 26: Stupid proximity limits on line notes prevent me from providing a full suggestion, but this nonsense can be dropped in favor of a straight `import configparser` above, as it was in `sopel/config/__init__.py`.

sopel/tools/web.py
```diff
@@ -154,13 +141,10 @@ def unquote(string):
...
-        This is a shim to make writing cross-compatible plugins for both
-        Python 2 and Python 3 easier.
+
+        This is a convenient shortcut for ``urllib.parse.unquote``.
     """
-    if sys.version_info.major < 3:
-        return urllib.unquote(string.encode('utf-8')).decode('utf-8')
-    else:
-        return urllib.parse.unquote(string)
+    return urllib.parse.unquote(string)
```
dgw, line 147:
```suggestion
    # TODO deprecated?
    return urllib.parse.unquote(string)
```

dgw (changes requested):
Surprisingly few line notes, considering the scope of this PR. Given that it's mostly deleting dead code and swapping imports, though, it's actually not _that_ surprising after all.

Hopefully LGTM will figure out that we don't care about py2 any more now that CI and everything has been reworked to target only 3.6+, and stop giving irrelevant warnings.

pytest_run.py
Exirel: I didn't know before looking into it, so here is the result of my research:

> The interaction of generators and StopIteration is currently somewhat surprising, and can conceal obscure bugs. [...] Currently, StopIteration raised accidentally inside a generator function will be interpreted as the end of the iteration by the loop construct driving the generator.

This quote is from the [PEP 479](https://www.python.org/dev/peps/pep-0479/) (you can read it for more details). The idea is that the `__future__` import prevents to stop a loop with a `StopIteration` exception within a generator: it'll be raised as a normal exception and handled up in the stack.

sopel/config/__init__.py
Exirel: I bet I forgot that one!

checkstyle.sh
Exirel: After thinking about it, I don't think it's important. Who will use `unicode` instead of ` str`? There is no `unicode` built-in anymore, and I don't see someone using it with no one noticing in review. I'm more worried of having yet another tool/script that could generate false positives, require maintenance, dependencies, etc.

sopel/tools/__init__.py
Exirel: Yup, agree, later.

checkstyle.sh
dgw: True, I suppose; this only checks _our_ code (core) anyway, and neither of us is going to approve any new patches that override or replace `str`. I'll just give this branch a once-over for any use of `unicode` that's not in a comment.

sopel/tools/__init__.py
dgw: Tossed on my to-deprecate list. 👍

sopel.py
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env python3
 # coding=utf-8
```
dgw, line 2: Missed a coding line.

```suggestion
```

pytest_run.py
```diff
@@ -11,7 +11,7 @@
```
dgw, line 11: Above this, in too-far-from-the-diff-to-make-suggestions territory, is a `coding=utf-8` line that should be removed.

setup.py
dgw: Missed removing this one, along with a couple others in `sopel.py` and `pytest_run.py` noted separately, and there's a different style of encoding cookie in `docs/source/conf.py` that could be deleted.

Several test files also define a `MOCK_MODULE_CONTENT` or similar that contains the coding cookie.

setup.cfg
```diff
@@ -50,17 +47,15 @@ ignore =
     W503,
     # Newline after binary operator. Ignored by default (which we want to keep)
     W504,
+    # No need for coding=utf-8 anymore
+    C101
```
dgw, line 51: Instead of this, change `accept-encodings: utf-8` below to [`no-accept-encodings: True`](https://github.com/tk0miya/flake8-coding#no-accept-encodings). That will also suss out any that were missed, except for the tests that include an encoding cookie as part of a mock module.

dgw:
I'm baffled after studying [LGTM's docs about Python version selection](https://lgtm.com/help/lgtm/analysis-faqs#how-python-version-identified). It _should_ default to Python 3, but isn't. Let's ignore the new alerts on this PR and look into it separately if they still appear on `master` after merging.

half-duplex:
Depending on the scope of this PR...
* requirements.txt and dev-requirements.txt should have the `python_version == '2.7'` etc lines removed. This also "fixes" CVE-2021-33503 (DoS), which sopel 7 will remain vulnerable to.
* docs/source/configuration.rst has a comment about py2.7
* README.rst says "Sopel requires Python 2.7.x or Python 3.3+ to run", etc
* The shebang of sopel/cli/run.py specifically uses 2.7 (dear god, why? does this even need a shebang?)
* sopel/formatting.py TODO
* conftest.py 2x "Python 3.3" comments
* sopel/plugins/rules.py has a 2.7 TODO
* sopel/irc/backends.py SSLAsynchatBackend.handle_connect

Exirel:
> Depending on the scope of this PR...

Indeed, more work to do! I've done what I could or think is important right now.

I left aside the TODOs in `sopel.plugins` (there are many!), as they will be dealt with in a future PR to rework/replace what was kept for older versions of Python.

dev-requirements.txt
```diff
@@ -1,31 +1,16 @@
...
+coveralls>=2.0
+flake8>=3.7.0,<3.8.0
 flake8-coding
 flake8-future-import<0.4.6
-flake8-import-order; python_version > '3.3'
-flake8-import-order<=1.18.1; python_version <= '3.3'
+flake8-import-order
 # transitive dependency of Sphinx
 # added f-strings in 2.x, but it's not worth doing a whole suite of version markers
 MarkupSafe<2.0
```
dgw, line 8: Shouldn't need this override any more.

requirements.txt
```diff
@@ -1,23 +1,11 @@
-xmltodict<0.12.0; python_version == '3.3'
-xmltodict==0.12; python_version != '3.3'
+xmltodict==0.12
 pytz
 praw>=4.0.0,<6.0.0
 # transitive dependency of praw; v0.18 introduced f-string syntax
```
dgw, line 4: This was about `update-checker`.

```suggestion
```

requirements.txt
```diff
@@ -1,23 +1,11 @@
...
-update-checker<0.18; python_version < '3.6'
-geoip2<3.0; python_version <= '3.5' and python_version != '2.7'
-geoip2>=3.0,<4.0; python_version == '2.7'
 geoip2>=4.0,<5.0; python_version >= '3.6'
```
dgw, line 5:
```suggestion
geoip2>=4.0,<5.0
```

requirements.txt
```diff
@@ -1,23 +1,11 @@
...
 # transitive dependency of geoip2; v2 dropped py2.7 & py3 < 3.6
```
dgw, line 6: This is about `maxminddb`, now removed from the file.

```suggestion
```

requirements.txt
```diff
@@ -1,23 +1,11 @@
...
-maxminddb<2.0; python_version < '3.6'
-ipaddress<2.0; python_version < '3.3'
 requests>=2.0.0,<3.0.0
 # transitive dependency of requests
```
dgw, line 8: Don't need to override it (`urllib3`) any more.

```suggestion
```

requirements.txt
```diff
@@ -1,23 +1,11 @@
...
-# 2.0 will drop EOL Python 2.7 & 3.5, just like Sopel 8 plans to
 urllib3<1.27; python_version != '3.3' and python_version != '3.4'
```
dgw, line 9: Don't need to override it (`urllib3`) any more.

```suggestion
```

sopel.py
```diff
@@ -1,6 +1,5 @@
-#!/usr/bin/env python3
-# coding=utf-8
-from __future__ import absolute_import, division, print_function, unicode_literals
+#!/usr/bin/env python
```
dgw, line 1: Is this change wise? Plenty of distros still default to Python 2 if `python3` isn't specified.

dgw (commented):
Some more stuff, mostly about requirements. Very close!

sopel.py
Exirel: Not safe, reverted.

Exirel:
Here we go. I haven't rebased/squashed yet. Other than that, it's ready for final review: everything new should go in new PRs, I think.

dgw (approved):
Squash the fixups and let's ship it. "Don't sweat the little stuff," etc., there's plenty of time to catch the stragglers and handle the various TODOs.

PR #2063, opened 2021-05-09
isup: fix double output, and add tests

### LINKED ISSUE #2061: isup responds twice
### Description
when I run .isup or .isupinsecure, the bot responds twice. first is the error (if any) then it says that the site is up
### Reproduction steps
1. setup a sopel bot using the master branch.
2. in the irc channel run .isup with some url that is down
3. it responds twice

### Expected behavior
only responds with an error or else up, not up and error

### Environment
- Sopel `.version`: [e.g. 7.0.0 or d416e19] master branch
- Sopel installed via: [apt, pip, `setup.py install`, source, ?] source
- Python version: [e.g. 3.6.9] 3.7
- Operating system: [e.g. Debian 10] debian buster
- IRCd `/version`: [e.g. InspIRCd 3.0.1] freenode
- Relevant plugins: [adminchannel, weather, custom\_thing.py, ?] isup

### Notes
seems to be because in https://github.com/sopel-irc/sopel/blob/master/sopel/modules/isup.py#L89 none of the except statements return and so the bot.say for "website is up" is always executed.

dgw:
### Description
Fixes #2061, a regression introduced in 5af0eea249238485d70c8e14743358a50115b0f4.

### Checklist


Exirel (changes requested):
I think it would be better to use `else`:

```python
try:
    # do something
except:
    # bot.reply error
except:
    # bot.reply a different error
else:
    # bot.say all is fine
```

dgw:
@Exirel Yeah, per our IRC chatter I did that locally. Haven't pushed yet; sparing Travis until I finish writing the test suite.

Exirel (approved):
Is it possible to refactor mocks to reduce the duplicate code in tests? Yes, for sure. Is it necessary to approve that PR? No, it's fine like it is.

Well done on your testing journey, I hope it was enjoyable. As discussed, some tweaks are required in the future to make it even easier, and I probably need to write some more doc about it.

For now, this is a go!

dgw:
@Exirel Reimplemented fixtures, along with a much more interesting use of `bot.running_triggers` that I think you'll appreciate:

https://github.com/sopel-irc/sopel/blob/559987df678d3f09da05173cec3c181d12134d88/sopel/bot.py#L959-L962

Early in the process, before figuring out that threading was the problem, I wrongly theorized that some multiple-instance voodoo might be messing up which `bot` (or at the time, `mockbot`) got the message vs. which one I checked for `message_sent`. Since my goal was to get _one_ test working before writing the rest based on it, I removed the fixtures just in case. Clearly, it wasn't necessary, but at the time it seemed like eliminating as many pseudo-magical bits of `pytest` as possible would help me track down why a seemingly innocuous example directly from one of your own docstrings simply refused to work. 😅

Exirel:
Even better now!

PR #2065, opened 2021-05-12
tests: `MockIRCServer` methods should block by default

dgw:
### Description
tl;dr: Threaded callables can cause race conditions during tests, but the `MockIRCServer` is in a perfect position to make sure they don't (and it now does so).

I had a very interesting time this past weekend, trying to write tests for #2063 that simply wouldn't work, no matter what I did, until eventually the idea came to me: "Try it with the plugin callable set to unthreaded." After that, and an IRC discussion wherein @Exirel reminded me that the `bot.running_triggers` property exists, those tests got written and merged.

Now this new feature inspired by the above process is ready to be "formally" proposed. I spent some time on the documentation for it, proved the concept by running the test suite on _three_ local Python versions (2.7, 3.3, and 3.8), and am reasonably confident that this is ready for the "bikeshedding" or "nitpicking" stage—that is, developed enough as a concept that I won't need to rewrite the whole patch based on feedback. 😉

I elected to go with the simpler Boolean argument value for the overriding (`blocking`) argument on each method. Also proposed was a numeric value that would allow a test writer to specify the `timeout` for `threading.Thread.join()`, but that seems needlessly complex. I can't see a use for tests where a call would _sometimes_ block and _sometimes_ time out—such a test would be prone to intermittent failures, and thus not very useful for CI. 😁

### Checklist

### Notes
Of course, if I've made any silly assumptions about things like never needing to set a `timeout` on `threading.Thread.join()` sometimes, that's what the PR discussion is for!

The `versionadded` directives are committed separately, so I can easily update them if for some reason this doesn't go into 7.1. Otherwise, I plan to squash them into the relevant code commits before merging.

### DIFF
--- a/sopel/tests/factories.py
+++ b/sopel/tests/factories.py
@@ -34,7 +34,7 @@ class BotFactory(object):
         every other plugin from ``preloads``::
 
             factory = BotFactory()
-            bot = factory.with_autoloads(settings, ['emoticons', 'remind'])
+            bot = factory.preloaded(settings, ['emoticons', 'remind'])
 
         .. note::
 
--- a/sopel/tests/factories.py
+++ b/sopel/tests/factories.py
@@ -108,8 +108,8 @@ class IRCFactory(object):
         The :func:`~sopel.tests.pytest_plugin.ircfactory` fixture can be used
         to create this factory.
     """
-    def __call__(self, mockbot):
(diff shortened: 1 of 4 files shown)


Exirel (approved):
Ship it.

PR #2014, opened 2021-01-09
find: format replacement text instead of "meant"

cottongin:
### Description
This simply applies the bold formatting to whatever the new replacement text is rather than the "meant" text.

### Checklist
- [x] I have read [CONTRIBUTING.md](https://github.com/sopel-irc/sopel/blob/master/CONTRIBUTING.md)
- [x] I can and do license this contribution under the EFLv2
- [x] No issues are reported by `make qa` (runs `make quality` and `make test`)
- [x] I have tested the functionality of the things this change touches

PR #2066, opened 2021-05-15
find: don't output useless bold formatting for empty replacement

dgw:
### Description
Tin. I did a `s/something //` replacement to delete an extra word recently and realized that the line contained two useless bold formatting bytes anyway, because that's how the logic was written and I didn't think of this case during review of #2014.

### Checklist

### DIFF
--- a/sopel/modules/find.py
+++ b/sopel/modules/find.py
@@ -163,9 +163,13 @@ def findandreplace(bot, trigger):
 
     sep = trigger.group('sep')
     old = trigger.group('old').replace('\\%s' % sep, sep)
-    new = bold(trigger.group('new')).replace('\\%s' % sep, sep)
+    new = trigger.group('new')
     me = False  # /me command
     flags = trigger.group('flags') or ''
+
+    # only clean/format the new string if it's non-empty
+    if new:
+        new = bold(new.replace('\\%s' % sep, sep))
 
     # If g flag is given, replace all. Otherwise, replace once.
     if 'g' in flags:


sopel/modules/find.py
```diff
@@ -163,10 +163,14 @@ def findandreplace(bot, trigger):
 
     sep = trigger.group('sep')
     old = trigger.group('old').replace('\\%s' % sep, sep)
-    new = bold(trigger.group('new')).replace('\\%s' % sep, sep)
+    new = trigger.group('new')
     me = False  # /me command
     flags = trigger.group('flags') or ''
 
+    # only clean/format the new string if it's non-empty
+    if new:
```
Exirel, line 171: I assume you don't want to have `    ` between bold either? Then I'd strip the string just for the condition:

```suggestion
    if new.strip():
```

Exirel (changes requested):
Yeah OK it's a good idea.

sopel/modules/find.py
dgw: I should test this with formatting codes in the replacement. IIRC `str.strip()` will remove some (but not all) of those.

dgw: As I thought:

```pycon
>>> from sopel import formatting
>>> for f in formatting.CONTROL_FORMATTING:
...     print(f.encode('unicode_escape'), f.strip().encode('unicode_escape'))
...
b'\\x0f' b'\\x0f'
b'\\x03' b'\\x03'
b'\\x04' b'\\x04'
b'\\x02' b'\\x02'
b'\\x1d' b''
b'\\x1f' b''
b'\\x1e' b''
b'\\x11' b'\\x11'
b'\\x16' b'\\x16'
```

So here again we run into the same issue with `str.strip()` that led to [the `_format_safe()` method in `choose`](https://github.com/sopel-irc/sopel/blob/b04c285d4fd780f8a70d738ad556093ac327770c/sopel/modules/choose.py#L26): its default set of characters to strip will remove _some_ (but not all) IRC formatting bytes. Of course it's not being used for output, but it would still cause some inconsistent behavior in the `find` plugin's output with certain user inputs.

Perhaps a better approach would be to skip applying bold formatting if the replacement contains _any_ formatting bytes? Otherwise yes, stripping it of whitespace would be OK here.

(Of course the idea of a monkey-patch to `str` here that adds some `str.safe_strip()` method crossed my mind… but there are _many_ reasons that's a bad idea, chief among them being that Sopel shouldn't mess with stdlib types.)

dgw:
@Exirel I don't mind having whitespace- or control-code-only changes as edge cases. For the intended class of uses (deleting repeated words or other extra characters), this works just fine, and doesn't _introduce_ any bold formatting where it wasn't already.

Exirel (approved):
Alright, let's move on!

sopel/modules/find.py
Exirel: I guess edge cases of an edge case are too edgy for today.

PR #2071, opened 2021-05-18
loader: use `inspect` to make callable docs consistent

dgw:
### Description
Our own `trim_docstring` function is OK, but definitely not perfect. What it returns can be inconsistently indented depending on whether the docstring passed in is single- or multi-line, or whether a multi-line docstring starts on line 0 or line 1 (next to or below the `"""`). It's nice to have consistent indentation.

I don't have any strong feelings about where to implement this. Maybe it would make more sense to apply `inspect.cleandoc()` inside the logic of `trim_docstring` instead of using `inspect.getdoc()` on the callable. Either way, our help infrastructure will be easier to improve if we eliminate anything like this that can cause inconsistent output.

### Checklist
  - ~~Not tested yet (see notes)~~ CI did its job
  - Well, I did when I wrote the patch last year…

### Notes
I know this has a merge conflict; it's an old branch I found while cleaning up now that most of 7.1.0's stuff is merged. I'll take care of that (or overwrite the branch with a different approach) when the time comes.

### DIFF
--- a/sopel/loader.py
+++ b/sopel/loader.py
@@ -12,12 +12,13 @@
 """
 from __future__ import absolute_import, division, print_function, unicode_literals
 
+import inspect
 import logging
 import re
 import sys
 
 from sopel.config.core_section import COMMAND_DEFAULT_HELP_PREFIX
-from sopel.tools import itervalues
+from sopel.tools import deprecated, itervalues
 
 
 if sys.version_info.major >= 3:
@@ -27,6 +28,11 @@ if sys.version_info.major >= 3:
 LOGGER = logging.getLogger(__name__)
 
 
+@deprecated(
+    reason="Replaced by simple logic using inspect.getdoc()",
+    version='7.1',
+    removed_in='8.0',
+)
 def trim_docstring(doc):
     """Get the docstring as a series of lines that can be sent.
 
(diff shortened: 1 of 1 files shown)


sopel/loader.py
```diff
@@ -44,7 +45,7 @@ def clean_callable(func, config):
     prefix = config.core.prefix
     help_prefix = config.core.help_prefix
     func._docs = {}
-    doc = trim_docstring(func.__doc__)
+    doc = trim_docstring(inspect.getdoc(func))
```
dgw, line 48: @Exirel, you think this is fine for 7.1? I'd also deprecate `trim_docstring()` for 7.1, and remove it in 8.0.

```suggestion
    doc = inspect.getdoc(func)
```

Exirel: With a bit more code, yes:

* `inspect.getdoc(func)` will return `None` if `func` doesn't have a docstring
* `inspect.getdoc(func)` will return a string, while `trim_docstring` will return a list

However, since `getdoc` (using `cleandoc`) will remove empty line before/after the docstring, you shouldn't need to use `trim_docstring` at all, and do this:

```suggestion
    doc = []
    docstring = inspect.getdoc(func)
    if docstring:
        doc = docstring.splitlines()
```

dgw: Adopted this approach in a70b1e6f653d4cac9ef8964d60b01943d284068f.

Exirel (approved):
Nice.

I had a look at our test suite, which checks what should be in `func._docs` and since they pass, I think we are all good.

Let's go, this is the last one before 7.1!

PR #2072, opened 2021-05-22
Substitute Libera Chat for all public-facing references to freenode

dgw:
### Description
Tin.

I elected to leave the tests alone, since those could just as well use `irc.example.com` and work exactly the same way.

Updated the NEWS entry for v6.6.0 just so future release archives and changelog pages on the website won't point people to an abandoned channel on freenode. Probably could leave it alone, but it's easy to update along with everything else.

### Checklist

### DIFF
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -1,5 +1,5 @@
 blank_issues_enabled: true
 contact_links:
   - name: Sopel's IRC channel
-    url: https://webchat.freenode.net/#sopel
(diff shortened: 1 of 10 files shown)


Exirel (approved):
If that's all the occurrence, then it's good for me!

dgw:
> If that's all the occurrence

tbh I was rather hoping someone would double-check my search results in case I missed something outside of `test/` 😁

Exirel:
Alright, I double checked locally, and I can't find more (except, as mentioned in your description, in the test folder but those are fine for now).

So: 🚢 🇮🇹

PR #2076, opened 2021-05-27
tld: ccTLD data moved to its own wiki page

dgw:
### Description
Instead of fetching one big wiki page with everything in it, we have to fetch the gTLD tables from the same place as before, and then get the ccTLD lists separately.

This _could be_ the first of several splits, since the "List of Internet top-level domains" page has a maintenance tag at the top decrying that it's long and hard to navigate. I've made it rather easy to add more pages if the different types of TLD do continue getting split out into their own pages.

#### Before/After

SopelTest is this branch; Kaede is current master (12f8640525d2e8eaa716c07abcfafd3a851604a4).

```irc
<~dgw> ;tld md
<&Kaede> [tld] The top-level domain 'md' exists, but no details about it could be found.
<~dgw> ,tld md
<SopelTest> [tld] Name: .md | Entity: Moldova | DNSSEC: Yes | SLD: Yes | Notes: Restricted to individuals
            or companies with a physical address in Moldova.
```

### Checklist

### DIFF
--- a/sopel/modules/tld.py
+++ b/sopel/modules/tld.py
@@ -40,8 +40,10 @@ DEFAULT_CACHE_TIME = '2000-01-01 00:00:00'
 DEFAULT_CACHE_DATETIME = _strptime_as_utc(DEFAULT_CACHE_TIME)
 
 IANA_LIST_URI = 'https://data.iana.org/TLD/tlds-alpha-by-domain.txt'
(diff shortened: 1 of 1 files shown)


Exirel (approved):
I haven't tested that myself, the code itself looks sound. So I'm OK with that.

PR #2078, opened 2021-05-30
gha: only run jobs when events are for 'sopel-irc' owned repos

HumorBaby:
### Description
As of right now, the GitHub Actions CI workflow will run on forks too.
This PR checks the workflow context to see if events are for the 'sopel-irc' owner, and only runs jobs if so. This check can/should be moved/updated as new jobs are defined, or if the event/branch filter for the CI workflow changes.

### Checklist
- [x] I have tested the functionality of the things this change touches<sup>1</sup>
> <sup>1</sup>As best as possible. Confirmed that adding this prevents from running on my fork.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,6 +14,7 @@ on:
 
 jobs:
   tests:
+    if: ${{ github.repository_owner == 'sopel-irc' }}
     name: Test suite
     runs-on: ubuntu-latest
     strategy:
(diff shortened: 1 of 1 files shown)


dgw:
You don't think `needs: tests` will block the Coveralls finisher from running without a condition?

Presumably the `pypi.yml` workflow will fail because of the missing secret, but maybe it should also have this condition.

I can't believe there's no way to specify a condition for an entire workflow… Putting it on _every_ job is ridiculous.

dgw (approved):
Really, I just want to merge this and further refine it later if needed. My question(s) above can be answered regardless of whether the PR remains unmerged. 😄

dgw:
Update: Workflow runs are still created on a fork; they just [immediately skip the job](https://github.com/dgw/sopel/actions/runs/891217575). Better than before, though still not ideal.

A true fix probably would have to come from the GitHub Actions team modifying this (asinine) default behavior to be more like a typical CI service—most of which seem to simply ignore forks.

PR #2084, opened 2021-06-02
config: fix BooleanAttribute default=True

half-duplex:
### Description
Makes `default=True` work correctly for `BooleanAttribute`s - see: url.py not working in 7.1.0.

I'm not sure why that stub was overridden, but I can't think of a reason BoolAttr would need to behave differently from the normal `BaseValidated._parse()`. @dgw?

Includes lazy test of default default and default=True. It might belong in test/config/test_config_types.py but that's harder and I'm out of time and brain to melt.

### Checklist

### DIFF
--- a/sopel/config/types.py
+++ b/sopel/config/types.py
@@ -404,9 +404,6 @@ class BooleanAttribute(BaseValidated):
             ]
         return bool(value)
 
-    def _parse(self, value, settings, section):
-        return self.parse(value)
-
     def __set__(self, instance, value):
         if value is None:
             instance._parser.remove_option(instance._section_name, self.name)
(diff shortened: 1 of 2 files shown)


dgw:
We're sending all PRs to `master` for now, and I'm cherry-picking bugfixes like this to `7.1.x` when needed. Will rebase so checks can run.

dgw (approved):
I have tested both the new test (by running it against the current state of `7.1.x`, where it fails) and the new behavior (by checking out this branch, where the new test now passes). LGTM 🚀

> I'm not sure why that stub was overridden, but I can't think of a reason BoolAttr would need to behave differently from the normal `BaseValidated._parse()`. @dgw?

It's a good question, and I don't have an answer. That override was part of #2044, and no one questioned it, so it got merged.

Exirel (approved):
🐑 🇮🇹

PR #2087, opened 2021-06-06
config, db: support `db_url` setting to override the individual options

### LINKED ISSUE #2057: Ability to specify full DB connection URL
### The problem

I'm running sopel on heroku. Heroku manages the database, and provides the connection info as an environment variable `$DATABASE_URL`. This URL has the user/pass/host all baked into it, ready to pass to sqlalchemy. But there's currently no way to tell sopel to just use this. Instead, you have to manually break it into pieces and hard-code each piece in the config file. This is (a) extra work, (b) insecure b/c you have to write down the password in plaintext, (c) liable to break whenever heroku's infrastructure changes.

### The solution

For now, I've hacked up my copy of sopel to use $DATABASE_URL if it exists, and otherwise fall back on the normal config options:

https://github.com/sopel-irc/sopel/compare/master...njsmith:njs-heroku-support

This is maybe a bit *too* simple to be a general solution, but it would be very nice if upstream sopel had something *like* this, so I don't have to maintain a fork :-).

dgw:
### Description
Specifying a `core.db_url` setting overrides the other `core.db_*` settings and just uses the provided URL directly.

Resolves #2057, more or less, though the envvar name isn't exactly what was mentioned in that feature request.

### Checklist
  - With my local `sqlite:` URL, retrieved from `bot.db` via `sopel-ipython`. Presumably if that works, it'll work with MySQL/PostgreSQL/whatever else.

### Notes
It should now be trivial to simply copy the value of the `DATABASE_URL` environment variable on Heroku into the `SOPEL_CORE_DB_URL` variable expected by Sopel, instead of splitting it into components.

Adding support for the envvar name `DATABASE_URL`, if it happens, can be done separately.

### DIFF
--- a/sopel/config/core_section.py
+++ b/sopel/config/core_section.py
@@ -373,6 +373,25 @@ class CoreSection(StaticSection):
 
     """
 
+    db_url = ValidatedAttribute('db_url')
+    """A raw database URL.
+
+    If this option is present, Sopel will ignore **all** other ``db_*``
+    settings and use this option's value only.
+
+    Any supported database type *other than SQLite* can be used this way.
+
+    .. note::
+
+        Specifying this option via the ``SOPEL_CORE_DB_URL`` :ref:`environment
+        variable <Overriding individual settings>` may prove especially useful
+        in certain cloud environments, avoiding the need to split a database
+        URI provided by the platform at runtime into its components with a
+        startup script.
+
+    .. versionadded:: 8.0
+    """
+
     db_user = ValidatedAttribute('db_user')
     """The user for Sopel's database.
 
--- a/sopel/config/core_section.py
+++ b/sopel/config/core_section.py
@@ -379,8 +379,6 @@ class CoreSection(StaticSection):
     If this option is present, Sopel will ignore **all** other ``db_*``
     settings and use this option's value only.
 
-    Any supported database type *other than SQLite* can be used this way.
-
     .. note::
 
         Specifying this option via the ``SOPEL_CORE_DB_URL`` :ref:`environment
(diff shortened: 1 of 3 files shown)


sopel/config/core_section.py
```diff
@@ -373,6 +373,25 @@ class CoreSection(StaticSection):
 
     """
 
+    db_url = ValidatedAttribute('db_url')
+    """A raw database URL.
+
+    If this option is present, Sopel will ignore **all** other ``db_*``
+    settings and use this option's value only.
+
+    Any supported database type *other than SQLite* can be used this way.
```
Exirel, line 382: I'm pretty sure SQLite can be used this way too: `sqlite:///relative/path/to/database.sqlite` should work.

The doc for [`make_url`](https://docs.sqlalchemy.org/en/14/core/engines.html#sqlite) seems to agree with me.

Exirel (changes requested):

sopel/config/core_section.py
dgw: This was definitely left over from a previous revision. Current one is indeed supposed to support SQLite too.

Whoops!

sopel/db.py
```diff
@@ -112,32 +112,36 @@ class SopelDB(object):
...
+            # it would be nice to validate this type name somehow. Shouldn't
+            # affect anything, since the only thing it's ever used for is
+            # checking whether the configured database is 'sqlite'.
+            self.type = self.url.drivername.split('+', 1)[0]
         else:
+            self.type = config.core.db_type
+
             query = {}
-            if self.type == 'mysql':
+            if self.type == 'sqlite':
```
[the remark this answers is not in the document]
dgw: Yeah, that was an awkward part of the previous implementation that I managed to make worse. See new changes and resolve the thread if you like that better!

Exirel (approved):

dgw:
@RustyBower Any chance of getting a sanity check on this from you in the next week or two?

RustyBower:
Actually looking at it right now :)

RustyBower (approved):
LGTM

Issue #2079, opened 2021-05-31
IRCv3 "bot" mode support


dgw:
### Background
Spec: https://ircv3.net/specs/extensions/bot-mode

This extension adds a new ISUPPORT token for servers to advertise a mode that `BOT` clients may set on themselves to indicate their non-human status. Clients with the bot mode enabled are supposed to have their messages tagged with `bot` (well, `draft/bot` until spec ratification) so other clients can act appropriately.

### Plans
I see ~~two~~ _three_ different, but related, things for Sopel to do here:

- [x] Set the bot mode specified in the `BOT` token from ISUPPORT, if given, unless the config explicitly overrides it (can be as simple as looking for the modechar in `core.modes` and skipping this behavior if it's present, [as Limnoria does](https://github.com/ProgVal/Limnoria/blob/711db72ad0224c3f5d094ccd3167a3362a1ce37e/src/irclib.py#L2271-L2272)). [#2088]
- [x] Start ignoring other clients whose messages are tagged as `bot`, so there will be less manual `.blocks add BotNick` work needed from Sopel admins whose instances share channels with other compliant bots and want to minimize the risk of mutual triggers. [#2089]
  - This will also require adding support for the `message-tags` CAP, since the `bot` tag is not sent to clients who have not requested `message-tags`. It shouldn't be much more work than adding it [here](https://github.com/sopel-irc/sopel/blob/269498a8ce29d75cf5235cf21ce3c4a851341154/sopel/coretasks.py#L897-L903), since there's already tag parsing in place for `server-time` and `account-tag` (the specs for which do _not_ depend on `message-tags` being enabled).
- [x] Should probably have a `@plugin.allow_bots` decorator like that allows a callable to receive messages tagged as coming from another bot. [#2244]

dgw:
Oops, yes. Updated the OP, and I guess this is ready to close.

PR #2088, opened 2021-06-06
coretasks: handle `BOT` ISUPPORT token & mode

dgw:
### Description
If the server advertises a `BOT` token in `005 ISUPPORT`, Sopel will set that mode on itself unless that modechar appears in the value of `core.modes` in its configuration.

Implements part of #2079.

### Checklist
  - By the power of `pytest`, yes!

### Notes
The default value of `core.modes` is `B`, which is the conventional modechar to advertise for `BOT`. Thus this won't do much on most networks unless the user overrides `core.modes` and doesn't know about this mode.

#### Support tables
Note to self: Prepped a support table update for IRCv3 [here](https://github.com/ircv3/ircv3.github.io/compare/master...dgw:sopel-bot-mode?expand=1).

### DIFF
--- a/sopel/coretasks.py
+++ b/sopel/coretasks.py
@@ -311,7 +311,9 @@ def startup(bot, trigger):
 @plugin.priority('medium')
 def handle_isupport(bot, trigger):
     """Handle ``RPL_ISUPPORT`` events."""
-    # remember if NAMESX is known to be supported, before parsing RPL_ISUPPORT
+    # remember if certain actionable tokens are known to be supported,
+    # before parsing RPL_ISUPPORT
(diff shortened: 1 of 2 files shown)


Exirel (changes requested):
I really like the code and the added tests. For completeness, I'd like to see 2 other test cases:

* one when `core.modes` contains more than one mode but none is the one in BOT
* and a second one, very similar, where `core.modes` contains more than one mode including the one in BOT

After all, the code does the `botmodes in config` and not `botmodes == config`.

dgw:
@Exirel Added. `@pytest.mark.parametrize` is a beautiful thing… again.

Exirel (approved):
👼

PR #2089, opened 2021-06-06
coretasks: activate `message-tags` and use it to ignore other bots' tagged messages

dgw:
### Description
Filter out `PRIVMSG` and `NOTICE` commands that have the `bot` (or `draft/bot`) tag. Other events should be left alone until there's a decorator to opt back into them, or Sopel itself will no longer even be aware of bot users' existence on networks that implement one of these tags (which per [the spec](https://ircv3.net/specs/extensions/bot-mode) "SHOULD" be added to all commands sent, _and_ numerics caused, by a bot client).

A later pull request can introduce the needed decorator and changes to Rule constructors for certain callables to allow `bot` messages (à la `@plugin.echo`). Trying to keep stuff in manageable chunks so nobody gets burned out on reviews. 😸

Part of #2079.

### Checklist
  - Once again, by the power of `pytest`!

### Notes
I really want to skip over the part where we have to support _only_ `draft/bot` and then come back later to add support for the final `bot` tag name. Technically we shouldn't do that if the spec isn't ratified by the time Sopel 8 goes stable, though.

### DIFF
--- a/sopel/coretasks.py
+++ b/sopel/coretasks.py
@@ -904,6 +904,7 @@ def receive_cap_ls_reply(bot, trigger):
         'cap-notify',
         'server-time',
         'userhost-in-names',
+        'message-tags',
     ]
     for cap in core_caps:
         if cap not in bot._cap_reqs:
--- a/sopel/plugins/rules.py
+++ b/sopel/plugins/rules.py
@@ -959,6 +959,13 @@ class Rule(AbstractRule):
         event = pretrigger.event
         intent = pretrigger.tags.get('intent')
         nick = pretrigger.nick
+        is_bot_message = (
+            (
+                'draft/bot' in pretrigger.tags or  # can be removed...someday
+                'bot' in pretrigger.tags
+            ) and
+            event in ["PRIVMSG", "NOTICE"]
+        )
         is_echo_message = (
             nick.lower() == bot.nick.lower() and
             event in ["PRIVMSG", "NOTICE"]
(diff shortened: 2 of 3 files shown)


dgw:
To @Exirel's point on IRC, I'm about to (force-)push a revision that prevents this from overriding the existing `@plugin.echo` behavior. The below console input is how I verified the approach:

```pycon
>>> for b in [True, False]:
...     for e in [True, False]:
...         for a in [True, False]:
...             res = ((not b or (e and a)) and (not e or a))
...             print('bot: {}, echo: {}, allowecho: {}, result: {}'.format(b, e, a, res))
...
bot: True, echo: True, allowecho: True, result: True
bot: True, echo: True, allowecho: False, result: False
bot: True, echo: False, allowecho: True, result: False
bot: True, echo: False, allowecho: False, result: False
bot: False, echo: True, allowecho: True, result: True
bot: False, echo: True, allowecho: False, result: False
bot: False, echo: False, allowecho: True, result: True
bot: False, echo: False, allowecho: False, result: True
```

Elegant? No. Still useful.

dgw:
Well I was _going_ to properly rebase this after clicking through GitHub's conflict-resolution flow, but [Coveralls is down](https://status.coveralls.io/incidents/ppvnbpd172gy) and [causing our builds to error](https://github.com/sopel-irc/sopel/runs/3641367626). Just another Saturday, I guess. 😅

Exirel (approved):

PR #2091, opened 2021-06-08
coretasks: fix bad WHO tracking endless loop

HumorBaby:
### Description
`who_reqs`, which is used to track `WHO` requests to the server, had flipped
keys/values and was never getting cleared upon `RPL_ENDOFWHO`.
Eventually, this would lead to an endless loop while the bot tried to
get an unused "querytype" (`randint`) in order to track `WHO` replies.

Now, key/values are in a more useful order (channel --map--> querytype).
Also, the unneccessary loop used to ensure unique values for the query type was
removed. A `RPL_WHOREPLY` includes the channel name in the response, so
confirming that the querytype for a channel matched is sufficient. The same
querytype could even be used for every channel without issue.

### Checklist

### DIFF
--- a/sopel/coretasks.py
+++ b/sopel/coretasks.py
@@ -28,7 +28,6 @@ import collections
 import datetime
 import functools
 import logging
-from random import randint
 import re
 import sys
 import time
@@ -45,8 +44,13 @@ if sys.version_info.major >= 3:
 
 LOGGER = logging.getLogger(__name__)
 
+CORE_QUERYTYPE = '999'
+"""WHOX querytype to indicate requests/responses from coretasks.
+
+Other plugins should use a different querytype.
+"""
+
 batched_caps = {}
-who_reqs = {}  # Keeps track of reqs coming from this plugin, rather than others
 
 
 def setup(bot):
@@ -698,14 +702,12 @@ def _remove_from_channel(bot, nick, channel):
 def _send_who(bot, channel):
     if 'WHOX' in bot.isupport:
         # WHOX syntax, see http://faerion.sourceforge.net/doc/irc/whox.var
-        # Needed for accounts in WHO replies. The random integer is a param
-        # to identify the reply as one from this command, because if someone
-        # else sent it, we have no fucking way to know what the format is.
-        rand = str(randint(0, 999))
-        while rand in who_reqs:
-            rand = str(randint(0, 999))
-        who_reqs[rand] = channel
-        bot.write(['WHO', channel, 'a%nuachtf,' + rand])
+        # Needed for accounts in WHO replies. The `CORE_QUERYTYPE` parameter
+        # for WHO is used to identify the reply from the server and confirm
+        # that it has the requested format. WHO replies with different
+        # querytypes in the response were initiated elsewhere and will be
+        # ignored.
+        bot.write(['WHO', channel, 'a%nuachtf,' + CORE_QUERYTYPE])
     else:
         # We might be on an old network, but we still care about keeping our
         # user list updated
@@ -1243,8 +1245,9 @@ def account_notify(bot, trigger):
 @plugin.priority('medium')
 def recv_whox(bot, trigger):
     """Track ``WHO`` responses when ``WHOX`` is enabled."""
-    if len(trigger.args) < 2 or trigger.args[1] not in who_reqs:
+    if len(trigger.args) < 2 or trigger.args[1] != CORE_QUERYTYPE:
         # Ignored, some plugin probably called WHO
+        LOGGER.debug("Ignoring WHO reply for channel '%s'; not queried by coretasks", trigger.args[1])
         return
     if len(trigger.args) != 8:
         LOGGER.warning(
@@ -1305,16 +1308,6 @@ def recv_who(bot, trigger):
     away = 'G' in status
     modes = ''.join([c for c in status if c in '~&@%+!'])
     _record_who(bot, channel, user, host, nick, away=away, modes=modes)
-
-
-@module.event(events.RPL_ENDOFWHO)
-@plugin.thread(False)
-@module.unblockable
-@plugin.priority('medium')
-def end_who(bot, trigger):
-    """Handle the end of a response to a ``WHO`` command (if needed)."""
-    if 'WHOX' in bot.isupport:
-        who_reqs.pop(trigger.args[1], None)
 
 
 @module.event('AWAY')


Exirel (approved):
Nice catch. This should be noted as a successful week of bug hunt!

HumorBaby:
Sorry, @Exirel, I have to _reject_ your _approval_.  
I added more :pleading_face:… 

See commit message for `coretask: update proper usage of WHOX querytype`. 
TLDR; querytype wasn't being used properly. Updated to spec.

Included the update with this PR since it is closely related to the bugfix, and a separate PR would immediately move/remove some of the fix changes. IMO, should be updated eventually, so _why not now, eh_?  
<sub>Also, I was extra-:rage: because, had it been used correctly to begin, this fix would not have been necessary.</sub>

I'm happy to discuss alternatives to this combo.

sopel/coretasks.py
```diff
@@ -698,14 +698,8 @@ def _remove_from_channel(bot, nick, channel):
 def _send_who(bot, channel):
     if 'WHOX' in bot.isupport:
         # WHOX syntax, see http://faerion.sourceforge.net/doc/irc/whox.var
-        # Needed for accounts in WHO replies. The random integer is a param
-        # to identify the reply as one from this command, because if someone
-        # else sent it, we have no fucking way to know what the format is.
```
dgw, line 703: I wouldn't delete quite _that_ much of the comment. At least the idea of "hav[ing] no fucking way to know what the format is" if the request was sent by other code should be preserved somewhere.

sopel/coretasks.py
```diff
@@ -1313,8 +1308,8 @@ def recv_who(bot, trigger):
 @plugin.priority('medium')
 def end_who(bot, trigger):
     """Handle the end of a response to a ``WHO`` command (if needed)."""
-    if 'WHOX' in bot.isupport:
-        who_reqs.pop(trigger.args[1], None)
+    # Currently, there is nothing to handle at the end of WHO reponse.
+    pass
```
dgw, line 1312: Might as well delete the whole function. No reason to keep an empty callable around eating up resources (especially since it's unthreaded) unless there's some action to perform.

We won't lose any magic incantations to the sands of time; the decorators on this are pure boilerplate.

dgw (changes requested):
I'm nitpicking. It's my job, no? 😁

Nothing here is part of Sopel's documented API, thankfully. This is _probably_ safe to release as part of 7.1.1, though I wish the "type" of WHO query could be non-numeric. `000` is something other people might already use in their plugins (though I did search GitHub for "sopel who 000" in all indexed Python code and found a total of 10 irrelevant hits—it's probably perfectly fine).

Oh well. Up to now it was possible to get Heisenbugs if the random integer happened to collide with some custom code's WHO requests. Now it'll be consistent, and therefore easier to find/fix.

HumorBaby:
>… though I wish the "type" of WHO query could be non-numeric. `000` …

_Wellll actually_, nowhere does it "officially" say it has to be numeric (where officially refers to IRCd docs, and not random forums on the interwebz). The pages cited in the comments (and commit message) use numbers in the example and one uses the term "digits" (which I ~believe~ hope was meant as characters :crossed_fingers:)

Looking at the source for IRCd's that implement `WHOX` (I checked ergo/oregano, InspIRCd, UnderNet's ircu2, Libera.chat's solanum, and freenode's ircd-seven) and they all store querytype as a language-specific string (e.g., `char[]` what have you). There is no enforcement of numeric querytypes. The only validation+enforment being: if your querytype is longer than three characters, they auto-default to `"0"` (still string).


… what to do?? `"Lmk"`?

dgw:
Yeah, you read the same Faerion page I did, which is where the "digits" thing came from. What annoys me about WHOX is that there seems to be no real spec to follow…which means we're stuck doing what you did, and what I presume the previous maintainers did: digging through IRCd source code. Not great.

In the absence of such a spec, though, I'd be wary of doing anything that might run afoul of older IRCds' logic. Your checks validate that being numeric isn't required _now_, but perhaps it _was_ in the past. A counterexample I found is from DareNET's [current documentation](https://www.darenet.org/help/guides/whox) ([snapshot](https://web.archive.org/web/20200926235756/https://www.darenet.org/help/guides/whox)) on WHOX: it specifies "value between 0 and 999".

Seriously, though. We can commiserate over the lack of a formal standard for WHOX all we want, but the endgame is always going to be "leave it as 000; it's fine". 😉

sopel/coretasks.py
```diff
@@ -45,8 +44,9 @@
 
 LOGGER = logging.getLogger(__name__)
 
+CORE_QUERYTYPE = '000'  # WHOX querytype to indicate requests/responses from coretasks, rather than others
```
Exirel, line 47:
```suggestion
CORE_QUERYTYPE = '000'
"""WHOX querytype to indicate requests/responses from coretasks.

Other plugins should uses a different querytype.
"""
```

Exirel (changes requested):
Same as dgw: let's remove that `end_who` plugin callable.

HumorBaby:
> In the absence of such a spec, though, I'd be wary of doing anything that might run afoul of older IRCds' logic. Your checks validate that being numeric isn't required _now_, but perhaps it _was_ in the past. A counterexample I found is from DareNET's [current documentation](https://www.darenet.org/help/guides/whox) ([snapshot](https://web.archive.org/web/20200926235756/https://www.darenet.org/help/guides/whox)) on WHOX: it specifies "<ins>**value between 0 and 999**</ins>".
> 
> Seriously, though. We can commiserate over the lack of a formal standard for WHOX all we want, but the endgame is always going to be "leave it as 000; it's fine".

@dgw, based on your reasoning, I have actually decided to change it to `"999"`. Sopel checks for correct querytype as a string match, despite being numeric. So, if an old IRCd casts a querytype with value < 100 to a numeric, it may end up causing false negatives when, for example, `"42" != "042"`.

In other news, I made the rest of the requested changes and squashed the original loop fix + set querytype constant commits.

dgw (approved):
🎉🥳🚀

> So, if an old IRCd casts a querytype with value < 100 to a numeric, it may end up causing false negatives when, for example, `"42" != "042"`.

That is a good point! It's no reason to amend the PR, but just so you know how silly I am: I probably would have picked something like "Sopel" encoded to [T9](https://en.wikipedia.org/wiki/T9_(predictive_text)) with the vowels removed ("775"), or a common "lucky" number like 777 or 888. 😁

Exirel (approved):
Let's merge that and publish a 7.1.1 asap.

PR #2093, opened 2021-06-08
cli: look harder for $PWD/cfg, don't automatically start config wizard

half-duplex:
### Description
This fixes two annoyances:
- Sopel automatically adds `.cfg` when searching the config dir, but it doesn't when searching the current working directory
- If the config file is not found, it currently automatically starts the configuration wizard.

Combining these two, a user (me) will (does, regularly) `cd` to their non-default Sopel config dir, run `sopel -c mybot` expecting it to load mybot.cfg and run, but instead have Sopel create ~/.sopel/ and ~/.sopel/mybot.cfg and start the config wizard.

Searching `$PWD/{config}.{ext}` is straightforward, but not automatically running the wizard is probably subject to opinions. I changed it to the below output, but I would also be content with "Do you want to run the configuration wizard? y/n", or at least not creating the folder/file until the end of the wizard so the user has an opportunity to ^C without having to then `rm -rf ~/.sopel`

Amended "couldn't find config":
```
$ sopel -c foobar
Welcome to Sopel!
I can't seem to find the configuration file.
To start the configuration wizard, run `sopel configure`
Unable to find the configuration file /home/user/.sopel/foobar
$
```

### Checklist

### DIFF
--- a/sopel/cli/run.py
+++ b/sopel/cli/run.py
@@ -344,11 +344,9 @@ def get_configuration(options):
     try:
         settings = utils.load_settings(options)
     except config.ConfigurationNotFound as error:
-        print(
-            "Welcome to Sopel!\n"
-            "I can't seem to find the configuration file, "
-            "so let's generate it!\n")
-        settings = utils.wizard(error.filename)
+        print("Welcome to Sopel!")
+        error.hint = "To start the configuration wizard, use `sopel configure`"
+        raise error
 
     settings._is_daemonized = options.daemonize
     return settings
--- a/sopel/config/__init__.py
+++ b/sopel/config/__init__.py
@@ -91,14 +91,20 @@ class ConfigurationNotFound(ConfigurationError):
     """Exception type for use when the configuration file cannot be found.
 
     :param str filename: file path that could not be found
+    :param str hint: supplemental information to show the user
     """
-    def __init__(self, filename):
+    def __init__(self, filename, hint=None):
         super(ConfigurationNotFound, self).__init__(None)
         self.filename = filename
         """Path to the configuration file that could not be found."""
+        self.hint = hint
+        """Supplemental information to show the user."""
 
     def __str__(self):
-        return 'Unable to find the configuration file %s' % self.filename
+        ret = 'Unable to find the configuration file %s' % self.filename
+        if self.hint:
+            ret += "\n" + self.hint
+        return ret
 
 
 class Config(object):
--- a/sopel/cli/utils.py
+++ b/sopel/cli/utils.py
@@ -236,6 +236,8 @@ def find_config(config_dir, name, extension='.cfg'):
     if os.path.isfile(name):
         return os.path.abspath(name)
     name_ext = name + extension
+    if os.path.isfile(name_ext):
+        return os.path.abspath(name_ext)
     for filename in enumerate_configs(config_dir, extension):
         if name_ext == filename:
             return os.path.join(config_dir, name_ext)
--- a/test/cli/test_cli_utils.py
+++ b/test/cli/test_cli_utils.py
@@ -112,7 +112,10 @@ def test_find_config_local(tmpdir, config_dir):
         assert found_config == working_dir.join('local.cfg').strpath
 
         found_config = find_config(config_dir.strpath, 'local')
-        assert found_config == config_dir.join('local').strpath
+        assert found_config == working_dir.join('local.cfg').strpath
+
+        found_config = find_config(config_dir.strpath, 'config')
+        assert found_config == config_dir.join('config.cfg').strpath
 
 
 def test_find_config_default(tmpdir, config_dir):


sopel/cli/run.py
```diff
@@ -346,9 +346,10 @@ def get_configuration(options):
     except config.ConfigurationNotFound as error:
         print(
             "Welcome to Sopel!\n"
-            "I can't seem to find the configuration file, "
-            "so let's generate it!\n")
-        settings = utils.wizard(error.filename)
+            "I can't seem to find the configuration file.\n"
+            "To start the configuration wizard, run `sopel configure`"
+        )
+        raise error
```
dgw, line 352: What would you say to printing the string representation of `error` instead of "I can't seem to find the configuration file."? Then the error cause and the name of the missing/expected file would be in one place, and Sopel wouldn't repeat itself.

dgw (commented):
This kind of behavior change isn't appropriate for our upcoming patch release series (7.1.x), so I'll have to tag it for 8.0. It will fit right in with some other CLI changes I remember Exi talking about wanting to do.

Just one idea in notes below.

sopel/cli/run.py
half-duplex: Done:
```
$ sopel -c baz
Welcome to Sopel!
Unable to find the configuration file /home/user/.sopel/baz
To start the configuration wizard, run `sopel configure`
$
```

dgw (approved):
Works for me.

sopel/cli/run.py
```diff
@@ -345,10 +345,12 @@ def get_configuration(options):
...
-            "Welcome to Sopel!\n"
-            "I can't seem to find the configuration file, "
-            "so let's generate it!\n")
-        settings = utils.wizard(error.filename)
+            "Welcome to Sopel!",
+            error,
+            "To start the configuration wizard, run `sopel configure`",
+            sep="\n",
+        )
+        sys.exit(2)
```
Exirel, line 353: That will be a hard no for me: `sys.exit` shouldn't appear here, and left to the caller to decide when to exit.

Exirel (changes requested):

sopel/cli/run.py
dgw: Problem is the caller needs the exception re-raised to know that it should exit, and re-raising the exception will print the error text again in both places where this func is called—exactly the redundancy I was trying to have mal eliminate.

https://github.com/sopel-irc/sopel/blob/a174d1838e295c8c0e0749629f944c69807454bf/sopel/cli/run.py#L410-L414

https://github.com/sopel-irc/sopel/blob/a174d1838e295c8c0e0749629f944c69807454bf/sopel/cli/run.py#L608-L612

Exirel: Yes? There is nothing that prevent that code to raise a new exception with a different error message, it looks like it'll work properly and do exactly what is expected.

half-duplex: Thoughts? Not redundant, no `exit()`, doesn't rely on context in a way that rubs me weird but exi says is fine.
```
$ sopel -c foobar
Welcome to Sopel! To start the configuration wizard, run `sopel configure`
Unable to find the configuration file /home/user/.sopel/foobar
```

dgw: Outputting the solution before stating the problem is a bit weird, but tolerable. There's no rule saying this can't be changed again later, after all.

Exirel: I still don't get why raising a fresh new exception is bad, but I guess it can be changed again later, after all.

half-duplex: See latest - Add `ConfNotFound.hint` for preferred order?

sopel/config/__init__.py
```diff
@@ -91,14 +91,20 @@ class ConfigurationNotFound(ConfigurationError):
     """Exception type for use when the configuration file cannot be found.
 
     :param str filename: file path that could not be found
+    :param str hint: supplimental information to show the user
```
dgw, line 94:
```suggestion
    :param str hint: supplemental information to show the user
```

sopel/config/__init__.py
```diff
@@ -91,14 +91,20 @@ class ConfigurationNotFound(ConfigurationError):
...
     """
-    def __init__(self, filename):
+    def __init__(self, filename, hint=None):
         super(ConfigurationNotFound, self).__init__(None)
         self.filename = filename
         """Path to the configuration file that could not be found."""
+        self.hint = hint
+        """Supplimental information to show the user"""
```
dgw, line 101:
```suggestion
        """Supplemental information to show the user."""
```

sopel/cli/run.py
```diff
@@ -344,11 +344,9 @@ def get_configuration(options):
     try:
         settings = utils.load_settings(options)
     except config.ConfigurationNotFound as error:
-        print(
-            "Welcome to Sopel!\n"
-            "I can't seem to find the configuration file, "
-            "so let's generate it!\n")
-        settings = utils.wizard(error.filename)
+        print("Welcome to Sopel!")
+        error.hint = "To start the configuration wizard, run `sopel configure`"
```
dgw, line 348: Regardless of how this message gets into the user's terminal, I can imagine that it would be useful to put the full command line, e.g. copy whatever `-c` arg was passed here and replicate it in the example. That might also be too much to do at once, and you're well within your rights to tell me "do that later".

dgw (commented):
Nitpicking the spelling and style. I'll let Exi come back and be Opinionated™ about the new exception approach.

sopel/cli/run.py
half-duplex: It seemed Annoying since argparse can't go backwards, but I guess it's probably okay to only bother with `-c` and `--config-dir`... Will do.

sopel/config/__init__.py
half-duplex: Damn it

sopel/cli/run.py
half-duplex: Actually... Only the parsed `options` object gets passed to `get_configuration()`, so I don't think I can get the defaults, so I don't think I can cleanly omit options the user didn't specify... `if "-c" not in sys.argv`? Or just let the user figure it out?

dgw: Hmm... You're right, that is Annoying. The hell with it, at least for now; just changing one word will be enough:

```suggestion
        error.hint = "To start the configuration wizard, use `sopel configure`"
```

While "run" implies "this is what you should type", "use" implies "this is the right tool" only.

Exirel (changes requested):
Alright, I've take time to read the code from `sopel/cli/run.py` and how `get_configuration` is used: in both case, it catches the generic configuration error to print it and exit.

Now, why does `get_configuration` exists? It does because it was a way to always get a configuration file and always run the wizard. Now, that's gone. Moreover, in #2118 I plan to simply remove the legacy mode.

So, I think adding the `hint` property is not the right solution here, and it's making thing more complicated than they should be. There are 3 options:

* remove the try/except and put a comment in #2118 to deal with a friendlier error message somehow
* raise a *new* configuration error (a generic one) with the proper message (I think I already suggested that)
* remove the try/except and catch the `ConfigurationNotFound` in command start (and forget about legacy, which is going down) to print a nice message (and also comment in #2118 to say good luck with the merge conflict)

In any case, I don't see the point of adding a `hint` attribute.

dgw:
I recommend taking another look at the behavior on `master` now that the legacy CLI is gone. If the reason for creating this patch is still present, please do look at @Exirel's suggested alternative solution in the above comment.

dgw:
Ping @half-duplex for reevaluation of `master` and/or responses to Exi's comments.

dgw:
Removing from release plans due to lack of attention over a period of months. Can be reopened if a second wind hits, as always.

PR #2095, opened 2021-06-08
irc: Use repr() and don't strip() rawlog, log charset guess failures

half-duplex:
### Description
Print rawlog with `repr()` so we don't need a hex editor to find problems with whitespace or non-printable characters.

Don't `.strip()` messages headed for rawlog, which currently obscures problems with whitespace.

Add logging (warning) and rawlogging of messages that couldn't be decoded, which were previously silently ignored. New rawlog prefix: `<<!`

Each part of this is a breaking change if any madman is parsing the rawlog, but I don't think we consider that part of the API...

### Concerns

QA passes, but this does change handling of IRC's message separators in ways that may have corner cases I haven't found:
- Switches asynchat to break messages on `b'\r\n'` instead of `b'\r'`. Do any real-world daemons use only `\n`?
- Puts the `\r\n` back on before decoding/logging. I'm not sure how far that goes before it hits another `.strip()`

I'm currently running it in prod (unreal 3 and inspircd 3) and haven't found any obvious problems.

### Checklist

### DIFF
--- a/sopel/irc/backends.py
+++ b/sopel/irc/backends.py
@@ -106,7 +106,7 @@ class AsynchatBackend(AbstractIRCBackend, asynchat.async_chat):
         AbstractIRCBackend.__init__(self, bot)
         asynchat.async_chat.__init__(self)
         self.writing_lock = threading.RLock()
-        self.set_terminator(b'\n')
+        self.set_terminator(b'\r\n')
         self.buffer = ''
         self.server_timeout = server_timeout or 120
         self.ping_interval = ping_interval or (self.server_timeout * 0.45)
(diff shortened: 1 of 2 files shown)


sopel/irc/backends.py
```diff
@@ -213,6 +213,8 @@ def collect_incoming_data(self, data):
         The incoming line is discarded (and thus ignored) if guessing the text
         encoding and decoding it fails.
         """
+        data = data + self.get_terminator()
```
Exirel, line 216: The problem is now that when you do `self.buffer += data`, `self.buffer` now contains `\r\n` while it didn't before.

half-duplex: That's ~intentional. I don't *think* the buffer goes anywhere but the log and the PreTrigger, and `PreTrigger.__init__()` immediately does `line = line.strip('\r\n')`. Everything gracefully handled the previous behavior, which unintentionally left a `\r` at the end of each message.

Exirel: Thinking about this. Indeed. It works because down the line, it's already handled. Thanks for your clarification!

Exirel (approved):

dgw (approved):
Looks sane. Labeled as a breaking change just because the change in message handling should be mentioned somewhere in the upgrade guide and/or changelog, to give people an idea of what happened if some noncompliant IRCd out there is sending incorrect line endings.

PR #2097, opened 2021-06-08
coretasks: Strip Unreal's extra empty MODE arg

half-duplex:
### Description
Reduce "possibly malformed MODE" logspam, and warn with useful info if we still have extra args.

I can't think of a case where we would need a legitimate empty arg, and the continued presence of the previous comments seems to agree, so... Let's just assume a trailing space is a meaningless trailing space.

(If we're *short* args, we'll... still get an uncaught `IndexError`. Should that be handled better?)

### Checklist

### DIFF
--- a/sopel/coretasks.py
+++ b/sopel/coretasks.py
@@ -496,10 +496,13 @@ def _parse_modes(bot, args):
         return
 
     channel = bot.channels[channel_name]
-    # Our old MODE parsing code checked for empty args. This would be a good
-    # place to re-implement that if necessary for a non-compliant IRCd, but for
-    # now just log malformed lines. After this we'll make a (possibly dangerous)
-    # assumption that the MODE message is more-or-less compliant.
+
+    # Unreal 3 sometimes sends an extraneous trailing space. If we're short an
+    # arg, we'll find out later.
+    if args[-1] == "":
+        args.pop()
+    # If any args are still empty, that's something we may not be prepared for,
+    # but let's continue anyway hoping they're trailing / not important.
     if len(args) < 2 or not all(args):
         LOGGER.debug(
             "The server sent a possibly malformed MODE message: %r", args)
@@ -598,6 +601,13 @@ def _parse_modes(bot, args):
             )
             _send_who(bot, channel_name)
             return
+
+    if param_idx != len(params):
+        LOGGER.warning(
+            "Too many arguments received for MODE: args=%r chanmodes=%r",
+            args,
+            chanmodes,
+        )
 
     LOGGER.info("Updated mode for channel: %s", channel.name)
     LOGGER.debug("Channel %r mode: %r", str(channel.name), channel.modes)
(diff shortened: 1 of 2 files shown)


sopel/coretasks.py
```diff
@@ -496,6 +496,11 @@ def _parse_modes(bot, args):
         return
 
     channel = bot.channels[channel_name]
+
+    # Unreal sometimes sends an extraneous trailing space. If we're short an
+    # arg, we'll find out later.
+    if args[-1] == "":
+        args.pop()
     # Our old MODE parsing code checked for empty args. This would be a good
     # place to re-implement that if necessary for a non-compliant IRCd, but for
```
dgw, line 505: > This would be a good place to re-implement that if necessary for a non-compliant IRCd

That's what you're doing. Feel free to reword this comment if you want, perhaps to talk about "more checks" for empty args other than at the end.

dgw (approved):
Sensible, and likely a safe bugfix to port back for 7.1.1. One meta-comment below, no barrier to merging if you don't get around to it.

sopel/coretasks.py
dgw: You did well with the last fixup earlier today. 😸 It's good as-is, once squashed so CI can run one last time.

sopel/coretasks.py
```diff
@@ -599,6 +602,14 @@ def _parse_modes(bot, args):
             _send_who(bot, channel_name)
             return
 
+    if param_idx != len(params):
+        LOGGER.warning(
+            "%s extra arguments received for MODE: args=%r chanmodes=%r",
```
Exirel, line 607:
```suggestion
            "%d extra arguments received for MODE: args=%r chanmodes=%r",
```

Exirel (approved):
Small nitpick that would be nice to have, otherwise LGTM.

sopel/coretasks.py
half-duplex: Do we care about "1 extra argument**s**"?

sopel/coretasks.py
```diff
@@ -599,6 +602,14 @@ def _parse_modes(bot, args):
             _send_who(bot, channel_name)
             return
 
+    if param_idx != len(params):
+        LOGGER.warning(
+            "%d extra arguments received for MODE: args=%r chanmodes=%r",
+            len(params) - param_idx,
```
dgw, line 608: To answer the question about `1 arguments`, kinda yeah.

```suggestion
    if param_idx != len(params):
        extra = len(params) - param_idx
        LOGGER.warning(
            "%d extra argument%s received for MODE: args=%r chanmodes=%r",
            extra,
            's' if extra != 1 else '',
```

sopel/coretasks.py
dgw: (addressed in another thread, so I could make a patch suggestion)

sopel/coretasks.py
half-duplex: We can figure that out from the other info provided, so I'll just keep it simple (and make translations easier).

dgw:
@half-duplex You'll need to leave that `caplog` test in a separate commit, since it won't work on old Pythons and this is supposed to be cherry-picked back to 7.1.x.

PR #2098, opened 2021-06-08
coretasks: Clear prior modes on RPL_CHANNELMODEIS

half-duplex:
### Description
RPL_CHANNELMODEIS should contain all channel modes, and we sometimes use `MODE {channel}` to sync, so we should discard our old knowledge of the modes when we get the response.

### Checklist

### DIFF
--- a/sopel/coretasks.py
+++ b/sopel/coretasks.py
@@ -483,11 +483,10 @@ def track_modes(bot, trigger):
 @module.unblockable
 def initial_modes(bot, trigger):
     """Populate channel modes from response to MODE request sent after JOIN."""
-    args = trigger.args[1:]
-    _parse_modes(bot, args)
+    _parse_modes(bot, trigger.args[1:], clear=True)
 
 
-def _parse_modes(bot, args):
+def _parse_modes(bot, args, clear=False):
     """Parse MODE message and apply changes to internal state."""
     channel_name = Identifier(args[0])
     if channel_name.is_nick():
@@ -517,6 +516,11 @@ def _parse_modes(bot, args):
         "Y": module.OPER,
     }
 
+    modes = {}
+    if not clear:
+        # Work on a copy for some thread safety
+        modes.update(channel.modes)
+
     # Process modes
     sign = ""
     param_idx = 0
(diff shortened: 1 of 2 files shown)


test/test_coretasks.py
```diff
@@ -156,6 +156,30 @@ def test_handle_rpl_channelmodeis(mockbot, ircfactory):
...
+        "+knlt",
+        "hunter2",
+        ":1",
+    ]
+    irc = ircfactory(mockbot)
+    irc.bot._isupport = isupport.ISupport(chanmodes=("b", "k", "l", "mnt", tuple()))
+    irc.channel_joined("#test", ["Alex", "Bob", "Cheryl"])
+    mockbot.on_message(" ".join(rpl_channelmodeis))
+    rpl_channelmodeis[4] = "+kl"
+    mockbot.on_message(" ".join(rpl_channelmodeis))
```
Exirel, line 175: I don't like using smart code like `rpl_channelmodeis[4] = "+kl"` because now I've to go back to the list, and check which one is the 4th index (which I'm bad at). I'd prefer to use a template string, like so:

```suggestion
    irc = ircfactory(mockbot)
    irc.bot._isupport = isupport.ISupport(chanmodes=("b", "k", "l", "mnt", tuple()))
    irc.channel_joined("#test", ["Alex", "Bob", "Cheryl"])

    rpl_channelmodeis = ":mercury.libera.chat 324 TestName #test {mode} hunter2 :1"
    mockbot.on_message(rpl_channelmodeis.format(mode='+knlt'))
    mockbot.on_message(rpl_channelmodeis.format(mode='+kl'))
```

This version highlight the `mode` part of the message, providing more intent.

While I'm at it, I prefer to have the IRC setup before the exact part that is tested here (the RPL_CHANNELMODEIS), for readability.

Exirel (approved):
A bit of nitpicking that doesn't prevent approval.

test/test_coretasks.py
half-duplex: Fair points, though I'll just split out all the mode bits.

sopel/coretasks.py
```diff
@@ -517,6 +516,11 @@ def _parse_modes(bot, args):
         "Y": module.OPER,
     }
 
+    if clear:
+        modes = {}
+    else:
+        modes = channel.modes
```
dgw, line 522: Is there some rationale for not directly resetting the channel's attribute?

```suggestion
    if clear:
        channel.modes = {}
```

_All_ of the changes below are unnecessary if this functions simply continues operating against `channel.modes`, as it has been. The callable is unthreaded, so no need to worry about races.

dgw (commented):
It seems @Exirel is more _laissez-faire_ (or _laisser faire_ for the 🤬🇫🇷 types 😁) than I. Got a question about why this patch was so big.

sopel/coretasks.py
half-duplex: I know less about threading than async, but couldn't a threaded callable started just before the RPL_CHANNELMODEIS unthreaded one end up reading modes[] while it's not yet rebuilt?

dgw: That could be solved by making it a `SopelMemory` for the mutex behavior, instead of a plain dict. In fact, it's kind of insane that `channel.modes` isn't already a thread-safe type.

half-duplex: As a general principle using `SopelMemory` would probably be good, but do you mean to manually take the lock on it while rebuilding? That would work, but I don't see the advantage, and then I'd need `if clear: [take/release lock]` on both ends of the loop.
If I don't manually take the lock: `modes={a,b}`, we `MODE #chan`, I replace the `SopelMemory` and begin rebuilding it. I `__additem__` `a` which takes and releases the lock, another thread checks `__contains__` `b` which takes and releases the lock, then I continue to add `b`, which lets the other callable think it was is absent when it never was.

Exirel:
```
modes = {}
if not clear:
    modes.update(channel.modes)

# replace channel.modes by modes
# then:
channel.modes = modes
```

Then you won't have the problem of race condition while updating `channel.modes`. The main problem here is that Sopel uses both async to deal with the socket and multi-threading to deal with command handling. And no, you can't complain as much as I do on that topic, no need to try.

half-duplex: Thanks!

dgw (approved):
I'm glad that this arrived at a state with reasonable thread-safety, even though I started that conversation for the wrong reason. 😸 Squash away!

Issue #1647, opened 2019-06-26
Eliminate/discourage "naive" datetime objects


dgw:
In short, we should try to eliminate things like this from Sopel's API:

https://github.com/sopel-irc/sopel/blob/b105fe4aaa6c1cd258337e60a4f17c1a0751ecb5/sopel/tools/time.py#L107

`datetime` objects, both in core and in plugins, should someday always be "aware" (have a timezone associated, even if it's just UTC), not "naive" (unaware of their own timezone).

I'm putting this in milestone 7.0.0 with intent to punt it if it isn't done by the first indication that Sopel 7 is ready to go. If this makes it in, great; if not, fine. I don't care, really; I just want to get it on our radar for planning and discussion purposes.

Follow-up to #1645.

dgw:
7.0 isn't "ready" yet, but I don't see us spending time on this in the next month.

dgw:
@Exirel What do you think, will it be enough to just change documentation, or do we need a deprecation cycle and warnings? Unless you have more concrete thoughts about it, I'm honestly not sure what to do about this yet, and it's probably best left to 8.0.

Exirel:
Last time I checked, I think we mostly use very basic timestamps, and we don't do any kind of timezone related manipulation. I mean, except for the `.time` command and stuff, but this part is already pretty good IIRC.

I guess I should dive a bit more into this issue and check if we need to do something or if we can just write something short and clear in the documentation.

dgw:
From #2005:

> This is related to #1647. It doesn't fix it but we may consider it done nonetheless.

We do. Let's let this one rest.

PR #2099, opened 2021-06-08
bot: ignore replayed messages on channel join

### LINKED ISSUE #1810: UnrealIRCd/InspIRCd Channel History repetition
### Description
I got around to uprading my UnrealIRCd to version 5, which supports more IRCv3 features. One of which, I was really excited about for my friends that connect is channel history.

https://www.unrealircd.org/docs/Channel_history

Essentially, you can set a channel to display a backlog of 50 messages that were said before JOIN.

The problem, is Sopel will process commands/URL, get rebooted and reply to the last 50 lines from before it was rebooted.

### Reproduction steps
Run a bot in a channel with Channel History enabled.

### Expected behavior
We should log timestamps in some manner to prevent the bot from repeating things it happened across bot reboots.

### Logs
```
If applicable, add logs to help us figure out what's happening. Raw logs are
super helpful! Logs are usually found in ~/.sopel/logs, depending on your
configuration.
```

### Environment
Any

### Notes
Screenshot:

99f4f678a27a6d4c66a201f7abbdc21aa8d6c701c9f5b3e13ceff95cd954

Note the timestamps while this bot was taken offline.

half-duplex:
### Description
Fixes #1810 (for servers with server-time cap)

I have no idea how to go about writing tests for this, but it works for me ™
Should it log when it ignores a message? On one hand, silently ignoring things is weird. On the other, that would get spammy fast.

### Checklist


sopel/tools/target.py
```diff
@@ -96,6 +96,12 @@ def __init__(self, name):
         self.last_who = None
         """The last time a WHO was requested for the channel."""
 
+        self.join_time = None
+        """The time the server received our JOIN message.
+
+        Requires server-time IRCv3 capability.
```
dgw, line 102:
```suggestion
        Will be ``None`` if the ``server-time`` IRCv3 capability is not available.
```

sopel/bot.py
```diff
@@ -898,6 +898,15 @@ def dispatch(self, pretrigger):
         user_obj = self.users.get(nick)
         account = user_obj.account if user_obj else None
 
+        if "time" in pretrigger.tags and pretrigger.sender in self.channels:
+            time = datetime.strptime(
+                pretrigger.tags["time"],
+                "%Y-%m-%dT%H:%M:%S.%f%z",
+            )
```
dgw, line 905: This pattern might benefit from an additional property on the `PreTrigger` and `Trigger` objects so plugins don't have to reinvent this wheel. (That part probably wouldn't be good to put in 7.1.1, regardless of the above `%z` discussion, as it'd be an API change.)

sopel/bot.py
```diff
@@ -898,6 +898,15 @@ def dispatch(self, pretrigger):
         user_obj = self.users.get(nick)
         account = user_obj.account if user_obj else None
 
+        if "time" in pretrigger.tags and pretrigger.sender in self.channels:
+            time = datetime.strptime(
+                pretrigger.tags["time"],
+                "%Y-%m-%dT%H:%M:%S.%f%z",
```
dgw, line 904: Is there some reason to parse `%z` instead of a literal `Z`? The wire format prescribed by [the `server-time` spec](https://ircv3.net/specs/extensions/server-time) specifically says UTC, with a `Z` suffix, no other timezone. Tested against `tilde.chat`, got tags like the following in Sopel's raw log:

`@time=2021-06-09T02:30:54.891Z`

The use of `%z` is wonky on Python 2.7, at least, as I found out when rewriting the caching logic for `tld`, and for obvious reasons this will have to wait until Sopel 8.0 (instead of 7.1.1) if it doesn't work properly on py2. (Though I had this thought before remarking that a new field is added to the `Channel` type… Hmm. Maybe this should just be an 8.0 feature.)

dgw (commented):
Not sure how to label/milestone this yet. We'll see where the discussions go.

sopel/bot.py
Exirel: Yup, go with literal Z, the spec is quite clear on that:

> The value of the tag MUST have the following syntax:
> 
>     <value> ::= YYYY-MM-DDThh:mm:ss.sssZ

Which is, indeed, ISO 8601:2004(E) 4.3.2, a commonly used format for the Zulu timezone (UTC+0). There is no reason to use `%z` in that context.

half-duplex: I forgot 7.1 still supports py2.7 and we're not free of it yet -.-
Using timezone-naïve objects is incorrect, and `datetime` sucks so the other ways to make the object timezone-aware (even UTC) are dumb. If this PR is going for 7.1 I'll change it, otherwise `%z` is best.

Exirel: > If this PR is going for 7.1 I'll change it, otherwise `%z` is best.

I disagree: go with literal `Z` in either case. The spec already tell you which timezone it should be (UTC+0). Forcing the timestamp to be in UTC+0 is a good thing, anything that isn't according to 

Also, what is dumb with the snippet of code?

```python
pytz.utc.localize(
    datetime.strptime(
        pretrigger.tags["time"], '%Y-%m-%dT%H:%M:%S.%fZ'))
```

I know you are annoyed with datetime, but right now, you are being stubborn about the wrong thing.

dgw: Was going to say, you can get an aware `datetime` object without having to use `%z`. The `tld` plugin does it in a slightly different (kludgier) way, but still:

https://github.com/sopel-irc/sopel/blob/e2d8439caae33ed3dfe08f0c164d1e8323fd6f7d/sopel/modules/tld.py#L33-L35

Anyway, I thought about it some more and this should go in 8.0. It's a significant behavior change, not appropriate for 7.1.1. Holding it also absolves @half-duplex of any need to support Python 2.7, and removes the risk of anyone getting mad at the project (meaning me) for adding a "new feature" (the added field on `Channel` objects) in a maintenance release.

dgw: > there should be a tzinfo= kwarg to strptime and many others

I think this every time I use `strptime()`. It's silly that there isn't one.

Maybe Python 4 will make datetime objects default to aware in the system's local TZ. We can dream.

sopel/bot.py
dgw: Bleh.

https://github.com/sopel-irc/sopel/blob/e2d8439caae33ed3dfe08f0c164d1e8323fd6f7d/sopel/trigger.py#L129-L134

Is there any reason to think stuff would break if 8.0 made that attribute into an _aware_ `datetime` in UTC?

dgw:
I love making `Channel.join_time` always available. That's consistent with how `trigger.time` works.

Breaking changes to object properties make me 🤔 though. While making `trigger.time` offset-aware makes it very easy to deal with in this patch, it will break an unknown amount of code outside of Sopel's core that expects `trigger.time` to be offset-naïve and uses it in comparisons against other naïve `datetime` objects.

That said, I also love the _idea_ of making `trigger.time` offset-aware (per #1647). I need either a plan for making the migration a bit more graceful (creative use of a new attr + `@property` + `@tools.deprecated` comes to mind), or to be convinced that no graceful migration is possible and we should rip off the band-aid.

half-duplex:
I thought about it, but I only see one use of `trigger.time` that's not in a project under dgw/ or sopel-irc/:
https://www.google.com/search?q=sopel+trigger.time+-sopel.chat+-inurl%3Agithub.com%2Fsopel-irc%2F

I can't think of a particularly graceful way to transition besides changing to something annoying like `trigger.utc_time`. Could we just `@deprecated` it in 7, with just a note that "This object will become timezone-aware in 8.0 and you will need to update your code"? Not great, but better than no warning if we want to keep `trigger.time`

dgw:
I think [searching GitHub](https://github.com/search?l=Python&q=sopel+trigger.time+-user%3Adgw+-user%3Asopel-irc&type=Code) is more useful, since it _only_ searches code. There isn't much there, either—most of the results are from core plugins in custom-bot repos where Sopel was used as a base.

We _could_ embark on a lengthy sequence of new/removed/renamed attributes over the next few major versions, but I agree that that sounds like _far_ more trouble than it's worth. Want to throw together a patch against `7.1.x` to implement that warning?

Exirel:
1. switching from naive to aware will break any operation because naive/aware datetime are not compatible with each other
2. `pretrigger.time` is UTC+0, which is never affected by timezone rules such as DST, so you are fine, you can use `datetime.utcnow()` and stop complaining
3. I really don't like having yet another property with a deprecated warning, I'm fine with having a breaking change between Sopel 7 and Sopel 8 by switching from naive to aware datetime for `pretrigger.time`

dgw:
An issue in sopel-weather reminded me of `tools.time.format_time()`, which I noticed expects a naïve `datetime` object in UTC:

image

If Sopel itself is going to start providing offset-aware `datetime` objects, is it worth enhancing the `time` tools to handle either? It's certainly possible for plugin devs to convert from aware to naïve on their own, but it's a bit annoying to have to remember.

> 3\. I really don't like having yet another property with a deprecated warning, I'm fine with having a breaking change between Sopel 7 and Sopel 8 by switching from naïve to aware datetime for `pretrigger.time`

The warning would be for users (bot owners and plugin devs) to find out in advance that their code is affected, since we know people don't read docs, release notes, etc. Is that somehow bad?

sopel/trigger.py
```diff
@@ -101,6 +101,9 @@ class PreTrigger(object):
         If the IRC server sent a message tag indicating when *it* received the
         message, that is used instead of the time when Sopel received it.
 
+        .. versionchanged:: 8.0
+            Now returns a timezone-aware ``datetime`` object.
```
dgw, line 105: It's not a function/method, so it can't "return" something.

```suggestion
            Now a timezone-aware ``datetime`` object.
```

dgw (changes requested):
If all I have is a documentation nitpick, we're doing well.

@Exirel I'm fine with this as-is, breaking change and no deprecation cycle, just ripping off the band-aid. Good to go besides this?

sopel/trigger.py
```diff
@@ -129,7 +132,10 @@ def __init__(self, own_nick, line, url_schemes=None):
         self.time = datetime.datetime.utcnow()
```
Exirel, line 132: If the server time is not provided, this would have been a naive datetime, which is bad.

```suggestion
        self.time = pytz.utc.localize(datetime.datetime.utcnow())
```

Exirel (changes requested):

Exirel (approved):
I like the tests. I don't understand why you don't want to use `pytz.utc`, but heh, it's functionally identical to using `datetime.timezone.utc`. But we already talk about how stubborn you can be. 😜

PR #2100, opened 2021-06-09
coretasks, irc: implement CertFP / SASL EXTERNAL authentication

dgw:
### Description
Tin. Followed Libera's instructions for generating a certificate and added its fingerprint to our official Sopel bot's account in NickServ, then stuck around in my human-operated client to observe. **Flawless victory:**

```
[00:53:18] -NickServ- SopelTest!~sopel@68.168.def.abc has just authenticated as you (Sopel)
                                       ^ my home IP ^
```

### Checklist

### Note to self
**@dgw pls update IRCv3 support info for Sopel when this is merged**

### Future plans
This lays some groundwork for expanding into SCRAM challenges, as implied by a new TODO comment, but implementing those will take some more research. I had some pseudocode for it in there, but obviously removed it before committing (otherwise linting would probably fail).

As far as actually getting SCRAM into Sopel goes, the options are:

- choose a library — possibly [`scramp`](https://pypi.org/project/scramp/) or [`tinysasl`](https://github.com/stan-janssen/tinysasl/blob/main/tinysasl/__init__.py) (no kidding; the whole library is ~110 lines of code), but I've barely searched for alternatives
- implement SCRAM ourselves (oy, IRC specs are hard enough amirite)
- steal someone else's implementation (e.g. BitBot has its own, but under GPLv2)

Using a library could be our first excuse to give Sopel setuptools extras. (`pip install sopel[scram]`, anyone?)

Meanwhile, CertFP is what was explicitly requested on May 29 (May 30 UTC) in our IRC channel, and it turns out that was fairly easy to get done, so it might as well get reviewed and merged without waiting for SCRAM nonsense.

### DIFF
--- a/sopel/config/core_section.py
+++ b/sopel/config/core_section.py
@@ -176,7 +176,9 @@ class CoreSection(StaticSection):
         * ``PLAIN`` if using the ``sasl`` :attr:`auth_method`
 
     The nickname of the NickServ service, or the name of the desired SASL
-    mechanism, if :attr:`auth_method` is set to one of these methods. This value
+    mechanism, if :attr:`auth_method` is set to one of these methods. For
+    SASL, the ``EXTERNAL`` option is available in case the IRC network
+    requires it (e.g. for CertFP using :attr:`client_cert_file`). This value
     is otherwise ignored.
 
     See :ref:`Authentication`.
@@ -261,6 +263,18 @@ class CoreSection(StaticSection):
         An unquoted ``#`` denotes a comment, which will be ignored by Sopel's
         configuration parser.
 
+    """
+
+    client_cert_file = FilenameAttribute('client_cert_file')
+    """Filesystem path to a certificate file for CertFP.
+
+    This is expected to be a ``.pem`` file containing both the certificate and
+    private key. Most networks that support CertFP will give instructions for
+    generating this, typically using OpenSSL.
+
+    Some networks may refer to this authentication method as SASL EXTERNAL.
+
+    .. versionadded:: 8.0
     """
 
     commands_on_connect = ListAttribute('commands_on_connect')
@@ -1068,7 +1082,12 @@ class CoreSection(StaticSection):
 
     :default: ``PLAIN``
 
+    ``EXTERNAL`` is also supported, e.g. for using :attr:`client_cert_file` to
+    authenticate via CertFP.
+
     .. versionadded:: 7.0
+    .. versionchanged:: 8.0
+        Added support for SASL EXTERNAL mechanism.
     """
 
     server_auth_username = ValidatedAttribute('server_auth_username')
--- a/sopel/coretasks.py
+++ b/sopel/coretasks.py
@@ -957,7 +957,6 @@ def receive_cap_ack_sasl(bot):
     if not password:
         return
 
-    mech = mech or 'PLAIN'
     available_mechs = bot.server_capabilities.get('sasl', '')
     available_mechs = available_mechs.split(',') if available_mechs else []
 
@@ -1025,10 +1024,23 @@ def auth_proceed(bot, trigger):
         be ignored. If none is set, then this function does nothing.
 
     """
-    if trigger.args[0] != '+':
-        # How did we get here? I am not good with computer.
+    if bot.config.core.auth_method == 'sasl':
+        mech = bot.config.core.auth_target or 'PLAIN'
+    elif bot.config.core.server_auth_method == 'sasl':
+        mech = bot.config.core.server_auth_sasl_mech or 'PLAIN'
+    else:
         return
-    # Is this right?
+
+    if mech == 'EXTERNAL':
+        if trigger.args[0] != '+':
+            # not an expected response from the server; abort SASL
+            token = '*'
+        else:
+            token = '+'
+
+        bot.write(('AUTHENTICATE', token))
+        return
+
     if bot.config.core.auth_method == 'sasl':
         sasl_username = bot.config.core.auth_username
         sasl_password = bot.config.core.auth_password
@@ -1036,11 +1048,22 @@ def auth_proceed(bot, trigger):
         sasl_username = bot.config.core.server_auth_username
         sasl_password = bot.config.core.server_auth_password
     else:
+        # How did we get here? I am not good with computer
         return
+
     sasl_username = sasl_username or bot.nick
-    sasl_token = _make_sasl_plain_token(sasl_username, sasl_password)
-    LOGGER.info("Sending SASL Auth token.")
-    send_authenticate(bot, sasl_token)
+
+    if mech == 'PLAIN':
+        if trigger.args[0] != '+':
+            # not an expected response from the server; abort SASL
+            token = '*'
+        else:
+            sasl_token = _make_sasl_plain_token(sasl_username, sasl_password)
+            LOGGER.info("Sending SASL Auth token.")
+            send_authenticate(bot, sasl_token)
+        return
+
+    # TODO: Implement SCRAM challenges
 
 
 def _make_sasl_plain_token(account, password):
@@ -1127,12 +1150,16 @@ def sasl_mechs(bot, trigger):
 def _get_sasl_pass_and_mech(bot):
     password = None
     mech = None
+
     if bot.config.core.auth_method == 'sasl':
         password = bot.config.core.auth_password
         mech = bot.config.core.auth_target
     elif bot.config.core.server_auth_method == 'sasl':
         password = bot.config.core.server_auth_password
         mech = bot.config.core.server_auth_sasl_mech
+
+    mech = 'PLAIN' if mech is None else mech.upper()
+
     return password, mech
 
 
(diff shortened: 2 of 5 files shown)


sopel/config/core_section.py
```diff
@@ -1048,6 +1048,16 @@ def homedir(self):
     silently from the triggering IRC user's perspective.
     """
 
+    sasl_cert_file = FilenameAttribute('sasl_cert_file')
+    """Filesystem path to a certificate file for CertFP.
+
+    This is expected to be a ``.pem`` file containing both the certificate and
+    private key. Most networks that support CertFP will give instructions for
```
half-duplex, line 1055: Is there a reason to not just provide an example command here? e.g. "`openssl req -new -newkey rsa:4096 -days 3650 -nodes -x509 -sha256 -out sopel.pem -keyout sopel.pem` and accept the defaults"
I guess the services setup instructions may differ?

half-duplex (commented):
LGTM with nits.
Is there a sane way in stock sopel to have a conversation *through* the bot? Notably, to command the bot to `PRIVMSG NickServ :CERT ADD` for enrollment without needing to configure a regular client or ZNC as the bot to do that. I've always just loaded a hackjob `.exec` plugin.

sopel/config/core_section.py
dgw: I figure it's safest to force users to look at the documentation from their network, so they 1) use the exact `openssl` command that network wants in case of weird algo/bitness requirements or maximum validity time, etc. and 2) will (more likely) see the instructions for registering the cert fingerprint with services.

Exirel: In my opinion it's best to let the service defines how to generate this file. We are not security expert, and the best we can do is to remind the bot's owner to read the doc of the service they are connecting to.

sopel/config/core_section.py
```diff
@@ -1048,6 +1048,16 @@ def homedir(self):
     silently from the triggering IRC user's perspective.
     """
 
+    sasl_cert_file = FilenameAttribute('sasl_cert_file')
```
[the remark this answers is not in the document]
dgw: Setting name is a good suggestion. Just be glad I got away from the original idea of shoehorning this setting into the existing auth stuff… 😝 (I also hate how some settings have different meanings depending on which auth method is in use, and 8.0 might be a time to fix.)

If I understand your last line correctly, I think Libera does. Their instructions say to enable SASL EXTERNAL if connecting via Tor (because SASL is required over Tor, no matter what), but otherwise it seems to just auth against the cert regardless of whether SASL negotiation happened.

sopel/coretasks.py
```diff
@@ -1127,12 +1150,19 @@ def sasl_mechs(bot, trigger):
...
         password = bot.config.core.auth_password
         mech = bot.config.core.auth_target
     elif bot.config.core.server_auth_method == 'sasl':
         password = bot.config.core.server_auth_password
         mech = bot.config.core.server_auth_sasl_mech
+
+    if mech is None:
+        mech = 'PLAIN'
+    else:
+        mech = mech.upper()
```
Exirel, line 1164: Sometimes I prefer one-liner for if/else:

```suggestion
    mech = 'PLAIN' if mech is None else mech.upper()
```

Or:

```suggestion
    mech = mech is None and 'PLAIN' or mech.upper()
```

sopel/irc/__init__.py
```diff
@@ -153,6 +153,8 @@ def get_irc_backend(self):
             if has_ssl:
                 backend_class = SSLAsynchatBackend
                 backend_kwargs.update({
+                    'certfile': self.settings.core.sasl_cert_file,
```
Exirel: It feels a bit wrong to have a "sasl" config related to a socket connection config.

sopel/irc/backends.py
```diff
@@ -278,10 +283,14 @@ def handle_connect(self):
         # version(s) it supports.
         if not self.verify_ssl:
             self.ssl = ssl.wrap_socket(self.socket,  # lgtm [py/insecure-default-protocol]
```
Exirel, line 285: I feel like it's time to fix that LGTM issue while we are playing with certfile.

sopel/config/core_section.py
half-duplex: Eugh. That's fair given the downright depressing state of updates/security on most networks, but I'd still kind of like to provide a reasonable default. "You can generate a certificate with `...`, but you should check your network's instructions for setup instructions and any special certificate requirements", since many networks won't have good documentation, or it'll be unnecessarily insecure, generate separate key/cert files, etc.
RSA4096 is secure and universal, SHA256 is secure and... widespread ☹️, and I can't imagine a reason for requiring short expiry of self-signed self-disenrolled difficult-to-automate certificates.

half-duplex:
References:
[IRCv3.2 SASL spec](http://ircv3.net/specs/extensions/sasl-3.2.html)
[SASL EXTERNAL RFC](https://datatracker.ietf.org/doc/html/rfc4422#appendix-A)

sopel/config/core_section.py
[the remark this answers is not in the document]
half-duplex: "Directly verify client certs" meaning that the IRCd itself ensures a client cert is signed by a CA it trusts, and if not, it disconnects. The `cafile` parameter in [inspircd3 ssl_gnutls.so](https://docs.inspircd.org/3/modules/ssl_gnutls/) implies that's supported. In these cases, SASL isn't necessarily involved at all, so the `sasl_` config name may not have made sense.

sopel/irc/backends.py
half-duplex: That might be better with other changes (dropping `ssl.wrap_socket()`, etc) that'd make a new PR.

dgw: I did notice the opportunity for switching from `ssl.wrap_socket()` to using an `SSLContext()`. Agree that such a refactoring should come separately (and not just because I don't feel like doing it right now 😝).

sopel/config/core_section.py
dgw: I also don't want to make it too easy for users who have no idea what they're doing (artist's rendition of one below) to make use of this feature. The ones who know what's up will have already visited their network's CertFP instructions to set it up for themselves, likely before checking to see if their bot (Sopel) supports it—perhaps even already generated a certificate, and just need to know how to use it.

<details>
<summary>Artist's rendition</summary>
clueless user
</details>

sopel/config/core_section.py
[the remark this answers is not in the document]
dgw: Fair. I think one of the best ways to find out if any networks validate client certs against CA roots is to ship this feature and move on to other projects, then wait for bug reports about it not working in that case. 😏

dgw:
@half-duplex re: sending commands as Sopel, the `.say` and `.me` commands (in `admin`) could use a sibling that handles raw commands. Maybe `.sendraw` or just `.raw`?

half-duplex:
Heh, this isn't the first time this has come up, I guess it is needed.
```irc
2019-03-20 06:41:51      ma1    idk if there's a .quote or .raw or something
2019-03-20 06:43:31     +dgw    hmm, feels like there should be
```

half-duplex:
Actually, thinking about it... `.say NickServ CERT ADD` is entirely sufficient, no need for a raw message. The half of the thought that isn't currently handled gracefully is that there's no way to have the bot send NickServ's responses back to you, either over IRC or in the terminal... :thinking:

sopel/irc/backends.py
```diff
@@ -260,12 +260,17 @@ class SSLAsynchatBackend(AsynchatBackend):
                             (default ``True``, for good reason)
     :param str ca_certs: filesystem path to a CA Certs file containing trusted
                          root certificates
+    :param str certfile: filesystem path to a certificate for SASL EXTERNAL
+                         authentication (CERTFP)
```
half-duplex, line 264: This is internal, but should probably also refer to it primarily as a client certificate.

half-duplex (commented):
The docstrings for `auth_target` and `server_auth_sasl_mech` should also have `EXTERNAL` added as a valid option. Maybe an example in configuration.rst too?

dgw:
@half-duplex Both of those comments should be addressed by the newest push.

half-duplex:
Well, here's a fun one: `sasl` is considered server-based, so that example won't work.
Does SASL make sense as server-based? Why do we distinguish between "server-based" and "nick-based"? Could we just allow an arbitrary set of authentications?
If you want to leave that for another time, I guess I'd just use a single-auth example.

dgw:
Hmm, I should probably just leave only the single-stage examples for CertFP.

> Why do we distinguish between "server-based" and "nick-based"? Could we just allow an arbitrary set of authentications?

This is something Exi and I talked about fairly recently. At some point, it would make a ton of sense to rewrite Sopel's auth system to be a bit more flexible. The context was more about adding pluggability, but I imagine that simplification of the settings would come along with it. No one's going to promise that for any particular timeframe, though, so I think it's best to ship CertFP as-is once we iron out the doc details.

True, this doesn't _fix_ the confusing auth crap, but it also doesn't make things appreciably worse. `client_cert_file` does its thing regardless of the SASL config, unless the _network_ chooses to require SASL EXTERNAL before authenticating the user.

half-duplex (approved):

PR #2103, opened 2021-06-10
safety: prevent UnicodeDecodeError on startup under py2

dgw:
### Description
One single comment in the StevenBlack list contains one single Unicode character that the 'ascii' codec doesn't like.

This will never be necessary on the master branch, which is about to permanently remove support for Python 2; py3 uses `utf-8` by default in its `str` constructor.

### Checklist
  - It pained me to run Sopel under py2 again, but I did it for science.

### DIFF
--- a/sopel/modules/safety.py
+++ b/sopel/modules/safety.py
@@ -108,7 +108,13 @@ def setup(bot):
         _download_domain_list(loc)
     with open(loc, 'r') as f:
         for line in f:
-            clean_line = unicode(line).strip().lower()
+            if sys.version_info.major < 3:
+                line = unicode(line, 'utf-8')
+            else:
+                line = unicode(line)
+
+            clean_line = line.strip().lower()
+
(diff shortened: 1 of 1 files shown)


Exirel (dismissed):

dgw:
This is merged, but GitHub doesn't know about it yet because I merged to a staging branch. Hopefully tomorrow, all items for 7.1.1 will be ready to go and I'll fast-forward the maintenance branch so Travis can test everything at once. (Also hopefully there will be no failed tests, since everything has been fine on `master`.)

dgw:
The previous patch actually failed under Python 3, which I'd never tested, with `TypeError: decoding str is not supported`. Rather than try to get fancy and write something that _will_ work on both platforms, I just stuck the offending operation in an `if`/`else` and called it a day.

Exirel (approved):

Exirel:
> and called it a day

This will do.

PR #2105, opened 2021-06-10
docs: config/auth: clarify auth_username, etc

half-duplex:
### Description
Add an example for having Sopel authenticate to ZNC, inspired by a user in IRC confused by `auth_username` not working with `auth_method=server`.

Also corrects/clarifies/creates a few other things:
- auth_password: used for server auth too, not just user
- please click the directives, I don't want to have to remember to update things in two places
- auth_method=userserv also uses auth_target
- when is auth_username ignored/required (previously: infinite docloop)

### Checklist

### DIFF
--- a/docs/source/configuration.rst
+++ b/docs/source/configuration.rst
@@ -399,11 +399,13 @@ And for **nick-based** methods:
 * ``Q``
 * ``userserv``
 
-These additional options can be used to configure the authentication method
-and the required credentials:
+Several additional options can be used to configure the authentication method
+and the required credentials. You can follow the link for each to find more
+details:
 
-* :attr:`~CoreSection.auth_username`: account's username, if required
-* :attr:`~CoreSection.auth_password`: account's password
+* :attr:`~CoreSection.auth_username`: account's username, if used by
+  the ``auth_method``
+* :attr:`~CoreSection.auth_password`: password for authentication
 * :attr:`~CoreSection.auth_target`: authentication method's target, if required
   by the ``auth_method``:
 
@@ -428,6 +430,16 @@ And here is an example of server-based authentication using SASL::
     auth_username = BotAccount     # your bot's username
     auth_password = SopelIsGreat!  # your bot's password
     auth_target = PLAIN            # default sasl mechanism
+
+Example of authentication to a ZNC bouncer::
+
+    [core]
+    auth_method = server           # select server-based auth
+    # auth_username is not used with server authentication, so instead
+    # we combine the ZNC username, network name, and password here:
+    auth_password = Sopel/libera:SopelIsGreat!
+
+Don't forget to configure your ZNC to log in to the real network!
 
 
 Multi-stage
--- a/sopel/config/core_section.py
+++ b/sopel/config/core_section.py
@@ -173,11 +173,12 @@ class CoreSection(StaticSection):
 
     :default:
         * ``NickServ`` if using the ``nickserv`` :attr:`auth_method`
+        * ``UserServ`` if using the ``userserv`` :attr:`auth_method`
         * ``PLAIN`` if using the ``sasl`` :attr:`auth_method`
 
-    The nickname of the NickServ service, or the name of the desired SASL
-    mechanism, if :attr:`auth_method` is set to one of these methods. This value
-    is otherwise ignored.
+    The nickname of the NickServ or UserServ service, or the name of the
+    desired SASL mechanism, if :attr:`auth_method` is set to one of these
+    methods. This value is otherwise ignored.
 
     See :ref:`Authentication`.
     """
@@ -185,7 +186,10 @@ class CoreSection(StaticSection):
     auth_username = ValidatedAttribute('auth_username')
     """The user/account name to use when authenticating.
 
-    May not apply, depending on :attr:`auth_method`. See :ref:`Authentication`.
+    Required for an :attr:`auth_method` of ``authserv``, ``Q``, or
+    ``userserv`` — otherwise ignored.
+
+    See :ref:`Authentication`.
     """
 
     auto_url_schemes = ListAttribute(


half-duplex:
For 8.0:
> <+dgw> Looks like clients that offer separate username & password fields for bouncers, like Colloquy, just concatenate the two together with a : for the PASS command.

I don't think it's worth adding `auth_method=bouncer` or anything, but can we think of a case where `if auth_method==server and auth_username: auth_password=auth_username+":"+auth_password` would cause problems?

docs/source/configuration.rst
```diff
@@ -413,6 +414,8 @@ and the required credentials:
   * ``userserv``: the service's nickame to send credentials to
     (``UserServ`` by default)
 
+Click a directive name above for more details about the available options.
```
Exirel, line 417: You'll confuse the hell out of beginners who'll try to understand what a "directive" is. Also, most people will stop their reading at the list, I'd suggest to update the line *before*:

> These additional options can be used to configure the authentication method
> and the required credentials (you can follow the link on each option for more details):

Or something similar.

docs/source/configuration.rst
```diff
@@ -429,6 +432,15 @@ And here is an example of server-based authentication using SASL::
     auth_password = SopelIsGreat!  # your bot's password
     auth_target = PLAIN            # default sasl mechanism
 
+Example of authentication to a ZNC bouncer::
+
+    [core]
+    auth_method = server           # select server auth
```
Exirel, line 438:
```suggestion
    auth_method = server           # select server-based auth
```

dgw (commented):
The only thing I don't understand about this patch is the addition of "allowed or" in `auth_username`'s description.

docs/source/configuration.rst
```diff
@@ -429,6 +432,15 @@ And here is an example of server-based authentication using SASL::
...
+    # auth_username can't be used with server authentication, so instead
```
Exirel, line 439: Can't, or isn't? It isn't clear to me why.

I think making it more explicit by saying "Sopel doesn't use auth_username with `server` auth method." would make it clear as to **why** the user **cannot** use `auth_username`: because it **isn't** used by Sopel.

The rest of the sentence "so instead [...]" looks very good, no comment on that part.

docs/source/configuration.rst
```diff
@@ -429,6 +432,15 @@ And here is an example of server-based authentication using SASL::
...
+    # we combine the ZNC username, network name, and password here:
+    auth_password = Sopel/libera:SopelIsGreat!
+    # Configure your ZNC to handle authentication for the real network.
```
Exirel, line 442: There is no option left, and this is part of the config. It feels a bit weird to me, like: is there something missing? Do I need to add an option in here? The example makes it look like there should be something here, even tho I think that's not what you intended.

sopel/config/core_section.py
```diff
@@ -173,19 +173,23 @@ class CoreSection(StaticSection):
...
 
     See :ref:`Authentication`.
     """
 
     auth_username = ValidatedAttribute('auth_username')
     """The user/account name to use when authenticating.
 
-    May not apply, depending on :attr:`auth_method`. See :ref:`Authentication`.
+    Ignored for an :attr:`auth_method` of ``sasl``, ``server``, or
+    ``nickserv``, required for ``authserv``, ``Q``, or ``userserv``.
```
Exirel, line 190: A bit of a nitpicking here:

```suggestion
    Required by the following auth methods: ``authserv``, ``Q``, or
    ``userserv``; ignored otherwise.
```

And maybe dgw will prefer this version:

```suggestion
    Required by the following auth methods: ``authserv``, ``Q``, or
    ``userserv``—ignored otherwise.
```

Exirel (changes requested):
Happy to see more/better documentation!

docs/source/configuration.rst
half-duplex: It _can't_, without inferring something like the colon-concatenation. The IRC protocol does not allow specifying an "authentication username", only your `USER`, `NICK`, and `PASS`, which are used for other things and configured elsewhere.

docs/source/configuration.rst
half-duplex: Intended as another nudge for "these are not your NickServ credentials, there is no place for those here, go put them in ZNC settings".

half-duplex:
dgw, I phrased it badly, but the intent was to remove the implication that "it's allowed, and sometimes it's required" since it's never optional.
Maybe we should log a warning when a username is provided with a method that ignores it?

docs/source/configuration.rst
Exirel: Then I confirm: "can't" is a consequence of the rule, not the rule itself, which is: it *is not* used.

```suggestion
    # auth_username is not used with server authentication, so instead
```

docs/source/configuration.rst
```diff
@@ -429,6 +431,15 @@ And here is an example of server-based authentication using SASL::
...
     auth_target = PLAIN            # default sasl mechanism
 
+Example of authentication to a ZNC bouncer::
+
+    [core]
+    auth_method = server           # select server-based auth
+    # auth_username can't be used with server authentication, so instead
+    # we combine the ZNC username, network name, and password here:
+    auth_password = Sopel/libera:SopelIsGreat!
+    # Don't forget to configure your ZNC to log in to the real network!
```
Exirel, line 441: Alright, it seems my comment wasn't clear: this part should not be within the example, as it may imply that it is expected to put more configuration *within the same file*:

```suggestion

Don't forget to configure your ZNC to log in to the real network!
```

Exirel (approved):

Exirel (approved):

dgw (approved):
I had already marked these changes as viewed, so I don't know why there wasn't also a (re-)review to go with that. Oh well, squash away.

PR #2112, opened 2021-06-20
docs: be even more specific that `owner` is ALWAYS required

dgw:
### Description
Noticed that while the docstring for `core_section.CoreSection.owner` specifies that it's always required, the "Owners & Admins" docs chapter didn't say anything about it, which could lead users to be confused.

### Checklist
  - Doesn't touch any `.py` files, so I didn't run those.
  - N/A, like the above

### Notes
Don't know yet if we'll need a 7.1.2 release, but I'm sure it'll happen. This is the first PR for it, apparently.

### DIFF
--- a/docs/source/configuration.rst
+++ b/docs/source/configuration.rst
@@ -136,6 +136,13 @@ A Sopel instance must have exactly one owner. This is configured either by
 or by :attr:`~CoreSection.owner`. If ``owner_account`` is set, ``owner`` will
 be ignored.
 
+.. important::
+
+    Even if ``owner_account`` is set, ``owner`` **must** still have a value.
+
+    This is silly, we know. The plan for Sopel 8 is to require only one value
+    or the other.
+
 The same instance can have multiple admins. Similarly, it can be configured
 by :attr:`~CoreSection.admin_accounts` or by :attr:`~CoreSection.admins`. If
 ``admin_accounts`` is set, ``admins`` will be ignored.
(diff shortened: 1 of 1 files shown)


Exirel:
Hm. I really don't like that. I mean, for 7.1.2 I'm fine, it's fixing a problem where we can fix it without breaking change.

However, I don't like forcing owner to set a config value that is ignored, it feels a bit weird. I'll see if I can think of something for Sopel 8.

docs/source/configuration.rst
```diff
@@ -155,6 +155,9 @@ Example owner & admin configurations::
     admin_accounts =
             Exirel
             HumorBaby
+    # ignored when owner_account is set,
+    # but MUST NOT be empty
+    owner = DoesNotMatter
```
Exirel, line 160:
```suggestion
    owner = dgw
```

I'd not let people put a "wrong" owner name, they might trip on that latter. You know how people work: they don't. They are broken and they make the most infuriating errors all the times. Let's not help them doing more harm.

Even tho I totally get the "DoesNotMatter". 😁

docs/source/configuration.rst
```diff
@@ -134,7 +134,7 @@ Owner & Admins
 A Sopel instance must have exactly one owner. This is configured either by
 :attr:`~CoreSection.owner_account` if the IRC server supports IRCv3 accounts,
 or by :attr:`~CoreSection.owner`. If ``owner_account`` is set, ``owner`` will
-be ignored.
+be ignored (but ``owner`` **must** still have a value).
```
Exirel, line 137: Maybe we should add a `.. note:` or `.. important:` here?

Exirel (changes requested):

dgw:
> I don't like forcing owner to set a config value that is ignored, it feels a bit weird.

Hard agree. I'm sure you already know this was never intended to be the final word, just a clarification for our stable branch.

To that end, you might like what I did with the `.. important::` block. Push incoming; rendered preview below.

image

Exirel (approved):
Squash away!

PR #2114, opened 2021-06-21
config: require `owner` only if no `owner_account` is set

dgw:
### Description
This shifts validation of the `core.owner` setting from the base `StaticSection` class to `CoreSection`, by unmarking it as required (removing `default=NO_DEFAULT`) and checking the values after loading and validating the section has succeeded via `super()`.

Comes with its own tests to validate the new behavior (a habit we should continue trying to keep up).

### Checklist

### Notes
Will need to undo #2112 once that's merged, and remove any other references to `owner` being required no matter what.

Would have loved to raise `ConfigurationError` instead of `ValueError`, but `config` imports `config.core_section`:

```
ImportError: cannot import name 'ConfigurationError' from partially initialized module 'sopel.config' (most
likely due to a circular import) (/mnt/c/Users/dgw/github/sopel/sopel/config/__init__.py)
```

### DIFF
--- a/sopel/config/core_section.py
+++ b/sopel/config/core_section.py
@@ -9,7 +9,6 @@ from sopel.config.types import (
     ChoiceAttribute,
     FilenameAttribute,
     ListAttribute,
-    NO_DEFAULT,
     SecretAttribute,
     StaticSection,
     ValidatedAttribute,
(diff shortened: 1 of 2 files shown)


dgw:
We actually have to keep requiring `core.owner` be set to the owner's nickname, even if `core.owner_account` is being used for auth. Sopel itself uses `core.owner` as the recipient for important messages/notices such as when a new version is available, or if something goes wrong during a netsplit and the bot loses its main nick. (Some third-party plugins surely also do similar things.)

I'm reworking the docs again instead, on top of the just-merged #2112; too bad I didn't think of this until after merging that.

PR #2117, opened 2021-06-25
Replace (almost) all usage of `sopel.module` with `sopel.plugin`

dgw:
### Description
Tin. `coretasks` and all tests (except `test_module.py`, for obvious reasons) now use the members of `sopel.plugin` only.

One more step on the road to #1738. This isn't actually part of the task list there, but it still feels good.

### Checklist

### Notes
Hopefully will not conflict with #2116, which already uses `plugin` for all its newly added decorator calls.

### DIFF
--- a/sopel/coretasks.py
+++ b/sopel/coretasks.py
@@ -32,7 +32,7 @@ import re
 import sys
 import time
 
-from sopel import loader, module, plugin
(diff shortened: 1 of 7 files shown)


Exirel (approved):
This does feel good.

PR #2118, opened 2021-06-26
cli: remove legacy mode for run script

### LINKED ISSUE #1584: cli: legacy argument before subcommand breaks parser
The current usage specifies: 
```
usage: sopel.py [options]                                                      
                                                                               
Sopel IRC Bot                                                                  
                                                                               
optional arguments:                                                            
 ...              
  -c filename, --config filename                                               
                        Use a specific configuration file                      

sub-commands:                                           
  List of Sopel's sub-commands                          
                                                        
  {start,configure,stop,restart}                        
    start               Start a Sopel instance          
    configure           Sopel's Wizard tool             
    stop                Stop a running Sopel instance   
    restart             Restart a running Sopel instance
```

If a user provides an option, then also specifies a subcommand (e.g., `sopel -c my_config.cfg start`), the parser may get confused and give an error like: <pre>error: unrecognized arguments: <i>name_of_subcommand</i></pre>

@Exirel pointed out on IRC that:
```
> it is due to Python 2's argparse. 
> to be more accurate: it's because of how argparse's subparsers in Python 2 are always required
> and they are not in Python 3
> so if you want (almost) the same behavior in both, you need to always provide a sub-parser action
> which you do only if someone provides an option first
```

Essentially, this "bug" (not-a-bug)™<sup> @dgw</sup>, is only due to legacy support that is being provided until (no later than) Sopel v8.0.0, at which point Sopel will stop supporting Python 2. 
> _Then, all of our problems will be solved_ :tada:
> --  _someone (probably...)_

Exirel:
### Description

Fix  #1584 by removing the legacy parser completely, fixing the doc and the help message. Options are no longer available without a subcommands, which is a breaking change. However, it makes everything else better (the doc, the usage, the code etc.) and it's Sopel 8 and it's the right way to do a command line tool.

### Checklist

### DIFF
--- a/sopel/cli/run.py
+++ b/sopel/cli/run.py
@@ -471,8 +471,20 @@ def main(argv=None):
 
     :param list argv: command line arguments
     """
-    # Step One: Parse The Command Line
+    # Build parser and handle default command
+    global_options = ['-h', '--help', '-V', '--version']
     parser = build_parser()
+
+    argv = argv or sys.argv[1:]
+    if not argv:
+        # No argument: assume start sub-command
+        argv = ['start']
+
+    elif argv[0].startswith('-') and argv[0] not in global_options:
+        # No sub-command and no global option
+        argv = ['start'] + argv
+
+    # Step One: Parse The Command Line
     opts = parser.parse_args(argv)
 
     # Handle "-V/--version" option
@@ -499,6 +511,7 @@ def main(argv=None):
         return command(opts)
     except KeyError:
         parser.print_usage()
+        return ERR_CODE
     except KeyboardInterrupt:
         print("\n\nInterrupted")
         return ERR_CODE
--- a/sopel/cli/run.py
+++ b/sopel/cli/run.py
@@ -484,7 +484,7 @@ def main(argv=None):
         # No sub-command and no global option
         argv = ['start'] + argv
 
-    # Step One: Parse The Command Line
+    # Parse The Command Line
     opts = parser.parse_args(argv)
 
     # Handle "-V/--version" option
@@ -493,14 +493,10 @@ def main(argv=None):
         return
 
     try:
-        # Step Two: "Do not run as root" checks
-        try:
-            check_not_root()
-        except RuntimeError as err:
-            tools.stderr('%s' % err)
-            return ERR_CODE
+        # Check "Do not run as root"
+        check_not_root()
 
-        # Step Three: Handle command
+        # Select command
         action = getattr(opts, 'action', None)
         command = {
             'start': command_start,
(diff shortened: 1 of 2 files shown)


Exirel:
Note: this should wait #2062 to be merged first, with a rebase, before merge.

dgw:
Have not dug in too deep yet, but would really like to keep `sopel start` as the default behavior. Bare `sopel` doesn't have to support arguments, but as so many people use only one instance with the default config file, taking away that convenience likely wouldn't go over well.

sopel/cli/run.py
```diff
@@ -703,15 +487,16 @@ def main(argv=None):
             return ERR_CODE
 
         # Step Three: Handle command
-        action = getattr(opts, 'action', 'legacy')
+        action = getattr(opts, 'action', None)
```
dgw, line 490:
```suggestion
        action = getattr(opts, 'action', 'start')
```

There was even a comment in the old version of this file:
``# TODO: `start` should be the default in Sopel 8``

dgw (commented):
Just following up on my previous comment now that I've looked at the changes properly.

sopel/cli/run.py
Exirel: I'll get back to that to fix that problem. The extra work required here is worth it.

Exirel:
> Bare `sopel` doesn't have to support arguments

That was way easier thanks to that. And so, it is done!

sopel/cli/run.py
```diff
@@ -544,159 +466,33 @@ def command_restart(opts):
...
-        # TODO: `start` should be the default in Sopel 8
-        argv = argv or sys.argv[1:]
-        if not argv:
-            argv = ['legacy']
-        elif argv[0].startswith('-') and argv[0] not in ['-h', '--help']:
-            argv = ['legacy'] + argv
-
-        opts = parser.parse_args(argv)
-
         # Step Two: "Do not run as root" checks
```
dgw, line 496: Step Two, the root check, doesn't need to be inside the outer `try`, does it? It's a perfect opportunity to flatten the nesting structure that, AFAICT, exists for no good reason. Moving this check to happen before handling of `--version` would match the current (7.1.x) behavior.

For previously-discussed (many times) GitHub reasons, I can't select and attach this note to the whole block. Two lines in the middle that are hidden in the diff view prevent doing so.

dgw (changes requested):
Just one move as long as this refactor (really, cleanup) is happening, to preserve original behavior.

I checked a number of things that were removed to make sure they still exist elsewhere, and didn't find any missing. We had a lot of redundancy from keeping this around, eh?

dgw (approved):
Let's ship it, sure. Plenty of time to catch omissions during feature work.

PR #2049, opened 2021-03-30
bot: deprecate (un)register_url_callback() functions

dgw:
### Description
Tin. They were a stopgap, and they have served their purpose.

### Checklist
- [x] I have read [CONTRIBUTING.md](https://github.com/sopel-irc/sopel/blob/master/CONTRIBUTING.md)
- [x] I can and do license this contribution under the EFLv2
- [x] No issues are reported by `make quality` (did not touch anything testable)
- [x] I have tested the functionality of the things this change touches

PR #2121, opened 2021-06-27
bot: remove `bot.memory['url_callbacks]` and use internal state

dgw:
### Description
Plugins doing this will receive warnings through the entire remaining lifetime of 7.x (at least several months, perhaps up to a year-ish), and it's high time to push plugin devs away from using anything but the `plugin.url` and `plugin.url_lazy` decorators.

As pointed out in a comment below, also deprecates `bot.search_url_callbacks()` (which should have been deprecated along with its register/unregister siblings).

### Checklist
  - Or rather, we have already tested that the functionality replacing it works as intended.

### Notes
Sopel 9.0 will drop the `(un)register_url_callback` functions, finally ending the era of plugins manually managing their own URL callbacks. That will feel _almost_ as good as dumping Python 2. 😁

This could have been one commit, and I didn't test the intermediate state. Initially I was only going to remove the stuff that said it could be removed in 8.0, then remembered that there was more to that story.

### DIFF
--- a/sopel/bot.py
+++ b/sopel/bot.py
@@ -53,6 +53,15 @@ class Sopel(irc.AbstractBot):
         self._rules_manager = plugin_rules.Manager()
         self._scheduler = plugin_jobs.Scheduler(self)
 
+        self._url_callbacks = tools.SopelMemory()
+        """Tracking of manually registered URL callbacks.
+
+        Should be manipulated only by use of :meth:`register_url_callback` and
+        :meth:`unregister_url_callback` methods, which are deprecated.
+
+        Remove in Sopel 9, along with the above related methods.
+        """
+
         self._times = {}
         """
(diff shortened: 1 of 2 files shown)


sopel/bot.py
```diff
@@ -1193,6 +1177,11 @@ def search_url_callbacks(self, url):
 
         .. versionadded:: 7.0
 
+        .. versionchanged:: 8.0
+
+            Searches for registered callbacks in an internal property instead
```
Exirel, line 1182: Wait a second... shouldn't this be deprecated and removed in Sopel 9, like the other methods?

dgw: Probably. Must have been an oversight in #2049.

dgw:
_Note: 3986227c4971e68530ae07113426187e95967046 had unwanted side effects: some parts of core still use it, making the deprecation warning spammy. See #2156._

PR #2125, opened 2021-06-28
docs: clarify relationship between `owner` and `owner_account` (again)

dgw:
### Description
We were going to make `core.owner` optional in Sopel 8, but that's not wise. The setting is used to send important notices to the bot's owner, and so it has to be set with a proper value even if `core.owner_account` is being used for verification (e.g. `plugin.require_owner`).

### Checklist

### Notes
**Can** be merged with a `[skip ci]`, as it only touches docs.
**Should** be cherry-picked back to 7.1.x with a `[skip ci]` so as not to burn Travis credits.

### DIFF
--- a/docs/source/configuration.rst
+++ b/docs/source/configuration.rst
@@ -144,15 +144,18 @@ Example owner & admin configurations::
 
     # Using nickname matching
     [core]
-    owner = dgw        # used for alerts and ownership verification
+    # Will be used for alerts and ownership verification
+    owner = dgw
     admins =
             Exirel
             HumorBaby
 
     # Using account matching
     [core]
-    owner = dgw                   # used for alerts only
-    owner_account = dgws_account  # used for ownership verification
+    # Will be used for alerts only
+    owner = dgw
+    # Will be used for ownership verification
+    owner_account = dgws_account
     admin_accounts =
             Exirel
             HumorBaby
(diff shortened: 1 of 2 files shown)


docs/source/configuration.rst
```diff
@@ -151,25 +144,30 @@ Example owner & admin configurations::
 
     # Using nickname matching
     [core]
-    owner = dgw
+    owner = dgw        # used for alerts and ownership verification
```
Exirel, line 147: Since I've seen people on IRC using our example directly, they had an issue with such comments: the config parser thinks that anything after the `=` is part of the content, **including** what we *think* is a comment, but isn't for the config parser!

So yeah, this line, for the config parser, means this:

```
>>> bot.settings.core.owner
'dgw # used for alerts and ownership verification'
```

I think we should not show example with this syntax, as it will (and it already did) confuse some readers.

Exirel (changes requested):
Inline comment in config example are a problem now. 😭

docs/source/configuration.rst
dgw: Phooey. If this was for 8.0 I'd say let's tweak the parser behavior (it's totally possible to support this syntax), but the patch is destined for 7.1.x maintenance…

I'll have to make a pass through the docs converting this style to comments on their own line(s). Not a big deal to fold it into this PR, but frustrating.

Exirel (approved):
👍

PR #2127, opened 2021-06-28
tools.events: convert to `Enum`

dgw:
### Description
Anyone using these as intended should see no ill effects, but using an `Enum` class will actively thwart plugins that try to do something silly like override a numeric name.

Really, it's the same thing as #2122 but for the numeric events instead.

### Checklist

### DIFF
--- a/sopel/tools/_events.py
+++ b/sopel/tools/_events.py
@@ -1,7 +1,9 @@
 from __future__ import generator_stop
 
+from enum import Enum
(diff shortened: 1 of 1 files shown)


Exirel (approved):

Exirel (approved):
Alright, I had second though about the whole `str, Enum` thing. I'm still not a big fan, but I get it.

PR #2130, opened 2021-06-30
pronouns: fetch list from the pronoun.is GitHub repo at start

dgw:
### Description
We don't have to wait until the 12th of Never for pronoun.is to have an API. Parsing tab-separated values is easy, and the file is already sorted according to approximate frequency for us according to their readme.

Includes a kill-switch config setting in case the repo ever gets deleted or reorganized, or the file format changes.

### Checklist

### Notes
Speaking of something that's over-complicated for what it does (as Exi said on IRC yesterday)… this also finds the shortest unique set of pronouns for each option to use in links. The only outlier cases now are `they/them/their/theirs/themsel(ves|f)`, which won't be abbreviated as `they/.../themsel(ves|f)`. I might or might not find the motivation to fix that.

### DIFF
--- a/sopel/modules/pronouns.py
+++ b/sopel/modules/pronouns.py
@@ -12,12 +12,32 @@ import logging
 import requests
 
 from sopel import plugin
+from sopel.config import types
 
 
 LOGGER = logging.getLogger(__name__)
 
 
+class PronounsSection(types.StaticSection):
+    fetch_complete_list = types.BooleanAttribute('fetch_complete_list', default=True)
+    """Whether to attempt fetching the complete list pronoun.is uses, at bot startup."""
+
+
+def configure(settings):
+    """
+    | name | example | purpose |
+    | ---- | ------- | ------- |
+    | fetch_complete_list | True | Whether to attempt fetching the complete pronoun list from pronoun.is at startup. |
+    """
+    settings.define_section('pronouns', PronounsSection)
+    settings.pronouns.configure_setting(
+        'fetch_complete_list',
+        'Fetch the current pronoun.is list at startup?')
+
+
 def setup(bot):
+    bot.config.define_section('pronouns', PronounsSection)
+
     # Copied from pronoun.is, leaving a *lot* out.
     # If ambiguous, the earlier one will be used.
     # This basic set is hard-coded to guarantee that the ten most(ish) common sets
@@ -34,6 +54,9 @@ def setup(bot):
         'it/it': 'it/it/its/its/itself',
         'ey/em': 'ey/em/eir/eirs/eirself',
     }
+
+    if not bot.config.pronouns.fetch_complete_list:
+        return
 
     # and now try to get the current one
     # who needs an API that might never exist?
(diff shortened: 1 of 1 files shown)


sopel/modules/pronouns.py
```diff
@@ -7,25 +7,133 @@
...
+
+            # go to the next node
+            cur = cur.children[el]
+
+    def insert_list(self, set_list):
+        """Load a list of pronoun sets all at once."""
+        for item in set_list:
+            self.insert(item)
+
+    def get_pairs(self, root=None, prefix=''):
```
Exirel, line 125: I think I hate the complexity of this method.

sopel/modules/pronouns.py
```diff
@@ -78,30 +186,30 @@ def set_pronouns(bot, trigger):
...
+    if len(split_request) < 5:
         matching = []
-        for known_pronoun_set in KNOWN_SETS.values():
-            known_pronoun_split = known_pronoun_set.split("/")
-            if known_pronoun_set.startswith(pronouns + "/") or (
-                len(requested_pronoun_split) == 3
+        for known_set in bot.memory['pronoun_sets'].values():
+            split_set = known_set.split("/")
+            if known_set.startswith(pronouns + "/") or (
+                len(split_request) == 3
```
Exirel, line 195: I don't really understand the change in names here. Sure, I don't really like "known_pronoun_set", I'd just use "known_pronouns".

Exirel: And yeah, both "split_request" and "request_pronoun_split" tries to encode too much information in one name. I'd just go with "requested_pronoun". 🤷

Exirel (changes requested):
I don't really get the tree aspect of the pronoun parsing. I feel like I'm missing the problem that required a solution.

I also tried to run the code on the current pronoun file, and the result is quite different from the default set of pronouns. For example, instead of `'they/.../themselves'`, the parser returns `'they/them/their/theirs/themselves'`. So, is it compatible? Why use `...` in the default value, knowing that the actual file don't contain them? Especially given that the file will replace the default value, so if there is a feature here, it'll be made inaccessible.

A lot of confusion!

dgw:
> I don't really get the tree aspect of the pronoun parsing. I feel like I'm missing the problem that required a solution.

Using a trie was to make it easier to match on prefixes. I think. It's been a few months. 😅

I've <em>trie</em>d (😏) to address most of the style comments in fixup commits. Haven't gone through another round of testing just yet.

Exirel (approved):
I still think this is a lot of work that I don't know how to justify. On the other hand, I didn't have to do anything, and it all looks good. Maybe I'm just annoyed there isn't any test for that yet, but there wasn't before anyway.

Let's move on.

dgw:
I have cleaned the history (squashing fixups) and fixed one last bug with the disambiguation output (in case the user specifies e.g. `.setpronouns they`, which matches multiple sets).

Testing also determined that it's not necessary to merge the fetched pronouns with the default set; `they/.../themself` (for example) works just fine without that entry existing in the dictionary. I dropped that change.

PR #2129, opened 2021-06-28
bot, coretasks: remove redundant `bot.privileges` data structure

dgw:
### Description
`bot.privileges` has been deprecated for over 5 years. There were some growing pains along the way, which is why `bot.channels` didn't take the front seat as quickly as it could have, but we're ready now.

### Checklist
- [x] I have read [CONTRIBUTING.md](https://github.com/sopel-irc/sopel/blob/master/CONTRIBUTING.md)
- [x] I can and do license this contribution under the EFLv2
- [x] No issues are reported by `make qa` (runs `make quality` and `make test`)
- [x] I have tested the functionality of the things this change touches

PR #2131, opened 2021-07-01
irc: rework parsing of modestrings

Exirel:
### Description

Following the work on #2098 I was able to refactor that part of the modestring parsing I had in mind since a long time ago. My main goal was to extract the parsing logic, and separate it from the application of new modes & privileges to users.

So now there is a `sopel.irc.modes` module that contains only modestring parsing (with the appropriate tests), and this module is then used in the `coretasks` plugin. There is close to 0 change in behavior, but there is at least two I can remember:

1. it now uses `deepcopy` to get the existing channel modes to prevent any mutability issues
2. before, when something went wrong (unknown mode type), `channel.modes = modes` wasn't done; this is now fixed

No tests on coretasks have been modified, although I used them locally to ensure that 1. it worked as intended, and 2. without breaking changes. And yes, at first, they didn't pass, and I had to fix what was broken in my code. So I'm pretty happy about all this. As said on IRC, this wasn't possible without the work of @half-duplex that relentlessly fixed the modeparsing before.

One last thing: I used type hint on `sopel.irc.modes` and validated it with `mypy`: so far, so good. I might do more of that later. I just hope it works fine with Py3.6 because I know for a fact that type hinting changed in later versions. Further conversation need to happen around Py3.10 too, because that version might not be easy to support while supporting 3.6.

### Checklist

### DIFF
--- a/sopel/coretasks.py
+++ b/sopel/coretasks.py
@@ -33,7 +33,7 @@ import time
 
 from sopel import loader, plugin
 from sopel.config import ConfigurationError
-from sopel.irc import isupport, modes as ircmodes
+from sopel.irc import isupport
 from sopel.irc.utils import CapReq, MyInfo
 from sopel.tools import events, Identifier, SopelMemory, target, web
 
@@ -55,8 +55,6 @@ MODE_PREFIX_PRIVILEGES = {
     "y": plugin.OPER,
     "Y": plugin.OPER,
 }
-
-MODE_PREFIXES = set(MODE_PREFIX_PRIVILEGES.keys())
 
 
 batched_caps = {}
@@ -342,6 +340,13 @@ def handle_isupport(bot, trigger):
 
     bot._isupport = bot._isupport.apply(**parameters)
 
+    # update bot's mode parser
+    if 'CHANMODES' in bot.isupport:
+        bot.modeparser.chanmodes = bot.isupport.CHANMODES
+
+    if 'PREFIX' in bot.isupport:
+        bot.modeparser.privileges = set(bot.isupport.PREFIX.keys())
+
     # was BOT mode support status updated?
     if not botmode_support and 'BOT' in bot.isupport:
         # yes it was! set our mode unless the config overrides it
@@ -541,15 +546,7 @@ def _parse_modes(bot, args, clear=False):
         LOGGER.debug(
             "The server sent a possibly malformed MODE message: %r", args)
 
-    privileges = MODE_PREFIXES
-    if 'PREFIX' in bot.isupport:
-        privileges = set(bot.isupport.PREFIX.keys())
-
-    modemessage = ircmodes.ModeParser(
-        bot.isupport.CHANMODES,
-        privileges=privileges,
-    )
-    modeinfo = modemessage.parse_modestring(args[1], tuple(args[2:]))
+    modeinfo = bot.modeparser.parse_modestring(args[1], tuple(args[2:]))
 
     # set or update channel's modes
     modes = {} if clear else copy.deepcopy(channel.modes)
@@ -608,7 +605,7 @@ def _parse_modes(bot, args, clear=False):
         LOGGER.warning(
             "Too many arguments received for MODE: args=%r chanmodes=%r",
             args,
-            modemessage.chanmodes,
+            bot.modeparser.chanmodes,
         )
 
     LOGGER.info("Updated mode for channel: %s", channel.name)
--- a/sopel/coretasks.py
+++ b/sopel/coretasks.py
@@ -558,6 +558,9 @@ def _parse_modes(bot, args, clear=False):
                 modes[mode].add(param)
             elif param in modes[mode]:
                 modes[mode].remove(param)
+                # remove mode if empty
+                if not modes[mode]:
+                    modes.pop(mode)
         elif letter == 'B':
             if is_added:
                 modes[mode] = param
--- a/sopel/coretasks.py
+++ b/sopel/coretasks.py
@@ -46,6 +46,19 @@ CORE_QUERYTYPE = '999'
 Other plugins should use a different querytype.
 """
 
+MODE_PREFIX_PRIVILEGES = {
+    "v": plugin.VOICE,
+    "h": plugin.HALFOP,
+    "o": plugin.OP,
+    "a": plugin.ADMIN,
+    "q": plugin.OWNER,
+    "y": plugin.OPER,
+    "Y": plugin.OPER,
+}
+
+MODE_PREFIXES = set(MODE_PREFIX_PRIVILEGES.keys())
+
+
 batched_caps = {}
 
 
@@ -528,23 +541,13 @@ def _parse_modes(bot, args, clear=False):
         LOGGER.debug(
             "The server sent a possibly malformed MODE message: %r", args)
 
-    mapping = {
-        "v": plugin.VOICE,
-        "h": plugin.HALFOP,
-        "o": plugin.OP,
-        "a": plugin.ADMIN,
-        "q": plugin.OWNER,
-        "y": plugin.OPER,
-        "Y": plugin.OPER,
-    }
-
-    privileges = mapping.keys()
+    privileges = MODE_PREFIXES
     if 'PREFIX' in bot.isupport:
-        privileges = bot.isupport.PREFIX.keys()
+        privileges = set(bot.isupport.PREFIX.keys())
 
     modemessage = ircmodes.ModeParser(
         bot.isupport.CHANMODES,
-        privileges=set(privileges),
+        privileges=privileges,
     )
     modeinfo = modemessage.parse_modestring(args[1], tuple(args[2:]))
 
@@ -585,7 +588,7 @@ def _parse_modes(bot, args, clear=False):
         # User privs modes, always have a param
         nick = Identifier(param)
         priv = channel.privileges.get(nick, 0)
-        value = mapping[privilege]
+        value = MODE_PREFIX_PRIVILEGES[privilege]
         if is_added:
             priv = priv | value
         else:
--- a/sopel/coretasks.py
+++ b/sopel/coretasks.py
@@ -527,7 +527,18 @@ def initial_modes(bot, trigger):
 
 
 def _parse_modes(bot, args, clear=False):
-    """Parse MODE message and apply changes to internal state."""
+    """Parse MODE message and apply changes to internal state.
+
+    Sopel, by default, doesn't know how to parse other types than A, B, C, and
+    D, and only a preset of privileges.
+
+    .. seealso::
+
+        Parsing mode messages can be tricky and complicated to understand. In
+        any case it is better to read the IRC specifications about channel
+        modes at https://modern.ircdocs.horse/#channel-mode
+
+    """
     channel_name = Identifier(args[0])
     if channel_name.is_nick():
         # We don't do anything with user modes
@@ -546,12 +557,16 @@ def _parse_modes(bot, args, clear=False):
         LOGGER.debug(
             "The server sent a possibly malformed MODE message: %r", args)
 
+    # parse the modestring with the parameters
     modeinfo = bot.modeparser.parse(args[1], tuple(args[2:]))
 
-    # set or update channel's modes
+    # set, unset, or update channel's modes based on the mode type
+    # modeinfo.modes contains only the valid parsed modes
+    # coretask can handle type A, B, C, and D only
     modes = {} if clear else copy.deepcopy(channel.modes)
     for letter, mode, is_added, param in modeinfo.modes:
         if letter == 'A':
+            # type A is a multi-value mode and always requires a parameter
             if mode not in modes:
                 modes[mode] = set()
             if is_added:
@@ -562,16 +577,19 @@ def _parse_modes(bot, args, clear=False):
                 if not modes[mode]:
                     modes.pop(mode)
         elif letter == 'B':
+            # type B is a single-value mode and always requires a parameter
             if is_added:
                 modes[mode] = param
             elif mode in modes:
                 modes.pop(mode)
         elif letter == 'C':
+            # type C is a single-value mode and requires a parameter when added
             if is_added:
                 modes[mode] = param
             elif mode in modes:
                 modes.pop(mode)
         elif letter == 'D':
+            # type D is a flag (True or False) and doesn't have a parameter
             if is_added:
                 modes[mode] = True
             elif mode in modes:
@@ -581,6 +599,7 @@ def _parse_modes(bot, args, clear=False):
     channel.modes = modes
 
     # update user privileges in channel
+    # modeinfo.privileges contains only the valid parsed privileges
     for privilege, is_added, param in modeinfo.privileges:
         # User privs modes, always have a param
         nick = Identifier(param)
@@ -598,6 +617,7 @@ def _parse_modes(bot, args, clear=False):
             "Unknown MODE message, sending WHO. Message was: %r",
             args,
         )
+        # send a WHO message to ensure we didn't miss anything
         _send_who(bot, channel_name)
 
     # log leftover parameters (too many arguments)
--- a/sopel/coretasks.py
+++ b/sopel/coretasks.py
@@ -546,7 +546,7 @@ def _parse_modes(bot, args, clear=False):
         LOGGER.debug(
             "The server sent a possibly malformed MODE message: %r", args)
 
-    modeinfo = bot.modeparser.parse_modestring(args[1], tuple(args[2:]))
+    modeinfo = bot.modeparser.parse(args[1], tuple(args[2:]))
 
     # set or update channel's modes
     modes = {} if clear else copy.deepcopy(channel.modes)
--- a/sopel/coretasks.py
+++ b/sopel/coretasks.py
@@ -24,6 +24,7 @@ from __future__ import generator_stop
 
 import base64
 import collections
+import copy
 import datetime
 import functools
 import logging
@@ -32,7 +33,7 @@ import time
 
 from sopel import loader, plugin
 from sopel.config import ConfigurationError
-from sopel.irc import isupport
+from sopel.irc import isupport, modes as ircmodes
 from sopel.irc.utils import CapReq, MyInfo
 from sopel.tools import events, Identifier, SopelMemory, target, web
 
@@ -527,9 +528,6 @@ def _parse_modes(bot, args, clear=False):
         LOGGER.debug(
             "The server sent a possibly malformed MODE message: %r", args)
 
-    modestring = args[1]
-    params = args[2:]
-
     mapping = {
         "v": plugin.VOICE,
         "h": plugin.HALFOP,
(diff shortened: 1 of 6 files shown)


sopel/coretasks.py
```diff
@@ -543,101 +541,91 @@ def _parse_modes(bot, args, clear=False):
...
+    # set or update channel's modes
+    modes = {} if clear else copy.deepcopy(channel.modes)
+    for letter, mode, is_added, param in modeinfo.modes:
+        if letter == 'A':
+            if mode not in modes:
+                modes[mode] = set()
+            if is_added:
+                modes[mode].add(param)
+            elif param in modes[mode]:
+                modes[mode].remove(param)
```
dgw, line 563: What about removing the mode when empty?

Just glancing at this for now. Full review to come after #2129 merges.

Exirel: Ha yes, the PR that will generate merge conflict on mine. 😁 

Yup, could be an option indeed, to remove the mode when empty.

sopel/irc/modes.py
```diff
@@ -0,0 +1,234 @@
...
+
+            >>> mm = ModeMessage({'A': tuple('beI'), 'B': tuple('k')}, {})
+            >>> mm.get_mode_type('b')
+            'A'
+            >>> mm.get_mode_type('k')
+            'B'
+
+        This method will raise a :exc:`ModeTypeUnknown` if the mode is unknown,
+        including the case where ``mode`` is actually a user privilege.
+        (such as ``v``)
```
dgw, line 123:
```suggestion
        including the case where ``mode`` is actually a user privilege such as
        ``v``.
```

sopel/irc/modes.py
```diff
@@ -0,0 +1,234 @@
...
+        """
+        for letter, modes in self.chanmodes.items():
+            if mode in modes:
+                return letter
+        raise ModeTypeUnknown(mode)
+
+    def get_mode_info(
+        self,
+        mode: str,
+        is_added: bool
```
dgw, line 133:
```suggestion
        is_added: bool,
```

We should have a linter rule that enforces such things, shouldn't we?

sopel/irc/modes.py
```diff
@@ -0,0 +1,234 @@
...
+            ('A', True, False)
+            >>> mm.get_mode_info('k', False)
+            ('B', False, False)
+            >>> mm.get_mode_info('v', True)
+            (None, True, True)
+
+        .. note::
+
+            A user privilege ``mode`` doesn't have a type so the first value
+            returned will be ``None`` in that case.
```
dgw, line 161: This is as good a place to ask as any: Is this because user privilege modes are given in `PREFIX` instead of `CHANMODES`, so they're not technically assigned to a category?

I wonder if `None` is really the best way to indicate such a thing, instead of another more meaningful value.

sopel/irc/modes.py
```diff
@@ -0,0 +1,234 @@
...
+
+        if letter not in self.type_params:
+            # we don't know how to handle this type of mode
+            raise ModeTypeImproperlyConfigured(mode, letter)
+
+        type_param = self.type_params[letter]
+        return letter, not type_param == PARAM_NEVER and (
+            type_param == PARAM_ALWAYS
+            or type_param == PARAM_ADDED and is_added
+            or type_param == PARAM_REMOVED and not is_added
```
dgw, line 180: This really makes me think that `type_params` should be a bitfield table. It should be much faster to bitwise-`&` the values than to do a string comparison, in theory.

I also just like having excuses to use bitwise logic…

sopel/irc/modes.py
```diff
@@ -0,0 +1,234 @@
...
+from collections import namedtuple
+import logging
+from typing import Dict, Generator, List, Optional, Set, Tuple
+
+
+LOGGER = logging.getLogger(__name__)
+
+PARAM_ALWAYS = 'always'
+PARAM_ADDED = 'added'
+PARAM_REMOVED = 'removed'
```
dgw, line 20: Is there such a thing? I only know of modes that always require a parameter; modes that need a parameter only when being set/added; and modes that forbid (never have) a parameter.

Sure you want to include code to handle this magical mode type that might never get real-world testing?

Because of this, I'm also wondering why the constructor for `ModeMessage` even needs to have the `type_param` argument passed in. Shouldn't it always be the same, because the types and their parameter requirements are standardized?

sopel/irc/modes.py
```diff
@@ -0,0 +1,234 @@
...
+
+ModeString = namedtuple('ModeString', [
+    'modes',
+    'privileges',
+    'ignored_modes',
+    'leftover_params',
+])
+
+
+class ModeMessage:
```
dgw, line 67: Hear me out: This class should be called `ModeParser` and what is now called `ModeString` should instead be named `ModeMessage`.

It took me until halfway through the class to realize that a `ModeMessage` isn't actually a message—but now that I noticed, I don't like it.

sopel/irc/modes.py
```diff
@@ -0,0 +1,234 @@
...
+        privileges: List = []
+        for mode, is_added in imodes:
+            param = None
+            try:
+                letter, required, is_priv = self.get_mode_info(mode, is_added)
+                if required:
+                    try:
+                        param = next(iparams)
+                    except StopIteration:
+                        # Not enough parameters: we have to stop here
```
dgw, line 204: Should log the event.

Having params left over after consuming all mode characters should also log that event.

sopel/coretasks.py
```diff
@@ -540,82 +538,72 @@ def _parse_modes(bot, args, clear=False):
...
-            if value is not None:
-                if sign == "+":
-                    priv = priv | value
-                else:
-                    priv = priv & ~value
-                channel.privileges[nick] = priv
-            param_idx += 1
+    privileges = mapping.keys()
+    if 'PREFIX' in bot.isupport:
+        privileges = bot.isupport.PREFIX.keys()
```
dgw, line 543: Is Python's bytecode compiler smart enough to optimize this so it won't call `mapping.keys()` unless necessary? I'm wary of overhead for things like this, which could hammer the bot e.g. after a netsplit is resolved.

sopel/coretasks.py
```diff
@@ -540,82 +538,72 @@ def _parse_modes(bot, args, clear=False):
...
+
+    modemessage = ircmodes.ModeMessage(
+        bot.isupport.CHANMODES,
+        privileges=set(privileges),
+    )
+    modeinfo = modemessage.parse_modestring(args[1], tuple(args[2:]))
```
dgw, line 549: This, too, is overhead of which I'm wary—constructing a fresh object every time the bot gets a `MODE` line.

The parameters to this constructor won't change once 005 `RPL_ISUPPORT` lines end, so I think it would be better for the bot to contain its own mode parser object that is rebuilt only when receiving lines that might change its behavior. Maybe you don't want to do it as part of this PR, but that should be a refactor goal.

sopel/coretasks.py
```diff
@@ -540,82 +538,72 @@ def _parse_modes(bot, args, clear=False):
...
+        _send_who(bot, channel_name)
 
-    channel.modes = modes
+    # log leftover parameters (too many arguments)
+    if modeinfo.leftover_params:
+        LOGGER.warning(
+            "Too many arguments received for MODE: args=%r chanmodes=%r",
+            args,
+            modemessage.chanmodes,
+        )
```
dgw, line 606: The mode parser should log this one, not coretasks.

Same could apply to the presence of `ignored_modes`, except coretasks needs that anyway to know that it should `_send_who()`.

dgw (commented):
OK, I finally went through it. Some pretty significant questions wound up in those line notes. I was surprised, not least because I'm not _that_ alert tonight and expected to miss stuff. x)

sopel/irc/modes.py
```diff
@@ -0,0 +1,234 @@
...
+    """Set of user privileges used by default."""
+
+    def __init__(
+        self,
+        chanmodes: Dict[str, Tuple[str, ...]],
+        type_params: Dict[str, str] = DEFAULT_MODETYPE_PARAM_CONFIG,
+        privileges: Set[str] = PRIVILEGES,
+    ) -> None:
+        self.chanmodes: Dict[str, Tuple[str, ...]] = chanmodes
+        """Map of mode types (``str``) with their list of modes (``tuple``).
```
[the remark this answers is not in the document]
Exirel: For each type there is one list of modes. If the grammatically correct syntax is "to their lists", then it is!

sopel/irc/modes.py
```diff
@@ -0,0 +1,234 @@
...
+
+        This map should come from ``ISUPPORT``, usually through
+        :attr:`bot.isupport.CHANMODES <sopel.irc.isupport.ISupport.CHANMODES>`.
```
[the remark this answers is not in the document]
Exirel: > a default mapping containing only the modes present in RFC

Agree. I need to check the RFC for that.

sopel/irc/modes.py
Exirel: Eeyup, that's why: it doesn't have a type. I use `None` for that, because that's what makes sense to me.

sopel/irc/modes.py
Exirel: That's because `type_param` should be symbols, not int or str value.

sopel/irc/modes.py
Exirel: The goal was to be able to handle any new type of modes, and for that I tried to designed `sopel.irc.modes` as a standalone library. At some point I'd love to just extract all that into its own package and distribute it as "here are some IRC tools for you". 😁 

So, for that to work, I made sure that one would be able to add mode types that are not the official ones (A, B, C, D): the spec allows it, even though servers should be very careful with those.

sopel/irc/modes.py
Exirel: If you look at how it is used, there *are* logs. Just not here where we parse everything and return a result with all the information (valid modes, leftover parameters, missing parameters, etc.).

sopel/coretasks.py
Exirel: There are less than 10 keys. I don't think this matter.

sopel/coretasks.py
Exirel: Oh yeah, that could totally be an attribute of the bot, indeed.

sopel/coretasks.py
Exirel: I really don't like having to depends on logging within the parser itself: the result is a stateless object that contains every bit of required information. For me, it's up to whatever code calls that to know what to do with the result, including dealing with errors.

sopel/irc/__init__.py
```diff
@@ -36,7 +36,7 @@
 from .utils import CapReq, safe
 
 
-__all__ = ['abstract_backends', 'backends', 'utils']
+__all__ = ['abstract_backends', 'backends', 'modes', 'utils']
```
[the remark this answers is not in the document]
Exirel: To be accessible from autodoc!

sopel/irc/modes.py
Exirel: Huh... this actually makes sense, now that you talk about it.

sopel/irc/modes.py
```diff
@@ -0,0 +1,234 @@
...
+ModeString = namedtuple('ModeString', [
+    'modes',
+    'privileges',
+    'ignored_modes',
+    'leftover_params',
+])
+
+
+class ModeMessage:
+    """Modestring parser for IRC's ``MODE`` messages for channel modes."""
```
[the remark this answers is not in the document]
Exirel: Actually... yes it could. I just read that user modes are of type D (no param), so that should be doable.

sopel/irc/__init__.py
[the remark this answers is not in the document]
Exirel: (or at least I think that's for that? I honestly can't remember)

sopel/irc/modes.py
```diff
@@ -0,0 +1,234 @@
...
+PARAM_ALWAYS = 'always'
+PARAM_ADDED = 'added'
+PARAM_REMOVED = 'removed'
+PARAM_NEVER = 'never'
+DEFAULT_MODETYPE_PARAM_CONFIG = {
+    'A': PARAM_ALWAYS,
+    'B': PARAM_ALWAYS,
+    'C': PARAM_ADDED,
+    'D': PARAM_NEVER,
+}
```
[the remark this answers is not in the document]
Exirel: The dict part is only the mapping between the letter and the "how to handle parameters for this type".

The `ParamType` ENUM is a good idea.

sopel/irc/modes.py
```diff
@@ -0,0 +1,234 @@
...
+
+
+class ModeTypeImproperlyConfigured(ModeException):
+    """Exception when the mode's type management is not configured."""
+    def __init__(self, mode: str, letter: str) -> None:
+        message = 'Type {mode} for mode {letter} is not properly configured.'
+        super().__init__(message.format(mode=mode, letter=letter))
+
+
+def _modes_is_added(modestring: str
```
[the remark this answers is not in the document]
Exirel: I get your point. I need to find a good name for that. Maybe `_modes_added`. I don't know.

Something like `_parse_modestring`? After all, that's what this function does.

sopel/coretasks.py
Exirel: No more optimization problem!

sopel/coretasks.py
```diff
@@ -45,6 +46,19 @@
...
+    "v": plugin.VOICE,
+    "h": plugin.HALFOP,
+    "o": plugin.OP,
+    "a": plugin.ADMIN,
+    "q": plugin.OWNER,
+    "y": plugin.OPER,
+    "Y": plugin.OPER,
+}
+
+MODE_PREFIXES = set(MODE_PREFIX_PRIVILEGES.keys())
```
dgw, line 59: This doesn't appear to be used anywhere. What am I missing?

sopel/irc/modes.py
```diff
@@ -0,0 +1,296 @@
...
+
+class ModeTypeImproperlyConfigured(ModeException):
+    """Exception when the mode's type management is not configured properly."""
+    def __init__(self, mode: str, letter: str) -> None:
+        message = 'Type {mode} for mode {letter} is not properly configured.'
+        super().__init__(message.format(mode=mode, letter=letter))
+
+
+def _parse_modestring(modestring: str
+                      ) -> Generator[Tuple[str, bool], None, None]:
```
dgw, line 59: I honestly hate this style for no particular reason. I can't tell you why. 😅

I also still think that wrapping at 79 chars is too narrow and we should be using 100 or 120, but eh. Suggestion in case you want to take pity on my sensibilities. It's a whopping 83 cols wide. 😝

```suggestion
def _parse_modestring(modestring: str) -> Generator[Tuple[str, bool], None, None]:
```

sopel/irc/modes.py
```diff
@@ -0,0 +1,296 @@
...
+        'D': tuple('Oimnpsrt'),
+    }
+    """Default CHANMODES per :rfc:`2811`.
+
+    .. note::
+
+        Mode ``a`` has been removed from the default list, as it appears
+        to be a relic of the past and is more commonly used as a privilege.
+
+        Mode ``q`` has been removed too, as it is communly used as a privilege.
```
dgw, line 132:
```suggestion
        Mode ``q`` has been removed too, as it is commonly used as a privilege.
```

sopel/irc/modes.py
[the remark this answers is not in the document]
dgw: This appears to be resolved by changes above here (including passing `CHANMODES` as the default value of `chanmodes` if the param isn't given, and the amusing note about servers that should advertise their features properly).

dgw (commented):
I'm back again! Loving the added snark peppered in the docs.

sopel/irc/modes.py
Exirel: Another common practice is to do that:

```suggestion
def _parse_modestring(modestring: str
) -> Generator[Tuple[str, bool], None, None]:
```

sopel/irc/modes.py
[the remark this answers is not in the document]
Exirel: 14 days later, and I think everyone saw that one coming: that will be (if ever) for another PR.

After all, parsing user mode is much easier by being all of type D, and since Sopel doesn't do anything with them, for now, I'll keep this PR focused on channel modes only.

sopel/irc/modes.py
```diff
@@ -0,0 +1,316 @@
+"""Mode management for IRC channels.
+
+The :class:`ModeParser` class is used internally by the bot to parse ``MODE``
+messages for channels. User modes are not parsed (yet), as the bot doesn't
+manage them.
+
+The goal of the parser is to return a :class:`ModeMessage` that contains the
+result: modes to add/remove with their parameter when necessary, privileges to
+add/remove for a user in a channel, and errors (ignored modes and unused
+parameters).
```
dgw, line 10: I like referring to these as "actions" rather than a "result". Whether the "errors" count as part of that or not is up for debate; I didn't include it in the list because it's mostly irrelevant unless you happen to use a non-compliant server. Ideally, almost no one will ever see non-empty ignored modes or unused parameters lists.

```suggestion
The goal of the parser is to return a :class:`ModeMessage` containing the
actions represented by the raw message:

* channel modes added/removed (including their parameters, if any)
* privileges added/removed for user(s) in a channel

Errors (ignored modes and unused parameters) are also included, mostly for
detecting when an IRC server is not conforming to specifications.
```

dgw (commented):
We must be getting close.

sopel/irc/modes.py
Exirel: Oh I like that! It's less verbose while being more accurate. Love it!

Exirel:
We are getting close.

sopel/irc/__init__.py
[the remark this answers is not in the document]
dgw: Resurrecting this thread because I noticed one of the force-pushes reverted this. Seems like an accidental change. Was it?

dgw (approved):
I deem this "close enough", mostly because the best way to find any remaining edge cases (or documentation mistakes) is going to be shipping this and having people use it in ~~production~~ testing environments. Squash the fixups and then bug @half-duplex (who wrote some of the logic this replaces) to take a look if he's willing.

sopel/irc/__init__.py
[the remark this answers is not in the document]
Exirel: 100% honest: I've no idea, last comment is like at least a few years old now. 😁 

No but seriously, I've no idea at this point.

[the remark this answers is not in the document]
Exirel: Reading the list of modified files, I think I went for the safer approach: I didn't modify `sopel/irc/__init__.py` so I don't have to think about adding other sub-modules. If that must be done, then it can be in a future PR.

dgw:
> I'm out of enumerable nits 😛

@Exirel I think this means you can squash. 🚀

Exirel:
> you can squash. :rocket: 

:rocket: landed.

dgw:
I'm sure CI will fail thanks to #2210, but 4 months in the PR list was enough for this patch. :shipit:

PR #2132, opened 2021-07-02
tools.time: handle aware datetimes too in `format_time()`

dgw:
### Description
As a side note, the absolute fallback default time format has been changed here to use `%z` (UTC offset) instead of `%Z` (zone name), since the zone name is almost never useful and the offset is actually handy to have in tests. And in the output, too...

### Checklist

### Notes
After faffing around _way_ too much with the tests, I'm not _entirely_ sure any more if their passing actually means anything. Sure, `tools.time.format_time()` seems to do what I expect in an interactive console, but… that's not a unit test.

I don't _think_ this warrants "Breaking Change" treatment since the only user-facing behavior change is the output when _no_ other time format source is passed to `format_time()` and it uses the tweaked default… but opinions may vary. Maybe aware `datetime` objects being legal now should be called out.

See also: https://github.com/sopel-irc/sopel/pull/2099#issuecomment-860932459

### DIFF
--- a/test/tools/test_tools_time.py
+++ b/test/tools/test_tools_time.py
@@ -4,6 +4,7 @@ from __future__ import generator_stop
 import datetime
 
 import pytest
+import pytz
 
 from sopel.tools import time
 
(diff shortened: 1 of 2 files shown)


test/tools/test_tools_time.py
```diff
@@ -46,6 +47,42 @@ def test_validate_format_none():
...
+            2021, 7, 4, 17, 7, 6,
+            tzinfo=UTC,
+        ),
+        '2021-07-04 - 17:07:06 +0000',
+    ),
+    (
+        # Aware, non-UTC
+        datetime.datetime(
+            2021, 7, 4, 17, 7, 6,
+            tzinfo=UTC,
```
Exirel, line 74: It looks very UTC to me.

Exirel (changes requested):

Exirel (approved):
Alright, it's getting better.

dgw:
Note for my future self: Labeled as Breaking Change after all, just to mention the new default in changelog.

PR #2133, opened 2021-07-02
Expose plugin versions, enable retrieval with `.version`

half-duplex:
### Description
Exposes `plugin._module.__version__` (or `None`) in plugin metadata, and adds `.version plugin_name`.

```irc
<!mal> .version
< Sopel> [version] Sopel v8.0.0.dev0 | Python: 3.9.5 | Commit: blahblah
<!mal> .version coretasks
< Sopel> [version] coretasks v8.0.0.dev0 (built in)
<!mal> .version mypackage
< Sopel> [version] mypackage v9.9.9
<!mal> .version test
< Sopel> [version] test (version unknown)
<!mal> .version fake
< Sopel> [version] I don't have a plugin named 'fake' loaded.
```
This will confuse any daring person with a sopel module named sopel, but that's a risk I'm willing to take.
It does work for individual .py files that contain a `__version__`.

### Checklist

### DIFF
--- a/sopel/modules/version.py
+++ b/sopel/modules/version.py
@@ -33,13 +33,34 @@ def git_info():
 
 
 @plugin.command('version')
+@plugin.example('.version [plugin_name]')
 @plugin.output_prefix('[version] ')
 def version(bot, trigger):
-    """Display the installed version of Sopel.
+    """Display the installed version of Sopel or a plugin.
 
     Includes the version of Python Sopel is installed on.
     Includes the commit hash if Sopel is installed from source.
     """
+    plugin = trigger.group(3)
+    if plugin and plugin.lower() != "sopel":
+        # Plugin version
+        if not bot.has_plugin(plugin):
+            bot.say("I don't have a plugin named %r loaded." % plugin)
+            return
+
+        meta = bot.get_plugin_meta(plugin)
+        if meta["version"] is None:
+            version = "(unknown)"
+        else:
+            version = "v" + str(meta["version"])
+
+        if meta["source"].startswith("sopel."):
+            version += " (built in)"
+
+        bot.say(plugin + " " + version)
+        return
+
+    # Sopel version
     parts = [
         'Sopel v%s' % release,
         'Python: %s' % platform.python_version()
--- a/sopel/plugins/handlers.py
+++ b/sopel/plugins/handlers.py
@@ -49,8 +49,9 @@ import importlib
 import inspect
 import itertools
 import os
+from typing import Optional
 
-from sopel import loader
+from sopel import __version__ as release, loader
 from . import exceptions
 
 try:
@@ -120,6 +121,15 @@ class AbstractPluginHandler(abc.ABC):
         * version: the plugin's version string if available, otherwise ``None``
         """
         # TODO: change return type to a TypedDict when dropping py3.7
+
+    @abc.abstractmethod
+    def get_version(self):
+        """Retrieve the plugin's version.
+
+        :return: the plugin's version string
+        :rtype: str
+        """
+        raise NotImplementedError
 
     @abc.abstractmethod
     def is_loaded(self) -> bool:
@@ -292,8 +302,19 @@ class PyModulePlugin(AbstractPluginHandler):
             'type': self.PLUGIN_TYPE,
             'name': self.name,
             'source': self.module_name,
-            'version': getattr(self._module, "__version__", None),
+            'version': self.get_version(),
         }
+
+    def get_version(self) -> Optional[str]:
+        """Retrieve the plugin's version.
+
+        :return: the plugin's version string
+        :rtype: Optional[str]
+        """
+        if hasattr(self._module, "__version__"):
+            return str(self._module.__version__)
+        if self.module_name.startswith("sopel."):
+            return release
 
     def load(self):
         """Load the plugin's module using :func:`importlib.import_module`.
--- a/sopel/plugins/handlers.py
+++ b/sopel/plugins/handlers.py
@@ -117,6 +117,7 @@ class AbstractPluginHandler(abc.ABC):
         * type: the plugin's type
         * source: the plugin's source
           (filesystem path, python import path, etc.)
+        * version: the plugin's version string if available, otherwise ``None``
         """
         # TODO: change return type to a TypedDict when dropping py3.7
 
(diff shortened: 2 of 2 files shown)


sopel/modules/version.py
```diff
@@ -33,22 +33,43 @@ def git_info():
 
 
 @plugin.command('version')
+@plugin.example('.version')
+@plugin.example('.version url')
```
dgw, line 37: Like the last time I commented on a PR of yours that added examples, these won't work as intended. Without using `user_help=True`, only the "first" (closest to the `def`) example will be shown in `.help` output.

In this case, since the example isn't being (ab)used to generate a test case, might as well write it like a usage string:

```suggestion
@plugin.example('.version [plugin_name]')
```

sopel/modules/version.py
```diff
@@ -33,22 +33,43 @@ def git_info():
...
+
+        meta = bot.get_plugin_meta(plugin)
+        if meta["source"].startswith("sopel."):
+            version = "v" + release + " (built in)"
+        elif meta["version"] is None:
+            version = "(version unknown)"
+        else:
+            version = "v" + meta["version"]
+
+        bot.say("{} {}".format(plugin, version))
```
dgw, line 72: I like to avoid unnecessary nesting and go for `if-return/do-more` rather than `if/else`. In this case it would allow leaving the original code alone and just inserting the new plugin logic above it.

dgw, line 72: Isn't this simpler?

```suggestion
        bot.say(' '.join(plugin, version))
```

dgw (changes requested):
Style nitpicks big and small. Which means they're not really all "nitpicks", I guess.

sopel/modules/version.py
Exirel:
```suggestion
        bot.say('%s %s, putain' % (plugin, version))
```

sopel/modules/version.py
half-duplex: The ordering was intended since `.version sopel` works too, but ack on brackets.

sopel/modules/version.py
half-duplex: @dgw, it's simpler, but it doesn't run :stuck_out_tongue:
I figured .format() was less gross than instantiating a whole list/tuple to .join().

sopel/plugins/handlers.py
```diff
@@ -114,6 +114,7 @@ def get_meta_description(self):
         * type: the plugin's type
         * source: the plugin's source
           (filesystem path, python import path, etc.)
+        * version: the plugin's version string if available, otherwise `None`
```
dgw, line 117:
```suggestion
        * version: the plugin's version string if available, otherwise ``None``
```

sopel/plugins/handlers.py
```diff
@@ -270,6 +271,7 @@ def get_meta_description(self):
         * label: see :meth:`~sopel.plugins.handlers.PyModulePlugin.get_label`
         * type: see :attr:`PLUGIN_TYPE`
         * source: the name of the plugin's module
+        * version: the version string of the plugin if available, otherwise `None`
```
dgw, line 274:
```suggestion
        * version: the version string of the plugin if available, otherwise ``None``
```

dgw (changes requested):
Looks good but for a couple of docstring-formatting nits that I missed the first time. (In rST, a single ` does something entirely different from what it does in Markdown; the equivalent is ``.)

I don't care enough about the performance to benchmark `.format()` vs. `%s` interpolation vs. `.join()` or anything, and it's not like this is going to run on every line, so I just marked that thread as resolved. 🙃

Exirel:
At first, I was rather hesitant regarding plugin version, because something didn't sit well with me. Now, I think plugin's version is an interesting information, so I put that aside for later.

After further consideration, I think it would be best if the Plugin Handler object had a `get_version` method, in order to override how we get the version number based on the plugin type. All the code used to extract the version from the meta-data would be gone and replaced by using the clean interface.

dgw:
@Exirel Is that "would be best" a request-changes or someone-should-follow-up-on-this? I'm ready to approve and let a plugin-handler `get_version()` method happen separately, fwiw.

@half-duplex Note that when you squash, it would be a good time to fix the typo "retreiving" in 1be8bbe48d285d41e898eef176b1ada37e026916

half-duplex:
Exi, what's the benefit of a .get_version() over just changing how get_meta_description() generates the version return attr?

Exirel:
> someone-should-follow-up-on-this?

I think I left this comment expecting a conversation, and that didn't happen a month ago.

> what's the benefit of a .get_version()

Getting a plugin's version can depends on the type of plugin, therefore each Plugin Handler may require an override on how to get said version. For example, there is no need for a `__version__` attribute in an Entry Point plugin, since you can get that information from pkg_resource or setuptools (I don't remember the details). Therefore, having one method that does that is more convenient.

Another reason is that it's easier to get a Plugin object, and ask for its version, without having to retrieve a dict and hope that it gets everything required. It means less code overall, so less chance of mistake.

Last but not least, it's easier to deprecate a behavior if you have a method with a single purpose for it. That's the reason for a proper interface.

dgw:
Still hoping for a refactor with that getter method Exi asked about.

dgw:
Refocusing on some 8.0 stuff now that 7.1.7 is out (and we weathered the Python 3.9.8 kerfuffle). Merged #2199, which opens the way to rebasing this patch on top of it.

Looks like some small difference in documentation was introduced when splitting the new `bot.plugins` property into its own PR?
image

sopel/bot.py
```diff
@@ -189,6 +190,15 @@ def hostmask(self) -> Optional[str]:
 
         return self.users.get(self.nick).hostmask
 
+    @property
+    def plugins(self) -> Mapping[str, plugins.handlers.AbstractPluginHandler]:
+        """A dict of the bot's currently loaded plugins.
+
+        :return: an immutable map of plugin name to plugin object
+        :rtype: Mapping[str, plugins.handlers.AbstractPluginHandler]
```
Exirel, line 198: This line is, indeed, not needed thanks to the signature that is picked up by Sphinx now.

dgw: Can confirm:

image
I love having [unstable docs](https://unstable.docs.sopel.chat/bot.html#sopel.bot.Sopel.plugins) built on every commit.

dgw (approved):
Whee

PR #2136, opened 2021-07-02
irc: add safe_text_length and use abc.ABC for AbstractBot

Exirel:
### Description

Tin. I was always a bit annoyed by the `say` method that does too much on its own, and I know that @half-duplex wanted a way to know the safe length for an IRC message without having to reinvent the wheel each time. I hope this will help, even if just a little.

Working on this part, I saw that `hostmask` is used but not defined by the abstract class, so I took this opportunity to use `abc.ABC` and the `abc.abstractmethod` decorator.

By doing so, I also replaced the `KeyError` when the hostmask is not defined to return `None` instead. This is a very minor breaking change. I'm not sure anyone is using `bot.hostmask`, and they probably don't use it when the bot is not connected/hasn't join a channel yet.

### Checklist

### DIFF
--- a/sopel/bot.py
+++ b/sopel/bot.py
@@ -15,6 +15,7 @@ import re
 import signal
 import threading
 import time
+from typing import Optional
 
 from sopel import irc, logger, plugins, tools
 from sopel.db import SopelDB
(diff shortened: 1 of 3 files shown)


dgw (approved):
I can think of no objections. It's mostly my code, just moved somewhere else. 😁

Exirel:
Your code, with more tests. :grin:

PR #2138, opened 2021-07-02
sopel: use python3 style class and abc.ABC

Exirel:
### Description

In Python 2, there was 2 styles:

* `class ClassName:`
* `class ClassName(object)`

And they behaved differently, and using `super()` is different too. However, in Python3, both are equivalent to doing the second style, and we can use `super()` directly. I remember having to work with Py2.4 compatible code in my career, and this "new style class" stuff was... a pure nightmare. I'm just glad we can get rid of this.

Also, there is now `abc.ABC` that allows to describe an abstract class as being abstract. This is interesting because it doesn't require to add `raise NotImplementedError` anymore: a few decorators and it's all good!

One last thing: I added few type hint here and there, and I had fun with `TypeVar`. It's looking good!

Note: I think I want a "housekeeping" or "cleaning code" label or something. :grin: 

### Checklist

### DIFF
--- a/sopel/plugins/handlers.py
+++ b/sopel/plugins/handlers.py
@@ -43,6 +43,7 @@ away from the rest of the application.
 # Licensed under the Eiffel Forum License 2.
 from __future__ import generator_stop
 
+import abc
 import imp
 import importlib
 import inspect
@@ -60,7 +61,7 @@ except AttributeError:
     reload = imp.reload
 
 
-class AbstractPluginHandler(object):
+class AbstractPluginHandler(abc.ABC):
     """Base class for plugin handlers.
 
     This abstract class defines the interface Sopel uses to
@@ -75,23 +76,25 @@ class AbstractPluginHandler(object):
     (commands, jobs, etc.), configuring it, and running any required actions
     on shutdown (either upon exiting Sopel or unloading that plugin).
     """
+
+    @abc.abstractmethod
     def load(self):
         """Load the plugin.
 
         This method must be called first, in order to setup, register, shutdown,
         or configure the plugin later.
         """
-        raise NotImplementedError
 
+    @abc.abstractmethod
     def reload(self):
         """Reload the plugin.
 
         This method can be called once the plugin is already loaded. It will
         take care of reloading the plugin from its source.
         """
-        raise NotImplementedError
 
-    def get_label(self):
+    @abc.abstractmethod
+    def get_label(self) -> str:
         """Retrieve a display label for the plugin.
 
         :return: a human readable label for display purpose
@@ -99,9 +102,9 @@ class AbstractPluginHandler(object):
 
         This method should, at least, return ``<module_name> plugin``.
         """
-        raise NotImplementedError
 
-    def get_meta_description(self):
+    @abc.abstractmethod
+    def get_meta_description(self) -> dict:
         """Retrieve a meta description for the plugin.
 
         :return: meta description information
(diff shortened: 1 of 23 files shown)


sopel/tools/_events.py
```diff
@@ -1,7 +1,7 @@
 from __future__ import generator_stop
 
 
-class events(object):
+class events:
```
dgw, line 4: Already touched by #2127. Can just drop this change, since it's now an actual `Enum`.

sopel/plugins/handlers.py
```diff
@@ -75,33 +76,35 @@ class AbstractPluginHandler(object):
...
         """Reload the plugin.
 
         This method can be called once the plugin is already loaded. It will
         take care of reloading the plugin from its source.
         """
-        raise NotImplementedError
 
-    def get_label(self):
+    @abc.abstractmethod
+    def get_label(self) -> str:
```
dgw, line 97: Seems like adding type hints should be mentioned in the commit message, at least.

sopel/plugins/handlers.py
```diff
@@ -115,9 +118,10 @@ def get_meta_description(self):
         * source: the plugin's source
           (filesystem path, python import path, etc.)
         """
-        raise NotImplementedError
+        # TODO: change return type to a TypedDict when dropping py3.7
```
dgw, line 121: https://pypi.org/project/typing-extensions/?

sopel/plugins/rules.py
```diff
@@ -591,8 +594,8 @@ def get_output_prefix(self):
             See the :class:`sopel.bot.SopelWrapper` class for more information
             on how the output prefix can be used.
         """
-        raise NotImplementedError
 
+    @abc.abstractmethod
     def match(self, bot, pretrigger):
```
dgw, line 599: No type hint?

sopel/plugins/rules.py
```diff
@@ -605,76 +608,76 @@ def match(self, bot, pretrigger):
...
+    def is_global_rate_limited(self) -> bool:
         """Tell when the rule reached the server's rate limit.
 
         :return: ``True`` when the rule reached the limit, ``False`` otherwise
         :rtype: bool
         """
-        raise NotImplementedError
 
+    @abc.abstractmethod
     def parse(self, text):
```
dgw, line 681: No type hint? Unless there isn't a sensible way to put one for a generator.

dgw (changes requested):
I could resolve the merge conflict manually (it's trivial), but there are a few other things.

sopel/plugins/handlers.py
Exirel: Hm. Good question. Might want to have a look at some point.

sopel/plugins/rules.py
Exirel: It's a more complex type hint, so I'll do that in a future PR (when I'll be *real* serious about type hint).

Exirel: Actually, I'm going to be a bit lazy and just do `-> Iterable[Any]`.

sopel/plugins/handlers.py
Exirel: Did that. I had to rebase to do so.

Exirel:
Fixed merge conflict.

dgw (approved):
Definitely leaned on GitHub to show me which files were changed since the last review. My little nitpicks are all taken care of, so let's mark this as ready to ship.

PR #2156, opened 2021-07-12
Revert "bot: deprecate `search_url_callbacks()` method"

RhinosF1:
This reverts commit 3986227c4971e68530ae07113426187e95967046.

```
 <+Exirel> 
Because Sopel still uses `search_url_callbacks`, and have to.
14:01:55 <RhinosF1> 
it does
14:02:21 <+Exirel> 
Yup. So yeah, we need to find a better solution than that.
14:04:19 <RhinosF1> 
I can hack beta if anything needs testing
14:07:20 <+Exirel> 
I think we should not trigger a warning on `search_url_callbacks` in 8.0, make it a no-op in 9.0, and remove it in 10.0.
```

### Description
Revert 3986227c4971e68530ae07113426187e95967046.

### Checklist

### DIFF
--- a/sopel/bot.py
+++ b/sopel/bot.py
@@ -1150,11 +1150,6 @@ class Sopel(irc.AbstractBot):
         except KeyError:
             pass
 
-    @deprecated(
-        reason='Issues with @url decorator have been fixed. Simply use that.',
-        version='8.0',
-        removed_in='9.0',
-    )
     def search_url_callbacks(self, url):
         """Yield callbacks whose regex pattern matches the ``url``.
 


sopel/bot.py
```diff
@@ -1175,11 +1170,6 @@ def search_url_callbacks(self, url):
             Searches for registered callbacks in an internal property instead
             of ``bot.memory['url_callbacks']``.
 
-        .. deprecated:: 8.0
```
Exirel, line 1178: Not that part! The documentation should stay in place. 😄

Exirel (changes requested):

RhinosF1:
> Notice: /Stage[main]/Sopel::Dev/Exec[rebuild sopel]: Triggered 'refresh' from 1 event
Deployed

Exirel:
For the record:

```
15:07:38 <+Exirel> Or just remove it.
15:08:42 <+Exirel> Because after all, it is useful only if you use the other methods, and those are not used by Sopel anymore and they are marked as deprecated.
15:08:48 <+Exirel> So I think that's enough warning.
15:09:02 <+Exirel> Beside, the deprecation note in the doc should be enough.
15:09:07 <+Exirel> No need to spam logs with that deprecation warning.
```

Exirel (approved):

dgw:
Why not use a `warning_in='9.0'` kwarg to `@deprecated`, plus a TODO comment? There should be a TODO anyway as a reminder to no-op this function in 9.0.

dgw:
N.B.: I elected not to worry about this spitting out a warning once the version hits 9.0-dev. Bikeshedding ain't worth it.

PR #2162, opened 2021-07-18
tell: do not lstrip leading formatting

### LINKED ISSUE #2157: tell: argument splitting/stripping is a bit broken
This code line removes most IRC formatting bytes from the beginning of the message:

https://github.com/sopel-irc/sopel/blob/f29c10dde2ff166dee15bc49f0cd43c93b93ba3e/sopel/modules/tell.py#L145

It's similar to another issue (#1877) we fixed in `choose` for 7.1 (#1965). The second `.lstrip()` with no arguments needs to be replaced with this plugin's own version of [`_format_safe()`](https://github.com/sopel-irc/sopel/blob/f29c10dde2ff166dee15bc49f0cd43c93b93ba3e/sopel/modules/choose.py#L19-L69) from `choose.py`.

In fact, it wouldn't take much at this point to convince me that Sopel's API should include its own formatting-safe versions of `strip()`, `lstrip()`, and `rstrip()`. Issues like this keep coming up, and stuff like `trigger.plain` won't help…

----

That line also misuses `.lstrip()` with the tellee's nick. That isn't a huge issue, because it's never going to go past the whitespace, but it could cause other quirks like the fact that punctuation gets through even though the code's _intent_ appears to be treating the tellee as a separate argument that isn't part of the message:

```irclog
<~dgw> ;tell Creator|TL, comma test
[…]
<&Kaede> [tell] Creator|TL: Monday, July 12, 2021 22:52:44 (IST) <dgw> tell Creator|TL , comma test
```

dgw:
### Description
Tin; bugfix to resolve #2157.

Also gets rid of the weird "lingering comma" behavior that that issue describes in passing, as a welcome side effect.

### Checklist

### Notes
Using a custom/bespoke function to do this for now. As mentioned in #2157, we probably should think about adding formatting-safe string-stripping functions like this and what `choose` uses to Sopel's API. Core plugins surely aren't the only ones to run into such edge cases with leading/trailing formatting bytes.

### DIFF
--- a/sopel/modules/tell.py
+++ b/sopel/modules/tell.py
@@ -14,8 +14,9 @@ import logging
 import os
 import threading
 import time
+import unicodedata
 
-from sopel import plugin, tools
+from sopel import formatting, plugin, tools
 from sopel.config import types
 from sopel.tools.time import format_time, get_timezone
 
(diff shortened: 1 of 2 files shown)


Exirel (approved):
Yet another PR that shows why we love `pytest.mark.parametrize`.

PR #2172, opened 2021-08-04
requirements: pin requests on py3.3 to fix install/build

dgw:
requests 2.26 switched from chardet to charset-normalizer on py3, and that seems to be [broken on py3.3](https://app.travis-ci.com/github/sopel-irc/sopel/builds/234445467) specifically. It installs, but requests throws an error on loading anyway for some reason. Easiest solution is just to pin the requests version, since especially on py3.3 we don't care about any of the new stuff.

### Checklist


Exirel (approved):
🤷‍♂️

PR #2174, opened 2021-08-14
Update to check for proper UTF8 on Windows

### LINKED ISSUE #2173: UTF-8 Check fails on windows in Powershell
### Description
The current version of sopel in pip (py3.9.6) and from git master both prompt this message upon running sopel even with a utf8 configuration:

> WARNING!!! You are running with a non-UTF8 locale environment variable (e.g. LC_ALL is set to "C"), which makes Python 3 do stupid things. If you get strange errors, please set it to something like "en_US.UTF-8".

Despite having a "English_United States", "utf8" locale from python.

### Reproduction steps
Pull into a fresh python dev environment on windows 10.
Install sopel via pip or from source.
Run sopel.

### Expected behavior
To not get a warning about UTF-8 since it's configured.

### Logs
```
(.venv) PS C:\Users\Michael\Documents\Visual Studio Projects\Python\sopel> $PSVersionTable     

Name                           Value
----                           -----
PSVersion                      7.1.4
PSEdition                      Core
GitCommitId                    7.1.4
OS                             Microsoft Windows 10.0.19041
Platform                       Win32NT
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0…}
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1
WSManStackVersion              3.0

(.venv) PS C:\Users\Michael\Documents\Visual Studio Projects\Python\sopel> Get-WinSystemLocale 

LCID             Name             DisplayName
----             ----             -----------
1033             en-US            English (United States)

(.venv) PS C:\Users\Michael\Documents\Visual Studio Projects\Python\sopel> echo $OutputEncoding

Preamble          :      
BodyName          : utf-8
EncodingName      : Unicode (UTF-8)
HeaderName        : utf-8
WebName           : utf-8
WindowsCodePage   : 1200
IsBrowserDisplay  : True
IsBrowserSave     : True
IsMailNewsDisplay : True
IsMailNewsSave    : True
IsSingleByte      : False
EncoderFallback   : System.Text.EncoderReplacementFallback
DecoderFallback   : System.Text.DecoderReplacementFallback
IsReadOnly        : True
CodePage          : 65001


(.venv) PS C:\Users\Michael\Documents\Visual Studio Projects\Python\sopel> py
Python 3.9.6 (tags/v3.9.6:db3ff76, Jun 28 2021, 15:26:21) [MSC v.1929 64 bit (AMD64)] on win32
>>> import locale
>>> locale.getlocale()
('English_United States', 'utf8')
>>> exit()
```

### Environment
- Sopel `.version`: 8.0.0
- Sopel installed via: pip && source
- Python version: 3.9.6
- Operating system: Windows 10
- IRCd `/version`: N/A
- Relevant plugins: N/A

Gizmokid2005:
The locale.getlocale() check currently looks for "UTF-8" only, however on windows with powershell and UTF8 enabled the return value is "utf8" so this check still fails.

### Description
Fixes #2173

### Checklist

### DIFF
--- a/sopel/__init__.py
+++ b/sopel/__init__.py
@@ -34,7 +34,7 @@ __all__ = [
 ]
 
 loc = locale.getlocale()
-if not loc[1] or 'UTF-8' not in loc[1]:
+if not loc[1] or ('UTF-8' not in loc[1] and 'utf8' not in loc[1]):
     print('WARNING!!! You are running with a non-UTF8 locale environment '
           'variable (e.g. LC_ALL is set to "C"), which makes Python 3 do '
           'stupid things. If you get strange errors, please set it to '


sopel/__init__.py
```diff
@@ -34,7 +34,7 @@
 ]
 
 loc = locale.getlocale()
-if not loc[1] or 'UTF-8' not in loc[1]:
+if not loc[1] or ('UTF-8' not in loc[1] and 'utf8' not in loc[1]):
```
RustyBower, line 37: Shouldn't this be or not and?

Gizmokid2005: No, because in an OR, one of them would always evaluate to true, because only one of the two conditions can be true at the same time.

Gizmokid2005:
> 
> 
> Before I approve, note that we usually use `sopel: [message]` instead of `__init__.py` for the commit message.
> 
> I was tempted to dismiss the PR because we don't officially support Windows (someone would have to actually test Sopel on Windows, and we don't do that), however the fix is rather simple, and shouldn't be a problem.
> 
> Also, I'm not even sure that this warning is legitimate anymore, as I don't recall Python 3.6+ doing anything out of the ordinary when the system is not utf-8. I mean... unless someone did something really wrong with stdout...

I misunderstood the contributing docs when I made the commit then, my apologies.
>Make your commit messages clear and explicative. Our convention is to place the name of the thing you're changing in at the beginning of the message, followed by a colon: the plugin name for plugins, "docs" for documentation files, "coretasks" for coretasks.py, "db" for the database feature, etc.

I normally don't use windows for things, but was trying to setup a quick touch dev environment and was running into this. I'm not sure of the validity of the Python warning though.

dgw:
Hopping around through some blog posts from random people eventually led me to find [PEP 540](https://www.python.org/dev/peps/pep-0540/), which talks about assuming UTF-8 for the C locale on Unix starting in Python 3.7. I've reached my limit for text encoding nonsense today, but it sounds like we do still need this check at present.

Python under Windows also allegedly uses different encoding modes depending on how it's run (lost the source link for this one mid-rabbit-hole, sadly), but we won't concern ourselves with that one.

Exirel (approved):

dgw (changes requested):
I would just amend the commit message myself, but since this commit is signed I'm sure @Gizmokid2005 would prefer to do it and keep the lovely green badge. Our usual format for this commit message would look like:

```
sopel: Update to check for proper UTF8 on Windows

The locale.getlocale() check currently looks for "UTF-8" only, however
on windows with powershell and UTF8 enabled the return value is "utf8"
so this check still fails.
```

Gizmokid2005:
> 
> 
> I would just amend the commit message myself, but since this commit is signed I'm sure @Gizmokid2005 would prefer to do it and keep the lovely green badge. Our usual format for this commit message would look like:
> 
> ```
> sopel: Update to check for proper UTF8 on Windows
> 
> The locale.getlocale() check currently looks for "UTF-8" only, however
> on windows with powershell and UTF8 enabled the return value is "utf8"
> so this check still fails.
> ```

Thanks! I'll get the commit amended today.

Gizmokid2005:
@dgw - Sorry about the noise (this is actually a first time I've amended a commit and well...Monday didn't go smoothly. Should be good now, let me know if you need anything else from me!

dgw (approved):
Oops, we lost the green "Verified" badge anyway, but it sounds like you probably learned something about amending. Not a total loss? 😸

PR #2176, opened 2021-08-18
reddit: handle post and comment links in one regex

### LINKED ISSUE #2175: reddit: comment links also output submission info
One of the bug fixes in 7.1.3 removed an anchor (`&`) from the end of post URLs. While it's now possible for Sopel to pick up reddit post links with e.g. `?sort=confidence` tacked on (the intended result), that also means Sopel will happily output the post info along with comments (unintended side effect).

The link matching for posts needs to exclude anything that looks like a comment ID, but still match on anything else.

dgw:
### Description
It's entirely impractical to try to handle post and comment links in separate callables, because comment links contain an entire post link. Before, this was handled by anchoring the post URL pattern at the end, but that broke Sopel handling e.g. `?sort=confidence` if someone sorted the comments before copying the link into IRC.

Lookaround assertions didn't appear to prevent matching the post link contained in every comment link, so this solution seemed like the only way forward. It's a nice tidy dispatcher pattern, really.

Fix #2175.

### Checklist

### DIFF
--- a/sopel/modules/reddit.py
+++ b/sopel/modules/reddit.py
@@ -27,10 +27,13 @@ PLUGIN_OUTPUT_PREFIX = '[reddit] '
 
 domain = r'https?://(?:www\.|old\.|pay\.|ssl\.|[a-z]{2}\.)?reddit\.com'
 subreddit_url = r'%s/r/([\w-]+)/?$' % domain
-post_url = r'%s/r/\S+?/comments/([\w-]+)(?:/[\w%%]+)?/?' % domain
+post_or_comment_url = (
+    domain +
+    r'/r/\S+?/comments/(?P<submission>[\w-]+)'
+    r'(?:/?(?:[\w%]+/(?P<comment>[\w-]+))?)'
+)
 short_post_url = r'https?://redd\.it/([\w-]+)'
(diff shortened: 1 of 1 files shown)


Exirel (approved):
If you say it works, I believe you, since I've no idea what these regex does. 😁

sopel/modules/reddit.py
```diff
@@ -27,10 +27,13 @@
 
 domain = r'https?://(?:www\.|old\.|pay\.|ssl\.|[a-z]{2}\.)?reddit\.com'
 subreddit_url = r'%s/r/([\w-]+)/?$' % domain
-post_url = r'%s/r/\S+?/comments/([\w-]+)(?:/[\w%%]+)?/?' % domain
+post_or_comment_url = (
+    domain +
+    r'/r/\S+?/comments/(?P<submission>[\w-]+)'
+    r'(?:/?(?:[\w%%]+/(?P<comment>[\w-]+))?)'
```
dgw, line 33: @Duckle29 pointed out on IRC that this probably isn't necessary if no `%` interpolation is happening.

```suggestion
    r'(?:/?(?:[\w%]+/(?P<comment>[\w-]+))?)'
```

PR #2178, opened 2021-08-28
docs: explicitly select both light and dark Pygments styles

dgw:
### Description
In the dark color scheme particularly, the default Pygments style ("native") has some weird choices in places like imports (underlined names), and bland colors.

Having the styles selected explicitly in the docs' `conf.py` also means we aren't subject to surprise default-value changes from Furo.

### Notes

I'm not _dead_ set on these choices. They're just the best-looking _to me_, out of the options I tested (about 60-70% of the available styles) at https://pygments.org/demo/. Got a suggestion that looks better? I'm all eyes.

In particular, I'd _love_ a more colorful style for light mode to match the choice I made in dark mode, but the light-mode styles that do use significant amounts of color are generally quite muted. The few that choose bolder colors come off as garish to me, or even unreadable (light cyan on white, for example—no thanks).

### Checklist
  - (No code changes.)
  - See comparisons below.

### Comparisons

**Light mode**

sphinx vs. friendly

**Dark mode**

native vs. monokai

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -105,7 +105,8 @@ exclude_patterns = []
 #show_authors = False
 
 # The name of the Pygments (syntax highlighting) style to use.
-pygments_style = 'sphinx'
+pygments_style = 'friendly'
+pygments_dark_style = 'monokai'
 
 # A list of ignored prefixes for module index sorting.
 #modindex_common_prefix = []


Exirel:
If you want an opinion on that:

* I slightly prefer the Sphinx default for light
* I kind of like Monokai more than the default for dark

So I think it's pretty good.

Exirel (approved):

dgw:
I'm honestly just going to call this "an improvement" and leave any better ideas to separate patches. Most of us who read the docs on a regular basis already use the dark theme, I suspect, and that side of things is 👌 already with this change.

PR #2179, opened 2021-08-29
sopel: reduce and reorganize imports for critical modules

Exirel:
### Description

Sopel has critical modules (`sopel.bot`, `sopel.coretasks`, etc.) and I wasn't really happy with their dependencies, starting with a dependency to `sopel.loader`. This is where I started: I tried to remove dependencies to loader when possible or necessary, and I followed that by reducing imports when possible, with few reorganization here and there.

This is not much, and some may argue that this isn't necessary. However, I'd rather take preventive action now against circular imports or dependency hell now than facing these issues later.

Also I had to extract privilege constants from `sopel.plugin` into their own module, which lead me to write some proper documentation about privileges.

### Checklist


sopel/privileges.py
```diff
@@ -0,0 +1,123 @@
+"""Constant for user privileges in channel.
```
dgw, line 1:
```suggestion
"""Constants for user privileges in channels.
```

sopel/privileges.py
```diff
@@ -0,0 +1,123 @@
...
+
+Privilege levels
+================
+
+Historically, there were two user privileges in channels:
+
+* :data:`OP`: channel operator, or chanop, set and unset by ``+o`` and ``-o``
+* :data:`VOICE`: give the privilege to send message to a channel with the
```
dgw, line 9:
```suggestion
* :data:`VOICE`: the privilege to send messages to a channel with the
```

sopel/privileges.py
```diff
@@ -0,0 +1,123 @@
...
+  ``+m`` mode, set and uset by ``+v`` and ``-v``
+
+Since then, other privileges have been used by IRC servers and clients:
```
dgw, line 12:
```suggestion
Since then, other privileges have been adopted by IRC servers and clients:
```

sopel/privileges.py
```diff
@@ -0,0 +1,123 @@
...
+.. important::
+
+    Not all IRC networks support these added privilege modes. If you are
+    writing a plugin for public distribution, ensure your code behaves sensibly
+    if only +v (voice) and +o (op) modes exist.
+
+Compare privileges
+==================
+
+Privileges are powers of two, the bigger the higher the privileges::
```
dgw, line 30:
```suggestion
This module represents privileges as powers of two, with higher values assigned
to higher-level privileges::
```

It was this or a semicolon, and this reads cleaner to me.

sopel/privileges.py
```diff
@@ -0,0 +1,123 @@
...
+
+In that case, ``priv`` contains both VOICE and OP privileges, but not HALFOP.
+"""
+from __future__ import generator_stop
+
+
+VOICE = 1
+"""Privilege level for the +v channel permission
+
+.. versionadded:: 4.1
```
dgw, line 61:
```suggestion
.. versionadded:: 4.1
.. versionchanged:: 8.0
   Moved into :module:`sopel.privileges`.
```

sopel/privileges.py
```diff
@@ -0,0 +1,123 @@
...
+"""
+
+HALFOP = 2
+"""Privilege level for the +h channel permission
+
+.. versionadded:: 4.1
```
dgw, line 67:
```suggestion
.. versionadded:: 4.1
.. versionchanged:: 8.0
   Moved into :module:`sopel.privileges`.
```

sopel/privileges.py
```diff
@@ -0,0 +1,123 @@
...
+    Not all IRC networks support this privilege mode. If you are writing a
+    plugin for public distribution, ensure your code behaves sensibly if only
+    ``+v`` (voice) and ``+o`` (op) modes exist.
+
+"""
+
+OP = 4
+"""Privilege level for the +o channel permission
+
+.. versionadded:: 4.1
```
dgw, line 80:
```suggestion
.. versionadded:: 4.1
.. versionchanged:: 8.0
   Moved into :module:`sopel.privileges`.
```

sopel/privileges.py
```diff
@@ -0,0 +1,123 @@
...
+"""
+
+ADMIN = 8
+"""Privilege level for the +a channel permission
+
+.. versionadded:: 4.1
```
dgw, line 86:
```suggestion
.. versionadded:: 4.1
.. versionchanged:: 8.0
   Moved into :module:`sopel.privileges`.
```

sopel/privileges.py
```diff
@@ -0,0 +1,123 @@
...
+    Not all IRC networks support this privilege mode. If you are writing a
+    plugin for public distribution, ensure your code behaves sensibly if only
+    ``+v`` (voice) and ``+o`` (op) modes exist.
+
+"""
+
+OWNER = 16
+"""Privilege level for the +q channel permission
+
+.. versionadded:: 4.1
```
dgw, line 99:
```suggestion
.. versionadded:: 4.1
.. versionchanged:: 8.0
   Moved into :module:`sopel.privileges`.
```

sopel/privileges.py
```diff
@@ -0,0 +1,123 @@
...
+
+"""
+
+OPER = 32
+"""Privilege level for the +y/+Y channel permissions
+
+Note: Except for these (non-standard) channel modes, Sopel does not monitor or
+store any user's OPER status.
+
+.. versionadded:: 7.0.0
```
dgw, line 115:
```suggestion
.. versionadded:: 7.0
.. versionchanged:: 8.0
   Moved into :module:`sopel.privileges`.
```

docs/source/plugin/privileges.rst
```diff
@@ -0,0 +1,169 @@
...
+
+    MODE #example +ov Nickname Nickname
+
+This will give both OP and Voice privileges to the user named "Nickname" in the
+"#example" channel (and only in this channel). When Sopel receives a MODE
+message it registers and updates its knowledge of a user privileges in a
+channel, which can be used by plugins in various way.
+
+Access right
+============
```
dgw, line 17:
```suggestion
Access rights
=============
```

docs/source/plugin/privileges.rst
```diff
@@ -0,0 +1,169 @@
...
+
+IRC users can have privileges in a **channel**, given by MODE messages such as:
+
+.. code-block:: irc
+
+    MODE #example +ov Nickname Nickname
+
+This will give both OP and Voice privileges to the user named "Nickname" in the
+"#example" channel (and only in this channel). When Sopel receives a MODE
+message it registers and updates its knowledge of a user privileges in a
```
dgw, line 13:
```suggestion
message it registers and updates its knowledge of a user's privileges in a
```

docs/source/plugin/privileges.rst
```diff
@@ -0,0 +1,169 @@
...
+
+    @plugin.require_bot_privilege(privileges.OP)
+    @plugin.require_chanmsg
+    @plugin.command('opbotcommand')
+    def change_topic(bot, trigger):
+        # only if the bot has OP privileges
+
+This way, this command cannot be used if the bot doesn't have the right
+privileges in the channel it is used, independently from the user's privileges
+who invokes the command.
```
dgw, line 60:
```suggestion
privileges in the channel where it is used, independent from the privileges of
the user who invokes the command.
```

docs/source/plugin/privileges.rst
```diff
@@ -0,0 +1,169 @@
...
+
+As per ``require_privilege`` you can provide an error message::
```
dgw, line 62:
```suggestion
As with ``require_privilege``, you can provide an error message::
```

docs/source/plugin/privileges.rst
```diff
@@ -0,0 +1,169 @@
...
+
+    @plugin.require_privilege(privileges.VOICE)
+    @plugin.require_bot_privilege(privileges.OP)
+    @plugin.require_chanmsg
+    @plugin.command('special')
+    def special_command(bot, trigger):
+        # only if the user has +v and the bot has +o (or above)
+
+This way, you will allow a less privileged user to access a command for a more
+privileged bot (this work for any combination of privileges).
```
dgw, line 78:
```suggestion
privileged bot (this works for any combination of privileges).
```

docs/source/plugin/privileges.rst
```diff
@@ -0,0 +1,169 @@
...
+the same plugin callable::
+
+    @plugin.require_privilege(privileges.VOICE)
+    @plugin.require_bot_privilege(privileges.OP)
+    @plugin.require_chanmsg
+    @plugin.command('special')
+    def special_command(bot, trigger):
+        # only if the user has +v and the bot has +o (or above)
+
+This way, you will allow a less privileged user to access a command for a more
```
dgw, line 77:
```suggestion
This way, you can allow a less privileged user to access a command for a more
```

docs/source/plugin/privileges.rst
```diff
@@ -0,0 +1,169 @@
...
+privileged bot (this work for any combination of privileges).
+
+Restrict to user account
+------------------------
+
+Sometimes, a command should be used only by properly authenticated users
+(through IRC services/NickServ), on IRC network that provides that information
+to IRC clients. This can be required with the
+:func:`~sopel.plugin.require_account` decorator::
```
dgw, line 86:
```suggestion
Sometimes, a command should be used only by users who are authenticated via
IRC services. On IRC networks that provide such information to IRC clients,
this is possible with the :func:`~sopel.plugin.require_account` decorator::
```

docs/source/plugin/privileges.rst
```diff
@@ -0,0 +1,169 @@
...
+        # only if the user has +v and has a registered account
+
+This has two consequences:
+
+1. this command cannot be used by users who are not authenticated
+2. this command cannot be used on an IRC network that doesn't allow
+   authentication or doesn't expose that information
+
+It makes your plugin safer to use and prevents the possibility to use it on
+insecure IRC networks.
```
dgw, line 102:
```suggestion
It makes your plugin safer to use and prevents the possibility to use it on
insecure IRC networks.

.. seealso::

   `IRCv3 account-tracking specifications <https://ircv3.net/irc/#account-tracking>`_.   
```

docs/source/plugin/privileges.rst
```diff
@@ -0,0 +1,169 @@
...
+
+
+Getting user privileges in a channel
+====================================
+
+Within a :term:`plugin callable`, you can get access to a user's privileges in
+a channel to check privileges manually. For example, you could adapt the level
+of information your callable provides based on said privileges.
+
+First you need a user's nick and a channel (e.g. from the trigger parameter)
```
dgw, line 112:
```suggestion
First you need a user's nick and a channel (e.g. from the trigger parameter),
```

docs/source/plugin/privileges.rst
```diff
@@ -0,0 +1,169 @@
...
+
+Another option is to use dedicated methods from the ``channel`` object::
+
+    if channel.is_voiced('Nickname'):
+        # user is voiced
+    elif channel.has_privileges('Nickname', plugin.VOICE):
+        # not voiced, but higher privileges
+        # like plugin.HALFOP or plugin.OP
+    else:
+        # no privilege
```
dgw, line 140: Document the _new_ location, right?

```suggestion
    from sopel import privileges

    if user_privileges & privileges.VOICE:
        # user is voiced
    elif user_privileges > privileges.VOICE:
        # not voiced, but higher privileges
        # like privileges.HALFOP or privileges.OP
    else:
        # no privilege

Another option is to use dedicated methods from the ``channel`` object::

    if channel.is_voiced('Nickname'):
        # user is voiced
    elif channel.has_privilege('Nickname', privileges.VOICE):
        # not voiced, but higher privileges
        # like privileges.HALFOP or privileges.OP
    else:
        # no privilege
```

The most important part of this suggestion is the change to `channel.has_privilege()` btw, since the actual method name has no 's'.

docs/source/plugin/privileges.rst
```diff
@@ -0,0 +1,169 @@
...
+
+You can also iterate over the list of users and filter them by privileges::
+
+    # get users with the OP privilege
+    op_users = [
+        user
+        for nick, user in channel.users
+        if channel.is_op(nick, plugin.OP)
```
dgw, line 148: Ditto above.

```suggestion
        if channel.is_op(nick, privileges.OP)
```

docs/source/plugin/privileges.rst
```diff
@@ -0,0 +1,169 @@
...
+    ]
+
+    # get users with OP privilege or above
+    op_or_higher_users = [
+        user
+        for nick, user in channel.users
+        if channel.has_privileges(nick, plugin.OP)
```
dgw, line 155: Ditto again.

```suggestion
        if channel.has_privileges(nick, privileges.OP)
```

dgw (changes requested):
OK, I pulled together a full review. Only took (I think) three sessions—which is shamefully inefficient on my part, but there has been a _lot_ going on IRL and it's been hard to find time to focus on this stuff. Should be more or less caught up on unreviewed PRs after this, though. 🥳

sopel/plugin.py
```diff
@@ -12,6 +12,10 @@
 import functools
 import re
 
+# import and expose privileges as shortcut
```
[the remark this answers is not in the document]
Exirel: I really like them being shortcut, so plugin author can keep using `from sopel import plugin` to use them, while Sopel's own source code can use `sopel.privileges` instead to prevent cyclic imports.

sopel/bot.py
```diff
@@ -17,13 +17,9 @@
...
 
-from sopel import irc, logger, plugins, tools
-from sopel.db import SopelDB
-import sopel.loader
-from sopel.module import NOLIMIT
+from sopel import db, irc, logger, plugin, plugins, tools
 from sopel.plugins import jobs as plugin_jobs, rules as plugin_rules
-from sopel.tools import deprecated, Identifier
-import sopel.tools.jobs
+from sopel.tools import deprecated, Identifier, jobs as tools_jobs
```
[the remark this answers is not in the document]
Exirel: One way would be to have `jobs` directly available through `tools`. Another option is to fusion both type of jobs into one single submodule. I suggest we do that in a future PR.

docs/source/plugin/privileges.rst
Exirel: Ditto @ using `sopel.plugin`  as shortcut intended for plugin authors.

Exirel:
Alright so, 2 topics to discuss that needs an action:

* I like `sopel.plugin` as a shortcut and I think it makes sense to present that to plugin authors and to document it as such (i.e. keep using `sopel.plugin` in the doc). Action: keep or switch to `sopel.privileges` in the doc?
* Plugin, handler, and name conflict: I wanted to prevent a conflict between `from sopel import plugins, plugin` and using the words plugins and plugin for variable name that would conflict with that (specifically in loops). Action: see below.

### Name conflict resolutions

Before even talking about semantics, there is still a name conflict to deal with. There are 3 options:

* keep it that way (handler, handlers for variable, plugin, plugins for module)
* replace handler by plugin_handler, and handlers by plugin_handlers, no more conflict
* use aliases `from sopel import plugin as plugin_, plugins as plugins_`  which I honestly hate but it exists

### Semantics

There is a question behind this, which is "what is a Plugin Handler?", and with it, "what does it do?". I argue that a Plugin Handler is *not* a plugin, it's a handler: it's a proxy between Sopel and its plugins. I argue that its actions are to load, setup, configure, and shutdown the plugin that it handles, and that "plugin" is its action's target. When I read `handler.load()` I read it as "handler, load your plugin", and not "handler, load yourself". So, for me, there is no confusion, however it looks like it is for others.

"Hold up!" said @dgw "why is PyModulePlugin named like that, and not PyModulePluginHandler?" AH AH. Great question my friend: it's because they are internal names, and their "handlerness" comes from their module: `handler.PyModulePlugin` already tells me that it's a Handler, not a "plugin". I tend not to do repetition too much, like `sopel.plugins.handlers.SopelPyModulePluginHandlerClass` (I know, cheeky me!).

However... nothing is that simple, because I know there are functions that call handlers "plugins", which create inconsistency in Sopel's codebase. That's not new, but that's not great either.

Exirel:
* changed "plugin" and "handler" to "plugin_handler" so it's less confusing for, hopefully, everyone
* plugins.jobs and tools.jobs will be handled in a separate PR (in an undefined future time)
* didn't rebase yet but there is a conflict to fix

ping @dgw for the follow up :)

docs/source/plugin/privileges.rst
```diff
@@ -0,0 +1,172 @@
...
+command ``chanopcommand`` in that channel: other users will be ignored by the
+bot. It is possible to tell these users why with the ``message`` parameter::
+
+    @plugin.require_privilege(privileges.OP, 'You need +o privileges.')
+
+.. note::
+
+    A command that requires channel privileges will probably be used for
+    channels only. For that purpose, you should use the
+    :func:`sopel.plugin.require_chanmsg` decorator.
```
dgw, line 43: This is more important than you might think: `require_privilege` behaves as if the given minimum privilege level is always met if the message isn't from a channel. I'd go stronger here:

```suggestion
.. important::

    A command that requires channel privileges will always execute if
    called from a private message to the bot. You can use the
    :func:`sopel.plugin.require_chanmsg` decorator to ignore the command
    if it's called in PMs.
```

docs/source/plugin/privileges.rst
```diff
@@ -0,0 +1,172 @@
...
+
+    @plugin.require_privilege(privileges.VOICE)
+    @plugin.require_bot_privilege(privileges.OP)
+    @plugin.require_chanmsg
+    @plugin.command('special')
+    def special_command(bot, trigger):
+        # only if the user has +v and the bot has +o (or above)
+
+This way, you can allow a less privileged user to access a command for a more
+privileged bot (this works for any combination of privileges).
```
dgw, line 78: Also important here: `require_bot_privilege` will always execute the command if called in PM, because there are no channel privileges. Needs a note about using `require_chanmsg` as appropriate (yes I see it in the examples; still, should have what Furo calls an "admonition").

dgw (commented):
Turns out I had more thoughts left than expected—and that was _after_ writing (then deleting) a whole thing about the `plugin_jobs`/`tools_jobs` thing that will be addressed in a separate patch.

docs/source/plugin/privileges.rst
dgw: @Exirel are you mucking with the rebase and this will be addressed in the next push? I'm nervous about seeing a resolved conversation with no corresponding change yet.

dgw (approved):
I say if any typo or other mistake has eluded both of us for this long, it deserves to make it into `master` and then get patched out in X months when we finally notice. 😁

PR #2181, opened 2021-09-09
clock: add unset functionality

RustyBower:
### Description
Adds functionality for user to unset their timezones if desired

### Checklist

### DIFF
--- a/sopel/modules/clock.py
+++ b/sopel/modules/clock.py
@@ -159,6 +159,14 @@ def update_user(bot, trigger):
         bot.reply('I now have you in the %s timezone.' % zone)
 
 
+@plugin.command('unsettz', 'unsettimezone')
+@plugin.example('.unsettz')
+def unset_user_tz(bot, trigger):
+    """Unset your preferred timezone."""
+    bot.db.delete_nick_value(trigger.nick, 'timezone')
+    bot.reply('Successfully unset timezone')
+
+
 @plugin.command('gettz', 'gettimezone')
 @plugin.example('.gettz Exirel', user_help=True)
 @plugin.example('.gettz', user_help=True)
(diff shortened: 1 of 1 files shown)


dgw:
ctz? tf/ctf?

dgw:
What about `.unsetctf`?

Also musing that maybe it's less awkward to call these "clear" instead of "unset". My mental jury is still out on whether to put that before or after the type (e.g. `.tzclear` vs. `.cleartz`).

half-duplex:
Personally, where a command can be reused without making it too confusing, I prefer that over a new command. `.settz -` or `.settz none`?

RustyBower:
> Personally, where a command can be reused without making it too confusing, I prefer that over a new command. `.settz -` or `.settz none`?

I disagree, I really try to avoid overloading a command where possible, and to @dgw's question, `.settz` and `.unsettz` are natural opposites, versus something like `.tzclear` which is antithetical to the format for `.settz`

dgw:
Welp, I can't laugh at "sorry, misclicked" people any more. Touchpads are something else…

Rusty's points are well taken, and I'm just stuck on the missing `.unsetchanneltf` command now.

half-duplex:
Eh... Personally, I'd still rather one `if` in the .setX than polluting .help by adding another four commands.
I only see one other '.unset*' command, and it's in admin.py for unsetting a config option, where 'none'/'-'/etc would potentially be valid values for .set.

RustyBower:
Good spot @dgw :)

RustyBower:
> Eh... Personally, I'd still rather one `if` in the .setX than polluting .help by adding another four commands.
> I only see one other '.unset*' command, and it's in admin.py for unsetting a config option, where 'none'/'-'/etc would potentially be valid values for .set.

I'll defer to @dgw on this, since he has to maintain this, but personally, I am adamantly against overloading commands (especially a set command with unset functionality) to have complex usage logic, rather than having a few extra commands in help.

RustyBower:
> I'll approve this as-is. @RustyBower please just squash to one commit.

Done

PR #2190, opened 2021-10-05
coretasks: backport SASL abort from #2187

dgw:
### Description
Backport of #2187 to the older SASL code in 7.x, so our stable branch will stop silently ignoring unexpected SASL replies.

### Checklist
- [x] I have read [CONTRIBUTING.md](https://github.com/sopel-irc/sopel/blob/master/CONTRIBUTING.md)
- [x] I can and do license this contribution under the EFLv2
- [x] No issues are reported by `make qa` (runs `make quality` and `make test`)
- [x] I have tested the functionality of the things this change touches

PR #2187, opened 2021-10-01
coretasks: correctly abort SASL PLAIN on invalid server reply

dgw:
### Description
Originally [flagged as](https://lgtm.com/projects/g/sopel-irc/sopel/snapshot/ecf4f1db80a7be5af44438e152aa3dfcf0d39818/files/sopel/coretasks.py?sort=name&dir=ASC&mode=heatmap#L1064) an unused local variable by LGTM, but turned out to be a real bug.

Could be worth cutting a 7.1.5 release.

### Checklist
- [x] I have read [CONTRIBUTING.md](https://github.com/sopel-irc/sopel/blob/master/CONTRIBUTING.md)
- [x] I can and do license this contribution under the EFLv2
- [x] No issues are reported by `make qa` (runs `make quality` and `make test`)
- [x] I have tested the functionality of the things this change touches

PR #2191, opened 2021-10-05
coretasks: emit log entry if SASL auth aborts

dgw:
### Description
Realized in #2190 that there was no logging output in #2187, and added it to that backported patch. Now here is a forward-port of the logging bit. We're all about messy history here at Sopel… 😅

### Checklist


Exirel (approved):

sopel/coretasks.py
```diff
@@ -1066,6 +1066,7 @@ def auth_proceed(bot, trigger):
             return
         else:
             # Not an expected response from the server
+            LOGGER.info("Aborting SASL: unexpected server reply '%s'" % trigger)
```
Exirel, line 1069: I was too quick to approve... I think this should be a warning instead.

```suggestion
            LOGGER.warning("Aborting SASL: unexpected server reply '%s'" % trigger)
```

Exirel (changes requested):

sopel/coretasks.py
dgw: Right, I'm gonna apply this to #2190 too (but not push it until I look into fixing that CI failure for py3.3).

Exirel (approved):

PR #2192, opened 2021-10-06
travis: remove expired DST Root CA X3 cert on trusty

dgw:
### Description
Forcibly remove the expired `mozilla/DST_Root_CA_X3.crt` file and update CA certificate bundle, only on Trusty distro (used for Python 3.3 tests). Also ignore certifi trust store on trusty+py3.3, since certifi maintainers are stubborn (see comments in diff).

Thanks to @half-duplex for helping with the monkey-patch step!

### Checklist
  - Painstakingly: both myself and @half-duplex ran tests in local VMs, and I pushed promising solutions to a `travis-` prefixed test branch to see how they performed on the real thing.
  - If this PR build fails despite all of our advance testing, I might actually cry.

### Notes
Blocks #2190, which can't pass its CI tests until this is fixed.

Considering an alternative approach that replaces the `cp` with an environment variable (`REQUESTS_CA_BUNDLE`).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,6 +39,18 @@ install:
   - pip install --upgrade "setuptools<=39.2.0; python_version < '3.9'"
   - pip install --upgrade -r requirements.txt -r dev-requirements.txt
   - python setup.py develop
+before_script:
+  # deal with Let's Encrypt's ISRG Root CA X1 being cross-signed by an expired root
+  # only needed on trusty - mark the bad cert disabled in conf, update ca-certificates
+  # bundle, then tell python-requests to use the system bundle instead of certifi's
+  # (as of 2021-10-05, certifi still stubbornly bundles the expired root;
+  # see https://github.com/certifi/python-certifi/pull/162
+  # and https://bugzilla.mozilla.org/show_bug.cgi?id=1733560 for updates)
+  - if [ "$TRAVIS_DIST" == "trusty" ]; then
+      sudo sed -re 's#^(mozilla/DST_Root_CA_X3.crt)$#!\1#' -i /etc/ca-certificates.conf;
+      sudo update-ca-certificates;
+      export REQUESTS_CA_BUNDLE=/etc/ssl/certs/ca-certificates.crt;
+    fi
 script:
   - make travis
 env:


.travis.yml
```diff
@@ -39,6 +39,20 @@ install:
...
+  # bundle, then remove the .crt file completely from the only place it should remain
+  # finally, monkeypatch certifi to use the system's fixed one (as of 2021-10-05,
+  # certifi still stubbornly bundles the expired root;
+  # see https://github.com/certifi/python-certifi/pull/162
+  # and https://bugzilla.mozilla.org/show_bug.cgi?id=1733560 for updates)
+  - if [ "$TRAVIS_DIST" == "trusty" ]; then
+      sudo sed -re 's#(mozilla/DST_Root_CA_X3.crt)#!\1#' -i /etc/ca-certificates.conf;
+      sudo update-ca-certificates;
+      sudo rm /usr/share/ca-certificates/mozilla/DST_Root_CA_X3.crt;
+      cp /etc/ssl/certs/ca-certificates.crt /home/travis/virtualenv/python3.3/lib/python3.3/site-packages/certifi/cacert.pem;
```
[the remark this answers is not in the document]
half-duplex: Yep, looks right.
I also explicitly tested `!mozilla/DST_Root_CA_X3.crt` and that does remove the file and make curl work, so the `rm` should be superfluous.

I don't have a sane case where this would matter, especially before 8.0, but I usually try to `^...$` my seds

.travis.yml
```diff
@@ -39,6 +39,20 @@ install:
...
+  # only needed on trusty - mark the bad cert disabled in conf, update ca-certificates
+  # bundle, then remove the .crt file completely from the only place it should remain
+  # finally, monkeypatch certifi to use the system's fixed one (as of 2021-10-05,
+  # certifi still stubbornly bundles the expired root;
+  # see https://github.com/certifi/python-certifi/pull/162
+  # and https://bugzilla.mozilla.org/show_bug.cgi?id=1733560 for updates)
+  - if [ "$TRAVIS_DIST" == "trusty" ]; then
+      sudo sed -re 's#(mozilla/DST_Root_CA_X3.crt)#!\1#' -i /etc/ca-certificates.conf;
+      sudo update-ca-certificates;
+      sudo rm /usr/share/ca-certificates/mozilla/DST_Root_CA_X3.crt;
```
half-duplex, line 53:
```suggestion
```

.travis.yml
```diff
@@ -39,6 +39,20 @@ install:
...
+before_script:
+  # deal with Let's Encrypt's ISRG Root CA X1 being cross-signed by an expired root
+  # only needed on trusty - mark the bad cert disabled in conf, update ca-certificates
+  # bundle, then remove the .crt file completely from the only place it should remain
+  # finally, monkeypatch certifi to use the system's fixed one (as of 2021-10-05,
+  # certifi still stubbornly bundles the expired root;
+  # see https://github.com/certifi/python-certifi/pull/162
+  # and https://bugzilla.mozilla.org/show_bug.cgi?id=1733560 for updates)
+  - if [ "$TRAVIS_DIST" == "trusty" ]; then
+      sudo sed -re 's#(mozilla/DST_Root_CA_X3.crt)#!\1#' -i /etc/ca-certificates.conf;
```
half-duplex, line 51:
```suggestion
      sudo sed -re 's#^(mozilla/DST_Root_CA_X3.crt)$#!\1#' -i /etc/ca-certificates.conf;
```

half-duplex (approved):
Nits, these and `REQUESTS_CA_BUNDLE=` are cleaner (and severable) but should be functionally identical and this patch dies with 7.x so :shrug:

Other thoughts:
- If anyone is actually using 3.3 in a venv they'll probably have this problem, if we care about fully supporting 3.3 (I don't) it should be documented.
- Why doesn't this affect newer python? My py3.9 also uses certifi's store in a venv, but works fine... openssl version?
- Nobody should be using python 3.3 anyway: I am not aware of [any supported distro/release without 3.4+ available](https://repology.org/project/python/versions)

PR #2193, opened 2021-10-09
bot: use `optionxform()`-ed setting names for existence check

dgw:
### Description
This makes the "setting is defined in section" check case-insensitive, which matches the parsing behavior of `RawConfigParser`.

The `ip` plugin has a `GeoIP_db_path` setting that will erroneously warn users who set it using `sopel-plugins configure`, because the parser writes out its normalized names (in this case, `geoip_db_path`) and then the previous check using `hasattr()` wouldn't find the normalized name.

### Checklist
  - `GeoIP_db_path`, `geoip_db_path`, and `i_dont_exist` settings under the `[ip]` section all behaved as expected (respectively: silent, silent, warning).

### Notes
Trivial to backport onto 7.1.x (I've already test-cherry-picked the commit locally; no conflicts), so might as well include it in the upcoming maintenance release we already plan to make.

A nicer future solution could be overriding `StaticSection.__getattr__()` to just always be case insensitive—but such a big change wouldn't fly in a maintenance release.

### DIFF
--- a/sopel/bot.py
+++ b/sopel/bot.py
@@ -9,6 +9,7 @@ from __future__ import generator_stop
 
 from ast import literal_eval
 from datetime import datetime
+import inspect
 import itertools
 import logging
 import re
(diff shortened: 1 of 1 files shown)


dgw:
Oh yeah, I'll have to make that not-a-set-comprehension for py2 when I backport this 😑

PR #2202, opened 2021-10-29
reddit: fix shortlinks

half-duplex:
### Description
reddit.py doesn't correctly handle redd.it shortlinks (anymore), and doesn't handle reddit.com shortlinks at all.

```irc
< xnaas> https://redd.it/qhoy7z
<+Sopel> Unexpected error (no such group) from xnaas at 2021-10-29 15:15:13.406136. Message was: https://redd.it/qhoy7z
```

This makes the "comment" group optional, and adds the "submission" group name and reddit.com domain to the shortlink regex.

### Checklist
- [x] I have read [CONTRIBUTING.md](https://github.com/sopel-irc/sopel/blob/master/CONTRIBUTING.md)
- [x] I can and do license this contribution under the EFLv2
- [x] No issues are reported by `make qa` (runs `make quality` and `make test`)
- [x] I have tested the functionality of the things this change touches

PR #2208, opened 2021-11-07
reddit: fix matching twice for reddit.com without www.

dgw:
### Description
Addresses a regression introduced by #2202 (itself a regression fix, ironically), which added `reddit.com` to the `short_post_url` pattern and made it possible for a post or comment URL to match BOTH that and `post_or_comment_url`.

Since URLs match individually, and not as part of the line, all we need to make sure a shortlink really is one is to ensure that the link ends after the submission ID and optional trailing slash.

### Checklist
  - Better than the author of #2202 did, no doubt.
  - This does not absolve me of failing to throw a battery of tests at the change myself. Past experience should have taught all of us by now that messing with `reddit.py`'s URL matching invites madness.

### Notes
To the PR template's question: This does not resolve any GitHub issues, because @xnaas doesn't "submit GitHub issues while naked in bed" and reported the bug on IRC instead.

### DIFF
--- a/sopel/modules/reddit.py
+++ b/sopel/modules/reddit.py
@@ -32,7 +32,7 @@ post_or_comment_url = (
     r'/r/\S+?/comments/(?P<submission>[\w-]+)'
     r'(?:/?(?:[\w%]+/(?P<comment>[\w-]+))?)'
 )
-short_post_url = r'https?://(redd\.it|reddit\.com)/(?P<submission>[\w-]+)'
+short_post_url = r'https?://(redd\.it|reddit\.com)/(?P<submission>[\w-]+)/?$'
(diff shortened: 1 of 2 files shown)


Exirel (approved):
If you say you tested it well... 😁

xnaas:
> To the PR template's question: This does not resolve any GitHub issues, because @xnaas doesn't "submit GitHub issues while naked in bed" and reported the bug on IRC instead.

The system works.

dgw:
OK, this py3.9 CI failure that magically appeared in #2209 definitely isn't related to that patch. @Exirel help, what could have possibly changed in the last two days to break ONLY this version of Python?

dgw:
CI failure discussion has its own issue now: #2210

xnaas:
༼ つ ◕_◕ ༽つ minor release

PR #2211, opened 2021-11-09
cli: fix legacy parser options for Py3.9.8 (Sopel 7.1.x)

### LINKED ISSUE #2210: `argparse` subparser defaults in Python 3.9+
Python seems to be making some [changes to `argparse`'s handling of subparser defaults](https://bugs.python.org/issue45235).

- The patch has [landed in Python 3.9.8](https://docs.python.org/release/3.9.8/whatsnew/changelog.html#library), and is causing CI failures for us now that GitHub Actions uses the new version.
- Python 3.10 shows the same patch in "next", meaning it will likely break whenever 3.10.1 comes out.
- [Python 3.11.0 alpha 1](https://docs.python.org/3.11/whatsnew/changelog.html?highlight=argparse#python-3-11-0-alpha-1) shipped this patch a month ago.

(All above verified via in-browser <kbd>Ctrl</kbd>-<kbd>F</kbd> for 'argparse' in the respective changelogs. The relevant patch is listed under the "Library" subheader.)

Our CI failures all come from numerous jobs in Sopel's `test_cli_run.py` suite, where in every case I looked at, the "custom" value passed in to validate the correct behavior is overridden by the option's default value. For example:

```
_______________________ test_build_parser_legacy_config ________________________
test/cli/test_cli_run.py:92: in test_build_parser_legacy_config
    assert options.config == 'custom'
E   AssertionError: assert 'default' == 'custom'
E     - default
E     + custom
```

Only two commits touched CPython's `Lib/argparse.py` after the tag date for 3.9.7, [one of which](https://github.com/python/cpython/commit/fb7203453855a43907048ff2b8f9b890305b3d15) appears to be irrelevant (it messes with some help stuff). That leaves https://github.com/python/cpython/commit/a18d52269ab6071a605d6c72f6af585a4c533ca4, a backport of the patch for [bpo-45235](https://bugs.python.org/issue45235).

Exirel:
### Description

The hack adds a prefix to the "dest" of the legacy options added to the main parser. It works because these options are not actually used by the legacy subparser: they exists only to trick the help message into something usable by end-users. It's rather ugly, but at this point we already know that the "legacy" subparser trick is ugly.

We are only fixing the issue introduced by Python 3.9.8 that "fixed" something in argparse that broke our use-case. Let's not delve into that for now.

Should fix #2210 

### Checklist

  * In theory that fixes the issue for Python 3.9.8 but I'll let Travis or Github Action ensure that

### DIFF
--- a/sopel/cli/run.py
+++ b/sopel/cli/run.py
@@ -118,7 +118,7 @@ def run(settings, pid_file, daemon=False):
     os._exit(0)
 
 
-def add_legacy_options(parser):
+def add_legacy_options(parser, prefix=''):
     """Add legacy options to the argument parser.
 
     :param parser: argument parser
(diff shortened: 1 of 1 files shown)


dgw:
FWIW, the patch to revert the change that makes this patch necessary is awaiting merge and backporting. python/cpython#29525

dgw:
The release of Python 3.9.9 made this hotfix unnecessary. Full details in #2210.

Let's keep discussion of whether to do this anyway in #2212 (the future-proof patch version). This can always be reopened.

PR #2212, opened 2021-11-10
cli: fix legacy parser options for Py3.9.8 (Sopel 8.x)

Exirel:
### Description

Same as #2211 but for the current development version (8.x). I hope Github Action will pass!

### Checklist

### DIFF
--- a/sopel/cli/run.py
+++ b/sopel/cli/run.py
@@ -120,7 +120,7 @@ def run(settings, pid_file, daemon=False):
     os._exit(0)
 
 
-def add_legacy_options(parser):
+def add_legacy_options(parser, prefix=''):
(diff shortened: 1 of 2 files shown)


dgw:
FWIW, the patch to revert the change that makes this patch necessary is awaiting merge and backporting. python/cpython#29525

dgw:
Open to suggestions for why we should do this anyway, even though Python 3.9.9 reverted the responsible change.

If the consensus is to skip this "hack" and wait for further news (my inclination), that's the reason for which I've left #2210 open.

Exirel:
Given that most of the parser will be reworked to remove deprecated options with a better "default action" system... I think we can close this and I'll make sure to include that question into the new version.

PR #2214, opened 2021-11-26
wiktionary: allow for multi-line etymologies

dgw:
### Description
Multi-paragraph etymologies no longer get truncated to only the _last_ line.

If the concatenated etymology is too long to fit in an IRC line, it's safely truncated by the existing logic to handle overly long output.

This patch definitely makes me want to give wiktionary the same kind of makeover that wikipedia got a while back. Which is to say: a proper parser-based implementation instead of...whatever you call this.

### Checklist


Exirel (approved):
I had a nitpick in mind that is not worth the effort, given we all wish a rewrite of whatever "that" is, as you said. Let's roll with that.

PR #2216, opened 2021-11-29
reddit, test: even more of the comments link format is optional

dgw:
### Description
Reddit will accept `/r/subreddit/comments/id`, `/comments/id`, or even just `/id`... The variations never end.

Follow-up to #2208. After merging I found some further tweaks that hadn't been pushed up yet, probably because of #2210. Could use input from @half-duplex, the person who pointed out these additional URL formats' existence.

While preparing this patch so it wouldn't conflict with the prior merge I noticed that my newly created test file for `reddit` used Windows line endings, because of course it did (probably a bad editor default on my newer laptop). Review 2bad1e1bcb53f6b060a1d2daacb191706304617b on its own for the meaningful changes; 7af388922b628dcd378f4a2ea44248a5a9c78616 contains only whitespace (correcting bad line endings).

### Checklist


test/modules/test_modules_reddit.py
```diff
@@ -1,76 +1,79 @@
...
+    'https://www.reddit.com/r/subname/?param=value',
+)
+
+
+@pytest.mark.parametrize('link', MATCHING_URLS)
+def test_url_matching(link, bot):
+    line = PreTrigger(bot.nick, ':User!user@irc.libera.chat PRIVMSG #channel {}'.format(link))
+    matches = bot.rules.get_triggered_rules(bot, line)
+
+    assert len([match for match in matches if match[0].get_plugin_name() == 'reddit']) == 1
```
Exirel, line 71: I think you should test *which* reddit rule is actually called, like `image_info` or `video_info`, etc.

Exirel (changes requested):
I like the tests. I just want more.

test/modules/test_modules_reddit.py
dgw: That was the original goal, but I simply couldn't find a way to get that information back from the list of triggered rules.

Exirel: Too busy at the time to really pay attention to details like that.

I think you can add a parameter to the `pytest.mark.parametrize` so that you get both the line and the expected name for the rule. You get the rule as the first parameter of match, and you can ask it its label through `get_rule_label`: that will return either the command name for a named rule, or the handler name for a generic rule (or the named label if `plugin.label` was used).

dgw: Haven't a clue why I thought that wouldn't work. Maybe I was hung up on checking precisely which regex matched (handlers like `post_or_comment_info` have multiple matching patterns defined). No matter… Amended tests pushed.

Exirel (approved):

PR #2218, opened 2021-12-02
url: bump user-agent

xnaas:
### Description

Update the user-agent string to Chrome 96, the 2nd most popular UA listed on
https://techblog.willshouse.com/2012/01/03/most-common-user-agents/

### Checklist
    - [x] I skimmed it, anyway.

### DIFF
--- a/sopel/modules/url.py
+++ b/sopel/modules/url.py
@@ -28,7 +28,7 @@ LOGGER = logging.getLogger(__name__)
 USER_AGENT = (
     'Mozilla/5.0 (Windows NT 10.0; Win64; x64) '
     'AppleWebKit/537.36 (KHTML, like Gecko) '
-    'Chrome/78.0.3904.108 Safari/537.36'
+    'Chrome/96.0.4664.45 Safari/537.36'
 )
 DEFAULT_HEADERS = {
     'User-Agent': USER_AGENT,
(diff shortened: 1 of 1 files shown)


xnaas:
In theory this might help get less "are you a robot?" prompts? Bloomberg is at least one site this could be testing against, probably.

dgw:
Don't let me forget about this before the next release, whether it's 7.1.8 or 8.0.0. But I don't want to merge it until we're close to a release, so the UA string doesn't get too stale again waiting for a new package.

We could also think about delegating this to a separate package users can update without needing a new Sopel release. PyPI has a surprisingly high number of packages dedicated to generating fake UAs.

xnaas:
Could also just be a bot owner configurable asset?

Like `.set url.ua Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/96.0.4664.45 Safari/537.36`? Dunno; food for thought.

dgw:
There's no `.set` command, but that's beside the point. Some kind of override could make sense if it's not pulled from something that can update as needed, independently of Sopel (or the eventual `sopel-url` plugin this should become).

xnaas:
> There's no `.set` command

*wondering how high dgw is*

dgw:
> > There's no `.set` command
> 
> _wondering how high dgw is_

_sleep-deprived enough to confuse #1986 for an IRC command patch_

xnaas:
Is that what you wanted, @dgw?? 🤔

dgw (approved):

PR #2220, opened 2021-12-07
plugins.rules: fix conflicting attributes in base classes

Exirel:
### Description

One PR, two issues:

* LGTM issues that are better fixed than ignored
* ~Type hint warning for documentation~
* Update documentation to replace the mixin by the abstract named rule class
* Fix a deprecated usage of an abstract method decorator in Rule

### Checklist

### DIFF
--- a/sopel/plugins/rules.py
+++ b/sopel/plugins/rules.py
@@ -465,7 +465,7 @@ class AbstractRule(abc.ABC):
 
     """
     @classmethod
-    @abc.abstractclassmethod
+    @abc.abstractmethod
     def from_callable(cls: Type[TypedRule], settings, handler) -> TypedRule:
         """Instantiate a rule object from ``settings`` and ``handler``.
 
@@ -1063,8 +1063,8 @@ class Rule(AbstractRule):
         return exit_code
 
 
-class NamedRuleMixin:
-    """Mixin for named rules.
+class AbstractNamedRule(Rule):
+    """Abstract base class for named rules.
 
     A named rule is invoked by using a specific word, and is usually known
     as a "command". For example, the command "hello" is triggered by using
@@ -1072,6 +1072,11 @@ class NamedRuleMixin:
 
     A named rule can be invoked by using one of its aliases, also.
     """
+    def __init__(self, name, aliases=None, **kwargs):
+        super().__init__([], **kwargs)
+        self._name = name
+        self._aliases = tuple(aliases) if aliases is not None else tuple()
+
     @property
     def name(self):
         return self._name
@@ -1133,8 +1138,15 @@ class NamedRuleMixin:
 
         return name
 
+    @abc.abstractmethod
+    def get_rule_regex(self):
+        """Make the rule regex for this named rule.
 
-class Command(NamedRuleMixin, Rule):
+        :return: a compiled regex for this named rule and its aliases
+        """
+
+
+class Command(AbstractNamedRule):
     """Command rule definition.
 
     A command rule (or simply "a command") is a named rule, i.e. it has a known
@@ -1196,11 +1208,9 @@ class Command(NamedRuleMixin, Rule):
                  help_prefix=COMMAND_DEFAULT_HELP_PREFIX,
                  aliases=None,
                  **kwargs):
-        super().__init__([], **kwargs)
-        self._name = name
+        super().__init__(name, aliases=aliases, **kwargs)
         self._prefix = prefix
         self._help_prefix = help_prefix
-        self._aliases = tuple(aliases) if aliases is not None else tuple()
         self._regexes = (self.get_rule_regex(),)
 
     def __str__(self):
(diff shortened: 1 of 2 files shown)


sopel/plugins/rules.py
```diff
@@ -1063,7 +1063,7 @@ def execute(self, bot, trigger):
         return exit_code
 
 
-class NamedRuleMixin:
+class AbstractNamedRule(Rule):
     """Mixin for named rules.
```
dgw, line 1067: It's not a Mixin any more.

sopel/plugins/rules.py
```diff
@@ -1063,7 +1063,7 @@ def execute(self, bot, trigger):
         return exit_code
 
 
-class NamedRuleMixin:
+class AbstractNamedRule(Rule):
```
dgw, line 1066: An abstract class that inherits from a concrete one? Seems like one could instantiate an `AbstractNamedRule` without errors, which is probably not intended.

dgw (changes requested):
Below is more of an idea. The "request changes" review status is about what I said in the line notes.

Surely there's a way to deal with this that doesn't require removing the mixin pattern? Off the top of my head, adding an `AnonymousRuleMixin` that provides `get_rule_label()` for `Rule`, leaving `NamedRuleMixin` to provide it for the command-type rules.

sopel/plugins/rules.py
Exirel: Hm, yeah, right, I need to change that somehow.

dgw:
Was troubleshooting for some other doc-related thing I'm trying to fix, and was checking to see if this patch affected that issue at all. It doesn't, but test-building the docs on this branch showed me that it breaks some of `docs\source\plugin\internals.rst` due to the reference left behind to a `NamedRuleMixin` class that no longer exists after this change.

sopel/plugins/rules.py
Exirel: Since there is now an abstract method, you can't do that anymore. That's better. :)

Exirel:
I stole the type-hint fix from #2223 and also the classmethod/abstractclassmethod warning since I'm touching the same files/area.

dgw:
> I stole the type-hint fix from #2223 and also the classmethod/abstractclassmethod warning since I'm touching the same files/area.

I wouldn't do that just because you're touching the same stuff. Let this PR be about fixing the code, and #2223 can be about making the docs better.

Exirel:
My main issue is that I need to check if the doc properly build, which it doesn't, and specifically on the classes I touch/modify. So maybe I just need #2223 to be merged before I can move forward.

As for the stolen part, don't worry: I did forgot the co-author tag, and that's my fault, I was a bit caught up into too many PRs at once, and I should have been more careful.

sopel/plugins/rules.py
```diff
@@ -1319,13 +1329,11 @@ def from_callable(cls, settings, handler):
         return cls(**kwargs)
 
     def __init__(self, nick, name, nick_aliases=None, aliases=None, **kwargs):
-        super().__init__([], **kwargs)
+        super().__init__(name, aliases=aliases, **kwargs)
         self._nick = nick
-        self._name = name
         self._nick_aliases = (tuple(nick_aliases)
                               if nick_aliases is not None
                               else tuple())
```
dgw, line 1336: Oof, I didn't notice before that every `NickCommand` is storing the bot's nick and aliases. Nothing to be done about it here, though. Don't even want to think about the kind of gymnastics that would be required to get rid of these attributes and the constructor parameters…

dgw (approved):
Yeah, I'm just dropping this line note for posterity, not because I want something to change as a result. Let's say this is ready.

sopel/plugins/rules.py
Exirel: Always has been. 😁 

After all, you *need* the bot's nick and its aliases from configuration in order to have `BotName: command trigger`. What bugs you about that?

dgw: Just seems like something that should be handled at a different layer, but as I said… it'd take some hoop-jumping, and offhand I don't see a way to do it without some annoying bidirectional references (bot <-> rule manager <-> rule).

Not worth it at the moment. Might be different if we decided to support runtime changes of the bot's nick.

dgw:
> My main issue is that I need to check if the doc properly build, which it doesn't, and specifically on the classes I touch/modify. So maybe I just need #2223 to be merged before I can move forward.

I spent a bit of time on that today and will shortly post a follow-up there with some things I noticed. But the bottom line for _this_ patch is going to be that I no longer think it's a good idea to mess with Sphinx in this PR. Let's get the code changes from this in, only, and then I'll resume work on #2223 to check for any other warnings/issues in the docs.

Exirel:
> I no longer think it's a good idea to mess with Sphinx in this PR

I kept the strict minimum that makes sense in this PR regarding the documentation: I replaced the autodoc for the mixin by the autodoc for the abstract class. I feel like it's a fair change of the documentation while not messing with Sphinx, i.e. not trying to fix too much.

dgw:
Ack, thanks Exirel. I'll approve and merge once squashed.

Exirel:
And done.

dgw (approved):
Remaining changes look sane, plus I know we have oodles of tests for this part of the code. 😏

PR #2221, opened 2021-12-08
adminchannel: fix unmatchable caret in regex

dgw:
### Description
`configureHostMask` has had an unmatchable caret in one of its regular expressions since 2014. The fix for master/8.0 will be more like a minor refactor; this is just a patch to fix the obvious bug for stable.

### Checklist
- [x] I have read [CONTRIBUTING.md](https://github.com/sopel-irc/sopel/blob/master/CONTRIBUTING.md)
- [x] I can and do license this contribution under the EFLv2
- [x] No issues are reported by `make qa` (runs `make quality` and `make test`)
- [x] I have tested the functionality of the things this change touches

### Notes
Merging to `backports-7.1.7`, which is where I've been staging all of the backported patches that were originally merged to `master`. Doing it this way (instead of sending this to `7.1.x`) lets me avoid an annoying rebase later.

PR #2222, opened 2021-12-08
adminchannel: fix, simplify, and comment hostmask utility function

dgw:
### Description
The "Big boy" version of #2221.

This is perhaps the most roundabout way to resolve a single LGTM alert in Sopel's history, but it improves a lot of stuff about the `adminchannel` plugin's `configureHostMask` helper function:

- Fix a broken regex (unmatchable caret, the thing LGTM complained about)
- Reduce the number of distinct cases
- Add comments explaining what each case is supposed to do
- Use exceptions to alert the caller about invalid input instead of a weak sentinel value of `''` (empty string)
- Add test cases to make sure that everything behaves as expected

Notably, adding tests also picked out a couple of edge cases that weren't handled, which I then also fixed.

The "Big boy" version of #2221.

### Checklist
  - Well yes, that's what the new file under `test/` is for.

### DIFF
--- a/sopel/modules/adminchannel.py
+++ b/sopel/modules/adminchannel.py
@@ -144,8 +144,8 @@ def configureHostMask(mask):
     if re.match(r'^[^!@\s]+![^!@\s]+@[^!@\s]+$', mask) is not None:
         return mask
 
-    # invalid format... empty-string is used as a sentinel value
-    return ''
+    # not a shortcut nor a valid hostmask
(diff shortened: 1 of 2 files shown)


Exirel (approved):
I don't understand everything, but there are tests, and they look way easier to understand, so I'm all for it. 👍

dgw:
Merged along with #2221 so we don't have a fixed bug in stable that's still unfixed on master.

PR #2224, opened 2021-12-19
db: add `forget_channel()` and `forget_plugin()` methods

dgw:
### Description
Tin, really. Up to now it's been possible to delete a _user_ (nick group) from the database, but not possible to forget about a channel or plugin without manually writing SQL for `db.execute()`. Seemed like deletion of value groupings should be as parallel as possible, just like management of individual values is between the three types.

I see these as most useful for cleaning up channels that the bot is leaving permanently, or for plugin authors to add graceful "uninstall preparation" features to clean up whatever detritus the plugin has left in the database.

### Checklist

### Notes
There's a deprecation & rename in here too, moving `delete_nick_group()` to `forget_nick_group()`, but I kept it in a separate commit so it's easy to drop if that seems like a bad idea to the rest of y'all.

I'm continuing to try and think of other useful cases that aren't covered by the existing database functions. Might follow up with methods to delete a given key from all nicks/channels. That also applies to the "I'm about to uninstall a plugin forever" situation.

### DIFF
--- a/sopel/db.py
+++ b/sopel/db.py
@@ -724,6 +724,28 @@ class SopelDB:
         finally:
             self.ssession.remove()
 
+    def forget_channel(self, channel):
+        """Remove all of a channel's stored values.
+
+        :param str channel: the name of the channel for which to delete values
+        :raise ~sqlalchemy.exc.SQLAlchemyError: if there is a database error
+
+        .. important::
+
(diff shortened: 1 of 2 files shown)


Exirel (approved):
It looks good to me. Even tho  I know for a fact that it might have a slight conflict with something I'm working on to implement the weird identifier logic.

PR #2225, opened 2021-12-20
wikipedia: refine no-consume logic for section hatnotes

dgw:
### Description
Sometimes note templates include inline style tags. Since their contents appear just like any other data to the parser, it will output the CSS unless told to ignore them.

### Checklist

### Notes
I'm not going to start a 7.1.8 milestone just for this, because it's quite the edge case. But if some other more important bug crops up that _is_ worth starting a milestone for, this should be included.


Exirel (approved):
It sounds sane. As sane as a parser can be.

PR #2227, opened 2021-12-24
sopel: drop support for Python 3.6 and require 3.7+

Exirel:
### Description

Tin. Python 3.6 died yesterday. I checked the warning for 3.6, they were written for Sopel 8, so we can either keep 3.6 or drop it now. Either way, this is not a big change, unlike Python 2.

I haven't modified the usage of `OrderedDict`, because there are slight difference between `dict` and `OrderedDict` in particular when considering equality. 

I haven't check the "future" import either, hence the "draft" status for now.

### Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,7 +20,6 @@ jobs:
     strategy:
       matrix:
         python-version:
-          - 3.6
           - 3.7
           - 3.8
           - 3.9
--- a/setup.cfg
+++ b/setup.cfg
@@ -16,14 +16,13 @@ classifiers =
     License :: Eiffel Forum License (EFL)
     License :: OSI Approved :: Eiffel Forum License
     Operating System :: POSIX :: Linux
-    Programming Language :: Python :: 3.6
     Programming Language :: Python :: 3.7
     Programming Language :: Python :: 3.8
     Programming Language :: Python :: 3.9
     Topic :: Communications :: Chat :: Internet Relay Chat
 
 [options]
-python_requires = >=3.6, <4
+python_requires = >=3.7, <4
 packages = find:
 zip_safe = false
 
--- a/setup.cfg
+++ b/setup.cfg
@@ -49,12 +49,10 @@ ignore =
     # These are forbidding certain __future__ imports. The future-import plugin
     # has errors both for having and not having them; we want to have these until
     # Sopel no longer supports Python versions that require them.
-    FI55,
+    FI58,
     # These would require future imports that are not needed any more on Sopel's
-    # oldest supported Python version (3.6).
-    FI10,FI11,FI12,FI13,FI14,FI16,FI17,
-    # Ignore "annotations" future import, since it's not available before 3.7
-    FI18
+    # oldest supported Python version (3.7).
+    FI10,FI11,FI12,FI13,FI14,FI15,FI16,FI17,
 exclude =
     docs/*,
     env/*,
--- a/sopel/bot.py
+++ b/sopel/bot.py
@@ -5,7 +5,7 @@
 #
 # Licensed under the Eiffel Forum License 2.
 
-from __future__ import generator_stop
+from __future__ import annotations
 
 from ast import literal_eval
 from datetime import datetime
--- a/sopel/bot.py
+++ b/sopel/bot.py
@@ -1157,8 +1157,8 @@ class Sopel(irc.AbstractBot):
             The Python documentation for the `re.search`__ function and
             the `match object`__.
 
-        .. __: https://docs.python.org/3.6/library/re.html#re.search
-        .. __: https://docs.python.org/3.6/library/re.html#match-objects
+        .. __: https://docs.python.org/3.7/library/re.html#re.search
+        .. __: https://docs.python.org/3.7/library/re.html#match-objects
 
         """
         for regex, function in self._url_callbacks.items():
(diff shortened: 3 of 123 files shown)


dgw:
Might as well take this opportunity to add 3.10 to CI, no?

Exirel:
>  Test suite (3.6) Expected — Waiting for status to be reported 

I've no idea what I missed here.

dgw:
> >  Test suite (3.6) Expected — Waiting for status to be reported 
> 
> I've no idea what I missed here.

It's because that check is marked as required in the repo settings. Before merging I would just have to update the list of expected checks.

.github/workflows/ci.yml
```diff
@@ -20,7 +20,6 @@ jobs:
     strategy:
       matrix:
         python-version:
-          - 3.6
           - 3.7
           - 3.8
           - 3.9
```
dgw, line 25:
```suggestion
          - 3.9
          - 3.10
```

setup.cfg
```diff
@@ -16,14 +16,13 @@ classifiers =
     License :: Eiffel Forum License (EFL)
     License :: OSI Approved :: Eiffel Forum License
     Operating System :: POSIX :: Linux
-    Programming Language :: Python :: 3.6
     Programming Language :: Python :: 3.7
     Programming Language :: Python :: 3.8
     Programming Language :: Python :: 3.9
```
dgw, line 21:
```suggestion
    Programming Language :: Python :: 3.9
    Programming Language :: Python :: 3.10
```

sopel/bot.py
```diff
@@ -1157,8 +1157,8 @@ def search_url_callbacks(self, url):
             The Python documentation for the `re.search`__ function and
             the `match object`__.
 
-        .. __: https://docs.python.org/3.6/library/re.html#re.search
-        .. __: https://docs.python.org/3.6/library/re.html#match-objects
+        .. __: https://docs.python.org/3.7/library/re.html#re.search
+        .. __: https://docs.python.org/3.7/library/re.html#match-objects
```
dgw, line 1161: Surely at least `re.search` could use an intersphinx reference instead of a ha(n|r)d-coded URL? It should be possible to <tt>:ref:\`match-objects\`</tt>, too (also in `sopel/plugins/rules.py`).

Perhaps should be changed in one of the doc cleanups you/we are planning, instead of this PR, but I'm still going to call attention to it because it's being touched here. 😁

dgw (commented):
Suggestions from me should be the only changes needed to add Python 3.10 to CI and reflect its "officially supported" status in the package once built.

sopel/bot.py
Exirel: Probably. I'll leave that to a future PR, one that fix all the "Pattern" and "Match" not found errors. 😁

Exirel:
So Github CI thinks it's "3.1" and not "3.10"... weird.

.github/workflows/ci.yml
```diff
@@ -20,10 +20,10 @@ jobs:
...
+          - 3.10
```
Exirel, line 26:
```suggestion
          - '3.10'
```

Let's try with a quote, because apparently YAML thinks it's a float... And people wonder why I hate YAML...

Exirel:
Alright, now we can see that Python 3.10 is not supported by `requests` yet. See also: https://pyreadiness.org/3.10/

dgw:
I see a lot of transitive requirements in the "not supported" category, too—stuff we don't use directly, but that things we require depend on.

Let's abandon adding 3.10 in this patch and try again after the next `requests` release. (2.27 is supposed to be the last that will support Python 2.7, too, so that will be time to revisit our 7.x branch.)

Exirel:
Weird, there is an isup test that fails in all three versions. I'm looking at potential cause, so here is the version installed:

- requests-2.27.1
- requests-mock~=1.9.3

I'll test locally to see if I can reproduce or not.

Exirel:
I can confirm that with `request` 2.26 the test pass, not with 2.27. This will need another PR to fix that.

Exirel:
Following [Python 3.7 documentation](https://docs.python.org/3.7/library/__future__.html), I removed the `generator_stop` and added the `annotation` future imports.

setup.cfg
```diff
@@ -50,12 +49,11 @@ ignore =
...
     # has errors both for having and not having them; we want to have these until
     # Sopel no longer supports Python versions that require them.
-    FI55,
+    FI58,
     # These would require future imports that are not needed any more on Sopel's
-    # oldest supported Python version (3.6).
-    FI10,FI11,FI12,FI13,FI14,FI16,FI17,
+    # oldest supported Python version (3.7).
+    FI10,FI11,FI12,FI13,FI14,FI15,FI16,FI17,
     # Ignore "annotations" future import, since it's not available before 3.7
```
dgw, line 56:
```suggestion
```

We're not ignoring this one any more. Don't know if trailing comma is allowed in this value, so you might need to remove the comma after `FI17` above.

dgw (commented):
One nit, for your INI-tweaking pleasure.

Exirel:
Yup, fixed it and squashed it. Didn't change anything else.

sopel/bot.py
Exirel: Future PR then?

dgw: Yeah, future is fine.

dgw (approved):
I guess passing CI is our proof that trailing commas in flake8 config don't matter. 😁

PR #2228, opened 2021-12-25
tools: deprecate more Python 2-like shims

dgw:
### Description
These four functions were for backward compatibility, as they all stand in for features removed in Python 3. Sopel no longer supports Python 2, and it's long past time for plugin authors to use the correct Python 3 functions instead of these.

Honestly, no one should be using these any more in almost-2022. Think of the deprecation warning(s) as a wake-up call to modernize your code!

### Checklist

### Notes
Excluded the now-deprecated module members from coverage reporting, in an attempt to keep our coverage percentage focused only on code we intend to keep. Also excluded these members from the docs. Compare [deploy preview](https://deploy-preview-2228--sopel-unstable-docs.netlify.app/api.html#sopel.tools.get_sendable_message) to [current unstable](https://unstable.docs.sopel.chat/api.html#sopel.tools.iteritems).

### DIFF
--- a/docs/source/api.rst
+++ b/docs/source/api.rst
@@ -9,6 +9,7 @@ sopel.tools
 
 .. automodule:: sopel.tools
    :members:
+   :exclude-members: iteritems, iterkeys, itervalues, raw_input
    :private-members: Identifier._lower, Identifier._lower_swapped
 
 sopel.tools.web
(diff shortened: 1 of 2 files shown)


Exirel (approved):
I like it.

PR #2231, opened 2021-12-30
irc: implement CASEMAPPING parameter for ISUPPORT

Exirel:
### Description

This PR contains various changes with the end goal to use the server's casemapping instead of the hardcoded one. Sadly, this includes breaking changes.

#### Non-breaking changes

* The bot has a new method `make_identifier`: it uses the `CASEMAPPING` parameter from isupport to select the appropriate casemapping function (from `sopel.tools.identifiers`)
* The `Identifier` class has several changes:
  * `Identifier` now lives its best life in `sopel.tools.identifiers`, mostly to prevent cyclic import errors. This comes with some more documentation.
  * `Identifier` doesn't override `str.__new__` anymore, I figured out a way to use `__init__` instead, and it works fine.
  * `Identifier` has a new argument: `casemapping`. It's a callable that **must** take one parameter (a string) and return a string.
  * The default value of `casemapping` is `rfc1459_lower`, making this argument optional (at the moment).
  * `self.casemapping` is now used in place of `Identifier._lower` to get a lowercase version of the identifier string
* Classes that create `Identifier` have a new argument: It's a callable that **must** take one parameter (a string) and return an instance of `Identifier` with properly configured casemapping.
  * `User` and `Channel` (both from `sopel.tools.target`)
  * `PreTrigger| (from `sopel.trigger`)
  * `SopelDB` (from `sopel.db`)
  * `SopelIdentifierMemory` (from `sopel.tools`)
* The new argument is used to instantiate `Identifier` when necessary
* `coretasks` and other modules are now using `bot.make_identifier` when they need an `Identifier`
* All method from `SopelDB` now accept a `str` (some used to accept `Identifier` only) since it can always convert them into an `Identifier` when necessary
* Various type hint here and there because why not

#### Breaking changes

* `Identifier._lower` doesn't lower all character as it used to do, now it uses the `rfc1459_lower` function which lowercase only [A-Z] characters. In theory, this is not a breaking change because IRC identifiers shouldn't accept non-ASCII characters anyway. However, in the case of IRC server supporting Unicode identifiers (for nicknames and/or channels), this is a breaking change. A solution to that would be to implement rfc7613 (PRECIS) as suggested by https://modern.ircdocs.horse/index.html#casemapping-parameter
* `core.nick` is now a `str` and not an `Identifier` anymore, meaning that any plugin using that config option will need to transform it into an `Identifier`; this is considered a breaking change, even tho it shouldn't be a problem considering that most Identifier-related operations are through objects that already convert `str` into `Identifier` when necessary

#### Going further

The next steps would be:

* to merge #2232
* to implement a PRECIS profile for casemapping—I think it would even work on current ascii-only identifier, making Sopel UTF-8 ready before servers can be
* to move memory classes into their own submodule
* and probably to move other things outside of `sopel/tools/__init__.py`
* to implement dynamic chantype for identifiers

### Checklist

### DIFF
--- a/docs/source/api.rst
+++ b/docs/source/api.rst
@@ -17,7 +17,6 @@ sopel.tools.identifiers
 
 .. automodule:: sopel.tools.identifiers
    :members:
-   :private-members: Identifier._lower, Identifier._lower_swapped
 
 
 sopel.tools.web
--- a/docs/source/api.rst
+++ b/docs/source/api.rst
@@ -10,7 +10,15 @@ sopel.tools
 .. automodule:: sopel.tools
    :members:
    :exclude-members: iteritems, iterkeys, itervalues, raw_input
+
+
+sopel.tools.identifiers
+-----------------------
+
+.. automodule:: sopel.tools.identifiers
+   :members:
    :private-members: Identifier._lower, Identifier._lower_swapped
+
 
 sopel.tools.web
 ---------------
--- a/docs/source/plugin/bot.rst
+++ b/docs/source/plugin/bot.rst
@@ -23,7 +23,8 @@ second argument::
 
     bot.say('The bot is now talking!', '#private-channel')
 
-Instead of a string, you can use an instance of :class:`sopel.tools.Identifier`.
+Instead of a string, you can use an instance of
+:class:`~sopel.tools.identifiers.Identifier`.
 
 If you want to reply to a user in a private message, you can use the trigger's
 :attr:`~sopel.trigger.Trigger.nick` attribute as destination::
@@ -261,8 +262,8 @@ which provides the following information:
            if not trigger.sender.is_nick():
                # this trigger is from a channel
 
-       See :meth:`Identifier.is_nick() <sopel.tools.Identifier.is_nick>` for
-       more information.
+       See :meth:`Identifier.is_nick() <sopel.tools.identifiers.Identifier.is_nick>`
+       for more information.
 
 Getting users in a channel
 --------------------------
--- a/sopel/bot.py
+++ b/sopel/bot.py
@@ -84,17 +84,18 @@ class Sopel(irc.AbstractBot):
         self.channels = tools.SopelIdentifierMemory()
         """A map of the channels that Sopel is in.
 
-        The keys are :class:`sopel.tools.Identifier`\\s of the channel names,
-        and map to :class:`sopel.tools.target.Channel` objects which contain
-        the users in the channel and their permissions.
+        The keys are :class:`~sopel.tools.identifiers.Identifier`\\s of the
+        channel names, and map to :class:`~sopel.tools.target.Channel` objects
+        which contain the users in the channel and their permissions.
         """
 
         self.users = tools.SopelIdentifierMemory()
         """A map of the users that Sopel is aware of.
 
-        The keys are :class:`sopel.tools.Identifier`\\s of the nicknames, and
-        map to :class:`sopel.tools.target.User` instances. In order for Sopel
-        to be aware of a user, it must share at least one mutual channel.
+        The keys are :class:`~sopel.tools.identifiers.Identifier`\\s of the
+        nicknames, and map to :class:`~sopel.tools.target.User` instances. In
+        order for Sopel to be aware of a user, it must share at least one
+        mutual channel.
         """
 
         self.db = db.SopelDB(config)
@@ -212,9 +213,10 @@ class Sopel(irc.AbstractBot):
             >>> bot.has_channel_privilege('#chan', plugin.VOICE)
             True
 
-        The ``channel`` argument can be either a :class:`str` or a
-        :class:`sopel.tools.Identifier`, as long as Sopel joined said channel.
-        If the channel is unknown, a :exc:`ValueError` will be raised.
+        The ``channel`` argument can be either a :class:`str` or an
+        :class:`~sopel.tools.identifiers.Identifier`, as long as Sopel joined
+        said channel. If the channel is unknown, a :exc:`ValueError` will be
+        raised.
         """
         if channel not in self.channels:
             raise ValueError('Unknown channel %s' % channel)
--- a/sopel/bot.py
+++ b/sopel/bot.py
@@ -22,7 +22,7 @@ from typing import Any, Dict, Iterable, Mapping, Optional, Tuple, Union
 from sopel import db, irc, logger, plugin, plugins, tools
 from sopel.irc import modes
 from sopel.plugins import jobs as plugin_jobs, rules as plugin_rules
-from sopel.tools import deprecated, Identifier, jobs as tools_jobs
+from sopel.tools import deprecated, jobs as tools_jobs
 from sopel.trigger import PreTrigger, Trigger
 
 
@@ -989,7 +989,7 @@ class Sopel(irc.AbstractBot):
             if not bad_nick:
                 continue
             if (re.match(bad_nick + '$', nick, re.IGNORECASE) or
-                    Identifier(bad_nick) == nick):
+                    self.make_identifier(bad_nick) == nick):
                 return True
         return False
 
--- a/sopel/bot.py
+++ b/sopel/bot.py
@@ -102,7 +102,7 @@ class Sopel(irc.AbstractBot):
         mutual channel.
         """
 
-        self.db = db.SopelDB(config)
+        self.db = db.SopelDB(config, identifier_factory=self.make_identifier)
         """The bot's database, as a :class:`sopel.db.SopelDB` instance."""
 
         self.memory = tools.SopelMemory()
--- a/sopel/bot.py
+++ b/sopel/bot.py
@@ -81,7 +81,9 @@ class Sopel(irc.AbstractBot):
         self.modeparser = modes.ModeParser()
         """A mode parser used to parse ``MODE`` messages and modestrings."""
 
-        self.channels = tools.SopelIdentifierMemory()
+        self.channels = tools.SopelIdentifierMemory(
+            identifier_factory=self.make_identifier,
+        )
         """A map of the channels that Sopel is in.
 
         The keys are :class:`~sopel.tools.identifiers.Identifier`\\s of the
@@ -89,7 +91,9 @@ class Sopel(irc.AbstractBot):
         which contain the users in the channel and their permissions.
         """
 
-        self.users = tools.SopelIdentifierMemory()
+        self.users = tools.SopelIdentifierMemory(
+            identifier_factory=self.make_identifier,
+        )
         """A map of the users that Sopel is aware of.
 
         The keys are :class:`~sopel.tools.identifiers.Identifier`\\s of the
--- a/sopel/db.py
+++ b/sopel/db.py
@@ -13,10 +13,12 @@ from sqlalchemy.exc import OperationalError, SQLAlchemyError
 from sqlalchemy.ext.declarative import declarative_base
 from sqlalchemy.orm import scoped_session, sessionmaker
 
-from sopel.tools import deprecated, Identifier
+from sopel.tools import deprecated
+from sopel.tools.identifiers import Identifier
 
 
 LOGGER = logging.getLogger(__name__)
+IdentifierFactory = typing.Callable[[str], Identifier]
 
 
 def _deserialize(value):
@@ -87,6 +89,8 @@ class SopelDB:
 
     :param config: Sopel's configuration settings
     :type config: :class:`sopel.config.Config`
+    :param identifier_factory: factory for
+                               :class:`~sopel.tools.identifiers.Identifier`
 
     This defines a simplified interface for basic, common operations on the
     bot's database. Direct access to the database is also available, to serve
@@ -106,7 +110,13 @@ class SopelDB:
 
     """
 
-    def __init__(self, config):
+    def __init__(
+        self,
+        config,
+        identifier_factory: IdentifierFactory = Identifier,
+    ) -> None:
+        self.make_identifier = identifier_factory
+
         if config.core.db_url is not None:
             self.url = make_url(config.core.db_url)
 
@@ -282,7 +292,7 @@ class SopelDB:
 
         """
         session = self.ssession()
-        slug = Identifier._lower(nick)
+        slug = self.make_identifier(nick).lower()
         try:
             nickname = session.query(Nicknames) \
                 .filter(Nicknames.slug == slug) \
@@ -333,17 +343,17 @@ class SopelDB:
             :meth:`unalias_nick`.
 
         """
-        alias_lower = Identifier._lower(alias)
+        slug = self.make_identifier(alias).lower()
         nick_id = self.get_nick_id(nick, create=True)
         session = self.ssession()
         try:
             result = session.query(Nicknames) \
-                .filter(Nicknames.slug == alias_lower) \
+                .filter(Nicknames.slug == slug) \
                 .filter(Nicknames.canonical == alias) \
                 .one_or_none()
             if result:
                 raise ValueError('Alias already exists.')
-            nickname = Nicknames(nick_id=nick_id, slug=alias_lower, canonical=alias)
+            nickname = Nicknames(nick_id=nick_id, slug=slug, canonical=alias)
             session.add(nickname)
             session.commit()
         except SQLAlchemyError:
@@ -460,7 +470,7 @@ class SopelDB:
             :meth:`delete_nick_value`.
 
         """
-        slug = Identifier._lower(nick)
+        slug = self.make_identifier(nick).lower()
         session = self.ssession()
         try:
             result = session.query(NickValues) \
@@ -494,7 +504,7 @@ class SopelDB:
             To *add* an alias for a nick, use :meth:`alias_nick`.
 
         """
-        alias_slug = Identifier._lower(alias)
+        slug = self.make_identifier(alias).lower()
         nick_id = self.get_nick_id(alias)
         session = self.ssession()
         try:
@@ -503,7 +513,7 @@ class SopelDB:
                 .count()
             if count <= 1:
                 raise ValueError('Given alias is the only entry in its group.')
-            session.query(Nicknames).filter(Nicknames.slug == alias_slug).delete()
+            session.query(Nicknames).filter(Nicknames.slug == slug).delete()
             session.commit()
         except SQLAlchemyError:
             session.rollback()
@@ -604,7 +614,7 @@ class SopelDB:
         databases/files, without regard for variation in case between
         different clients and/or servers on the network.
         """
-        slug = Identifier._lower(chan)
+        slug = self.make_identifier(chan).lower()
         session = self.ssession()
         try:
             count = session.query(ChannelValues) \
@@ -943,7 +953,7 @@ class SopelDB:
             # The goal is to call this method with an existing Identifier
             # and not a simple string. This is a convenient/shortcut method
             # and should probably not be used with a simple string anyway.
-            name = Identifier(name)
+            name = self.make_identifier(name)
 
         if name.is_nick():
             return self.get_nick_value(name, key, default)
--- a/sopel/db.py
+++ b/sopel/db.py
@@ -5,6 +5,7 @@ import json
 import logging
 import os.path
 import traceback
+import typing
 
 from sqlalchemy import Column, create_engine, ForeignKey, Integer, String
 from sqlalchemy.engine.url import make_url, URL
@@ -256,13 +257,12 @@ class SopelDB:
 
     # NICK FUNCTIONS
 
-    def get_nick_id(self, nick, create=False):
+    def get_nick_id(self, nick: str, create: bool = False):
         """Return the internal identifier for a given nick.
 
         :param nick: the nickname for which to fetch an ID
-        :type nick: :class:`~sopel.tools.Identifier`
-        :param bool create: whether to create an ID if one does not exist
-                            (set to ``False`` by default)
+        :param create: whether to create an ID if one does not exist
+                       (set to ``False`` by default)
         :raise ValueError: if no ID exists for the given ``nick`` and ``create``
                            is set to ``False``
         :raise ~sqlalchemy.exc.SQLAlchemyError: if there is a database error
@@ -282,7 +282,7 @@ class SopelDB:
 
         """
         session = self.ssession()
-        slug = nick.lower()
+        slug = Identifier._lower(nick)
         try:
             nickname = session.query(Nicknames) \
                 .filter(Nicknames.slug == slug) \
@@ -317,7 +317,7 @@ class SopelDB:
         finally:
             self.ssession.remove()
 
-    def alias_nick(self, nick, alias):
+    def alias_nick(self, nick: str, alias: str):
         """Create an alias for a nick.
 
         :param str nick: an existing nickname
@@ -333,18 +333,17 @@ class SopelDB:
             :meth:`unalias_nick`.
 
         """
-        nick = Identifier(nick)
-        alias = Identifier(alias)
+        alias_lower = Identifier._lower(alias)
         nick_id = self.get_nick_id(nick, create=True)
         session = self.ssession()
         try:
             result = session.query(Nicknames) \
-                .filter(Nicknames.slug == alias.lower()) \
+                .filter(Nicknames.slug == alias_lower) \
                 .filter(Nicknames.canonical == alias) \
                 .one_or_none()
             if result:
                 raise ValueError('Alias already exists.')
-            nickname = Nicknames(nick_id=nick_id, slug=alias.lower(), canonical=alias)
+            nickname = Nicknames(nick_id=nick_id, slug=alias_lower, canonical=alias)
             session.add(nickname)
             session.commit()
         except SQLAlchemyError:
@@ -353,11 +352,11 @@ class SopelDB:
         finally:
             self.ssession.remove()
 
-    def set_nick_value(self, nick, key, value):
+    def set_nick_value(self, nick: str, key: str, value: typing.Any):
         """Set or update a value in the key-value store for ``nick``.
 
-        :param str nick: the nickname with which to associate the ``value``
-        :param str key: the name by which this ``value`` may be accessed later
+        :param nick: the nickname with which to associate the ``value``
+        :param key: the name by which this ``value`` may be accessed later
         :param mixed value: the value to set for this ``key`` under ``nick``
         :raise ~sqlalchemy.exc.SQLAlchemyError: if there is a database error
 
@@ -374,7 +373,6 @@ class SopelDB:
             :meth:`delete_nick_value`.
 
         """
-        nick = Identifier(nick)
         value = json.dumps(value, ensure_ascii=False)
         nick_id = self.get_nick_id(nick, create=True)
         session = self.ssession()
@@ -398,7 +396,7 @@ class SopelDB:
         finally:
             self.ssession.remove()
 
-    def delete_nick_value(self, nick, key):
+    def delete_nick_value(self, nick: str, key: str):
         """Delete a value from the key-value store for ``nick``.
 
         :param str nick: the nickname whose values to modify
@@ -413,8 +411,6 @@ class SopelDB:
             :meth:`get_nick_value`.
 
         """
-        nick = Identifier(nick)
-
         try:
             nick_id = self.get_nick_id(nick)
         except ValueError:
@@ -437,11 +433,16 @@ class SopelDB:
         finally:
             self.ssession.remove()
 
-    def get_nick_value(self, nick, key, default=None):
+    def get_nick_value(
+        self,
+        nick: str,
+        key: str,
+        default: typing.Optional[typing.Any] = None
+    ):
         """Get a value from the key-value store for ``nick``.
 
-        :param str nick: the nickname whose values to access
-        :param str key: the name by which the desired value was saved
+        :param nick: the nickname whose values to access
+        :param key: the name by which the desired value was saved
         :param mixed default: value to return if ``key`` does not have a value
                               set (optional)
         :raise ~sqlalchemy.exc.SQLAlchemyError: if there is a database error
@@ -459,12 +460,12 @@ class SopelDB:
             :meth:`delete_nick_value`.
 
         """
-        nick = Identifier(nick)
+        slug = Identifier._lower(nick)
         session = self.ssession()
         try:
             result = session.query(NickValues) \
                 .filter(Nicknames.nick_id == NickValues.nick_id) \
-                .filter(Nicknames.slug == nick.lower()) \
+                .filter(Nicknames.slug == slug) \
                 .filter(NickValues.key == key) \
                 .one_or_none()
             if result is not None:
@@ -493,7 +494,7 @@ class SopelDB:
             To *add* an alias for a nick, use :meth:`alias_nick`.
 
         """
-        alias = Identifier(alias)
+        alias_slug = Identifier._lower(alias)
         nick_id = self.get_nick_id(alias)
         session = self.ssession()
         try:
@@ -502,7 +503,7 @@ class SopelDB:
                 .count()
             if count <= 1:
                 raise ValueError('Given alias is the only entry in its group.')
-            session.query(Nicknames).filter(Nicknames.slug == alias.lower()).delete()
+            session.query(Nicknames).filter(Nicknames.slug == alias_slug).delete()
             session.commit()
         except SQLAlchemyError:
             session.rollback()
@@ -523,7 +524,6 @@ class SopelDB:
             you want to do this.
 
         """
-        nick = Identifier(nick)
         nick_id = self.get_nick_id(nick)
         session = self.ssession()
         try:
@@ -564,8 +564,8 @@ class SopelDB:
         Plugins which define their own tables relying on the nick table will
         need to handle their own merging separately.
         """
-        first_id = self.get_nick_id(Identifier(first_nick), create=True)
-        second_id = self.get_nick_id(Identifier(second_nick), create=True)
+        first_id = self.get_nick_id(first_nick, create=True)
+        second_id = self.get_nick_id(second_nick, create=True)
         session = self.ssession()
         try:
             # Get second_id's values
@@ -591,7 +591,7 @@ class SopelDB:
 
     # CHANNEL FUNCTIONS
 
-    def get_channel_slug(self, chan):
+    def get_channel_slug(self, chan: str):
         """Return the case-normalized representation of ``channel``.
 
         :param str channel: the channel name to normalize, with prefix
@@ -604,8 +604,7 @@ class SopelDB:
         databases/files, without regard for variation in case between
         different clients and/or servers on the network.
         """
-        chan = Identifier(chan)
-        slug = chan.lower()
+        slug = Identifier._lower(chan)
         session = self.ssession()
         try:
             count = session.query(ChannelValues) \
@@ -909,7 +908,12 @@ class SopelDB:
 
     # NICK AND CHANNEL FUNCTIONS
 
-    def get_nick_or_channel_value(self, name, key, default=None):
+    def get_nick_or_channel_value(
+        self,
+        name: Identifier,
+        key: str,
+        default=None
+    ):
         """Get a value from the key-value store for ``name``.
 
         :param str name: nick or channel whose values to access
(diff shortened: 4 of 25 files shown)


sopel/db.py
```diff
@@ -393,7 +401,7 @@ def set_nick_value(self, nick, key, value):
         finally:
             self.ssession.remove()
 
-    def delete_nick_value(self, nick, key):
+    def delete_nick_value(self, nick: str, key: str):
```
dgw, line 404: Don't forget to remove the redundant types in the `:param:` declarations below.

sopel/db.py
```diff
@@ -312,7 +322,7 @@ def get_nick_id(self, nick, create=True):
         finally:
             self.ssession.remove()
 
-    def alias_nick(self, nick, alias):
+    def alias_nick(self, nick: str, alias: str):
```
dgw, line 325: Don't forget redundant type declarations in `:param:`s below.

sopel/db.py
```diff
@@ -481,7 +493,7 @@ def unalias_nick(self, alias):
             To *add* an alias for a nick, use :meth:`alias_nick`.
```
dgw, line 493: …which got type hints added to its definition, but `unalias_nick` didn't.

Neither did `forget_nick_group` and several other related functions below. I don't understand why some touched functions have hinting added and some don't.

sopel/db.py
```diff
@@ -896,7 +906,12 @@ def forget_plugin(self, plugin):
 
     # NICK AND CHANNEL FUNCTIONS
 
-    def get_nick_or_channel_value(self, name, key, default=None):
+    def get_nick_or_channel_value(
+        self,
+        name: Identifier,
+        key: str,
+        default=None
+    ):
```
dgw, line 914: Redundant (and at least one mismatched) types in `:param:` declarations below.

sopel/db.py
```diff
@@ -921,7 +936,13 @@ def get_nick_or_channel_value(self, name, key, default=None):
             :meth:`get_channel_value`.
 
         """
-        name = Identifier(name)
+        if not isinstance(name, Identifier):
+            # TODO: log warning/debug?
+            # The goal is to call this method with an existing Identifier
+            # and not a simple string. This is a convenient/shortcut method
+            # and should probably not be used with a simple string anyway.
```
dgw, line 943: Is there a reason it shouldn't be callable with a simple `str`? The standalone methods allow it. For API consistency I would _really_ rather not have this be the only place where passing a `str` instead of `Identifier` throws an error, or even a warning.

sopel/tools/target.py
```diff
@@ -57,29 +67,35 @@ class Channel:
...
-        self.privileges = {}
+        self.privileges: Dict[identifiers.Identifier, int] = {}
         """The permissions of the users in the channel.
 
-        This maps nickname :class:`~sopel.tools.Identifier`\\s to bitwise
-        integer values. This can be compared to appropriate constants from
-        :mod:`sopel.plugin`.
+        This maps nickname :class:`~sopel.tools.identifiers.Identifier`\\s to
+        bitwise integer values. This can be compared to appropriate constants
+        from :mod:`sopel.plugin`.
```
dgw, line 93: As long as you're here?

```suggestion
        from :mod:`sopel.privileges`.
```

(Which now that I think about it, why isn't that `sopel.irc.privileges`? And why isn't the new location for `Identifier` in `sopel.irc.identifiers`? So many questions!)

sopel/tools/target.py
```diff
@@ -193,13 +213,13 @@ def is_oper(self, nick):
             sensibly if only ``+v`` (voice) and ``+o`` (op) modes exist.
 
         """
-        return self.privileges.get(Identifier(nick), 0) & privileges.OPER
+        nick_id = self.make_identifier(nick)
+        return bool(self.privileges.get(nick_id, 0) & privileges.OPER)
```
dgw, line 217: OK, this (and the numerous similar changes to other methods below) made me think: Why not just make `Channel.privileges` a `SopelIdentifierMemory`?

sopel/trigger.py
```diff
@@ -106,30 +110,40 @@ class PreTrigger:
...
     intent_regex = re.compile('\x01(\\S+) ?(.*)\x01')
 
-    def __init__(self, own_nick, line, url_schemes=None):
+    def __init__(
+        self,
+        own_nick: tools.Identifier,
+        line: str,
+        url_schemes: Optional[Sequence] = None,
+        identifier_factory: IdentifierFactory = tools.Identifier,
+    ):
```
dgw, line 119: `tools.Identifier` still? Is this a proof that the move hasn't affected backward compatibility? 😂

sopel/trigger.py
```diff
@@ -106,30 +110,40 @@ class PreTrigger:
...
                 else:
                     self.tags[tag[0]] = None
 
+        # Client time or server time
         self.time = datetime.datetime.utcnow().replace(
             tzinfo=datetime.timezone.utc
         )
         if 'time' in self.tags:
+            # ensure "time" is a string (typecheck)
+            tag_time = self.tags['time'] or ''
```
dgw, line 143: This will catch if `self.tags['time']` is `False` or `0`, maybe, but not if it's some other kind of object that isn't falsy. What am I missing about the utility of this check?

(Plus these changes feel a bit out of place in this patch.)

dgw (commented):
I made it through. 😩 Expected to break and finish tomorrow, but nope.

While I'm (understandably?) annoyed by all the extra constructor parameters that had to happen to make this work, I get that it's impossible—or at least, _very_ code-smell-y—to do it another way.

docs/source/plugin/bot.rst
```diff
@@ -261,8 +262,8 @@ which provides the following information:
            if not trigger.sender.is_nick():
                # this trigger is from a channel
 
-       See :meth:`Identifier.is_nick() <sopel.tools.Identifier.is_nick>` for
-       more information.
+       See :meth:`Identifier.is_nick() <sopel.tools.identifiers.Identifier.is_nick>`
+       for more information.
```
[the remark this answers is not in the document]
Exirel: Yes! That's for another PR, that will follow the exact same idea: link Identifiers to the Bot's ISUPPORT parameters.

sopel/tools/time.py
```diff
@@ -140,9 +140,10 @@ def _check(zone):
 
     if zone:
         tz = _check(zone)
-        if not tz:
+        if not tz and zone:
             tz = _check(
-                db.get_nick_or_channel_value(zone, 'timezone'))
+                db.get_nick_value(zone, 'timezone')
+                or db.get_channel_value(zone, 'timezone'))
```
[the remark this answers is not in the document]
Exirel: Hm. It was back when I didn't know if I could pass an Identifier maker to SopelDB or not, and I thought: hey, `get_nick_or_channel_value` should only accept an Identifier, but it's functionally identical to call one then the other in that case.

However, this is not true anymore, I can revert that change.

test/test_coretasks.py
```diff
@@ -315,6 +315,24 @@ def test_handle_isupport(mockbot):
...
 
+def test_handle_isupport_casemapping(mockbot):
+    # Set bot's nick to something that needs casemapping
+    mockbot.settings.core.nick = 'Test[a]'
+    mockbot._nick = mockbot.make_identifier(mockbot.settings.core.nick)
+
+    assert mockbot.nick.lower() == 'test{a}'
+    assert str(mockbot.nick) == 'Test[a]'
+
+    mockbot.on_message(
```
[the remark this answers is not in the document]
Exirel: Fair enough!

sopel/bot.py
```diff
@@ -81,23 +81,28 @@ def __init__(self, config, daemon=False):
         self.modeparser = modes.ModeParser()
         """A mode parser used to parse ``MODE`` messages and modestrings."""
 
-        self.channels = tools.SopelIdentifierMemory()
+        self.channels = tools.SopelIdentifierMemory(
+            identifier_factory=self.make_identifier,
+        )
```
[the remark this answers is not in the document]
Exirel: Eeyup. I didn't want to make a breaking change here. However, should you choose to accept it, I can make it mandatory.

And yes, the *TBD* description will contain such information, and the doc will come too!

sopel/tools/identifiers.py
```diff
@@ -0,0 +1,171 @@
...
+        self.casemapping: Casemapping = casemapping
+        """Casemapping function to lower the identifier."""
+        self._lowered = self.casemapping(identifier)
+
+    def lower(self):
+        """Get the RFC 2812-compliant lowercase version of this identifier.
+
+        :return: RFC 2812-compliant lowercase version of the
+                 :py:class:`Identifier` instance
+        :rtype: str
```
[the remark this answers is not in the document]
Exirel: Oh yeah, that's part of the "doc overhaul".

sopel/tools/identifiers.py
```diff
@@ -0,0 +1,171 @@
...
+        :param str identifier: the identifier (nickname or channel) to convert
+        :return: RFC 2812-compliant lowercase version of ``identifier``
+        :rtype: str
+        """
+        if isinstance(identifier, Identifier):
+            return identifier.lower()
+        return rfc1459_lower(identifier)
+
+    @staticmethod
+    def _lower_swapped(identifier: str):
```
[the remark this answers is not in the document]
Exirel: Oh, yeah, right. I know why I didn't do it yet: because I want to move the `deprecated` function somewhere else, if you agree: I think `sopel.deprecated` or `sopel.utils` would be good place for that one. I actually think that `from sopel.deprecated import deprecated` works really nicely!

sopel/irc/__init__.py
```diff
@@ -123,6 +124,19 @@ def hostmask(self) -> Optional[str]:
...
+    def make_identifier(self, name: str) -> identifiers.Identifier:
+        """Instantiate an Identifier using the bot's context."""
+        casemapping = identifiers.rfc1459_lower
+
+        if 'CASEMAPPING' in self.isupport:
+            casemapping = {
+                'ascii': identifiers.ascii_lower,
+                'rfc1459': identifiers.rfc1459_lower,
+                'rfc1459-strict': identifiers.rfc1459_strict_lower,
+            }.get(self.isupport['CASEMAPPING'], casemapping)
```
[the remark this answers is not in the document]
Exirel: Oh but then I don't even need the `if` in`, I can just use `self.isupport.get('CASEMAPPING')`.

The reason for that `if in` is to have the proper structure in the future for `UTF8MAPPING`:

```
casemapping = identifiers.rfc1459_lower

if 'CASEMAPPING' in self.isupport:
    ...
elif 'UTF8MAPPING' in self.isupport:
    # change default
    casemapping = identifiers.ascii_lower

if 'UTF8MAPPING' in self.isupport:
    # apply the usual one then Precis, in that order
```

But that's for another PR, we can simplify here for now.

sopel/db.py
Exirel: Yes that's not very thorough: I tend to type hint what I need to understand or modify, and not everything. Given how much I've changed in SopelDB, I'll go through every methods to fix that.

sopel/db.py
Exirel: Yeah you are right, now that SopelDB has an identifier maker, that's doable.

sopel/tools/target.py
Exirel: Long term planning: you can't move things around without causing cyclic import, thus breaking changes. I'm OK with that if you are, but I don't expect you to be (yet).

sopel/tools/target.py
Exirel: Hm, I need to check that.

sopel/trigger.py
Exirel: That's the goal here, yes!

Exirel: But truly I should change to `identifiers.Identifier`, would be better.

sopel/trigger.py
Exirel: The answer is in the comment above: it ensures that "time" is a string, for type-checking purpose, because `self.tags` has the following type:

    Dict[str, Optional[str]]

It means that any value in `self.tags` can be either a `str` (what we want), or `None` (what we don't want): no integer, no boolean, nothing else but string and `None`. As a result, all I'm making sure here is that I do have a string in the end.

See? Type. It all makes sense.

docs/source/api.rst
```diff
@@ -10,8 +10,16 @@ sopel.tools
...
    :members:
    :exclude-members: iteritems, iterkeys, itervalues, raw_input
+
+
+sopel.tools.identifiers
+-----------------------
+
+.. automodule:: sopel.tools.identifiers
+   :members:
    :private-members: Identifier._lower, Identifier._lower_swapped
```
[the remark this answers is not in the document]
Exirel: Oh. Bad rebase conflict.

sopel/tools/target.py
Exirel: Yes, this can be done, at a cost: we need to move `SopelIdentifierMemory` from `sopel.tools` to `sopel.tools.memory`, to prevent circular import.

If you want, I will be happy to do that in that PR, or into a future PR. Whatever you prefer!

Exirel:
I did a rebase and squashed fixup commits. There was some conflicts a bit complicated to sort out properly, I'm glad I made atomic commit so it was easier to test every single one with py.test and flake8 (except for one commit but heh it's squashed now).

I added one commit to update `SopelIdentifierMemory`'s docstring, as requested (I almost forgot about it, so nice catch!).

What's left is:

* to decide if you want me to move memory classes into their own submodule of `sopel.tools` in that PR or not
* to write the PR's description, that must contain a list of all the breaking changes and things to consider for the upgrade

I won't write a migration script in that PR, that's too big of a job, and I'm not even sure how to properly do it at the moment. I know it's doable, because I had to write a database migration system into one of my application a few years ago that was using SQLAlchemy and Alembic, so it's clearly doable.

sopel/tools/target.py
Exirel: And this lead me to #2236 and #2237, the latter using `SopelIdentifierMemory` in `Channel`.

Exirel:
> What's left is:
> 
> * to decide if you want me to move memory classes into their own submodule of `sopel.tools` in that PR or not
> * to write the PR's description, that must contain a list of all the breaking changes and things to consider for the upgrade

* See PR #2237 
* Done

This is ready for a last round of review.

sopel/irc/__init__.py
```diff
@@ -47,11 +48,11 @@ class AbstractBot(abc.ABC):
     """Abstract definition of Sopel's interface."""
     def __init__(self, settings):
         # private properties: access as read-only properties
-        self._nick = tools.Identifier(settings.core.nick)
         self._user = settings.core.user
         self._name = settings.core.name
         self._isupport = ISupport()
         self._myinfo = None
+        self._nick = self.make_identifier(settings.core.nick)
```
dgw, line 55: I don't like moving this to the bottom. The bot's nick is arguably its most important attribute, and it should stay grouped with the other identity-related stuff (user and name).

sopel/irc/__init__.py
```diff
@@ -330,12 +343,21 @@ def on_error(self):
         self.last_error_timestamp = datetime.utcnow()
         self.error_count = self.error_count + 1
 
-    def change_current_nick(self, new_nick):
+    def rebuild_nick(self) -> None:
+        """Rebuild nick as a new identifier.
+
+        This method exists to update the casemapping rules for the
+        :class:`~sopel.tools.identifiers.Identifier` that represents the bot's
+        nick.
```
dgw, line 351:
```suggestion
        nick, e.g. after ISUPPORT info is received.
```

Good idea to mention _why_ one might want to use this? 😸 Can link to some relevant doc section about ISUPPORT.

I also wonder if, perhaps, the method name should be `_rebuild_nick` since it's not really intended for public use?

sopel/modules/adminchannel.py
```diff
@@ -112,7 +112,7 @@ def kick(bot, trigger):
         channel = opt
         reasonidx = 3
     reason = ' '.join(text[reasonidx:])
-    if nick != bot.config.core.nick:
+    if nick != bot.make_identifier(bot.config.core.nick):
```
dgw, line 115: Some future patch making `bot.config.core.nick` already return an `Identifier` would probably be very helpful. Should you agree, I look forward to (what I expect will be) a discussion about how annoying that would be to implement. 😛

test/test_db.py
```diff
@@ -38,41 +38,33 @@ def teardown_function(function):
...
-        assert nick_id == new_id
+    assert nick_id == db.get_nick_id(name)
 
     # Even if the case is different
-    for test in tests:
-        nick_id = test[0]
-        # still no `create`, because the nick ID should already exist now
-        new_id = db.get_nick_id(Identifier(test[2].upper()))
-        assert nick_id == new_id
+    assert nick_id == db.get_nick_id(variant)
```
dgw, line 66: Not a huge fan of losing the comments about using/not using the `create` argument. Granted I'm biased because I wrote them, but still… Alternative would be to comment above (where `create=True`) that it defaults to `False`.

test/test_db.py
```diff
@@ -38,41 +38,33 @@ def teardown_function(function):
...
 
-def test_get_nick_id(db):
+@pytest.mark.parametrize('name, slug, variant', (
+    # Check case insensitive with ASCII only
+    ('Embolalia', 'embolalia', 'eMBOLALIA'),
+    # Ensures case conversion is handled properly
+    ('[][]', '{}{}', '[}{]'),
+    # Unicode, just in case
+    ('EmbölaliÅ', 'embölaliÅ', 'EMBöLALIÅ'),
+))
```
dgw, line 48: Doesn't switching to this remove the verification that different nicks (that aren't equivalent once casemapped) do indeed get different IDs? I think that's why the test used a loop structure before.

sopel/trigger.py
```diff
@@ -163,21 +178,25 @@ def __init__(self, own_nick, line, url_schemes=None):
...
+            # Unless we're messaging the bot directly, in which case that
+            # second arg will be our bot's name.
+            if target.lower() == own_nick.lower():
+                target = self.nick
 
-        # Unless we're messaging the bot directly, in which case that second
-        # arg will be our bot's name.
-        if target and target.lower() == own_nick.lower():
-            target = self.nick
         self.sender = target
```
dgw, line 200: I've been staring at this for 5 minutes (not counting however long I stared at it in the last review) and… Well, it'll be faster just to ask what you're trying to accomplish with the restructuring. 😅

sopel/tools/target.py
```diff
@@ -61,33 +71,46 @@ class Channel:
     """A representation of a channel Sopel is in.
 
     :param name: the channel name
-    :type name: :class:`~.tools.Identifier`
+    :type name: :class:`~sopel.tools.identifiers.Identifier`
+    :param identifier_factory: A factory to create
+                               :class:`~sopel.tools.identifiers.Identifier`
```
dgw, line 76:
```suggestion
                               :class:`~sopel.tools.identifiers.Identifier`\\s
```

❔

Also I don't think this is the only place where such a docstring appears.

sopel/tools/target.py
```diff
@@ -265,13 +292,13 @@ def is_admin(self, nick):
             sensibly if only ``+v`` (voice) and ``+o`` (op) modes exist.
 
         """
-        return self.privileges.get(Identifier(nick), 0) & privileges.ADMIN
+        nick_id = self.make_identifier(nick)
+        return bool(self.privileges.get(nick_id, 0) & privileges.ADMIN)
```
dgw, line 296: Double take alert… I thought this was touching the database (!) for some unknown reason. This might reduce the WTF factor for future developers trekking through this part of the code:

```suggestion
        identifier = self.make_identifier(nick)
        return bool(self.privileges.get(identifier, 0) & privileges.ADMIN)
```

Or one could K.I.S.S. instead (if type-checking doesn't get mad about this):

```suggestion
        nick = self.make_identifier(nick)
        return bool(self.privileges.get(nick, 0) & privileges.ADMIN)
```

_Consider this same note applied to all the methods below using the same local variable name, `nick_id`._

sopel/tools/__init__.py
```diff
@@ -620,24 +501,51 @@ class SopelIdentifierMemory(SopelMemory):
     with both ``Identifier`` and :class:`str` objects, taking advantage of the
     case-insensitive behavior of ``Identifier``.
 
+    As it works with :class:`~.identifiers.Identifier`, it accepts an
+    identifier factory. This factory usually comes from a
+    :class:`bot instance<sopel.bot.Sopel>`, like in the example of a plugin
```
dgw, line 506: Low priority: ``:class:`bot instance<sopel.bot.Sopel>`,`` will render "`bot instance`" in fixed-width font (see [deploy preview](https://deploy-preview-2231--sopel-unstable-docs.netlify.app/api.html#sopel.tools.SopelIdentifierMemory)). This is not _broken_ per se, but it's one recurring annoyance that I have with Sphinx docs. Sometimes you want to link a class or function with custom, normal text. 😤

Something like this could work?

```suggestion
    :ref:`bot instance<py:class:sopel.bot.Sopel>`, like in the example of a plugin
```

sopel/tools/identifiers.py
```diff
@@ -0,0 +1,266 @@
...
+        :return: IRC-compliant lowercase version used for case-insensitive
+                 comparisons
+
+        The behavior of this method depends on the identifier's casemapping
+        function, which should be selected based on the ``CASEMAPPING``
+        parameter from ``ISUPPORT``.
+
+        .. versionchanged:: 8.0
+
+            Now use the :attr:`casemapping` function to lower the identifier.
```
dgw, line 156:
```suggestion
            Now uses the :attr:`casemapping` function to lower the identifier.

```

sopel/tools/identifiers.py
```diff
@@ -0,0 +1,266 @@
...
+        You shouldn't use this unless you need to migrate stored values from
+        the previous, incorrect "lowercase" representation to the correct one.
+
+        :meta public:
+
+        .. versionadded: 7.0
+
+            This method was added to ensure migration of improperly lowercased
+            data: it reverts the data back to the previous lowercase rules.
+        """
```
dgw, line 204:
```suggestion

        """
```

sopel/tools/identifiers.py
```diff
@@ -0,0 +1,266 @@
...
+    """
+    return text.translate(RFC1459_TABLE)
+
+
+def rfc1459_strict_lower(text: str) -> str:
+    """Lower ``text`` according to :rfc:`1459` (strict version).
+
+    As per :rfc:`1459#section-2.2`:
+
+        Because of IRC's scandanavian origin, the characters ``{}|`` are
```
dgw, line 88: I'll be over here, twitching, because I have to leave this error alone in the interest of faithfully quoting the source. 🙈

sopel/tools/identifiers.py
```diff
@@ -0,0 +1,266 @@
...
+
+    .. __: https://modern.ircdocs.horse/index.html#casemapping-parameter
+    """
+    return text.translate(RFC1459_TABLE)
+
+
+def rfc1459_strict_lower(text: str) -> str:
+    """Lower ``text`` according to :rfc:`1459` (strict version).
+
+    As per :rfc:`1459#section-2.2`:
```
dgw, line 86: Anywhere a specific section is linked like this, it would be useful to override the display text, e.g. "RFC 1459 § 2.2", rather than having the `#section-x.y` fragment identifier included in the final docs.

I am, in fact, about to research how annoying it would be to define our own custom role that takes an RFC number _and_ section number, and formats it appropriately without having to muck with manual overrides every time.

dgw (commented):
Hm, I found a surprising number of things left to say, despite the many previous comments and despite only re-reviewing about half the files (mostly the ones that "Changed since your last review"). Nobody would ever accuse me of being a code-review Yes Man, huh?

sopel/irc/__init__.py
Exirel: It needs at least `self._isupport` first, so I figured I could put it there, given that it might, in the end, require more than that.

sopel/irc/__init__.py
Exirel: I was thinking about having that private. However, I also realized that a plugin may want to override coretasks's behavior, or have some way to manage the bot's nick and its rules. This method is idempotent, there are very few risk of anything bad happening (if any), and it doesn't accept any parameter, reducing further the risk of anything wrong done to the bot.

sopel/modules/adminchannel.py
Exirel: This is simply not possible: the configuration is evaluated at startup, before the bot even exists in the first place. This is the string put in the configuration file, and I'd prefer the configuration to stay that way if possible.

test/test_db.py
Exirel: The previous version thought that they were the same, because it used `str.lower()`, which is not ascii, rfc1459, nor rfc1459-strict. 😁 

This is the part that is "breaking change", which should affect server that allows UTF-8 nicknames and channel names, which will be solved by a future PR trying to implement something like UTF8MAPPING and what is suggested in RFC 7564 (and its upgraded version RFC 8264).

sopel/trigger.py
Exirel: The core idea is that it doesn't make sense to have a second `if target` if we can put that directly within the first `if` that ensure it's not `None`:

1. The target is None by default, with a type hint for an optional Identifier.
2. If it's not a `QUIT` event, it becomes an Identifier (before: it would either become an Identifier or be `None`)
3. If we just created the `target`, we can check if it's the bot's nick or not

The diff is not the most readable things to be honest.

sopel/tools/target.py
Exirel: I knew I should have wrote "instances of" and not bother with that `\\s`.

sopel/tools/target.py
Exirel: Type checking will get extra mad at this, yes.

sopel/tools/__init__.py
Exirel: Oh... you are right. I'll try that, otherwise I think I know a way.

sopel/tools/identifiers.py
Exirel: Oh oh oh, good luck with that. I know what it would take, and I'm not super into it. 😁

Exirel:
Alright this is good, the last review is all about a better presentation and attention to details, I really like it. 👍

sopel/irc/__init__.py
dgw: Maybe "needs" is a bit strong at this point in the process? Sure, `self.isupport` will _exist_ but it won't be populated yet. Is there a meaningful, functional difference between this and the old way (with updated import)?

sopel/irc/__init__.py
dgw: True enough, I suppose. My only real argument against making it public is that it slightly clutters the docs for new users trying to grok the basics—but then, that's what the other current project (expanding the prose-style docs about writing plugins) is meant to address.

sopel/modules/adminchannel.py
dgw: Oof, right. Changing the type after bootstrapping (even though `Identifier` is basically `str`) wouldn't be great.

But I still have another question after looking at the context: Why not just use `if nick != bot.nick`?

test/test_db.py
dgw: Seems I didn't quite get my question across, probably due to attaching this comment to lines in the wrong place.

I meant that there doesn't seem to be any equivalent in the new test code below that matches this old check:

```py
    # Check that each nick ended up with a different id
    assert len(set(test[0] for test in tests)) == len(tests)
```

Each parametrized version of the test is going to start with a fresh `db` object, right? So unless I've missed something (always possible 😅), the test has been downgraded to only making sure that `get_nick_id(somenick, create=True)` retrieves an ID, rather than making sure it creates a _new_ ID when expected.

Put another way, nothing about this new test that I can see makes sure that e.g. the ID for the nick `reblochon`, when seen for the first time, is different from the ID for `brie` (also seen for the first time).

sopel/tools/target.py
dgw: So I hear that the type-checker doesn't like changing the type of a variable, fine. First version, then? I'm just looking for an alternative name that isn't `nick_id`. 😺

sopel/tools/identifiers.py
dgw: Just fix the rendering for this PR by overriding the display text, and let me worry about scratching the Sphinx-extension itch if I start feeling it really strongly. 😉

In this case, the below would implement the example I wrote above:

```suggestion
    As per :rfc:`RFC 1459 § 2.2<1459#section-2.2>`:
```

Which renders a link to `https://tools.ietf.org/html/rfc1459.html#section-2.2` like this (on Alabaster):

image

(Link rendering shouldn't differ between Alabaster and Furo, so I didn't bother changing branches.)

sopel/irc/__init__.py
Exirel: In theory yes: if we change the default value for ISupport's CASEMAPPING, we may end up with a different result, which wasn't the case before. In fact, I should even move that *after* `settings` is set, so `make_identifiers()`  could take advantage of the configuration file.

However, I'm not sure to see the problem here: protected and undocumented attributes, all defined in the same bloc of code, very early in `__init__()`. It's not like there are 100 lines before the `self._nick` here.

sopel/irc/__init__.py
Exirel: > that's what the other current project (expanding the prose-style docs about writing plugins) is meant to address.

Exactly yes!

sopel/modules/adminchannel.py
Exirel: That's actually a good question, why was it like that? I don't know. I don't think I should change that in *that* PR, but it is worth thinking about it.

sopel/tools/identifiers.py
Exirel: I find it hilarious to be honest.

sopel/tools/target.py
Exirel: Yup, `identifier` is good!

test/test_db.py
Exirel: Bonus point: I've added some comment about why `create=True`. However, I didn't comment on `create=False`, because I'm not documenting the past here, only what *is* as the moment.

Exirel:
That should be good now, @dgw you can read my replies and see if I need to take action or if I can rebase & squash away.

test/test_db.py
```diff
@@ -39,40 +46,66 @@ def teardown_function(function):
...
+    session = db.session()
+
+    # Attempt to get NickIDs: it is not created by default
+    with pytest.raises(ValueError):
+        db.get_nick_id(nick)
+
+    # Create the NickIDs
+    nick_id = db.get_nick_id(nick, create=True)
+
+    # Check that one and only one Nicknames exist with that ID
```
dgw, line 60: I recall now the danger of asking you to write comments sometimes: the nemesi**S** can strike 😹

```suggestion
    # Attempt to get NickID: it is not created by default
    with pytest.raises(ValueError):
        db.get_nick_id(nick)

    # Create the NickID
    nick_id = db.get_nick_id(nick, create=True)

    # Check that one and only one Nickname exists with that ID
```

test/test_db.py
```diff
@@ -39,40 +46,66 @@ def teardown_function(function):
...
+    ('[][]', '{}{}', '[}{]'),
+    # Unicode, just in case
+    ('EmbölaliÅ', 'embölaliÅ', 'EMBöLALIÅ'),
+))
+def test_get_nick_id_casemapping(db, name, slug, variant):
+    """Test get_nick_id is case-insensitive through an Identifier."""
+    session = db.session()
+    nick = Identifier(name)
+
+    # Create the NickIDs
```
dgw, line 83:
```suggestion
    # Create the NickID
```

test/test_db.py
```diff
@@ -39,40 +46,66 @@ def teardown_function(function):
...
 
     # Even if the case is different
-    for test in tests:
-        nick_id = test[0]
-        # still no `create`, because the nick ID should already exist now
-        new_id = db.get_nick_id(Identifier(test[2].upper()))
-        assert nick_id == new_id
+    assert nick_id == db.get_nick_id(variant)
+
+    # And no other NickIDs was created (even with create=True)
```
dgw, line 99:
```suggestion
    # And no other NickIDs are created (even with create=True)
```

test/test_db.py
```diff
@@ -39,40 +46,66 @@ def teardown_function(function):
...
+    assert nick_id == db.get_nick_id(name, create=True)
+    assert nick_id == db.get_nick_id(variant, create=True)
+    assert 1 == session.query(NickIDs).count()
+
+    # But a very different name means a new NickIDs
```
dgw, line 104:
```suggestion
    # But a truly different name means a new NickID
```

Now I've added "splitting semantic hairs" to the ongoing battle against the nemesi**S** 😀

sopel/tools/identifiers.py
```diff
@@ -0,0 +1,268 @@
...
+Nick and channel are defined by their names, which are "identifiers": their
+names are used to differentiate users from each others, channels from each
+others. To ensure that two channels or two users are the same, their
+identifiers must be processed to be compared properly. This process depends on
+which RFC and how that RFC is implemented by the server: IRC being an old
+protocol, different RFCs have differents version of that process:
+
+* :rfc:`1549#section-2.2`: ASCII characters, and ``[]\\`` are mapped to ``{}|``
+* :rfc:`2812#section-2.2`: same as in the previous RFC, adding ``~`` mapped to
+  ``^``
```
dgw, line 12:
```suggestion
* :rfc:`RFC 1459 § 2.2<1459#section-2.2>`: ASCII characters, and ``[]\\`` are
  mapped to ``{}|``
* :rfc:`RFC 2812 § 2.2<2812#section-2.2>`: same as in the previous RFC, adding
  ``~`` mapped to ``^``
```

Bonus catch: transposed digits. We're not talking about [PPP in HDLC Framing](https://datatracker.ietf.org/doc/html/rfc1549.html). 😄

sopel/tools/identifiers.py
```diff
@@ -0,0 +1,268 @@
...
+    equivalent (``[a-z]``). Non-ASCII characters are kept unmodified.
+    """
+    return text.translate(ASCII_TABLE)
+
+
+def rfc1459_lower(text: str) -> str:
+    """Lower ``text`` according to :rfc:`2812`.
+
+    Similar to :func:`rfc1459_strict_lower`, but also maps ``~`` to ``^``, as
+    per :rfc:`2812#section-2.2`:
```
dgw, line 66:
```suggestion
    per :rfc:`RFC 2812 § 2.2<2812#section-2.2>`:
```

sopel/tools/identifiers.py
```diff
@@ -0,0 +1,268 @@
...
+    :param casemapping: a casemapping function (optional keyword argument)
+    :type casemapping: Callable[[:class:`str`], :class:`str`]
+
+    When used as normal ``str`` objects, case will be preserved.
+    However, when comparing two Identifier objects, or comparing an Identifier
+    object with a ``str`` object, the comparison will be case insensitive.
+
+    This case insensitivity uses the provided ``casemapping`` function,
+    following the rules for the `CASEMAPPING parameter`__ from ISUPPORT. By
+    default, it uses :func:`rfc1459_lower`, following :rfc:`2812#section-2.2`.
```
dgw, line 111:
```suggestion
    default, it uses :func:`rfc1459_lower`, following
    :rfc:`RFC 2812 § 2.2<2812#section-2.2>`.
```

dgw (commented):
Just one conversation from before left unresolved—but I started some new ones. Hopefully they won't make you 🙁 too much, since it's about fixing details & consistency in the new changes.

test/test_db.py
Exirel: The classes are named `NickIDs` and `Nicknames` tho. 😁

dgw: The test isn't checking for instances of the classes, though. The classes represent database tables, but the test is checking for values. Maybe clearer in the `NickIDs` case to just replace it with `nick ID`, and lowercase `nickname`?

Exirel: Yes, as per our IRC discussion!

Exirel:
And once again! But this time, no new docstrings. :grin:

dgw:
I looked, all good. Squash what you're going to squash when ready.

dgw (approved):
I'm going to do what I should have done earlier and actually approve this.

You can squash the fixup into the commit SHA I gave and trust me (and CI) that everything will work out post-merge, or rebase on `master` if you like.

PR #2232, opened 2021-12-31
sopel: move deprecated() to sopel.lifecycle

Exirel:
### Description

The `deprecated()` function is used everywhere, whenever we need to deprecate something. Sometimes, that somethings is right where this function is defined: in `sopel.tools`, causing cyclic import errors.

In an effort to reduce cyclic import errors, and to organize the code a bit better (mostly by removing everything from `sopel/tools/__init__.py`), this function is moved to its own sopel's submodule: `sopel.lifecycle`.

Documentation has been updated accordingly, and every part of Sopel that used `sopel.tools.deprecated` is now using `sopel.lifecycle.deprecated`.

In order to prevent yet another cyclic import error, `deprecated` no longer uses `sopel.tools.get_logger` to get the logger for an external plugins (that isn't using the namespace), but tries to mimic its behavior for namespace plugins only.

**Edit**: After discussion, I moved from `sopel.deprecated` to `sopel.lifecycle`. This module can be used, in the **future** to manage more version & lifecycle related things. But for now, I'd rather see this merged than trying to do too much and be frustrated because I've other things in mind that are blocked by this change.

### Checklist

### DIFF
--- a/docs/source/lifecycle.rst
+++ b/docs/source/lifecycle.rst
@@ -1,0 +1,31 @@
+====================
+Lifecycle Management
+====================
+
+As Sopel grows and evolves, sometimes it needs to say goodbye to obsolete code
+and replace it. The solution can be new (and hopefully better) code, switching
+to an external library, or maybe putting the feature into an external plugin.
+
+In any case, Sopel will always try to mark as deprecated what will be removed,
+indicating when it was deprecated, and when it should be removed. The goal is
+to help Sopel developers and plugin authors understand what must be done to
+adapt their code to a future version of Sopel::
+
+    Deprecated since 8.0, will be removed in 9.0: sopel.module has been
+    replaced by sopel.plugin
+      File "/home/exirel/dev/sopel/sopel/module.py", line 46, in <module>
+        deprecated(
+
+This example above shows a deprecation warning for the ``sopel.module`` module
+that is deprecated: it shows when it was deprecated (Sopel 8.0), when it will
+be removed (Sopel 9.0), and how to fix that warning (use :mod:`sopel.plugin`).
+
+Although this feature is primarily for use by Sopel developers, plugin authors
+can also take advantage of it with the :func:`sopel.lifecycle.deprecated`
+function documented below.
+
+sopel.lifecycle
+===============
+
+.. automodule:: sopel.lifecycle
+   :members:
--- a/sopel/lifecycle.py
+++ b/sopel/lifecycle.py
@@ -1,0 +1,180 @@
+"""Deprecation module for Sopel developers and plugin authors.
+
+.. versionadded:: 8.0
+
+    Previously in :mod:`sopel.tools`, the :func:`deprecated` function has been
+    moved to this newly created module, as it can be used in every part of the
+    Sopel codebase, including :mod:`sopel.tools` itself.
+
+"""
+from __future__ import annotations
+
+import functools
+import inspect
+import logging
+import traceback
+from typing import Callable, Optional
+
+from pkg_resources import parse_version
+
+from sopel import __version__
+
+
+def deprecated(
+    reason: Optional[str] = None,
+    version: Optional[str] = None,
+    removed_in: Optional[str] = None,
+    warning_in: Optional[str] = None,
+    stack_frame: int = -1,
+    func: Optional[Callable] = None,
+):
+    """Decorator to mark deprecated functions in Sopel's API
+
+    :param reason: optional text added to the deprecation warning
+    :param version: optional version number when the decorated function
+                    is deprecated
+    :param removed_in: optional version number when the deprecated function
+                       will be removed
+    :param warning_in: optional version number when the decorated function
+                       should start emitting a warning when called
+    :param stack_frame: optional stack frame to output; defaults to
+                        ``-1``; should almost always be negative
+    :param func: deprecated function
+    :return: a callable that depends on how the decorator is called; either
+             the decorated function, or a decorator with the appropriate
+             parameters
+
+    Any time the decorated ``func`` is called, a deprecation warning will be
+    logged, with the last frame of the traceback. The optional ``warning_in``
+    argument suppresses the warning on Sopel versions older than that, allowing
+    for multi-stage deprecation timelines.
+
+    The decorator can be used with or without arguments::
+
+        from sopel.lifecycle import deprecated
+
+        @deprecated
+        def func1():
+            print('func 1')
+
+        @deprecated()
+        def func2():
+            print('func 2')
+
+        @deprecated(reason='obsolete', version='7.0', removed_in='8.0')
+        def func3():
+            print('func 3')
+
+    which will output the following in a console::
+
+        >>> func1()
+        Deprecated: func1
+        File "<stdin>", line 1, in <module>
+        func 1
+        >>> func2()
+        Deprecated: func2
+        File "<stdin>", line 1, in <module>
+        func 2
+        >>> func3()
+        Deprecated since 7.0, will be removed in 8.0: obsolete
+        File "<stdin>", line 1, in <module>
+        func 3
+
+    The ``stack_frame`` argument can be used to choose which stack frame is
+    logged along with the message text. By default, this decorator logs the
+    most recent stack frame (the last entry in the list, ``-1``), corresponding
+    to where the decorated function itself was called. However, in certain
+    cases such as deprecating conditional behavior within an object
+    constructor, it can be useful to show a less recent stack frame instead.
+
+    .. note::
+
+        This decorator can be also used on callables that are not functions,
+        such as classes and callable objects.
+
+    .. versionadded:: 7.0
+        Parameters ``reason``, ``version``, and ``removed_in``.
+
+    .. versionadded:: 7.1
+        The ``warning_in`` and ``stack_frame`` parameters.
+
+    .. versionchanged:: 8.0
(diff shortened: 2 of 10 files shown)


sopel/deprecated.py
```diff
@@ -0,0 +1,178 @@
...
+            module_name = None
+            if mod:
+                module_name = mod.__name__
+            if module_name:
+                if module_name.startswith('sopel.'):
+                    # core, or core plugin
+                    logger = logging.getLogger(module_name)
+                else:
+                    # probably a plugin; use Sopel's public API for getting the
+                    # logger for a plugin
```
dgw, line 154: Not true any more. This just uses `logging` directly.

```suggestion
                    # probably a plugin; try to handle most cases sanely
```

sopel/deprecated.py
```diff
@@ -0,0 +1,178 @@
...
+
+        There is nothing that prevents this decorator to be used on a class's
+        method, or on any existing callable.
+
+    .. versionadded:: 7.0
+        Parameters ``reason``, ``version``, and ``removed_in``.
+
+    .. versionadded:: 7.1
+        The ``warning_in`` and ``stack_frame`` parameters.
+
```
dgw, line 100: For total completeness, one could add this:

```suggestion

.. versionchanged:: 8.0
    Moved out of :mod:`sopel.tools` to solve circular dependency issues.

```

docs/source/deprecated.rst
```diff
@@ -0,0 +1,32 @@
+==========
+Deprecated
+==========
+
+As Sopel grows and evolves, it needs to say goodbye to obsolete pieces of code
+and to replace them with the appropriate solution: sometimes with new code,
+sometimes with new dependencies, sometimes by putting the feature into an
+external plugins.
```
dgw, line 8:
```suggestion
As Sopel grows and evolves, sometimes it needs to say goodbye to obsolete code
and replace it. The solution can be new (and hopefully better) code, switching
to an external library, or maybe putting the feature into an external plugin.
```

docs/source/deprecated.rst
```diff
@@ -0,0 +1,32 @@
+==========
+Deprecated
+==========
```
dgw, line 3:
```suggestion
============
Deprecations
============
```

\+ rename the file

I'm wondering if the module wouldn't be better named `sopel.deprecation`, actually.

dgw (commented):
I've had another thought since we talked about this, regarding the new module name. What if we called it instead `sopel.lifecycle`?

At the very least, that makes more sense as a title for the doc page, and it means we could have other API lifecycle stuff in there later. Fancy a decorator for alerting users to planned function signature changes, for example? (That's a quick idea, so don't judge it too hard. 😝)

If you don't like this idea, see line notes for how to improve the docs for how things are in the PR at present.

sopel/tools/__init__.py
```diff
@@ -15,20 +15,15 @@
...
 import re
 import sys
 import threading
-import traceback
-
-from pkg_resources import parse_version
-
-from sopel import __version__
 
+from sopel.deprecated import deprecated
```
dgw, line 24: Forgot to add this before submitting the review. 😑

```suggestion
from sopel.deprecated import deprecated  # Don't delete; maintains backward compatibility with pre-8.0 API
```

docs/source/deprecated.rst
dgw: > + rename the file

What about this?

dgw:
I'm still looking for a general reaction to calling this module `sopel.lifecycle` or something else like that.

Exirel:
Hm. Lifecycle could be interesting if we want to add subclass of `Warning` to be used with Python's warning system, subclasses that could be turned into exception via en env var if we wanted to (like when running test to see what must disappear).

Well, I guess it's time for me to think about that a bit more.

Exirel:
Rebased & squashed because conflicts, and then I moved to `sopel.lifecycle`. I don't think I can come up with a better name, and the prospect of managing version and lifecycle related things into that module is very interesting. So let's go!

docs/source/lifecycle.rst
```diff
@@ -0,0 +1,31 @@
...
+      File "/home/exirel/dev/sopel/sopel/module.py", line 46, in <module>
+        deprecated(
+
+This example above shows a deprecation warning for the ``sopel.module`` module
+that is deprecated: it shows when it was deprecated (Sopel 8.0), when it will
+be removed (Sopel 9.0), and how to fix that warning (use :mod:`sopel.plugin`).
+
+If this feature is primarily used by Sopel developers, Plugin authors, too, can
+take advantage of that feature with the :func:`sopel.lifecycle.deprecated`
+function documented below.
```
dgw, line 25:
```suggestion
Although this feature is primarily for use by Sopel developers, plugin authors
can also take advantage of it with the :func:`sopel.lifecycle.deprecated`
function documented below.
```

sopel/lifecycle.py
```diff
@@ -0,0 +1,180 @@
+"""Deprecation module for Sopel developers and Plugin authors.
```
dgw, line 1: I'm not a fan of German noun style, lol.

```suggestion
"""Deprecation module for Sopel developers and plugin authors.
```

sopel/lifecycle.py
```diff
@@ -0,0 +1,180 @@
...
+"""
+from __future__ import annotations
+
+import functools
+import inspect
+import logging
+import traceback
+from typing import Callable, Optional
+
+from pkg_resources import parse_version
```
dgw, line 18: This made me realize we don't actually have an explicit dependency on `pkg_resources` in either requirements file, and we should—_especially_ if it's used at runtime like this.

While it's _probably_ going to be available if people have `setuptools` available to install Sopel in the first place, we shouldn't implicitly rely on anything. Zen of Python, etc.

sopel/lifecycle.py
```diff
@@ -0,0 +1,180 @@
...
+
+    :param reason: optional text added to the deprecation warning
+    :param version: optional version number when the decorated function
+                        is deprecated
+    :param removed_in: optional version number when the deprecated function
+                           will be removed
+    :param warning_in: optional version number when the decorated function
+                           should start emitting a warning when called
+    :param stack_frame: optional stack frame to output; defaults to
+                            ``-1``; should almost always be negative
```
dgw, line 41:
```suggestion
    :param version: optional version number when the decorated function
                    is deprecated
    :param removed_in: optional version number when the deprecated function
                       will be removed
    :param warning_in: optional version number when the decorated function
                       should start emitting a warning when called
    :param stack_frame: optional stack frame to output; defaults to
                        ``-1``; should almost always be negative
```

I didn't look to see if this weird indentation was in the original `tools` definition, but… I'm gonna fix it blind.

sopel/lifecycle.py
```diff
@@ -0,0 +1,180 @@
...
+    :param func: deprecated function
+    :return: a callable that depends on how the decorator is called; either
+             the decorated function, or a decorator with the appropriate
+             parameters
+
+    Any time the decorated ``func`` is called, a deprecation warning will be
+    printed to ``sys.stderr``, with the last frame of the traceback. The
```
dgw, line 48: It uses `logging` now, not `sys.stderr`.

I'm sure this is a holdover from the older implementation(s), but it's one we should fix as long as we're touching this stuff.

sopel/lifecycle.py
```diff
@@ -0,0 +1,180 @@
...
+    printed along with the message text. By default, this decorator prints the
+    most recent stack frame (the last entry in the list, ``-1``),
+    corresponding to where the decorated function itself was called. However,
+    in certain cases such as deprecating conditional behavior within an object
+    constructor, it can be useful to show a less recent stack frame instead.
+
+    .. note::
+
+        There is nothing that prevents this decorator to be used on a class's
+        method, or on any existing callable.
```
dgw, line 93: Kind of confused why this note exists.

sopel/lifecycle.py
```diff
@@ -0,0 +1,180 @@
...
+
+    .. versionadded:: 7.0
+        Parameters ``reason``, ``version``, and ``removed_in``.
+
+    .. versionadded:: 7.1
+        The ``warning_in`` and ``stack_frame`` parameters.
+
+.. versionchanged:: 8.0
+    Moved out of :mod:`sopel.tools` to solve circular dependency issues.
```
dgw, line 102:
```suggestion
    .. versionchanged:: 8.0
        Moved out of :mod:`sopel.tools` to resolve circular dependency issues.
```

dgw (changes requested):
Why does it feel like I didn't even look at the code before? Maybe you changed some of this stuff in the rename from `sopel.deprecated` to `sopel.lifecycle`, idk.

Also note, you'll want to either squash the second "renale" commit or fix that spelling error. 😸

sopel/lifecycle.py
Exirel: Yeah that's weird.

sopel/lifecycle.py
Exirel: Because when people think about decorators, they tend to think about functions only, but they can totally use that on classes and other callable objects.

Exirel:
Yup yup applied & fixed. No rebase nor squash yet, of course.

sopel/lifecycle.py
dgw: Yep, to be replaced by `importlib.metadata` per my personal tickler list.

dgw (approved):
Time to squash. Really wish there was a "squash" emoji; either the food or the sport would do.

PR #2234, opened 2022-01-06
db, test: do not create nick ID by default

dgw:
Having `get_nick_id()` default to `create=True` may cause unexpected behavior if invalid nicknames are passed to it via plugins' use of the getter and setter methods.

We are taking this opportunity to fix the API design itself, with no period of deprecation and warnings, since this particular method is mostly used only by the database object itself, and isn't really intended for direct use by plugins.

I've also added some more assertions and error-case checking to some of the related tests for `sopel.db`. The tests required minimal changes to continue passing; only cases where some manual setup was using `get_nick_id()` as a shortcut to make sure the nick would exist in the database before running the test case needed to be changed.

### Checklist

### Notes
A code search on GitHub for `get_nick_id` returned very few results that weren't either a Sopel plugin or a clone of Sopel itself being used as the base for a custom bot. Of the results that were actually plugin code, only one out of a dozen or so might be problematic (and I will open an issue on that repo to let the author know about this planned change).

As I thought, not many people actually use this function directly. At least half the plugin results were from _my_ plugins or modified clones thereof (which already used `create=False` explicitly).

### DIFF
--- a/sopel/db.py
+++ b/sopel/db.py
@@ -256,18 +256,23 @@ class SopelDB:
 
     # NICK FUNCTIONS
 
-    def get_nick_id(self, nick, create=True):
+    def get_nick_id(self, nick, create=False):
         """Return the internal identifier for a given nick.
 
         :param nick: the nickname for which to fetch an ID
         :type nick: :class:`~sopel.tools.Identifier`
         :param bool create: whether to create an ID if one does not exist
+                            (set to ``False`` by default)
         :raise ValueError: if no ID exists for the given ``nick`` and ``create``
                            is set to ``False``
         :raise ~sqlalchemy.exc.SQLAlchemyError: if there is a database error
 
         The nick ID is shared across all of a user's aliases, assuming their
         nicks have been grouped together.
+
+        .. versionchanged:: 8.0
+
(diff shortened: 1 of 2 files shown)


dgw:
This CI failure is what you were talking about in #2227, eh, @Exirel?

Exirel:
> This CI failure is what you were talking about in #2227, eh, @Exirel?

Yes. I'm trying to fix it, it's almost done.

Exirel:
Oh, this will conflict with #2231 a lot. 😁

Exirel (approved):

dgw:
> Oh, this will conflict with #2231 a lot. 😁

Heh, #2231 already has conflicts from merging #2233, so what's a little more? 😏 (And it looks to me like most of the conflict will come from type-hinting, not the actual changes to support "smart" identifiers, so really you did it to yourself. 😛)

dgw:
Figured if I merge sooner, you can assess + fix the conflicts sooner. 😸

Exirel:
Is that a permission to rebase right now? 😁

dgw:
"As they say in the Temporal Mechanics department, 'There's no time like the present.'"

PR #2236, opened 2022-01-15
irc: implement CHANTYPES parameter for ISUPPORT

Exirel:
### Description

Tin.

This is based on #2231 which made it very easy to do: all commits are squashed into one that must be removed once that PR is merged.

### Checklist

### DIFF
--- a/sopel/coretasks.py
+++ b/sopel/coretasks.py
@@ -348,6 +348,7 @@ def handle_isupport(bot, trigger):
     namesx_support = 'NAMESX' in bot.isupport
     uhnames_support = 'UHNAMES' in bot.isupport
     casemapping_support = 'CASEMAPPING' in bot.isupport
+    chantypes_support = 'CHANTYPES' in bot.isupport
 
     # parse ISUPPORT message from server
     parameters = {}
@@ -368,9 +369,15 @@ def handle_isupport(bot, trigger):
     if 'PREFIX' in bot.isupport:
         bot.modeparser.privileges = set(bot.isupport.PREFIX.keys())
 
-    # was CASEMAPPING support status updated?
-    if not casemapping_support and 'CASEMAPPING' in bot.isupport:
-        # Re-create the bot's nick with the proper identifier+casemapping
+    # rebuild nick when CASEMAPPING and/or CHANTYPES are set
+    if any((
+        # was CASEMAPPING support status updated?
+        not casemapping_support and 'CASEMAPPING' in bot.isupport,
+        # was CHANTYPES support status updated?
+        not chantypes_support and 'CHANTYPES' in bot.isupport,
+    )):
+        # these parameters change how the bot makes Identifiers
+        # since bot.nick is an Identifier, it must be rebuilt
         bot.rebuild_nick()
 
     # was BOT mode support status updated?
--- a/sopel/tools/identifiers.py
+++ b/sopel/tools/identifiers.py
@@ -49,6 +49,7 @@ RFC1459_STRICT_TABLE = str.maketrans(
     string.ascii_uppercase + '[]\\',
     string.ascii_lowercase + '{}|',
 )
+DEFAULT_CHANTYPES = ('#', '&', '+', '!')
 
 
 def ascii_lower(text: str) -> str:
@@ -93,9 +94,6 @@ def rfc1459_strict_lower(text: str) -> str:
     return text.translate(RFC1459_STRICT_TABLE)
 
 
-_channel_prefixes = ('#', '&', '+', '!')
-
-
 class Identifier(str):
     """A ``str`` subclass which acts appropriately for IRC identifiers.
 
@@ -120,7 +118,7 @@ class Identifier(str):
 
     .. versionchanged:: 8.0
 
-        The ``casemapping`` parameter has been added.
+        The ``casemapping`` and ``chantypes`` parameters have been added.
 
     .. __: https://modern.ircdocs.horse/index.html#casemapping-parameter
     """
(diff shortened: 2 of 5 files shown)


sopel/tools/identifiers.py
```diff
@@ -137,10 +136,13 @@ def __init__(
         identifier: str,
         *,
         casemapping: Casemapping = rfc1459_lower,
+        chantypes: tuple = DEFAULT_CHANTYPES,
     ) -> None:
         super().__init__()
         self.casemapping: Casemapping = casemapping
         """Casemapping function to lower the identifier."""
+        self.chantypes = chantypes
+        """Tuple of prefix used for channels."""
```
dgw, line 145:
```suggestion
        """Tuple of prefixes used for channels."""
```

What do you call a nemesi**S** that also contains `e`? 🤔

test/tools/test_tools_identifiers.py
```diff
@@ -166,13 +166,28 @@ def test_identifier_compare_invalid(wrong_type):
...
+    ('&Exirel', identifiers.DEFAULT_CHANTYPES),
+    ('@Exirel', ('#', '@')),
+))
+def test_identifier_is_nick_channel(name, chantypes):
+    assert not identifiers.Identifier(name, chantypes=chantypes).is_nick()
 
 
 def test_identifier_is_nick_empty():
     assert not identifiers.Identifier('').is_nick()
+    assert not identifiers.Identifier('', chantypes=('',)).is_nick()
```
dgw, line 193: You know, this whole block now makes me wonder why the behavior exists. It [goes all the way back](https://github.com/sopel-irc/sopel/commit/f1182ffe6c3f2431a2058613a91a0b91b00747c6#diff-600b52f13ab78aec82ba0dedb4f3434275ba5d68b5de8f708ced1a49828f697f) to the Willie days when `Identifier` was still called `Nick`, some 8 years ago.

Even then, it was questionably useful—but now it's (at best) a dangerous pitfall, one left undocumented. One might reasonably assume that an `Identifier` for which `is_nick()` returns `False` is, therefore, a channel name… But an empty identifier represents neither.

Surely this isn't the PR in which to do it, but maybe one of us should consider making `Identifier('')` a `ValueError`. Is there ever a real-world case when an `Identifier` might be empty? I can't think of one. Channels must have names (and a hypothetical server that allowed a channel with no name would surely enforce at least the prefix character). A nickname of `''` makes no sense, since the server would reply `ERR_NONICKNAMEGIVEN` if a client tried to use it.

dgw (commented):
I have looked at only the 3 post-2231 commits, hoping that the pending rebase will not affect those changes. You might find my comment on the Identifier tests the most thought-provoking, since the rest is just your garden-variety nitpicking/grammar stuff.

test/tools/test_tools_identifiers.py
Exirel: I agree with you. It's time to add a `ValueError` when trying to create an `Identifier` from an empty string.

Exirel: Also, I think `Identifier` should have a `.is_channel()` method.

dgw: Couldn't have said it better myself! Should I put that patch on my to-do list for tonight? 😁

Exirel: Alright, I tried to change that and raise an exception, however there are numerous side effects:

* When creating a PreTrigger, Sopel always set PreTrigger.nick, even if empty
* Setting PreTrigger.nick = None fix that but...
* Rule matching check if pretrigger.nick is the bot's nick, which fail if pretrigger.nick is None (could be fix by being less stupid with `Identifier.__eq__`)
* Fixing that leads to another error down the chain, and so on, and so on...

So, given the scope of this PR, I think it's best to keep that for a future PR. The test is here to check the current behavior, it'll be helpful.

Also, this will be yet another breaking change for Sopel 8, but we are used to that by now, aren't we?

sopel/coretasks.py
```diff
@@ -368,9 +369,15 @@ def handle_isupport(bot, trigger):
...
-        # Re-create the bot's nick with the proper identifier+casemapping
+    # rebuild nick when CASEMAPPING and/or CHANTYPES are set
+    if any((
+        # was CASEMAPPING support status updated?
+        not casemapping_support and 'CASEMAPPING' in bot.isupport,
+        # was CHANTYPES support status updated?
+        not chantypes_support and 'CHANTYPES' in bot.isupport,
+    )):
+        # these parameters change how the bot makes Identifiers and since
+        # bot.nick is an Identifier, it must be rebuild
```
dgw, line 380: I decided to tackle style in addition to the typo that first drew my eye.

```suggestion
        # these parameters change how the bot makes Identifiers
        # since bot.nick is an Identifier, it must be rebuilt
```

test/tools/test_tools_identifiers.py
dgw: I never expected such a change to be added in this PR, haha. Adding a `.is_channel()` method will be a patch of its own (probably from me, assuming I don't forget about it in the next few days), and some kind of overhaul to make `''` an invalid `Identifier` should also be a patch of its own—probably after the simpler method addition.

dgw (changes requested):
So damn close! I'm not one to let a comment typo slide just because the code is ready, though. 😁

sopel/coretasks.py
Exirel: Ha, I wasn't sure about that one!

Exirel:
And squashed.

PR #2240, opened 2022-01-21
adminchannel, coretasks: improve handling of bot's current nick

dgw:
### Description
Tin, kind of.

There's a bit more esoteric detail in the commit messages if anyone's interested, but the short version is that we really should be moving toward a world where `bot.nick` always represents the bot's _current_ nick in use—even if that doesn't match the config.

While we aren't there yet, this is a step (or two) forward on that path. I've made it so that `coretasks`' `NICK` event handler updates the bot's nick if it changes, and only warns for a change _away_ from the config value.

### Checklist

### Notes
On the way to writing this patch, I discovered that `bot.change_current_nick()` would send a `NICK` command to the server along with updating the `bot._nick` value. That's probably a bad idea, actually, and I didn't use it (also because it's redundant to send `NICK thenickwealreadyhave`). We should have another look at that method and rethink it…

### DIFF
--- a/sopel/coretasks.py
+++ b/sopel/coretasks.py
@@ -672,19 +672,32 @@ def track_nicks(bot, trigger):
 
     # Give debug message, and PM the owner, if the bot's own nick changes.
     if old == bot.nick and new != bot.nick:
-        privmsg = (
-            "Hi, I'm your bot, %s. Something has made my nick change. This "
-            "can cause some problems for me, and make me do weird things. "
-            "You'll probably want to restart me, and figure out what made "
-            "that happen so you can stop it happening again. (Usually, it "
-            "means you tried to give me a nick that's protected by NickServ.)"
-        ) % bot.nick
-        debug_msg = (
-            "Nick changed by server. This can cause unexpected behavior. "
-            "Please restart the bot."
-        )
-        LOGGER.critical(debug_msg)
-        bot.say(privmsg, bot.config.core.owner)
+        # Is this the original nick being regained?
+        # e.g. by ZNC's keepnick module running in front of Sopel
+        if old != bot.config.core.nick and new == bot.config.core.nick:
+            LOGGER.info(
+                "Regained configured nick. Restarting is still recommended.")
+        else:
+            privmsg = (
+                "Hi, I'm your bot, %s. Something has made my nick change. This "
+                "can cause some problems for me, and make me do weird things. "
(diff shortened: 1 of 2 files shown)


Exirel:
> We should have another look at that method and rethink it…

Does this mean that the bot should change `bot.nick` only when the *server* tells that the nick has been changed? Does the server always reply back to the client after it sends a NICK command?

If yes: sure, we can change that behavior. If no: then we still need this method.

In any case, we also need a different, separate method that only set the current nick, something like this:

```
def set_current_nick(nick: str) -> None:
    self._nick = self.make_identifier(nick)
```

And keep `change_current_nick()` as a way to send a NICK command.

sopel/coretasks.py
```diff
@@ -672,19 +672,32 @@ def track_nicks(bot, trigger):
...
+            )
+            LOGGER.critical(debug_msg)
+            bot.say(privmsg, bot.config.core.owner)
+
+        # Always update bot.nick anyway so Sopel doesn't lose its self-identity.
+        # This should cut down the number of "weird things" that happen while
+        # the active nick doesn't match the config, but it's not a substitute
+        # for regaining the expected nickname.
+        LOGGER.info("Updating bot.nick property with server-changed nick.")
+        bot._nick = new
```
Exirel, line 700: I think it's time we have a `bot.set_current_nick()` method!

```suggestion
        bot.set_current_nick(new)
```

Exirel (approved):
It works as it is. However, if we are heading toward that direction, I suggest to add a `set_current_nick()` method to the bot (probably on `sopel.irc.AbstractBot`).

dgw:
The way forward is probably to keep `bot.change_current_nick()` as-is, add the new `bot.set_current_nick()` method, and add `NICK` error handling to revert the bot's internal state if the server rejects a `bot.change_current_nick()`. I'd love to cleanly separate `change` and `set`, but there's no specified behavior in RFCs 1459 or 2812 about what the server should send (if anything) when a client changes nicks except for the error numerics. Modern/horse docs just has a "may" (not even in caps) in reference to servers sending a `NICK` command back to the originating user confirming the change. Sigh.

This PR is already way beyond the original scope, though, so I'm going to work on that stuff in a new branch based on this.

Exirel:
> This PR is already way beyond the original scope, though, so I'm going to work on that stuff in a new branch based on this.

Heh, you already got my approval, merge if you feel like it!

PR #2243, opened 2022-01-25
db: adapt SopelDB to SQLAlchemy 2.x new style

Exirel:
### Description

SQLAlchemy 1.4 (used by Sopel) defines a cross-compatible new style to query the database using its ORM, its engine, and its connection:

* Engine cannot execute SQL directly
* Connection can be used as context manager
* Session can be used as context manager
* Instead of doing *.query, use *.execute(stmt)
* SQL statements are easier to create with select, update, and delete function from `sqlalchemy.sql`
* And many other things

One of the key change here is the usage of session object as context manager instead of using try/except block: less code, same feature!

We can't set `future=True` on the Engine yet, because we need the non-compatible and now deprecated behavior of direct SQL execution from the Engine object: `SopelDB.execute` has been deprecated and should be removed in Sopel 8.1, so we can make Sopel compatible with both SQLAlchemy 1.4 and 2.0, as planned by SQLAlchemy migration strategy.

In order to lower @dgw 's blood pressure, there are new tests, existing tests have been updated as well, and the coverage has increased from ~70% to ~85%.

Also, more doc, because I always write more doc.

PS: yes, some tests have been moved around to follow the same order of declaration as the method they are testing.

### Checklist

### DIFF
--- a/sopel/db.py
+++ b/sopel/db.py
@@ -1,3 +1,18 @@
+"""Sopel database module: management and tools around Sopel's datamodel.
+
+This module defines a datamodel, using `SQLAlchemy ORM's mapping`__:
+
+* :class:`NickIDs` and :class:`Nicknames` are used to track users
+* :class:`NickValues` is used to store arbitrary values for users
+* :class:`ChannelValues` is used to store arbitrary values for channels
+* :class:`PluginValues` is used to store arbitrary values for plugins
+
+These models are made available through the :class:`SopelDB` class and its
+convenience methods, such as :meth:`~SopelDB.get_nick_value` or
+:meth:`~SopelDB.get_channel_value`.
+
+.. __: https://docs.sqlalchemy.org/en/14/orm/tutorial.html#declare-a-mapping
+"""
 from __future__ import annotations
 
 import errno
@@ -9,9 +24,9 @@ import typing
 
 from sqlalchemy import Column, create_engine, ForeignKey, Integer, String
 from sqlalchemy.engine.url import make_url, URL
-from sqlalchemy.exc import OperationalError, SQLAlchemyError
-from sqlalchemy.ext.declarative import declarative_base
-from sqlalchemy.orm import scoped_session, sessionmaker
+from sqlalchemy.exc import OperationalError
+from sqlalchemy.orm import declarative_base, scoped_session, sessionmaker
+from sqlalchemy.sql import delete, func, select, update
 
 from sopel.tools import deprecated
 from sopel.tools.identifiers import Identifier
@@ -94,8 +109,8 @@ class SopelDB:
     :type: Callable[[:class:`str`], :class:`str`]
 
     This defines a simplified interface for basic, common operations on the
-    bot's database. Direct access to the database is also available, to serve
-    more complex plugins' needs.
+    bot's database. Direct access to the database is also available through
+    its :attr:`engine` attribute, to serve more complex plugins' needs.
 
     When configured to use SQLite with a relative filename, the file is assumed
     to be in the directory named by the core setting ``homedir``.
@@ -115,6 +130,12 @@ class SopelDB:
         :class:`~sopel.tools.identifiers.Identifier` when dealing with Nick or
         Channel names.
 
+    .. seealso::
+
+        For any advanced usage of the ORM, refer to the
+        `SQLAlchemy documentation`__.
+
+    .. __: https://docs.sqlalchemy.org/en/14/
     """
 
     def __init__(
@@ -185,6 +206,29 @@ class SopelDB:
                            database=db_name, query=query)
 
         self.engine = create_engine(self.url, pool_recycle=3600)
+        """SQLAlchemy Engine used to connect to Sopel's database.
+
+        .. seealso::
+
+            Read `SQLAlchemy engine`__'s documentation to know how to use it.
+
+        .. __: https://docs.sqlalchemy.org/en/14/core/connections.html
+
+        .. important::
+
+            Introduced in Sopel 7, Sopel uses SQLAlchemy 1.4+. This version of
+            SQLAlchemy deprecates various behaviors and methods, to prepare the
+            migration to its future 2.0 version, and the new 2.x style.
+
+            Sopel doesn't enforce the new 2.x style yet. This will be modified
+            in Sopel 9 by using the ``future=True`` flag on the engine.
+
+            You can read more about the `migration guide from 1.x to 2.x`__, as
+            Sopel will ensure in a future version that it is compatible with
+            the new style.
+
+        .. __: https://docs.sqlalchemy.org/en/14/changelog/migration_20.html
+        """
 
         # Catch any errors connecting to database
         try:
@@ -196,7 +240,8 @@ class SopelDB:
         # Create our tables
         BASE.metadata.create_all(self.engine)
 
-        self.ssession = scoped_session(sessionmaker(bind=self.engine))
+        self.ssession = scoped_session(
+            sessionmaker(bind=self.engine, future=True))
 
     def connect(self):
         """Get a direct database connection.
@@ -250,6 +295,7 @@ class SopelDB:
         """
         return self.ssession()
 
+    @deprecated('Use SopelDB.engine directly', version='8.0', removed_in='9.0')
     def execute(self, *args, **kwargs):
         """Execute an arbitrary SQL query against the database.
 
@@ -258,10 +304,37 @@ class SopelDB:
 
         The ``Result`` object returned is a wrapper around a ``Cursor`` object
         as specified by :pep:`249`.
+
+        .. deprecated:: 8.0
+
+            This method will be removed in Sopel 9, following the deprecation
+            of SQLAlchemy's :meth:`sqlalchemy.engine.Engine.execute`.
+
+            To perform a raw SQL query, use the :class:`~SopelDB.engine`
+            attribute as per the migration guide from SQLAlchemy::
+
+                from sqlalchemy.sql import text
+
+                def my_command(bot, trigger):
+                    raw_sql = ' ... '  # your raw SQL
+                    # get a connection as a context manager
+                    with bot.db.engine.connect() as conn:
+                        res = conn.execute(text(raw_sql))
+                        data = res.fetchall()
+
+                    # do something with your data here
+
+        .. seealso::
+
+            Read the `migration guide from 1.x style to 2.x style`__ by
+            SQLAlchemy to learn more about using SQLALchemy's engine and
+            connection.
+
+        .. __: https://docs.sqlalchemy.org/en/14/changelog/migration_20.html
         """
         return self.engine.execute(*args, **kwargs)
 
-    def get_uri(self):
+    def get_uri(self) -> URL:
         """Return a direct URL for the database.
 
         :return: the database connection URI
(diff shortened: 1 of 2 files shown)


sopel/db.py
```diff
@@ -185,6 +206,29 @@ def __init__(
...
+
+            Introduced in Sopel 7, Sopel uses SQLAlchemy 1.4+. This version of
+            SQLAlchemy deprecates various behaviors and methods, to prepare the
+            migration to its future 2.0 version, and the new 2.x style.
+
+            Sopel doesn't enforce the new 2.x style yet. This will be modified
+            in Sopel 8.1 by using the ``future=True`` flag on the engine.
+
+            You can read more about the `migration guide from 1.x to 2.x`__, as
+            Sopel will ensure in future version that it is compatible with
```
dgw, line 227:
```suggestion
            Sopel will ensure in a future version that it is compatible with
```

sopel/db.py
```diff
@@ -185,6 +206,29 @@ def __init__(
...
+        .. __: https://docs.sqlalchemy.org/en/14/core/connections.html
+
+        .. important::
+
+            Introduced in Sopel 7, Sopel uses SQLAlchemy 1.4+. This version of
+            SQLAlchemy deprecates various behaviors and methods, to prepare the
+            migration to its future 2.0 version, and the new 2.x style.
+
+            Sopel doesn't enforce the new 2.x style yet. This will be modified
+            in Sopel 8.1 by using the ``future=True`` flag on the engine.
```
dgw, line 224: What about the use of `future=True` with the `sessionmaker` below? Not a compatibility concern for plugins, I take it?

sopel/db.py
```diff
@@ -258,10 +304,37 @@ def execute(self, *args, **kwargs):
 
         The ``Result`` object returned is a wrapper around a ``Cursor`` object
         as specified by :pep:`249`.
+
+        .. deprecated:: 8.0
+
+            This method will be removed in Sopel 8.1, following the deprecation
+            of SQLAlchemy's :meth:`sqlalchemy.engine.Engine.execute`.
+
+            To perform raw SQL query, use the :class:`~SopelDB.engine`
```
dgw, line 313:
```suggestion
            To perform a raw SQL query, use the :class:`~SopelDB.engine`
```

sopel/db.py
```diff
@@ -258,10 +304,37 @@ def execute(self, *args, **kwargs):
...
+            attribute as per the migration guide from SQLAlchemy::
+
+                from sqlalchemy.sql import text
+
+                def my_command(bot, trigger):
+                    raw_sql = ' ... '  # your raw SQL
+                    # get a connection as a context manager
+                    with bot.db.engine.connect() as conn:
+                        res = conn.execute(text(raw_sql))
+                        data = res.fetchall()
```
dgw, line 323: This works even if SQLAlchemy is being used on top of SQLite, right? I ask because SQLite doesn't really have "connections" like other RDBMS options do.

sopel/db.py
```diff
@@ -332,11 +406,6 @@ def get_nick_id(self, nick: str, create: bool = False) -> int:
                 session.add(nickname)
                 session.commit()
             return nickname.nick_id
-        except SQLAlchemyError:
-            session.rollback()
-            raise
```
dgw, line 337: I understand removing the `finally` block below (its function is now served by the context manager), but what am I missing that replaces this error handler?

This applies to other similar functions below, but I only need the question answered once. 😁

test/test_db.py
```diff
@@ -99,17 +131,56 @@ def test_get_nick_id_casemapping(db, name, slug, variant):
...
 
-    session.close()
+    with db.session() as session:
+        assert 2 == session.scalar(
+            select(func.count()).select_from(NickIDs)
+        )
+
+
+def test_get_nick_id_migration(db: SopelDB):
+    """Test nick with wrong casemapping are properly migrated."""
```
dgw, line 151:
```suggestion
    """Test nicks with wrong casemapping are properly migrated."""
```

test/test_db.py
```diff
@@ -129,105 +200,154 @@ def test_alias_nick(db):
...
+    'string-value',
+    123456789,
+    'unicode-välûé',
+    ['structured', 'value'],
+))
+def test_set_nick_value(db: SopelDB, value):
+    nick = 'TestNick'
+    db.set_nick_value(nick, 'testkey', value)
+    assert db.get_nick_value(nick, 'testkey') == value, (
+        'The value retrieve must be exactly the same that was stored.')
```
dgw, line 213:
```suggestion
        'The value retrieved must be exactly what was stored.')
```

test/test_db.py
```diff
@@ -129,105 +200,154 @@ def test_alias_nick(db):
...
 
+
+def test_unalias_nick_one_or_none(db: SopelDB):
+    # this will create the first version of the nick
+    db.get_nick_id('Exirel', create=True)
+
+    # assert you can't unalias a unique nick
     with pytest.raises(ValueError):
-        db.unalias_nick('Mister_Bradshaw')
+        db.unalias_nick('Exirel')
```
dgw, line 343: Aw, can't we keep the Monty Python reference? 🥺

dgw (commented):
I realized that there are some unreviewed PRs, and also that I had the time to look at one.

sopel/db.py
Exirel: The [context manager does both](https://docs.sqlalchemy.org/en/14/orm/session_transaction.html#managing-transactions) now!

sopel/db.py
Exirel: It should, yes. And also, that's why I can't use the `future=True` on the engine/session, because the `execute` method exists. And that's why I deprecate it in Sopel 8 and I want to remove in 8.1. Because no one should use that!

test/test_db.py
Exirel: Oh, I didn't get it, so I didn't left it here, given all other nicks are from Sopel dev or users. But I could put it back, sure.

sopel/db.py
Exirel: As said in the (now deprecated) `execute` method, I can't use `future=True` because of it. Once it's removed in Sopel 8.1, we should be able to set that flag.

test/test_db.py
dgw: I kinda want to do more indirect references/jokes like your lists of cheeses, rather than people's nicks. So when I write new tests (I think I wrote this one?) I try to insert those instead of using dev/user names.

sopel/db.py
dgw: Set it on the engine, yes. But R244 already sets `future=True` for the `sessionmaker`. That won't cause compatibility problems for plugins, even though setting it on the engine could (because of `execute`)?

test/test_db.py
Exirel: Oh, I see! Then I can change them all in one commit, that would make sense.

sopel/db.py
Exirel: Eeyup, that's what I tested, IIRC: one was fine, the other wasn't, because of `execute`.

sopel/db.py
dgw: The only thing I have to say about this (thanks to the reminder I got when we chatted on IRC today) is that I'd rather only _remove_ things in a .0 version—in this case, wait until 9.0.

No objection to deprecating in 8.0, but deleting stuff in a "minor" release always rubs me the wrong way (even if other packages seem happy to throw things out right and left even in "patch" versions 😑).

Exirel: Alright, done in a future commit.

test/test_db.py
Exirel: Done.

Exirel:
Yeah, this one's good for a last round of review!

test/test_db.py
```diff
@@ -336,20 +336,20 @@ def test_unalias_nick(db: SopelDB):
...
+    db.get_nick_id('MrEricPraline', create=True)
 
     # assert you can't unalias a unique nick
     with pytest.raises(ValueError):
-        db.unalias_nick('Exirel')
+        db.unalias_nick('MrEricPraline')
 
     # and you can't either with a non-existing nick
     with pytest.raises(ValueError):
-        db.unalias_nick('xnaas')
```
xnaas: 💔

Exirel:
Squashed!

dgw:
\*stares in Klingon at @RustyBower\* (not really to rush you, but to give a status update in-band since we've only talked about your ongoing review on IRC)

sopel/db.py
```diff
@@ -595,28 +648,36 @@ def merge_nick_groups(self, first_nick: str, second_nick: str):
...
         first_id = self.get_nick_id(first_nick, create=True)
         second_id = self.get_nick_id(second_nick, create=True)
-        session = self.ssession()
-        try:
+        with self.session() as session:
             # Get second_id's values
-            res = session.query(NickValues).filter(NickValues.nick_id == second_id).all()
+            results = session.execute(
+                select(NickValues).where(NickValues.nick_id == second_id)
+            ).scalars()
```
RustyBower, line 655: Just looking at https://docs.sqlalchemy.org/en/14/changelog/migration_20.html#migration-orm-usage, it looks like they recommend .all() becomes .scalars().all()

sopel/db.py
```diff
@@ -595,28 +648,36 @@ def merge_nick_groups(self, first_nick: str, second_nick: str):
...
-                first_res = session.query(NickValues) \
-                    .filter(NickValues.nick_id == first_id) \
-                    .filter(NickValues.key == row.key) \
-                    .one_or_none()
+            for row in results:
+                first_res = session.execute(
+                    select(NickValues)
+                    .where(NickValues.nick_id == first_id)
+                    .where(NickValues.key == row.key)
+                ).scalar_one_or_none()
```
RustyBower, line 663: This is just a style thing, but would it make more sense to do .scalars().one_or_none() rather than use the compound function?

sopel/db.py
```diff
@@ -595,28 +648,36 @@ def merge_nick_groups(self, first_nick: str, second_nick: str):
...
-            session.query(Nicknames) \
-                .filter(Nicknames.nick_id == second_id) \
-                .update({'nick_id': first_id})
+                    self.set_nick_value(
+                        first_nick, row.key, _deserialize(row.value))
+
+            session.execute(
+                delete(NickValues)
+                .where(NickValues.nick_id == second_id)
+                .execution_options(synchronize_session="fetch")
```
RustyBower, line 672: Were you running into session state issues with the execution options?

RustyBower (changes requested):

sopel/db.py
Exirel: Eeyup. It happens all the time when you delete/update things with SQLAlchemy.

sopel/db.py
Exirel: The method `all()` will return a `list`, which I don't need: I'll directly iterate over `results` the line below.

sopel/db.py
Exirel: The doc says it's identical.

RustyBower: For sure, but just given we use .scalars() earlier in the code, was just wondering if it made more sense to keep that same style. I don't care either way, just wanted to call it out.

Exirel: OK, I realized what you were talking here: `scalars()` (notice the S) is not the same as `scalar`. Everywhere else where it was possible, I used `scalar_one_or_none()`, and when necessary I used `scalars()` . So no inconsistency here.

Exirel:
@dgw all good now!

dgw:
I think you meant to ping @RustyBower for final thumbs-up, lol. This already got a ✔️ from me. 😁

Exirel:
Hm, I'm not sure we are going to get that review from @RustyBower. I'm pretty confident in my change in any case.

RustyBower:
Still LGTM :)

Exirel:
> Still LGTM :)

Well the PR says  "@RustyBower requested changes" so... 😁

RustyBower (approved):

PR #2244, opened 2022-01-26
loader, plugin, plugins.rules: add `allow_bots` decorator

dgw:
### Description
You know, it's fun when you find an unmerged branch hanging around during a local cleanup.

It needs rebasing, which I'll handle in a moment, but it will fill a gap in the plugin decorator library. After #2089 we're missing a way for plugin authors to opt back into getting messages from other bots. Most will probably appreciate the filtering Sopel now does on their behalf, but I'm sure there are a few who [relied on the old behavior](https://xkcd.com/1172/)…

### Checklist
  - ~~I will make sure of this after rebasing.~~ It was true! And I even remembered to upgrade the dev-reqs before running tests.

### DIFF
--- a/docs/source/plugin/anatomy.rst
+++ b/docs/source/plugin/anatomy.rst
@@ -124,6 +124,32 @@ Example::
 A rule with rate-limiting can return :const:`sopel.plugin.NOLIMIT` to let the
 user try again after a failed command, e.g. if a required argument is missing.
 
+Bypassing restrictions
+----------------------
+
+By default, a :term:`Rule` will not trigger on messages from Sopel itself,
+other users that are flagged as bots, or users who are
+:ref:`ignored <Ignore User>` or :ref:`rate-limited <Rate limiting>`. In
+certain cases, it might be desirable to bypass these defaults using one or
+more of these decorators:
+
+* :func:`sopel.plugin.allow_bots`: the rule will accept events from other
+  users who are flagged as bots (like Sopel itself)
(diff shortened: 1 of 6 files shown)


Exirel:
Is there a section of the documentation about non-rule decorators, such as echo and now allow_bot? If so, it should be updated. Otherwise, it means a future PR will be needed to talk about that.

dgw:
Do you mean a section on the `plugin/anatomy` page? Many of the (what I'll call) "conditional" decorators are documented as subsections of how to define rules, but not the (again, coining a term) "filter" decorators like `echo` and this new `allow_bots` feature are mentioned yet. I don't mind adding that here, if you don't. It's relevant.

And of course the decorators themselves are autodoc'd at `plugin/decorators`.

Exirel:
> [...] "filter" decorators like `echo` and this new `allow_bots` feature are mentioned yet. [...] It's relevant.

Great! And yes, I was thinking about that, so that would be perfect thank you. ❤️

dgw:
I took a stab at the documentation update.

My IRC client is currently closed because my ongoing network problems are so distracting. I might stick to GitHub comments for a while, just for my own sanity's sake.

Exirel (approved):
A nitpick, and it's more or less a question/suggestion. The doc looks dope.

dgw:
Happy to get a second look before I squash, @Exirel. I almost committed a nemesi**S** of my own. 😁

Exirel:
Let's. Go. 🚢

PR #2245, opened 2022-02-04
reddit: handle image preview links

half-duplex:
### Description
The reddit plugin currently does not handle preview links like this, which users sometimes mistakenly post ("Copy image link" instead of "Copy link").
https://preview.redd.it/rgcetuc1rmf81.jpg?width=576&auto=webp&s=61de9945ce109cefe95cda589c18d434b2d23553

It falls through to url.py, which fails to find a title.
This changes the reddit.py image link handling to also catch preview links, and show both the comment and full-res links in that case.

pytest is currently exploding on master and I don't know why so uh. CI is doing my QA for me

### Checklist

### DIFF
--- a/sopel/modules/reddit.py
+++ b/sopel/modules/reddit.py
@@ -34,7 +34,7 @@ post_or_comment_url = (
 )
 short_post_url = r'https?://(redd\.it|reddit\.com)/(?P<submission>[\w-]+)/?$'
 user_url = r'%s/u(?:ser)?/([\w-]+)' % domain
-image_url = r'https?://i\.redd\.it/\S+'
+image_url = r'https?://(?P<subdomain>i|preview)\.redd\.it/(?P<image>[^?\s]+)'
 video_url = r'https?://v\.redd\.it/([\w-]+)'
 gallery_url = r'https?://(?:www\.)?reddit\.com/gallery/([\w-]+)'
 
@@ -88,6 +88,9 @@ def get_is_cakeday(entrytime):
 @plugin.output_prefix(PLUGIN_OUTPUT_PREFIX)
 def image_info(bot, trigger, match):
     url = match.group(0)
+    preview = match.group("subdomain") == "preview"
+    if preview:
+        url = "https://i.redd.it/{}".format(match.group("image"))
     results = list(
         bot.memory['reddit_praw']
         .subreddit('all')
@@ -98,7 +101,7 @@ def image_info(bot, trigger, match):
     except IndexError:
         # Fail silently if the image link can't be mapped to a submission
         return plugin.NOLIMIT
-    return say_post_info(bot, trigger, oldest.id, False, True)
+    return say_post_info(bot, trigger, oldest.id, preview, True)
 
 
 @plugin.url(video_url)


sopel/modules/reddit.py
```diff
@@ -87,7 +87,11 @@ def get_is_cakeday(entrytime):
 @plugin.url(image_url)
 @plugin.output_prefix(PLUGIN_OUTPUT_PREFIX)
 def image_info(bot, trigger, match):
-    url = match.group(0)
+    preview = match.group("subdomain") == "preview"
+    if preview:
+        url = "https://i.redd.it/{}".format(match.group("image"))
+    else:
+        url = match.group(0)
```
dgw, line 94: I think @Exirel would agree that it would be better to leave the original `url = match.group(0)` assignment, and only add an `if preview:` block. We try to avoid having `else:` in cases like this, just for style reasons.

(Which is to say, we try to avoid _adding_ an `else:` in cases like this, because preexisting code certainly has plenty of them.)

dgw (commented):
Grudgingly, I have to admit these are worth handling. Reddit is awful for making it so hard to get at the original media, but it's not users' fault. Just got a style comment.

sopel/modules/reddit.py
dgw: Regarding what you said on IRC about avoiding the function call: The performance impact is meaningless in practice (these tests used `timeit`'s default of 1,000,000 iterations), but not zero.

<details>
<summary>Test script</summary>

```python
import timeit

no_else_i = timeit.timeit('''
url = m.group(0)
if preview:
    url = 'https://i.redd.it/{}'.format(m.group('image'))
''',
r'''
import re
r = re.compile(r'https?://(?P<subdomain>i|preview)\.redd\.it/(?P<image>[^?\s]+)')
m = r.match('https://i.redd.it/afheihwoi61.png')
preview = m.group('subdomain') == 'preview'
''')

no_else_preview = timeit.timeit('''
url = m.group(0)
if preview:
    url = 'https://i.redd.it/{}'.format(m.group('image'))
''',
r'''
import re
r = re.compile(r'https?://(?P<subdomain>i|preview)\.redd\.it/(?P<image>[^?\s]+)')
m = r.match('https://preview.redd.it/afheihwoi61.png?some=get&params=here')
preview = m.group('subdomain') == 'preview'
''')

yes_else_i = timeit.timeit('''
if preview:
    url = 'https://i.redd.it/{}'.format(m.group('image'))
else:
    url = m.group(0)
''',
r'''
import re
r = re.compile(r'https?://(?P<subdomain>i|preview)\.redd\.it/(?P<image>[^?\s]+)')
m = r.match('https://i.redd.it/afheihwoi61.png')
preview = m.group('subdomain') == 'preview'
''')

yes_else_preview = timeit.timeit('''
if preview:
    url = 'https://i.redd.it/{}'.format(m.group('image'))
else:
    url = m.group(0)
''',
r'''
import re
r = re.compile(r'https?://(?P<subdomain>i|preview)\.redd\.it/(?P<image>[^?\s]+)')
m = r.match('https://preview.redd.it/afheihwoi61.png?some=get&params=here')
preview = m.group('subdomain') == 'preview'
''')

print('if/else:')
print('  i.       {}'.format(yes_else_i))
print('  preview. {}'.format(yes_else_preview))
print('if only:')
print('  i.       {}'.format(no_else_i))
print('  preview. {}'.format(no_else_preview))
```

</details>

Result:

```
if/else:
  i.       0.04802380000000994
  preview. 0.17630289999988236
if only:
  i.       0.0444231999999829
  preview. 0.23507989999961865
```

half-duplex: Additionally, `preview` is the uncommon case, so eh. I'll go for tidy.

dgw (approved):
Thought this was approved, whoops.

sopel/modules/reddit.py
Exirel: (I agree with @dgw )

PR #2246, opened 2022-02-05
backends: allow setting TLS version and ciphers

half-duplex:
### Description
This resolves a TODO comment and two LGTM ignores regarding failing to specify a minimum SSL/TLS version for connections. It also introduces a new core config option, `tls_ciphers`, since python 3.10 tightens the defaults.

### Checklist

### DIFF
--- a/sopel/config/core_section.py
+++ b/sopel/config/core_section.py
@@ -1,5 +1,7 @@
 from __future__ import annotations
 
+from ssl import TLSVersion
+
 from sopel.config.types import (
     BooleanAttribute,
     ChoiceAttribute,
@@ -19,6 +21,20 @@ COMMAND_DEFAULT_HELP_PREFIX = '.'
 """Default help prefix used in commands' usage messages."""
 URL_DEFAULT_SCHEMES = ['http', 'https', 'ftp']
 """Default URL schemes allowed for URLs."""
+
+
+def _parse_ssl_version(var: str) -> TLSVersion:
+    """Parse an ssl_version config variable.
+
+    :param var: The input string, e.g. "TLSv1_3".
+    :return: The TLSVersion object for that version.
+    """
+    try:
+        return TLSVersion[var]
+    except KeyError:
+        raise ValueError(
+            "'{}' is not a valid TLSVersion. Try e.g. TLSv1_3'".format(var)
+        )
 
 
 def configure(config):
@@ -1184,6 +1200,75 @@ class CoreSection(StaticSection):
 
         So for a timeout of 120s it's a PING every 54s. For a timeout of 250s
         it's a PING every 112.5s.
+
+    """
+
+    ssl_ciphers = ListAttribute(
+        'ssl_ciphers',
+        default=[
+            "ECDHE-ECDSA-AES128-GCM-SHA256",
+            "ECDHE-RSA-AES128-GCM-SHA256",
+            "ECDHE-ECDSA-AES256-GCM-SHA384",
+            "ECDHE-RSA-AES256-GCM-SHA384",
+            "ECDHE-ECDSA-CHACHA20-POLY1305",
+            "ECDHE-RSA-CHACHA20-POLY1305",
+            "DHE-RSA-AES128-GCM-SHA256",
+            "DHE-RSA-AES256-GCM-SHA384",
+        ],
+    )
+    """The cipher suites enabled for SSL/TLS connections.
+
+    :default: ``ECDHE-ECDSA-AES128-GCM-SHA256
+                ECDHE-RSA-AES128-GCM-SHA256
+                ECDHE-ECDSA-AES256-GCM-SHA384
+                ECDHE-RSA-AES256-GCM-SHA384
+                ECDHE-ECDSA-CHACHA20-POLY1305
+                ECDHE-RSA-CHACHA20-POLY1305
+                DHE-RSA-AES128-GCM-SHA256
+                DHE-RSA-AES256-GCM-SHA384``
+
+    The default is the `Mozilla Intermediate configuration`__ as of February
+    2022. This parameter is in OpenSSL format; see the `ciphers manual page`__
+    for your version. This setting cannot be used to configure TLS 1.3.
+
+    .. __: https://wiki.mozilla.org/Security/Server_Side_TLS
+    .. __: https://www.openssl.org/docs/manmaster/man1/ciphers.html
+
+    .. code-block:: ini
+
+        ssl_ciphers =
+            ECDSA+AES256
+            !SHA1
+
+    .. note::
+
+        Cipher selection is also subject to the available SSL/TLS versions;
+        see :attr:`ssl_minimum_version`.
+
+    """
+
+    ssl_minimum_version = ValidatedAttribute(
+        'ssl_minimum_version',
+        default=TLSVersion.TLSv1_2,
+        parse=_parse_ssl_version,
+        serialize=lambda ver: ver.name,
+    )
+    """The minimum version allowed for SSL/TLS connections.
+
+    :default: ``TLSv1_2``
+
+    You should not set this lower than the default unless the server does not
+    support modern versions. Set this to a :class:`~ssl.TLSVersion` value,
+    e.g. ``TLSv1`` or ``TLSv1_3``.
+
+    .. code-block:: ini
+
+        ssl_minimum_version = TLSv1_3
+
+    .. note::
+
+        To use insecure SSL/TLS versions, you may also need to enable
+        insecure cipher suites. See :attr:`ssl_ciphers`.
 
     """
 
--- a/sopel/irc/backends.py
+++ b/sopel/irc/backends.py
@@ -8,7 +8,7 @@ import logging
 import signal
 import ssl
 import threading
-from typing import Dict, Optional, Tuple, TYPE_CHECKING
+from typing import Dict, List, Optional, Tuple, TYPE_CHECKING
 
 from .abstract_backends import AbstractIRCBackend
 
@@ -54,6 +54,8 @@ class AsyncioBackend(AbstractIRCBackend):
                        ``use_ssl`` is not ``True``
     :param ca_certs: optional location to the CA certificates; ignored if
                     ``verify_ssl`` is ``False``
+    :param ssl_ciphers: the OpenSSL cipher suites to use
+    :param ssl_minimum_version: the lowest SSL/TLS version to accept
     """
     def __init__(
         self,
@@ -68,6 +70,8 @@ class AsyncioBackend(AbstractIRCBackend):
         keyfile: Optional[str] = None,
         verify_ssl: bool = True,
         ca_certs: Optional[str] = None,
+        ssl_ciphers: Optional[List[str]] = None,
+        ssl_minimum_version: Optional[ssl.TLSVersion] = None,
         **kwargs,
     ):
         super().__init__(bot)
@@ -80,6 +84,8 @@ class AsyncioBackend(AbstractIRCBackend):
         self._keyfile: Optional[str] = keyfile
         self._verify_ssl: bool = verify_ssl
         self._ca_certs: Optional[str] = ca_certs
+        self._ssl_ciphers: str = ":".join(ssl_ciphers)
+        self._ssl_minimum_version: ssl.TLSVersion = ssl_minimum_version
 
         # timeout configuration
         self._server_timeout: float = float(server_timeout or 120)
@@ -263,6 +269,8 @@ class AsyncioBackend(AbstractIRCBackend):
 
         if self._use_ssl:
             ssl_context = ssl.create_default_context(ssl.Purpose.SERVER_AUTH)
+            ssl_context.minimum_version = self._ssl_minimum_version
+            ssl_context.set_ciphers(self._ssl_ciphers)
             if self._certfile is not None:
                 # load_cert_chain requires a certfile (cannot be None)
                 ssl_context.load_cert_chain(
--- a/sopel/irc/__init__.py
+++ b/sopel/irc/__init__.py
@@ -231,6 +231,8 @@ class AbstractBot(abc.ABC):
             keyfile=self.settings.core.client_cert_file,
             verify_ssl=self.settings.core.verify_ssl,
             ca_certs=self.settings.core.ca_certs,
+            ssl_ciphers=self.settings.core.ssl_ciphers,
+            ssl_minimum_version=self.settings.core.ssl_minimum_version,
         )
 
     def run(self, host: str, port: int = 6667) -> None:


half-duplex:
I don't know what to flag this, since it moves from feature towards bugfix as 3.10 becomes more widely available. I put it on 8.0 since that's when we drop old-python, but sopel 7 will have these default changes applied anyway by py3.10, with no knobs to dial it back.
This draft has **potentially-breaking changes**:
- Sets minimum TLS version to 1.2, even in python versions that don't already do that
- Sets default cipher suites to sane forward-secret ones, even in python versions that don't already do that

CNAME handling needs to be re-added, since it seems that does exist in the real world (#570).

RFC: Should I add a knob to change the minimum SSL/TLS version?

half-duplex:
Thinking more about it, are CNAMEs used for anything where the user can't just update their config? `irc.twitch.tv` -> `irc.chat.twitch.tv`

sopel/irc/backends.py
```diff
@@ -251,67 +250,89 @@ class SSLAsynchatBackend(AsynchatBackend):
...
     def handle_connect(self):
         """Handle potential TLS connection."""
-        # TODO: Refactor to use SSLContext and an appropriate PROTOCOL_* constant
-        # See https://lgtm.com/rules/1507225275976/
-        # These warnings are ignored for now, because we can't easily fix them
-        # while maintaining compatibility with py2.7 AND 3.3+, but in Sopel 8
-        # the supported range should narrow sufficiently to fix these for real.
-        # Each Python version still generally selects the most secure protocol
-        # version(s) it supports.
+
```
dgw, line 275: Style nit: We never add a blank line after a method docstring.

sopel/config/core_section.py
```diff
@@ -1213,6 +1213,43 @@ def homedir(self):
...
+    """The cipher suites enabled for SSL/TLS connections.
+
+    :default: ``ECDHE-ECDSA-AES128-GCM-SHA256:ECDHE-RSA-AES128-GCM-SHA256:ECDHE-ECDSA-AES256-GCM-SHA384:ECDHE-RSA-AES256-GCM-SHA384:ECDHE-ECDSA-CHACHA20-POLY1305:ECDHE-RSA-CHACHA20-POLY1305:DHE-RSA-AES128-GCM-SHA256:DHE-RSA-AES256-GCM-SHA384``
+
+    The default is the `Mozilla Intermediate configuration`_ as of 2022-02.
+    This parameter is in OpenSSL format, see the `ciphers manual page`_ for
+    your version. This setting cannot be used to configure TLS 1.3.
+
+    .. _`Mozilla Intermediate configuration`: https://wiki.mozilla.org/Security/Server_Side_TLS
+    .. _`ciphers manual page`: https://www.openssl.org/docs/manmaster/man1/ciphers.html
```
dgw, line 1240: [Anonymous hyperlinks](https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#anonymous-hyperlinks) are fine, even preferred (since it eliminates the chance of creating reference-name-collision errors later).

Also note preference for a prose-form month/year, & fixing of a comma splice.

```suggestion
    The default is the `Mozilla Intermediate configuration`__ as of February 2022.
    This parameter is in OpenSSL format; see the `ciphers manual page`__ for
    your version. This setting cannot be used to configure TLS 1.3.

    .. __: https://wiki.mozilla.org/Security/Server_Side_TLS
    .. __: https://www.openssl.org/docs/manmaster/man1/ciphers.html
```

Finally: What does it mean that "This setting cannot be used to configure TLS 1.3"?

dgw (commented):
I've reached my limit of digging through SSL/TLS docs for today.

Some knob to override the minimum TLS version would probably be useful, given that IRC networks often seem to fall behind on encryption support vs. the much more common HTTP services we all use every day. No, I would not suggest making this knob support anything ridiculous like SSLv3.

I know you referenced the [shocking number of servers](https://www.ircstats.org/server-features/tls) that still don't have any TLS support at all as of December 2021. Presumably TLSv1.2 is a sane default (it enjoys support by 54% of surveyed servers, vs. 56% for "any SSL protocol"), but we'd probably do well to allow people the option to explicitly enable at least TLSv1.1 if their network of choice doesn't support anything newer. I can take or leave TLSv1.0.

sopel/config/core_section.py
half-duplex: TIL that exists, thanks. I wish it had [1] style footnotes like markdown though, I could see this getting annoying when there's more than a couple links or if any are duplicated.
The TLS 1.3 comment is from the python `ssl` docs: https://docs.python.org/3/library/ssl.html#ssl.SSLContext.set_ciphers and https://docs.python.org/3/library/ssl.html#tls-1-3
This is because TLS 1.3 replaces specific cipher suites where you needed to list every permutation (e.g. "ECDHE-ECDSA-AES128-GCM-SHA256") with a caps-like system ("I support enc=AES128/CHACHA20, mode/mac=GCM/POLY1305, hash=SHA256").

half-duplex:
I don't like storing the min_ver string, but I'm not sure if the validation belongs there, but "invalid version TLSv5" is much more useful than "invalid version None"... :thinking:

sopel/config/core_section.py
dgw: There are _many_ different ways to make links in rST, including a footnote-y style where you give <tt>\`link text <_target-name>\`_</tt> and then define the link target named `_target-name` later. The anonymous style is just much simpler to write and it's more than adequate for our purposes where the intended use is as rendered HTML.

All of that said… I honestly prefer Markdown's link syntaxes. But Sphinx is built for rST, so we're kind of stuck with that in docstrings for now.

Would it be overly complicated to expose a separate option for TLSv1.3 cipher suites, then? I don't deny that Python's defaults are probably better than whatever I'd come up with independently, but I can imagine there could be cases where the defaults Just Don't Work™ for certain servers.

sopel/config/core_section.py
```diff
@@ -1232,24 +1232,49 @@ def homedir(self):
...
-    .. _`Mozilla Intermediate configuration`: https://wiki.mozilla.org/Security/Server_Side_TLS
-    .. _`ciphers manual page`: https://www.openssl.org/docs/manmaster/man1/ciphers.html
+    .. __: https://wiki.mozilla.org/Security/Server_Side_TLS
+    .. __: https://www.openssl.org/docs/manmaster/man1/ciphers.html
 
     .. note::
 
-        Cipher selection is also subject to the available SSL/TLS versions.
-        Sopel currently sets the minimum version to TLS 1.2.
+        Cipher selection is also subject to the available SSL/TLS versions,
```
dgw, line 1244: This is another comma splice like the one I fixed before. Should be a semicolon.

sopel/config/core_section.py
```diff
@@ -1213,6 +1213,68 @@ def homedir(self):
...
+    )
+    """The minimum version allowed for SSL/TLS connections.
+
+    :default: ``TLSv1_2``
+
+    You should not set this lower than the default unless the server does not
+    support modern versions. Set this to a `TLSVersion`__ value, e.g. ``TLSv1``
+    or ``TLSv1_3``.
+
+    .. __: https://docs.python.org/3/library/ssl.html#ssl.TLSVersion
```
dgw, line 1265: Use Intersphinx refs for this kind of thing.

```suggestion
    You should not set this lower than the default unless the server does not
    support modern versions. Set this to a :class:`~ssl.TLSVersion` value, e.g.
    ``TLSv1`` or ``TLSv1_3``.
```

dgw:
Would you agree that putting the ini example code before the `.. note` would look better? I went to the deploy preview to verify that the TLSVersion xref works properly and, well… there's always something more. 🙈

I wish English had a reliable linter, too.

sopel/config/core_section.py
half-duplex: Those settings aren't exposed in Python, so rather inexpedient to implement :smile_cat:
If you needed tight control over the ciphers used, you could set the *max* tls version and tweak 1.2, but barring some "the server selects chacha20 and has a broken implementation" nonsense there's no reason to muck with it anyway.

dgw: Presuming that if Python does expose them in the future, we'll still need to differentiate between 1.3 and 1.2 ciphers because of the different formats… want to rename the option so it's clear that it's only for 1.2?

half-duplex: I figure there's a substantial chance openssl or python ends up shoving 1.3 config into the same cipher spec string, and I'd rather carry `tls_ciphers`+`tls_caps` cruft than `tls12_and_older_ciphers`+"wait no actually it's all of them" cruft

sopel/config/core_section.py
```diff
@@ -1213,6 +1213,66 @@ def homedir(self):
 
     """
 
+    tls_ciphers = ValidatedAttribute(
```
Exirel, line 1216: You can replace that by a `ListAttribute`, and then use it with a join like:

    `ssl_ctx.set_ciphers(':'.join(self.tls_ciphers))

If someone provide only one line with `:`, it will still work, so there is only advantages here.

sopel/config/core_section.py
```diff
@@ -1213,6 +1213,66 @@ def homedir(self):
...
+    The default is the `Mozilla Intermediate configuration`__ as of February
+    2022. This parameter is in OpenSSL format; see the `ciphers manual page`__
+    for your version. This setting cannot be used to configure TLS 1.3.
+
+    .. __: https://wiki.mozilla.org/Security/Server_Side_TLS
+    .. __: https://www.openssl.org/docs/manmaster/man1/ciphers.html
+
+    .. code-block:: ini
+
+        tls_ciphers = ECDSA+AES256:!SHA1
```
Exirel, line 1244: You'll need to change that example if you switch to a `ListAttribute`.

Exirel:
I'm OK with the default minimum TLS version. I think it's fair anyway, at least if we target Sopel 8 for that (as we should, I think).

sopel/config/core_section.py
half-duplex: Good call

half-duplex:
I'm not sure what you mean about `SSLContext.load_verify_locations` - that's only for certs, and we only manually check hostname (and I'm not sure we should be, vs "if you find a cname, update your config").

Exirel:
> I'm not sure we should be, vs "if you find a cname, update your config").

The more I think about it, the less I understand why Sopel does that. Certificates are not supposed to care about CNAME or anything, they only care about the domain name. If someone has a certificate, they must ensure it's valid for all the domain names they want to use, and that includes A records and CNAME, and any other DNS records anyway.

And yes, I was thinking of just letting the SSLContext do its work and never bother with manual check of hostname.

Note: #2256 embed parts of that implementation (without the TLS min version) and you should check it, it's pretty cool IMO.

dgw:
Are you two going to make me decide whether to merge this and then immediately replace it with the `asyncio` version, or would you just work together on the TLS implementation in #2256 and avoid the conflict/resolution cycle altogether? 😛

Exirel:
I didn't implement TLS specifically to let @half-duplex see if what I did with SSLContext was in line with his work and if he would like to add TLS on top on my PR. But I don't think it's a good idea to merge this PR, given that the asyncio one is really changing a lot of things.

dgw:
> But I don't think it's a good idea to merge this PR, given that the asyncio one is really changing a lot of things.

That's why I asked if you guys were planning to work together on it and avoid the conflicts before they start. 😺

half-duplex:
Rebased on (so depends on) #2256

sopel/irc/backends.py
```diff
@@ -80,6 +84,21 @@ def __init__(
...
+        self._ssl_ciphers = ":".join(ssl_ciphers)
+
+        self._ssl_minimum_version = ssl_minimum_version
+        if isinstance(ssl_minimum_version, str):
+            self._ssl_minimum_version = getattr(
+                ssl.TLSVersion,
+                ssl_minimum_version,
+                None,
+            )
+        if self._ssl_minimum_version is None:
```
Exirel, line 96: I'm not sure it's relevant if the bot doesn't use SSL/TLS:

```suggestion
        if self._use_ssl and self._ssl_minimum_version is None:
```

sopel/irc/backends.py
```diff
@@ -80,6 +84,21 @@ def __init__(
         self._keyfile: Optional[str] = keyfile
         self._verify_ssl: bool = verify_ssl
         self._ca_certs: Optional[str] = ca_certs
+        self._ssl_ciphers = ":".join(ssl_ciphers)
+
+        self._ssl_minimum_version = ssl_minimum_version
+        if isinstance(ssl_minimum_version, str):
```
Exirel, line 90: I'd rather invert the if here:

```suggestion
        if not isinstance(ssl_minimum_version, ssl.TLSVersion):
```

sopel/irc/backends.py
```diff
@@ -80,6 +84,21 @@ def __init__(
...
+            self._ssl_minimum_version = getattr(
```
Exirel, line 91: This will confuse mypy/the type hint system, because you redefine what `self._ssl_minimum_version` can be:

* at first it's `Optional[str]`
* then if the getattr succeed, it'll be either "any" or `Optional[TLSVersion]`

Exirel: You also have another option to deal with that:

```python
try:
    self._ssl_minimum_version = ssl.TLSVersion[ssl_minimum_version]
except KeyError:
    LOGGER.error("invalid min version")
    raise ValueError('...')
```

Exirel (changes requested):
I love it! So simple in the end to add these settings.

sopel/irc/backends.py
half-duplex: True!

sopel/irc/backends.py
half-duplex: Oops. Hm. If I'm reading the docs correctly, could I just use `ValidatedAttribute(parse=...), I think? Then the backend should only ever see the TLSVersion

Exirel: You can absolutely do that, yes.

half-duplex: New push seem reasonable? That also fixes the weirdness I mentioned [above](https://github.com/sopel-irc/sopel/pull/2246#issuecomment-1030744713).
If None actually makes it through for ssl_min_ver or ssl_ciphers, `ssl` won't like it, how much do we care about that? Could 'if not None' both of them when they're used...

sopel/config/core_section.py
```diff
@@ -1233,7 +1249,9 @@ def homedir(self):
 
     ssl_minimum_version = ValidatedAttribute(
         'ssl_minimum_version',
-        default="TLSv1_2",
+        default=TLSVersion.TLSv1_2,
+        parse=_parse_ssl_version,
+        serialize=lambda ver: ver.name,
```
Exirel, line 1254: If you want, you can do `operator.attrgetter('name')` instead of the lambda. You'll need to add `import operator`.

half-duplex: Hmm. Neat, but is that cleaner or more understandable than the lambda?

Exirel: Lambda has a weird place in Python. It's both loved and hated. Flake8, for instance, will tell you not to use them when you can use a `def` statement, but otherwise won't complain. Pylint may complain, but its rules are not clear to me yet.

I love `attrgetter` because it's a very functional approach. If you like it, use it. If you don't, leave the lambda. I'm only offering an option here. 😄

dgw:
@Exirel Is this good to go for real? Time to squash?

Exirel (approved):
Yeah all good!

PR #2249, opened 2022-02-08
url: bring additional sanity and calm to log output

dgw:
### Description
The sky is not falling if this plugin can't connect to a found URL, nor is it The End Of The World As We Know It if a title isn't found.

These are debug/info events, not errors/warnings.

### Checklist
- [x] I have read [CONTRIBUTING.md](https://github.com/sopel-irc/sopel/blob/master/CONTRIBUTING.md)
- [x] I can and do license this contribution under the EFLv2
- [x] No issues are reported by `make qa` (runs `make quality` and `make test`)
- [x] I have tested the functionality of the things this change touches

### Notes
This is not marked as a resolution of #2207 because it doesn't technically address the core problem reported there. If sent to IRC as described, the "Traceback" can still become "part of [the] url". I will touch `IrcLoggingHandler` in a separate PR to address that.

Issue #2207, opened 2021-11-06
url: traceback can become part of url


RhinosF1:
### Description
The broken url given in tracebacks to IRC is not always correct

### Reproduction steps
1. Setup sopel
2. Connect to a server, join channel and set exceptions to go to IRC
3. Paste a broken url
4. Read exception

### Expected behavior
url not include traceback on the end

### Logs
```
08:56:42 <RhinosF1> 08:54:54 <MirahezeBot> metawiki * [[Special:Log/block]] reblock  * RhinosF1 *  changed block settings for [[User:45.14.71.0/24]] with an expiration time of 1 year (anonymous users only, account creation disabled, email disabled, cannot edit own talk page): VPN - See [[NOP]] (ProtonVPN per https://whois-
08:56:42 <MirahezeBot> [2021-11-06 08:56:41,711] sopel.modules.url    ERROR    - Unable to reach URL: https://whois-Traceback (most recent call last):  File "/srv/sopelbots/prodvenv/lib/python3.7/site-packages/urllib3/connection.py", line 175, in _new_conn    (self._dns_host, self.port), self.timeout, **extra_kw  File "/srv/sopelbots/prodvenv/lib/python3.7/site-packages/urllib3/util/connection.py", line 73, in create_connection    for res in socket.getaddr
08:56:44 <MirahezeBot> [2021-11-06 08:56:41,714] sopel.modules.url    WARNING  - No title found; ignoring URL: https://whois-
```

### Environment
- Sopel `.version`: 7.1.5
- Sopel installed via: pip
- Python version: 3.7.3
- Operating system: Debian 10
- IRCd `/version`: 09:01:29 solanum-1.0-dev(20211009-430833dc). molybdenum.libera.chat eIKMpS6 TS6ow
- Relevant plugins: url

### Notes
Should we tell users that their URLs are broken rather than an exception anyway?

Exirel:
What was the exact URL expected?

RhinosF1:
[...]

PR #2250, opened 2022-02-08
logger: convert newlines to spaces in IRC logging handler

dgw:
### Description
Tin. Stupid-simple "fix" for #2207.

/cc @RhinosF1

### Checklist

### Notes
I discovered while trying to test this on top of #2249 that setting `logging_channel_level = DEBUG` causes some serious problems. That probably warrants its own issue, but since @Exirel will probably be the one to review this I wanted to see if he has any advice first before I go digging around in things to write up a real report.

### DIFF
--- a/sopel/logger.py
+++ b/sopel/logger.py
@@ -33,7 +33,7 @@ class IrcLoggingHandler(logging.Handler):
             return
 
         try:
-            msg = self.format(record)
+            msg = self.format(record).replace('\n', ' ')
             self._bot.say(msg, self._channel)
         except (KeyboardInterrupt, SystemExit):
             raise


Exirel (commented):
Maybe we should also consider sending this in multiple messages? IIRC `bot.say` doesn't allow multi-line messages.

dgw:
I thought about putting it in multiple messages, but decided not to for two main reasons:

  1. This is a stable-version patch and I don't want to make a significant behavior change
  2. Excessive output already seems to cause problems if the IRC logger is set to `DEBUG` (see OP notes section), so it doesn't currently seem like a good idea to spit out potentially 5-10 lines per log message

Exirel:
> Excessive output already seems to cause problems

Maybe it's time to deprecate such logger...

> if the IRC logger is set to `DEBUG` (see OP notes section)

Also we should probably filter out DEBUG in any case for that logger (that's easy to do with logging config).

Exirel (approved):

dgw:
> > Excessive output already seems to cause problems
> 
> Maybe it's time to deprecate such logger...

The use case is valid, so I wouldn't go that far.

> > if the IRC logger is set to `DEBUG` (see OP notes section)
> 
> Also we should probably filter out DEBUG in any case for that logger (that's easy to do with logging config).

Probably yes, logging to IRC should not allow the more verbose log levels. Too much output in my testing with `logging_channel_level = DEBUG` led to a bunch of messages about flooding via [`handleError`](https://docs.python.org/3.8/library/logging.html#logging.Handler.handleError), though I don't have the exact output available to copy and paste into this comment rn.

PR #2252, opened 2022-02-10
config: deprecate using a comma instead of a newline in ListAttribute

Exirel:
### Description

Tin, according to the step for 8.x of #1626.

### Checklist

### DIFF
--- a/sopel/config/types.py
+++ b/sopel/config/types.py
@@ -27,11 +27,15 @@ from __future__ import annotations
 
 import abc
 import getpass
+import logging
 import os.path
 import re
 
 from sopel.lifecycle import deprecated
 from sopel.tools import get_input
+
+
+LOGGER = logging.getLogger(__name__)
 
 
 class NO_DEFAULT:
@@ -423,8 +427,8 @@ class ListAttribute(BaseValidated):
 
     :param str name: the attribute name to use in the config file
     :param strip: whether to strip whitespace from around each value
-                  (optional; applies only to legacy comma-separated lists;
-                  multi-line lists are always stripped)
+                  (optional, deprecated; applies only to legacy comma-separated
+                  lists; multi-line lists are always stripped)
     :type strip: bool
     :param default: the default value if the config file does not define a
                     value for this option; to require explicit configuration,
@@ -468,18 +472,20 @@ class ListAttribute(BaseValidated):
 
         **About:** backward compatibility with comma-separated values.
 
-        A :class:`ListAttribute` option allows to write, on a single line,
-        the values separated by commas. As of Sopel 7.x this behavior is
-        discouraged. It will be deprecated in Sopel 8.x, then removed in
-        Sopel 9.x.
+        A :class:`ListAttribute` option used to allow to write, on a single
+        line, the values separated by commas. It is still technically possible
+        while raising a deprecation warning.
 
-        Bot owners are encouraged to update their configurations to use
-        newlines instead of commas.
+        In Sopel 7.x this behavior was discouraged; as of Sopel 8.x it is now
+        deprecated with warnings, and it will be removed in Sopel 9.x.
+
+        Bot owners should update their configurations to use newlines instead
+        of commas.
 
         The comma delimiter fallback does not support commas within items in
         the list.
     """
-    DELIMITER = ','
+    DELIMITER = ','  # Deprecated, will be removed in Sopel 9
     QUOTE_REGEX = re.compile(r'^"(?P<value>#.*)"$')
     """Regex pattern to match value that requires quotation marks.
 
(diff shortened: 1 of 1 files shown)


sopel/config/types.py
```diff
@@ -423,7 +427,7 @@ class ListAttribute(BaseValidated):
     :param str name: the attribute name to use in the config file
     :param strip: whether to strip whitespace from around each value
                   (optional; applies only to legacy comma-separated lists;
-                  multi-line lists are always stripped)
+                  multi-line lists are always stripped; deprecated)
```
dgw, line 430: I would suggest putting "deprecated" earlier in the string, close to "optional", as both are very important statuses.

Maybe "(optional, deprecated; applies […]"?

----

Also, since I can't line-note `__init__` because it's outside the diff… Have you thought ahead to what it will be like in 9.0 when we actually remove the behavior? Got any more sophisticated plan for removing `strip` than just letting the constructor `TypeError` if a plugin tries to use this unexpected kwarg after removal?

sopel/config/types.py
```diff
@@ -506,6 +510,11 @@ def parse(self, value):
 
             When modified and saved to a file, items will be stored as a
             multi-line string (see :meth:`serialize`).
+
+        .. versionchanged:: 8.0
+
+            When the value contains a delimiter without newline, it warns user
+            to switch to multi-line value, without a delimiter.
         """
```
dgw, line 518: I'm gonna nitpick.

```suggestion

        """
```

sopel/config/types.py
```diff
@@ -515,10 +524,17 @@ def parse(self, value):
                 for item in value.splitlines())
         else:
             # this behavior will be:
-            # - Discouraged in Sopel 7.x (in the documentation)
```
dgw, line 518: Removing this seems unnecessary. Why not keep the history around as a reminder to those of us actually maintaining the code? It won't appear in docs.

sopel/config/types.py
```diff
@@ -506,6 +510,11 @@ def parse(self, value):
 
             When modified and saved to a file, items will be stored as a
             multi-line string (see :meth:`serialize`).
+
+        .. versionchanged:: 8.0
+
+            When the value contains a delimiter without newline, it warns user
+            to switch to multi-line value, without a delimiter.
```
dgw, line 517:
```suggestion
            When the value contains a delimiter without newline, it warns the
            user to switch to a multi-line value, without a delimiter.
```

sopel/config/types.py
```diff
@@ -468,9 +472,9 @@ class SpamSection(StaticSection):
         **About:** backward compatibility with comma-separated values.
 
         A :class:`ListAttribute` option allows to write, on a single line,
```
dgw, line 474: "initially allowed writing, on a single line,"?

Sorry I can't make this one into a suggestion unified with the changes below & nicely rewrapped for you; GitHub has blinders on and can only see those deleted lines in the middle, so it won't let me. 😑

dgw (commented):
About those tweaked docs: It turns out that I'm a nitpicker. Who knew?

sopel/config/types.py
Exirel: I think we can raise a warning if `strip is not None` in Sopel 9, and remove it in Sopel 10.

sopel/config/types.py
Exirel: Yeah sure, at some point the whole block will be removed anyway.

sopel/config/types.py
Exirel: Hm. You can still write this on a single line, even in Sopel 8.

dgw: Technically yes, but it's going to generate a warning if you do now. Maybe this isn't the exact phrasing to use, but I would really like to avoid use of present tense when talking about _deprecated_ support for comma-as-delimiter.

sopel/config/types.py
```diff
@@ -467,18 +471,20 @@ class SpamSection(StaticSection):
...
-        A :class:`ListAttribute` option allows to write, on a single line,
-        the values separated by commas. As of Sopel 7.x this behavior is
-        discouraged. It will be deprecated in Sopel 8.x, then removed in
-        Sopel 9.x.
+        A :class:`ListAttribute` option used to allow to write, on a single
+        line, the values separated by commas. It is still technically possible
+        while raising a deprecation warning.
+
+        In Sopel 7.x this behavior was discouraged, as of Sopel 8.x it is now
+        deprecated and raises warnings, and it will be removed in Sopel 9.x.
```
dgw, line 479:
```suggestion
        In Sopel 7.x this behavior was discouraged; as of Sopel 8.x it is now
        deprecated with warnings, and it will be removed in Sopel 9.x.
```

Comma splices 💢

dgw (commented):
Just one last nitpick. Feel free to squash-and-apply in one step.

dgw (approved):
As usual, any typos/etc. that eluded me this long _deserve_ to make it into `master`.

Exirel:
> squash-and-apply

It has been done.

PR #2261, opened 2022-02-26
core: replace `pkg_resources` with `importlib.metadata`

dgw:
### Description
Tin. Performance of `pkg_resources` can be quite slow (just search all of GitHub issues for "pkg_resources slow"); [its use is discouraged](https://setuptools.pypa.io/en/latest/pkg_resources.html) by the `setuptools` project itself, in favor of the newer modules such as `importlib.metadata`.

Needs a backport package until we no longer support Python 3.9. stdlib gained `importlib.metadata` in Python 3.8, but the ability to "select" in `entry_points()` won't be added until Python 3.10. I don't mind this, because it takes an implied dependency (`setuptools` was required to install Sopel but not listed in `requirements.txt`) and makes it explicit until we can use the stdlib version. My requirement specifier gives the minimum version of the `importlib_metadata` backport that supports selectors.

I didn't see a direct equivalent to `pkg_resources.parse_version()` in `importlib.metadata`; looking at a couple other projects that have already made the move away from `pkg_resources` told me that `packaging.version.parse()` is the way to go.

### Checklist
  - I will eagerly await @Exirel telling me what I did wrong when updating the tests. 😁
  - In addition to the test suite, I also installed and used `sopel-dns` on my local (`-e .`) SopelTest instance with no issues.

### Notes
We devs/hackers have been installing `importlib_metadata` for a while without thinking about it, because [Sphinx explicitly requires it](https://github.com/sphinx-doc/sphinx/blob/b8789b4cb6ca034ed7209b9adc91095b0b9b763c/setup.py#L32) for install on py<3.10.

### DIFF
--- a/sopel/plugins/__init__.py
+++ b/sopel/plugins/__init__.py
@@ -12,7 +12,7 @@ and only one plugin per unique name, using a specific order:
 
 * extra directories defined in the settings
 * homedir's ``plugins`` directory
-* ``sopel.plugins`` setuptools entry points
+* ``sopel.plugins`` entry point group
 * ``sopel_modules``'s subpackages
 * ``sopel.modules``'s core plugins
 
@@ -98,15 +98,15 @@ def find_sopel_modules_plugins():
 
 
 def find_entry_point_plugins(group='sopel.plugins'):
-    """List plugins from a setuptools entry point group.
+    """List plugins from an entry point group.
 
-    :param str group: setuptools entry point group to look for
-                      (defaults to ``sopel.plugins``)
+    :param str group: entry point group to search in (defaults to
+                      ``sopel.plugins``)
     :return: yield instances of :class:`~.handlers.EntryPointPlugin`
-             created from setuptools entry point given ``group``
+             created from each entry point in the ``group``
 
-    This function finds plugins declared under a setuptools entry point; by
-    default it uses the ``sopel.plugins`` entry point.
+    This function finds plugins declared under an entry point group; by
+    default it looks in the ``sopel.plugins`` group.
     """
     for entry_point in importlib_metadata.entry_points(group=group):
         yield handlers.EntryPointPlugin(entry_point)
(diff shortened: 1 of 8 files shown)


sopel/plugins/handlers.py
```diff
@@ -537,11 +537,11 @@ class EntryPointPlugin(PyModulePlugin):
 
     And this plugin can be loaded with::
 
-        >>> from pkg_resources import iter_entry_points
+        >>> from importlib_metadata import entry_points
```
Exirel, line 540: Is that still a "`setuptools` entry point object" in that case? If not, then the docstring above need some adaptation, I think.

sopel/plugins/__init__.py
```diff
@@ -33,7 +33,12 @@
 import itertools
 import os
 
-import pkg_resources
+try:
+    import importlib_metadata
+except ImportError:
+    # TODO: use stdlib only when possible, after dropping py3.9
+    # stdlib does not support `entry_points(group='filter')` until py3.10
```
Exirel, line 40: 🤦‍♂️ 🤦‍♂️ 🤦‍♂️

sopel/plugins/handlers.py
dgw: There are a lot of `setuptools` mentions around the place, honestly. I'm personally not yet clear whether an "entry point" is something that `setuptools` invented, or if it's a Python standard that should be referred to generically. The packaging documentation [seems to hint](https://packaging.python.org/en/latest/specifications/entry-points/) that it's a standard now and we should stop using `setuptools` as a descriptor. What say you?

Also, note to self about docs: there's at least one place that refers to `'sopel.plugins'` as the name of an entry point, rather than a group. Fix that.

dgw:
@Exirel I went a bit ham on updating the docs instead of just tweaking the one `setuptools` reference you pointed out. 😅

sopel/plugins/handlers.py
Exirel: Oh you're right, it's just "entry point" now. Good!

Exirel (approved):
I'm clearly not going to nitpick at grammar, this looks pretty good to me.

dgw:
Followed up with #2268, since the `importlib_metadata` reality isn't quite as elegant as the theory this PR was based on.

PR #2265, opened 2022-03-07
seen: modify/fix stored value & delta calculation for Aware `trigger.time`

dgw:
### Description
Fixes a bug on `master` reported on IRC by monaco (@hedho), related to `trigger.time` changes from #2099.

I chose to eliminate the bug by using Unix timestamps instead of `datetime` objects. `<datetime_obj>.timestamp()` [has been around since Python 3.3](https://docs.python.org/3.9/library/datetime.html#datetime.datetime.timestamp), so we're safe to use it with 8.x's minimum requirement of Python 3.7.

### Checklist

### DIFF
--- a/sopel/modules/seen.py
+++ b/sopel/modules/seen.py
@@ -9,7 +9,6 @@ https://sopel.chat
 """
 from __future__ import annotations
 
-import datetime
 import time
 
 from sopel import plugin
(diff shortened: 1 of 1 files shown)


Exirel (changes requested):
I think it would be best to add this change (line 68):

```diff
- bot.db.set_nick_value(nick, 'seen_timestamp', time.time())
+ bot.db.set_nick_value(nick, 'seen_timestamp', trigger.time.timestamp())
```

Exirel:
Yeah, but it's a good opportunity to be more coherent with the feature. `trigger.time` is timezone sensitive, and I feel like it makes more sense to use it here as well. Beside, if the event comes from the IRC history (if that ever get implemented), we won't have to update the `seen` plugin (at least, not that part).

dgw:
> `trigger.time` is timezone sensitive

Well, this is annoying, but not because of what you said. It's annoying that `time.time()` isn't always in UTC.

Give me a moment to update the patch.

Exirel (approved):

PR #2268, opened 2022-03-13
deps: forget about `python_version` specifier for `importlib_metadata`

dgw:
### Description
My clean install of Python 3.10 on Ubuntu 20.04 (via deadsnakes/ppa) came with an old version of `importlib_metadata` that tripped up the fallback logic I so carefully included.

Rather than contort the import logic, let's just use the backport until Sopel no longer supports Python versions where stdlib can't replace it entirely.

### Checklist

### Notes
The test suite passes on my py3.10 env with this patch, except for a few `DeprecationWarning`s (from third-party code) that I'd like to try to mitigate _without_ adding transitive deps into our requirements again. Did too much of that in 7.x.

`1311 passed, 8 xfailed, 4 warnings` is still a great sight to see. I'd just _rather_ see `1311 passed, 8 xfailed` instead. 😛

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,5 +5,5 @@ geoip2>=4.0,<5.0
 requests>=2.24.0,<3.0.0
 dnspython<3.0
 sqlalchemy>=1.4,<1.5
-importlib_metadata>=3.6; python_version < '3.10'
+importlib_metadata>=3.6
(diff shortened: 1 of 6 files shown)


Exirel (approved):
I like the "when practical".

PR #2272, opened 2022-04-26
rules: drop `draft/bot` tag

dgw:
### Description
Bot Mode spec has been ratified (ircv3/ircv3-specifications#495). Since we haven't published a release with support for this message tag yet, it's a very easy decision to take *right now* that we'll only support the finalized spec and not the WIP version.

### Checklist

### DIFF
--- a/sopel/plugins/rules.py
+++ b/sopel/plugins/rules.py
@@ -982,10 +982,7 @@ class Rule(AbstractRule):
         intent = pretrigger.tags.get('intent')
         nick = pretrigger.nick
         is_bot_message = (
-            (
-                'draft/bot' in pretrigger.tags or  # can be removed...someday
-                'bot' in pretrigger.tags
-            ) and
+            'bot' in pretrigger.tags and
             event in ["PRIVMSG", "NOTICE"]
         )
         is_echo_message = (
--- a/test/plugins/test_plugins_rules.py
+++ b/test/plugins/test_plugins_rules.py
@@ -658,16 +658,10 @@ def test_rule_match_privmsg_bot_tag(mockbot):
     regex = re.compile(r'.*')
     rule = rules.Rule([regex])
 
-    line = '@draft/bot :TestBot!sopel@example.com PRIVMSG #sopel :Hi!'
-    pretrigger = trigger.PreTrigger(mockbot.nick, line)
-    assert not list(rule.match(mockbot, pretrigger)), (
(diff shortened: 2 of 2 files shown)


half-duplex:
What implements the draft version? How safe is the assumption that they'll all switch to the final spec/tag promptly? Otherwise I'm not sure there's a reason to drop the draft one.

dgw:
The top comment of the linked IRCv3 PR is a non-exhaustive list of existing implementations that may or may not have yet switched to the final tag name. There are 8 servers/clients listed, including Sopel itself. However, since the server adds the message tag, the client implementations are irrelevant to this decision. That leaves just 3 IRCds with draft impls to consider.

Pushing to ratify the spec this week was because InspIRCd wanted to ship the feature in a release on Friday, 4/29. It's safe to assume they will use the final tag name. The other servers will almost certainly release new versions with this feature in its final state before Sopel 8.0 is done. (They all have much better release cadences than we do…)

In this specific case there are no meaningful differences between the draft and final specs, but _in general_ it's improper to implement a WIP spec as I did in #2089, where it responds to both the prefixed and unprefixed tag names.

sopel/plugins/rules.py
```diff
@@ -982,10 +982,7 @@ def match_preconditions(self, bot, pretrigger):
         intent = pretrigger.tags.get('intent')
         nick = pretrigger.nick
         is_bot_message = (
-            (
-                'draft/bot' in pretrigger.tags or  # can be removed...someday
-                'bot' in pretrigger.tags
-            ) and
+            'bot' in pretrigger.tags and
             event in ["PRIVMSG", "NOTICE"]
         )
```
half-duplex, line 987:
```suggestion
        is_bot_message = 'bot' in pretrigger.tags and event in ["PRIVMSG", "NOTICE"]
```
Github has made this confusing but yeah the whole conditional fits in 85 chars now

half-duplex (approved):
Fair enough. We also have +B, and that's probably all anyone should be relying on until they've got the final bot mode cap.

sopel/plugins/rules.py
dgw: Oh, hmm. I was going for "don't touch the pieces that aren't changing", but let's see what @Exirel thinks about the style.

Exirel: 85 is bigger than 79.

Exirel (approved):

PR #2278, opened 2022-05-13
config: don't manually hunt for CA paths

half-duplex:
### Description
It was agreed in #2256 that we should probably not be hunting for CA paths ourselves. This removes that, and allows `ssl` to use its default (openssl's default). This _should_ be an invisible change.

To rebase this on top of or beneath #2256: discard this PR's backends.py changes (`wrap_socket()` had no sane default for ca_certs)

### Checklist
Contemporary caveats apply.

### DIFF
--- a/sopel/config/core_section.py
+++ b/sopel/config/core_section.py
@@ -1,7 +1,5 @@
 from __future__ import annotations
 
-import os.path
-
 from sopel.config.types import (
     BooleanAttribute,
     ChoiceAttribute,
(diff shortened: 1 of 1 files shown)


dgw:
Since #2256 is approved, let's give @Exirel a chance to squash it down and then rebase this on top of it. We could even change the base branch of this PR to that one and go for the "stacked PRs" workflow, but uh… I don't think anyone wants to mess with that. 🙃

Exirel (approved):
It looks good, yet I haven't tested it myself.

sopel/config/core_section.py
```diff
@@ -228,17 +203,16 @@ class CoreSection(StaticSection):
 
     """
 
-    ca_certs = FilenameAttribute('ca_certs', default=_find_certs())
-    """The path to the CA certs ``.pem`` file.
+    ca_certs = FilenameAttribute('ca_certs')
+    """The path to the CA certs ``pem`` file.
```
dgw, line 207: Shouldn't this be `PEM`? I get eliminating the `.` because that's not always the file extension (👀 Debian), but the expected _format_ is probably PEM.

PR #2279, opened 2022-05-14
safety: rework

half-duplex:
### Description
This fixes (or works towards fixing) a bunch of issues with safety.py:
- `.safety` doesn't say what the current state is
- Uses VirusTotal API v2 (though they have no current plans to deprecate it) instead of v3
- Doesn't find URLs in messages containing anything else ("`foo http://example.com`" fails)
- Doesn't make local cache accessible to other plugins
- Doesn't use already-cached VT results in local mode
- Doesn't allow configuration of local or strict defaults
- url.py was not checking the domain blacklist and was also failing to handle "`foo http://example.com`"
- Other tweaks and improvements

### Checklist

### DIFF
--- a/sopel/modules/safety.py
+++ b/sopel/modules/safety.py
@@ -211,7 +211,7 @@ def shutdown(bot: Sopel):
 @plugin.priority('high')
 @plugin.output_prefix(PLUGIN_OUTPUT_PREFIX)
 def url_handler(bot: SopelWrapper, trigger: Trigger):
-    """Checks for malicious URLs"""
+    """Checks for malicious URLs."""
     mode = bot.db.get_channel_value(
         trigger.sender,
         "safety",
@@ -279,7 +279,7 @@ def virustotal_lookup(
 
     :param url: The URL to look up
     :param local_only: If set, only check cache, do not make a new request
-    :param max_cache_age: If set, don't use cache older than this value.
+    :param max_cache_age: If set, don't use cache older than this value
     :returns: A dict containing information about findings, or None if not found
     """
     if url.startswith("hxxp"):
@@ -365,7 +365,7 @@ def virustotal_lookup(
 @plugin.example(".virustotal hxxps://malware.wicar.org/")
 @plugin.output_prefix("[safety][VirusTotal] ")
 def vt_command(bot: SopelWrapper, trigger: Trigger):
-    """Look up VT results on demand"""
+    """Look up VT results on demand."""
     if not bot.settings.safety.vt_api_key:
         bot.reply("Sorry, I don't have a VirusTotal API key configured.")
         return
@@ -421,7 +421,7 @@ def vt_command(bot: SopelWrapper, trigger: Trigger):
 @plugin.example(".safety on")
 @plugin.output_prefix(PLUGIN_OUTPUT_PREFIX)
 def toggle_safety(bot: SopelWrapper, trigger: Trigger):
-    """Set safety setting for channel"""
+    """Set safety setting for channel."""
     if not trigger.admin and bot.channels[trigger.sender].privileges[trigger.nick] < plugin.OP:
         bot.reply('Only channel operators can change safety settings')
         return
--- a/sopel/modules/safety.py
+++ b/sopel/modules/safety.py
@@ -34,6 +34,9 @@ if TYPE_CHECKING:
 LOGGER = logging.getLogger(__name__)
 PLUGIN_OUTPUT_PREFIX = '[safety] '
 
+SAFETY_CACHE_KEY = "safety_cache"
+SAFETY_CACHE_LOCK_KEY = SAFETY_CACHE_KEY + "_lock"
+SAFETY_CACHE_LOCAL_KEY = SAFETY_CACHE_KEY + "_local"
 SAFETY_MODES = ["off", "local", "local strict", "on", "strict"]
 VT_API_URL = "https://www.virustotal.com/api/v3/urls"
 CACHE_LIMIT = 512
@@ -99,10 +102,10 @@ def setup(bot: Sopel):
             "config: enabled_by_default is deprecated, please use default_mode=off",
         )
 
-    if 'safety_cache' not in bot.memory:
-        bot.memory['safety_cache'] = tools.SopelMemory()
-    if 'safety_cache_lock' not in bot.memory:
-        bot.memory['safety_cache_lock'] = threading.Lock()
+    if SAFETY_CACHE_KEY not in bot.memory:
+        bot.memory[SAFETY_CACHE_KEY] = tools.SopelMemory()
+    if SAFETY_CACHE_LOCK_KEY not in bot.memory:
+        bot.memory[SAFETY_CACHE_LOCK_KEY] = threading.Lock()
     for item in bot.settings.safety.known_good:
         known_good.append(re.compile(item, re.I))
 
@@ -195,13 +198,13 @@ def update_local_cache(bot: Sopel, init: bool = False):
             if '.' in domain:
                 # only publicly routable domains matter; skip loopback/link-local stuff
                 unsafe_domains.add(domain)
-    bot.memory["safety_cache_local"] = unsafe_domains
+    bot.memory[SAFETY_CACHE_LOCAL_KEY] = unsafe_domains
 
 
 def shutdown(bot: Sopel):
-    bot.memory.pop('safety_cache', None)
-    bot.memory.pop('safety_cache_local', None)
-    bot.memory.pop('safety_cache_lock', None)
+    bot.memory.pop(SAFETY_CACHE_KEY, None)
+    bot.memory.pop(SAFETY_CACHE_LOCAL_KEY, None)
+    bot.memory.pop(SAFETY_CACHE_LOCK_KEY, None)
 
 
 @plugin.rule(r'(?u).*(https?://\S+).*')
@@ -233,7 +236,7 @@ def url_handler(bot: SopelWrapper, trigger: Trigger):
             if any(regex.search(hostname) for regex in known_good):
                 continue  # explicitly trusted
 
-            if hostname in bot.memory["safety_cache_local"]:
+            if hostname in bot.memory[SAFETY_CACHE_LOCAL_KEY]:
                 LOGGER.debug("[local] domain in blocklist: %r", hostname)
                 positives += 1
                 total += 1
@@ -291,10 +294,10 @@ def virustotal_lookup(
     oldest_cache = datetime(1970, 1, 1, 0, 0, tzinfo=timezone.utc)
     if max_cache_age is not None:
         oldest_cache = datetime.now(timezone.utc) - max_cache_age
-    cache = bot.memory["safety_cache"]
+    cache = bot.memory[SAFETY_CACHE_KEY]
     if url in cache and cache[url]["fetched"] > oldest_cache:
         LOGGER.debug("[VirusTotal] Using cached data for %r", safe_url)
-        return bot.memory["safety_cache"].get(url)
+        return bot.memory[SAFETY_CACHE_KEY].get(url)
     if local_only:
         return None
 
@@ -351,8 +354,8 @@ def virustotal_lookup(
         "fetched": fetched,
         "virustotal_data": vt_data["data"]["attributes"],
     }
-    bot.memory["safety_cache"][url] = result
-    if len(bot.memory["safety_cache"]) >= (2 * CACHE_LIMIT):
+    bot.memory[SAFETY_CACHE_KEY][url] = result
+    if len(bot.memory[SAFETY_CACHE_KEY]) >= (2 * CACHE_LIMIT):
         _clean_cache(bot)
     return result
 
@@ -456,30 +459,30 @@ def _clean_cache(bot: Sopel):
 
     update_local_cache(bot)
 
-    if bot.memory['safety_cache_lock'].acquire(False):
+    if bot.memory[SAFETY_CACHE_LOCK_KEY].acquire(False):
         LOGGER.debug('Starting safety cache cleanup...')
         cutoff = datetime.now(timezone.utc) - timedelta(days=7)
         try:
             # clean up by age first
             old_keys = []
-            for key, data in bot.memory['safety_cache'].items():
+            for key, data in bot.memory[SAFETY_CACHE_KEY].items():
                 if data['fetched'] <= cutoff:
                     old_keys.append(key)
             for key in old_keys:
-                bot.memory['safety_cache'].pop(key, None)
+                bot.memory[SAFETY_CACHE_KEY].pop(key, None)
 
             # clean up more values if the cache is still too big
-            overage = len(bot.memory["safety_cache"]) - CACHE_LIMIT
+            overage = len(bot.memory[SAFETY_CACHE_KEY]) - CACHE_LIMIT
             if overage > 0:
                 extra_keys = sorted(
                     (data.fetched, key)
                     for (key, data)
-                    in bot.memory['safety_cache'].items())[:overage]
+                    in bot.memory[SAFETY_CACHE_KEY].items())[:overage]
                 for (_, key) in extra_keys:
-                    bot.memory['safety_cache'].pop(key, None)
+                    bot.memory[SAFETY_CACHE_KEY].pop(key, None)
         finally:
             # No matter what errors happen (or not), release the lock
-            bot.memory['safety_cache_lock'].release()
+            bot.memory[SAFETY_CACHE_LOCK_KEY].release()
 
         LOGGER.debug('Safety cache cleanup finished.')
     else:
--- a/sopel/modules/safety.py
+++ b/sopel/modules/safety.py
@@ -7,75 +7,112 @@ This plugin uses virustotal.com
 """
 from __future__ import annotations
 
+from base64 import urlsafe_b64encode
 import json
 import logging
 import os.path
 import re
 import threading
 import time
+from typing import Dict, Optional
 from urllib.parse import urlparse
 from urllib.request import urlretrieve
-
 
 import requests
 
 from sopel import formatting, plugin, tools
-from sopel.config import types
+from sopel.bot import Sopel
+from sopel.config import Config, types
+from sopel.trigger import Trigger
 
 
 LOGGER = logging.getLogger(__name__)
 PLUGIN_OUTPUT_PREFIX = '[safety] '
 
-vt_base_api_url = 'https://www.virustotal.com/vtapi/v2/url/'
-malware_domains = set()
+SAFETY_MODES = ["off", "local", "local strict", "on", "strict"]
+VT_API_URL = "https://www.virustotal.com/api/v3/urls"
+CACHE_LIMIT = 512
 known_good = []
-cache_limit = 512
 
 
 class SafetySection(types.StaticSection):
-    enabled_by_default = types.BooleanAttribute('enabled_by_default', default=True)
-    """Whether to enable URL safety in all channels where it isn't explicitly disabled."""
+    enabled_by_default = types.BooleanAttribute("enabled_by_default", default=True)
+    """Deprecated: Sets default_mode to "off" or "on"."""
+    default_mode = types.ValidatedAttribute("default_mode")
+    """Which mode to use in channels without a mode set."""
     known_good = types.ListAttribute('known_good')
     """List of "known good" domains to ignore."""
     vt_api_key = types.ValidatedAttribute('vt_api_key')
     """Optional VirusTotal API key (improves malicious URL detection)."""
+    domain_blocklist_url = types.ValidatedAttribute("domain_blocklist_url")
+    """Optional hosts-file formatted domain blocklist to use instead of StevenBlack's."""
 
 
-def configure(config):
+def configure(settings: Config):
     """
     | name | example | purpose |
     | ---- | ------- | ------- |
-    | enabled\\_by\\_default | True | Enable URL safety in all channels where it isn't explicitly disabled. |
-    | known\\_good | sopel.chat,dftba.net | List of "known good" domains to ignore. |
+    | default\\_mode | on | Which mode to use in channels without a mode set. |
+    | known\\_good | sopel.chat,dftba.net | List of "known good" domains or regexes to ignore. This can save VT API calls. |
     | vt\\_api\\_key | 0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef | Optional VirusTotal API key to improve malicious URL detection |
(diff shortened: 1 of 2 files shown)


sopel/modules/url.py
```diff
@@ -303,14 +303,20 @@ def title_auto(bot, trigger):
...
-            return
-
-    urls = web.search_urls(
+    unchecked_urls = web.search_urls(
         trigger, exclusion_char=bot.config.url.exclusion_char, clean=True)
 
+    urls = []
+    for url in unchecked_urls:
+        # Avoid fetching known malicious links
+        if "safety_cache" in bot.memory and url in bot.memory["safety_cache"]:
```
Exirel, line 312: You can simplify this with:

```suggestion
        if url in bot.memory.get("safety_cache", {}):
```

sopel/modules/url.py
```diff
@@ -303,14 +303,20 @@ def title_auto(bot, trigger):
...
+            if bot.memory["safety_cache"][url]["positives"] > 0:
+                continue
+        netloc = urlparse(url).netloc.lower()
+        if "safety_cache_local" in bot.memory and netloc in bot.memory["safety_cache_local"]:
```
Exirel, line 316: Same simplification here:

```suggestion
        if netloc in bot.memory.get("safety_cache_local", {}):
```

sopel/modules/safety.py
```diff
@@ -26,56 +27,107 @@
...
         bot.memory['safety_cache_lock'] = threading.Lock()
-    for item in bot.config.safety.known_good:
+    for item in bot.settings.safety.known_good:
         known_good.append(re.compile(item, re.I))
 
-    old_file = os.path.join(bot.config.homedir, 'malwaredomains.txt')
+    update_local_cache(bot, init=True)
+
+
+def update_local_cache(bot, init: bool = False):
```
Exirel, line 122: Since you are adding type hint, I invite you to also type `bot`: it can be either a `sopel.bot.Sopel` or a `sopel.bot.SopelWrapper`.

Exirel (changes requested):

sopel/modules/safety.py
```diff
@@ -26,56 +27,107 @@
 LOGGER = logging.getLogger(__name__)
 PLUGIN_OUTPUT_PREFIX = '[safety] '
 
-vt_base_api_url = 'https://www.virustotal.com/vtapi/v2/url/'
-malware_domains = set()
+SAFETY_MODES = ["off", "local", "local strict", "on", "strict"]
+VT_API_URL = b"https://www.virustotal.com/api/v3/urls"
```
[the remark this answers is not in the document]
half-duplex: To make combining with the b64'd domain not require bytes.decode() just for requests to str.encode()

sopel/modules/safety.py
```diff
@@ -26,56 +27,107 @@
...
 known_good = []
 cache_limit = 512
+requests_session = requests.Session()
```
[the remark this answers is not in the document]
half-duplex: Ah, true. https://github.com/psf/requests/issues/2766

sopel/modules/safety.py
```diff
@@ -26,56 +27,107 @@
...
     )
 
 
 def setup(bot):
-    bot.config.define_section('safety', SafetySection)
+    bot.settings.define_section("safety", SafetySection)
+
+    if bot.settings.safety.default_mode is None:
+        bot.settings.safety.default_mode = "on"
+    # migrate from enabled_by_default to default_mode
```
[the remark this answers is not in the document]
half-duplex: I'm not sure what you're suggesting - If there was a way to use BooleanAttribute as optional I'd have done that and checked `is None` and `== False`, but if no default is specified we get `== False`. No way I saw to allow default_mode to override enabled_by_default without making the latter a pretend-bool.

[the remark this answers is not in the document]
Exirel: What I'm suggesting is to not bother removing `enabled_by_default` at all: keep it a boolean (with a default value set to `True`), and do the following:

* set `enabled_by_default`'s default value to `True`
* if `default_mode` is set, use that, and ignore `enabled_by_default`
* otherwise, if `enabled_by_default` is `True`, then set `default_mode` to `on` (or whatever appropriate value we want as the default)
* if it is set to `False`, then set `default_mode` to `off` or something similar to deactivate it

No need to transform `enabled_by_default` into a pseudo-boolean value with `None` for default, since you don't have to care about it. Silently do the migration, document it, and it'll be fine.

sopel/modules/safety.py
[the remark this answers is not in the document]
Exirel: Not a big fan of dealing with bytes in the code. It works, yet I don't feel very at ease with that. Oh well. Doesn't matter.

sopel/modules/safety.py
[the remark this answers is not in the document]
half-duplex: Ah, so migrate then *save*? I was going to avoid an automatic save since last I saw that clobbers comments, formatting, ordering, etc. in the config file.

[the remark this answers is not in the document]
Exirel: Nah, no need to save. When people will upgrade to Sopel 9, having set `enabled_by_default` to `False`, they will get a warning that `enabled_by_default` is not a recognized setting. Given it's a major upgrade, people should be a bit more careful. The worst that can happen here is that they get safety activated by default instead. Not a big deal in my opinion.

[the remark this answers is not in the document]
half-duplex: Ehh. I'm strongly against sending every URL the bot sees to VirusTotal by surprise, but I suppose if that's not wanted, they probably don't have a VT api key configured.
This could also be made more simple by being able to call BooleanAttribute.parse("on"), but eh.
See new push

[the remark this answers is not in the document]
Exirel: This new version is pretty good! And better than I suggested.

Exirel (dismissed):

sopel/modules/safety.py
```diff
@@ -7,75 +7,112 @@
...
+            "Which mode should be used in channels that haven't specifically set one?"
+            "\n({})".format("/".join(SAFETY_MODES))
+        ),
+        default="on",
     )
-    config.safety.configure_setting(
+    settings.safety.configure_setting(
         'known_good',
-        'Enter any domains to whitelist',
+        "Enter any domains to allowlist",
```
dgw, line 71: Using "allowlist" as a verb is stylistically icky. Using it as a single word in any sense is icky, too, for that matter.

I think changes like this come across as performative, and [ignore the original derivation](https://seattlecollegian.com/op-ed-blacklist-and-whitelist-arent-racist-words-you-are/) of the terms, but that's not a good reason to block this whole PR. I'm not that petty. Just… If you're dead set on eliminating the word "whitelist" let's please at least change it to something that sounds like English. Some alternatives:

```suggestion
        "Enter any domains to trust",
```

```suggestion
        "Enter any domains to consider trusted",
```

sopel/modules/safety.py
```diff
@@ -7,75 +7,112 @@
...
-    old_file = os.path.join(bot.config.homedir, 'malwaredomains.txt')
+    update_local_cache(bot, init=True)
+
+
+def update_local_cache(bot: Sopel, init: bool = False):
+    """Download the current malware domain list and load it into memory.
+
+    :param init: Load the file even if it's unchanged
+    """
+
```
dgw, line 112:
```suggestion
```

One day I will get around to installing/configuring/whatever the flake8 plugin/rule that flags this as an error.

sopel/modules/safety.py
```diff
@@ -85,13 +122,18 @@ def setup(bot):
...
-    else:
-        _download_domain_list(loc)
+    loc = os.path.join(bot.settings.homedir, "unsafedomains.txt")
+    if not os.path.isfile(loc) or os.path.getmtime(loc) < time.time() - 24 * 60 * 60:
+        # File doesn't exist or is older than one day — update it
+        url = bot.settings.safety.domain_blocklist_url
+        if url is None or not url.startswith("http"):
+            url = "https://raw.githubusercontent.com/StevenBlack/hosts/master/hosts"
+        LOGGER.info("Downloading malicious domain list from %s", url)
+        # TODO: Can we use a cache header to avoid the download if it's unmodified?
```
dgw, line 132: Is the possibility of doing so really in question, or is this actually a TODO asking some future person to implement checking said header?

Checking the output of `curl --head $stevenblack_url` shows that GitHub serves an [ETag header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/ETag), which can be saved and compared before making future requests for the full file body.

```suggestion
        # TODO: Use ETag header to avoid the download if it's unmodified
```

sopel/modules/safety.py
```diff
@@ -110,129 +152,180 @@ def setup(bot):
...
+        return
+    local_only = "local" in mode or bot.settings.safety.vt_api_key is None
+    strict = "strict" in mode
 
-    try:
-        netloc = urlparse(trigger.group(1)).netloc
-    except ValueError:
-        return  # Invalid IPv6 URL
+    for url in tools.web.search_urls(trigger):
+        safe_url = "hxx" + url[3:]
```
dgw, line 180: This got me to check what `search_urls()` actually finds. The default list of schemes includes ftp:

https://github.com/sopel-irc/sopel/blob/2e25645487c87108330a992f68336a6b2da42188/sopel/tools/web.py#L251

Unless this plugin handles FTP links, it should search for URLs as below.

```suggestion
    for url in tools.web.search_urls(trigger, schemes=['http', 'https']):
        safe_url = "hxx" + url[3:]
```

dgw (commented):
Non-blocking style/correctness nits.

sopel/modules/safety.py
half-duplex: Mostly a suggestion for a future improvement, since it was beyond the scope of my original unfscking goals. I guess that could be easily stuck in the DB.

sopel/modules/safety.py
half-duplex: VT doesn't seem to take FTP links, but I don't mind running them through the hosts blocklist. Will .startswith() it

sopel/modules/safety.py
half-duplex: I'd changed the last bit in one place, but not in the other two. Why are plugin config options documented in three separate places? I understand doc phrasing vs prompt phrasing, but where's the table used?

dgw: The table gets slurped up by the part of the website that generates [plugin configuration info](https://sopel.chat/usage/plugin-configuration/). Is there a better way? Absolutely there is, but no one has bothered refactoring this bit yet. 🤷‍♂️

half-duplex:
I meant to keep this as at least two separate commits, but accidentally squashed them together :sob:
Re-requesting reviews because of nontrivial logic changes and new command.

image

sopel/modules/safety.py
```diff
@@ -212,80 +250,167 @@ def url_handler(bot: Sopel, trigger: Trigger):
             )
             bot.say(
                 "{} {} of {} engine{} flagged a link {} posted as malicious".format(
-                    formatting.bold(formatting.color("WARNING:", "red")),
+                    bold(color("WARNING:", colors.RED)),
                     positives,
                     total,
                     "" if total == 1 else "s",
-                    formatting.bold(trigger.nick),
+                    bold(trigger.nick),
```
dgw, line 257: Kinda gratuitous refactoring but OK.

sopel/modules/safety.py
```diff
@@ -122,19 +113,67 @@ def update_local_cache(bot: Sopel, init: bool = False):
...
+    updated = download_domain_list(bot, path)
+    if not os.path.isfile(path):
+        LOGGER.warning("Could not load unsafe domain list")
+        return
+
+    if not updated and not init:
         return
 
-    with open(loc, 'r') as f:
+    LOGGER.debug("Loading new unsafedomains list")
```
dgw, line 174: Missing a space? Weird to have different logger calls use "unsafedomains list" and "unsafe domain list" interchangeably.

sopel/modules/safety.py
```diff
@@ -122,19 +113,67 @@ def update_local_cache(bot: Sopel, init: bool = False):
...
+    malware_domains = set()
```
dgw, line 175: Probably should rename this variable if the list is going to be treated/logged/configured as "unsafe domains".

sopel/modules/safety.py
```diff
@@ -212,80 +250,167 @@ def url_handler(bot: Sopel, trigger: Trigger):
...
+        LOGGER.debug("[VirusTotal] Scan failed or unfinished for %r", safe_url)
+        return None
+    fetched = datetime.now(timezone.utc)
     # Only count strong opinions (ignore suspicious/timeout/undetected)
     result = {
         "positives": last_analysis["malicious"],
         "total": last_analysis["malicious"] + last_analysis["harmless"],
         "fetched": fetched,
-        "virustotal_data": vt_data,
+        # Subject to change formats!
```
dgw, line 346: Don't quite understand what this comment is telling me.

sopel/modules/safety.py
```diff
@@ -212,80 +250,167 @@ def url_handler(bot: Sopel, trigger: Trigger):
...
+@plugin.example(".virustotal hxxps://malware.wicar.org/")
+@plugin.output_prefix("[safety][VirusTotal] ")
+def vt_command(bot: Sopel, trigger: Trigger):
+    """Look up VT results on demand"""
+    if not bot.settings.safety.vt_api_key:
+        bot.reply("Sorry, I don't have a VirusTotal API key configured.")
+        return
+
+    url = trigger.group(2)
+    zwsp_safe_url = zwsp("hxx" + url[3:])
```
dgw, line 366: I am very confused. Why replace the scheme prefix _and_ insert ZWSP between every character? Seems excessive, especially because `zwsp(thing)` as defined above will return `4 * len(thing) - 3` bytes when encoded to UTF-8 for sending to IRC (`'\u200b'.encode('utf-8')` ➡️ `b'\xe2\x80\x8b'`).

For nontrivial real-world URLs (not just the domain part) that might actually link to malware, this approach will quickly make the output of this command longer than the 510 total bytes available for one IRC line.

dgw (commented):
OK, I have been nudged into "actual review" territory.

sopel/modules/safety.py
half-duplex: Was originally trying to not muck around inside that block much, but yeah.

sopel/modules/safety.py
half-duplex: It started feeling silly with `formatting.bold(formatting.color("foo bar", formatting.colors.RED))`

sopel/modules/safety.py
half-duplex: If VT changes how it returns data, the structure of that dict will change, with or without a sopel version change.

sopel/modules/safety.py
half-duplex: The length concern is fair. The idea was that I don't know what clients use to decide what's a link or not (whether they'd catch just `example.com` out of `hxxp://example.com/`, and when quoting a possibly-malware domain we don't want it to ever be linkified.
I also didn't want to not refer to the link somehow when this may wait 10 sec for a scan & be confusing in a busy channel, though the same applies to the regular "WARNING:" lines.

```irc
<+mal> does everybody's client properly delinkify this? hxxp://example.com/
< RhinosF1> If I share it, it becomes a normal link
< RhinosF1> The protocol gets stripped
```

So, I think it's important to invalidate the link in at least one more way. I'll switch it to only ZWSP the periods.

sopel/modules/safety.py
dgw: We should have a comment like this next to every use of API data in any plugin, then.

Of course the API data format _could_ change, but it _won't_ if VT is actually serious about the API version number.

sopel/modules/safety.py
half-duplex: No more ZWSP, just hxxp and [.]ing the .s and :s

dgw:
I'm not going to dig into this much more. Just noting that I added another flake8 plugin to the todo list in #1765 after seeing the mess of imports that type-hinting this plugin added.

sopel/modules/safety.py
```diff
@@ -7,75 +7,107 @@
...
 
 
-def setup(bot):
-    bot.config.define_section('safety', SafetySection)
+def setup(bot: Sopel):
+    bot.settings.define_section("safety", SafetySection)
+
+    if bot.settings.safety.default_mode is None:
+        bot.settings.safety.default_mode = "on"
+    # migrate from enabled_by_default to default_mode. remove in v8.1 or v9
```
dgw, line 95: This should contain a `TODO` so we'll find it later when looking for such things.

sopel/modules/safety.py
```diff
@@ -7,75 +7,107 @@
...
+    if not bot.settings.safety.enabled_by_default:
+        bot.settings.safety.default_mode = "off"
+        LOGGER.info(
```
dgw, line 98: Shouldn't this be a warning?

sopel/modules/safety.py
```diff
@@ -85,14 +117,68 @@ def setup(bot):
...
+
+    r = requests.get(url, stream=True)
+    try:
+        r.raise_for_status()
+        with open(path + ".new", "wb") as f:
+            for data in r.iter_content(None):
+                f.write(data)
+    except Exception:
+        # don't bother handling, we'll try again tomorrow
+        LOGGER.warning("Unsafe domain list download failed, using cache")
```
dgw, line 157: Nittiest of nits.

```suggestion
        LOGGER.warning("Unsafe domain list download failed; using cache")
```

sopel/modules/safety.py
```diff
@@ -85,14 +117,68 @@ def setup(bot):
...
+    path = os.path.join(bot.settings.homedir, "unsafedomains.txt")
+    updated = download_domain_list(bot, path)
+    if not os.path.isfile(path):
+        LOGGER.warning("Could not load unsafe domain list")
+        return
+
+    if not updated and not init:
+        return
+
+    LOGGER.debug("Loading new unsafe domain list")
```
dgw, line 179: Unclear why this is a `.debug()` but most messages related to the domain list fetch/load process are `.info()`s.

dgw (commented):
I actually thought this was ready to go, but apparently not. Now that I've gone through it once more it's time to find out why GH still shows it in my "Approved 8.0.0" view, and make it not do that.

sopel/modules/safety.py
half-duplex: also added one on the malwaredomains migration

sopel/modules/safety.py
half-duplex: Changed a few, since there's a higher chance of caring about an actual update than it being loaded. Also rebased for merge conflict.

sopel/modules/url.py
```diff
@@ -313,7 +319,7 @@ def title_auto(bot: SopelWrapper, trigger: Trigger):
             # Guard against responding to other instances of this bot.
             if message != trigger:
                 bot.say(message)
-        bot.memory["last_seen_url"][trigger.sender] = url
+        bot.memory['last_seen_url'][trigger.sender] = url
```
dgw, line 322: Why the no-op quote change? It's not even consistent with the above rewrite.

dgw (approved):
All good from my perspective. Probably still wait for re-review from @Exirel to be sure everything he requested is satisfactory too.

sopel/modules/url.py
half-duplex: Huh. Don't know. I may have gotten confused in trying to keep the diff tight? Fixed

sopel/modules/url.py
```diff
@@ -297,14 +297,20 @@ def title_auto(bot: SopelWrapper, trigger: Trigger):
...
 
+    urls = []
+    for url in unchecked_urls:
+        # Avoid fetching known malicious links
+        if url in bot.memory.get("safety_cache", {}):
+            if bot.memory["safety_cache"][url]["positives"] > 0:
+                continue
+        hostname = urlparse(url).hostname.lower()
+        if hostname in bot.memory.get("safety_cache_local", {}):
+            continue
```
Exirel, line 311: Each `dict.get` or `dict[...]` call cost a little more than doing this:

```suggestion
    urls = []
    safety_cache = bot.memory.get("safety_cache", {})
    safety_cache_local = bot.memory.get("safety_cache_local", {})
    for url in unchecked_urls:
        # Avoid fetching known malicious links
        if url in safety_cache and safety_cache[url]["positives"] > 0:
            continue
        hostname = urlparse(url).hostname.lower()
        if hostname in safety_cache_local:
            continue
```

And yes, it could be a list-comprehension. Or put that in a function that would `yield url`... so many options. A for-loop is fine.

Exirel (changes requested):
While we are upgrading the `safety` plugin, I think we should replace all the `"safety_cache"` hard-coded key into a constant, so `bot.memory["satefy_cache"]` would become `bot.memory[SAFETY_CACHE_KEY]`.

They are also some probable nitpick around docstrings that lack a period here and there, but I couldn't be bothered to check them all.

half-duplex:
Should url.py import safety.SAFETY_CACHE_KEY, or just use the string?
Also considering whether I should just find a way to merge safety_cache and safety_cache_local...

Exirel:
> Should url.py import safety.SAFETY_CACHE_KEY, or just use the string?

You can import it if you want, both option are fine by me. We can accept that URL totally depends on the safety plugin, because it kind of does. As long as safety is part of core, anyway. Which shouldn't change yet.

half-duplex:
I don't think URL *does* depend on the safety plugin - if safety is absent, url just skips those checks. Importing the constant would require a conditional to keep the dependency separate.

half-duplex:
Ok, if something uses safety.py's key without following the format there will be issues, but url.py doesn't fail to perform any of its  functions if safety.py is absent. I can add two more .get()s if you want, but either way I think I prefer using the string.

dgw:
If the code only even tries to read the safety info if the safety plugin is available, you minimize the chance of stepping on something else that has coopted its key name. That, to me, would be the main advantage of e.g.

```py
try:
    from sopel.modules.safety import SAFETY_CACHE_KEY
except ImportError:
    SAFETY_CACHE_KEY = None

# ... other code ...

# in the relevant function
    if SAFETY_CACHE_KEY is not None:
        # use it
```

Is that perfect? No, because `safety` might not be enabled. But there's also `bot.has_plugin('safety')` for additional sanity checking if you want that.

Exirel:
> both option are fine by me

I'm still OK with that, as I don't think it matters too much at the moment. The Safety feature of Sopel would probably be better with a proper interface dedicated to that, instead of a Plugin. So for now, a string is fine by me. I'm just glad we could talk a bit about it, even tho it's nothing more than nitpicking at this point.

Exirel (approved):

Issue #2281, opened 2022-05-16
search/check_url_callbacks() ignore disable_commands


half-duplex:
`url.py` uses both `bot.search_url_callbacks()` and `bot.rules.check_url_callback()` to see if a URL has been handled by another plugin.

These functions ignoring `config[channel].disable_commands` means that if a channel has disabled e.g. `{"reddit":["post_or_comment_info"]}` and a user sends a reddit post link, url.py will not auto-title it.

In some cases this may be desirable ("we have a bot that does something entirely different with these URLs"), but in some it may not be ("we don't like the colors in sopel-youtube" (fixed, but you get the idea)). I'm not sure which way to go on that, unless we make `.urlexclude` per-channel.

More impactfully, this also makes it difficult for url.py's `.title` command to behave as expected. Currently, it lets Sopel's url handling code dispatch the callback to e.g. reddit.py, and skips any handling of that URL - but it doesn't know that the callback has been disabled in this channel, and that it should dispatch the callback itself. The current behavior is that `.title` for a disabled-command URL is a no-op.

### Environment
- Sopel `.version`: cea42e12f42ac653336ac139cf80ea0bbd876f7d
- Sopel installed via: git/pip

dgw:
The old `bot.register_url_callback()` method and its friends are deprecated, and according to GitHub code search results for `org:sopel-irc register_url_callback`, neither core plugins nor first-party plugins use them any more.

Rather than spend any time fixing this, I motion we just let it be until #2436 kills off the last vestiges of an obsolete system.

PR #2282, opened 2022-05-16
url: fix .title vs url callback plugins

### LINKED ISSUE #2054: url: `exclude` setting also applies to `.title` command
### Description
Before Sopel 7.1, the only sane way to disable auto-title would have been setting the config option `url.exclude` to `.*` (matching any possible URL). The only problem is, doing that _also_ makes `.title` not work.

At least as far back as Willie 4.0.0, when the responsible code was restructured (in d1acb126a51a6353bd559ab1d9b36c1e4b9e5c83), and probably even longer, doing e.g. `.title http://thisurlis.excluded` would yield silence from the bot.

Recently, the silence has changed to an error message, since it's impossible for the code added in #1380 (then re-added in #2029) to tell the difference between a failed fetch, an excluded link, or a link handled by some other plugin.

### Reproduction steps
1. Set `url.exclude` to `.*`
2. Run `.title http://any.link/of/your/choosing`
3. Get an error: "Sorry, fetching that title failed. Make sure the site is working."

### Expected behavior
I expect Sopel to try and fetch the link's title anyway, because it's been explicitly requested.

### Environment
- Sopel `.version`: 7.1.0-dev
- Sopel installed via: `pip install -e`
- Python version: 3.8.5
- Operating system: Ubuntu 18.04
- Relevant plugins: url

### Notes
Since this behavior has been around for _so long_, I elected to hold any change until 8.0, even though any fix is probably trivial enough to include in 7.1 still. We'd be changing the meaning of a setting.

It seems easiest to fix both this and #2055 at the same time.

### LINKED ISSUE #2055: url: `.title` command shows an error for links handled by another plugin
### Description
Links handled by another plugin (for example links to GitHub issues, if the `github` plugin is installed) are skipped intentionally in the `.title` command, to avoid giving duplicate output. However, the error message code from #2029 can't tell the difference between a link that was intentionally skipped and a link for which a title couldn't be fetched.

### Reproduction steps
1. Command Sopel, `.title http://some.link/that/another/plugin/handles`
   - With core plugins only, a Reddit post is a good way to trigger this.
2. `url` responds with an error: "Sorry, fetching that title failed. Make sure the site is working."
   AND the plugin that knows how to handle the link will also reply

### Expected behavior
The `url` plugin should be silent in this case, unless multiple URLs are passed to `.title` and at least one of them isn't handled by another plugin.

### Environment
- Sopel `.version`: 7.1.0-dev
- Sopel installed via: `pip install -e`
- Python version: 3.8.5
- Operating system: Ubuntu 18.04
- Relevant plugins: url

### Notes
Likely easiest to fix this in conjunction with #2054. I'm therefore marking this for 8.0 even though it would be nice to fix it in 7.1. (The target milestone is, as always, subject to change after discussion.)

half-duplex:
- .title for e.g. reddit would say "Sorry, couldn't title that" followed by the reddit.py callback output
- last_seen_url (therefore `.title` sans args) was missing callback'd/excluded URLs
- tighten some security dumbs (except the safety part in #2279)
  - no point in checking ip without DNS lookups if an attacker has $5 and a namecheap account
- add type hints
- some other tidying.

Related for some remaining issues: #2281
Fixes #2054
Fixes #2055

### Checklist

### DIFF
--- a/sopel/modules/url.py
+++ b/sopel/modules/url.py
@@ -10,9 +10,10 @@ https://sopel.chat
 """
 from __future__ import annotations
 
-import ipaddress
+from ipaddress import ip_address
 import logging
 import re
+from typing import TYPE_CHECKING
 from urllib.parse import urlparse
 
 import dns.resolver
@@ -23,6 +24,12 @@ from sopel import plugin, tools
 from sopel.config import types
 from sopel.tools import web
 
+if TYPE_CHECKING:
+    from typing import Generator, List, Optional, Tuple
+
+    from sopel.bot import Sopel, SopelWrapper
+    from sopel.config import Config
+    from sopel.trigger import Trigger
 
 LOGGER = logging.getLogger(__name__)
 USER_AGENT = (
@@ -39,8 +46,6 @@ DEFAULT_HEADERS = {
 # world's best way to do this, but it'll do for now.
 TITLE_TAG_DATA = re.compile('<(/?)title( [^>]+)?>', re.IGNORECASE)
 QUOTED_TITLE = re.compile('[\'"]<title>[\'"]', re.IGNORECASE)
-# This is another regex that presumably does something important.
-RE_DCC = re.compile(r'(?i)dcc\ssend')
 # This sets the maximum number of bytes that should be read in order to find
 # the title. We don't want it too high, or a link to a big file/stream will
 # just keep downloading until there's no more memory. 640k ought to be enough
@@ -62,13 +67,10 @@ class UrlSection(types.StaticSection):
     """If greater than 0, the title fetcher will include a TinyURL version of links longer than this many characters."""
     enable_private_resolution = types.BooleanAttribute(
         'enable_private_resolution', default=False)
-    """Enable URL lookups for RFC1918 addresses"""
-    enable_dns_resolution = types.BooleanAttribute(
-        'enable_dns_resolution', default=False)
-    """Enable DNS resolution for all domains to validate if there are RFC1918 resolutions"""
+    """Enable requests to private and local network IP addresses"""
 
 
-def configure(config):
+def configure(config: Config):
     """
     | name | example | purpose |
     | ---- | ------- | ------- |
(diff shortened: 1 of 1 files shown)


sopel/modules/url.py
```diff
@@ -154,7 +149,7 @@ def shutdown(bot):
 @plugin.example('.urlpexclude example\\.com/\\w+', user_help=True)
 @plugin.example('.urlexclude example.com/path', user_help=True)
 @plugin.output_prefix('[url] ')
-def url_ban(bot, trigger):
+def url_ban(bot: Sopel, trigger: Trigger):
```
Exirel, line 152: This is not an instance of `Sopel` but of `SopelWrapper`, true for the other commands & rules.

half-duplex: Ah, thanks. IIRC in the safety PR you said to use either - same there? Or what's different?

Exirel: Plugin callables need a SopelWrapper. Job, Setup, Teardown: Sopel. Everything else: it depends.

half-duplex:
RE_DCC - Seems to strip "dcc send". Should this ever matter? DCC is CTCP only, and title responses should never be sent as CTCP, right?

Exirel:
Given that you are fixing an issues, that's pretty impressive work. I like that every time you work on something, you make it more clear for me and give me better ideas than I had for future refactoring.

So all in all, good job.

sopel/modules/url.py
```diff
@@ -311,55 +303,68 @@ def title_auto(bot, trigger):
...
+
+    If a callback was dispatched, only the url and dispatched=True will be set.
+
+    For titles explicitly requested by the user, exclusion_char and excludes
+    are skipped.
+
+    :param bot: Sopel instance
+    :param trigger: The trigger object for this event
+    :param urls: The URLs detected in the triggering message
+    :param requested: Whether the title was explicitly requested (vs automatic)
```
Exirel, line 336: The format used in other part of the code is usually in that order:

* one line synopsis
* parameters
* long description

So this bloc of 4 lines should be above the long description, and not at the end.

Exirel (changes requested):
Mostly style nitpicks. All the complains I have about `process_urls` come from the complexity of the feature as outgrow the old interface that was put in place (probably by me at some point).

So this is now in my "refactoring" bucket, and clearly not in the near future because, as I already said: this PR is a damn good work.

Exirel:
Is there a reason for this PR to still be a draft? It looks pretty good to me. Did I forget something?

half-duplex:
I was going to remove the DCC thing if we can't figure out why it's there, but I guess that's stretching the scope a bit.

Exirel (approved):

dgw:
I wouldn't mind removing the seemingly useless DCC regex, but I'll merge this regardless of whether that's included by the time I get to its place in the test/review queue.

dgw:
> by the time I get to its place in the test/review queue.

I did. It passed my gauntlet. 🎉

PR #2285, opened 2022-05-25
url: improve private IP protection

### LINKED ISSUE #2284: url: private_resolution/dns_resolution useless
### Description
The url.enable_private_resolution and url.enable_dns_resolution settings do not work as advertised, and the concept of the latter is fatally flawed.

The current `url.py` private-address protection logic is as follows:
```python
if not enable_private_resolution:
    if host is ip and ip is private:
        return E_NAUGHTY
    if enable_dns_resolution and resolve(host) is private:
        return E_NAUGHTY
post_title(url)
```
This has many problems, as demonstrated below.

### Reproduction steps
Oh lordy.

#### Scenario 1: Attacker has $2
1. Buy a domain, e.g. `haha.bad`
2. Create a DNS record: `haha.bad A 127.0.0.1`
3. Post a link in IRC: `http://haha.bad/csrf/me/daddy`

If enable_dns_resolution is disabled, this bypasses all private-address protection.

#### Scenario 2: Attacker doesn't have $2
1. Configure webserver to respond to requests with `302 http://127.0.0.1/csrf/me/daddy`
2. Post a link to `http://$webserver_ip/`

The posted link is a public IP, so Sopel will happily load it and follow the redirect to the private address.

#### Scenario 3: Attacker has $2, but the above is fixed
1. Buy a domain, e.g. `haha.bad`
2. Set the nameservers for `haha.bad` to an IP you control
3. Run a script on that IP with the following behavior:
```
attack = False
if attack := not attack:
    return dns_record(type="A", address="1.2.3.4", ttl=0)
return dns_record(type="A", address="127.0.0.1", ttl=0)
```
In the checking stage, Sopel will see the address 1.2.3.4. When performing the request, 127.0.0.1 will be used without validation. ([TOCTOU](https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use))

#### Scenario 4: Attacker has $2 and target is IPv6
1. Buy a domain, e.g. `haha.bad`
2. Point `haha.bad` to the target IPv6 address
3. Post a link to `http://haha.bad/csrf/me/daddy`

`dns.resolver.resolve()` does not request AAAA records, so any combination of DNS and IPv6 passes validation.

### Expected behavior
As explained on IRC, this behavior (especially the first two parts) is objectively broken and gives a false sense of security. We should either remove those config options and not pretend, or we should make url.py _unable_ to talk to private IPs.

### Environment
- Sopel `.version`: Since before the name "Sopel" to present.
- Relevant plugins: url.py

### Notes
Ping @dgw and @Exirel, who requested this be an issue instead of IRC comments and a PR.

### LINKED ISSUE #2348: url: Unexpected error on nonexistent domains
The following code line raises an unhandled `dns.resolver.NXDOMAIN` for URLs at nonexistent domains:

https://github.com/sopel-irc/sopel/blob/eff60e9b49a14e1f3c0c4dab5bb4d2ddbf0098a5/sopel/modules/url.py#L377

Some variant of this operation has been in the code for a while, but the context has changed so much during 8.x that I decided not to follow `git blame` all the way back to the first addition of `dns.resolver`. Seemed like a waste of time.

half-duplex:
### Description
Atop #2282
Fixes #2284

Draft implementation should resolve #2348.

- [x] Document TOCTOU

### Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -47,7 +47,6 @@ dependencies = [
     "xmltodict>=0.12,<0.14",
     "pytz",
     "requests>=2.24.0,<3.0.0",
-    "dnspython<3.0",
     "sqlalchemy>=1.4,<1.5",
     "importlib_metadata>=3.6",
     "packaging>=23.2",
--- a/sopel/builtins/url.py
+++ b/sopel/builtins/url.py
@@ -10,22 +10,24 @@ https://sopel.chat
 """
 from __future__ import annotations
 
+from email.message import EmailMessage
 from ipaddress import ip_address
 import logging
 import re
+from socket import getaddrinfo, IPPROTO_TCP
 from typing import NamedTuple, Optional, TYPE_CHECKING
 from urllib.parse import urlparse
 
-import dns.resolver
 import requests
 from urllib3.exceptions import LocationValueError  # type: ignore[import]
 
 from sopel import plugin, privileges, tools
 from sopel.config import types
-from sopel.tools import web
 
 if TYPE_CHECKING:
     from collections.abc import Generator
+    from typing import Iterable
+
     from sopel.bot import Sopel, SopelWrapper
     from sopel.config import Config
     from sopel.trigger import Trigger
@@ -73,7 +75,14 @@ class UrlSection(types.StaticSection):
     """If greater than 0, the title fetcher will include a TinyURL version of links longer than this many characters."""
     enable_private_resolution = types.BooleanAttribute(
         'enable_private_resolution', default=False)
-    """Enable requests to private and local network IP addresses"""
+    """Allow all requests to private and loopback networks.
+
+    If disabled (the default), obvious attempts to load pages from loopback and
+    private IP addresses will be blocked. If this matters for your security you
+    must use additional protections like a firewall and CSRF tokens, since an
+    attacker can change which IP address a domain name refers to between when
+    Sopel checks it and when the HTTP request is made.
+    """
 
 
 def configure(config: Config):
@@ -84,7 +93,7 @@ def configure(config: Config):
     | exclude | https?://git\\\\.io/.* | A list of regular expressions for URLs for which the title should not be shown. |
     | exclusion\\_char | ! | A character (or string) which, when immediately preceding a URL, will stop the URL's title from being shown. |
     | shorten\\_url\\_length | 72 | If greater than 0, the title fetcher will include a TinyURL version of links longer than this many characters. |
-    | enable\\_private\\_resolution | False | Enable requests to private and local network IP addresses. |
+    | enable\\_private\\_resolution | False | Allow all requests to private IP addresses. Leaving this disabled only blocks obvious attempts, use a firewall! |
     """
     config.define_section('url', UrlSection)
     config.url.configure_setting(
@@ -107,7 +116,7 @@ def configure(config: Config):
     )
     config.url.configure_setting(
         'enable_private_resolution',
-        'Enable requests to private and local network IP addresses?'
+        'Allow all requests to private (local network) IP addresses?'
     )
 
 
@@ -288,7 +297,7 @@ def title_command(bot: SopelWrapper, trigger: Trigger):
         urls = [bot.memory["last_seen_url"][trigger.sender]]
     else:
         # needs to be a list so len() can be checked later
-        urls = list(web.search_urls(trigger))
+        urls = list(tools.web.search_urls(trigger))
 
     for url, title, domain, tinyurl, ignored in process_urls(
         bot, trigger, urls, requested=True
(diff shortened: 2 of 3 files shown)


half-duplex:
- Passing bot.memory to find_title feels weird, but it's needed for the safety checks, which need to be done each redirect
- Both methods of getting the IPs for a hostname ignore whether a system has working IPv6. This is a really annoying problem I didn't foresee.
  - Check for a global IPv6 address? (But some working v6 doesn't have that)
  - Check for IPv6 routing? (Feels dirty
  - If TTL is over a second or two, assume it'll be cached and accept the edge case TOCTOUs?
    - Needs fallback. Fall back to only supporting IPv4? Gross, but only needed when TTL<2, so would work 99.999% of the time.

dgw:
With #2282 merged, presumably this can proceed toward Ready status after a rebase.

Any idea why this affects VCR tests? Somehow it is causing the example request for google.com to appear as a direct IP instead.

half-duplex:
If the test checks for the address actually contacted, it'll be getting something different from what it used to in some cases. I figured I'd look into that after seeing if I'm even continuing with this redirect handling strategy because of the above questions.

dgw:
Assuming this PR graduates from draft status, please remove `dnspython` from requirements. These changes remove its use from `url.py` (🥳), and nothing else uses it.

half-duplex:
Rebased, but still looking for feedback on the above questions.

sopel/modules/url.py
```diff
@@ -436,32 +419,107 @@ def check_callbacks(bot: SopelWrapper, url: str, use_excludes: bool = True) -> b
     )
 
 
-def find_title(url: str, verify: bool = True) -> Optional[str]:
-    """Return the title for the given URL.
+def find_title(
+    url: str,
+    verify: bool = True,
+    allow_local: bool = False,
+    memory: tools.SopelMemory = {},
```
Exirel, line 426: I'd replace that by 2 parameters:

* `unsafe_urls`: a list of unsafe URLs
* `unsafe_domains`: a list of unsafe domain name

You would then need to call the function as follow:

```python
unsafe_urls = [
    unsafe_url
    for unsafe_url, safe_data in bot.memory.get("safety_cache", {}).items()
    if safe_data.get('positives', 0) > 0
]
unsafe_domains = bot.memory.get("safety_cache_local", {}).values()

result = find_title(url, verify, allow_local, unsafe_urls, unsafe_domains)
```

No need for a SopelMemory anymore, no need to pass the bot around, etc.

sopel/modules/url.py
```diff
@@ -436,32 +419,107 @@ def check_callbacks(bot: SopelWrapper, url: str, use_excludes: bool = True) -> b
...
+    session.mount("https://", HostAdapter())
+    session.headers = dict(DEFAULT_HEADERS)
+    while redirects_left > 0:
+        redirects_left -= 1
+        parsed_url = urlparse(url)
+
+        # Avoid fetching known malicious links
+        safety_cache = bot.memory.get("safety_cache", {})
+        safety_cache_local = bot.memory.get("safety_cache_local", {})
+        if url in safety_cache and safety_cache[url]["positives"] > 0:
```
Exirel, line 447: Note that you are using `bot.memory` but your argument is `memory`.

sopel/modules/url.py
```diff
@@ -436,32 +419,107 @@ def check_callbacks(bot: SopelWrapper, url: str, use_excludes: bool = True) -> b
...
+                LOGGER.debug("Ignoring unsafe URL: %r", url)
+                return None
```
Exirel, line 449:
```suggestion
            LOGGER.debug("Ignoring unsafe URL: %r", url)
            return None
```

Exirel (changes requested):

dgw:
@half-duplex Did you still want to get this in? Haven't seen any activity for a while.

half-duplex:
For when I look at this next: https://github.com/requests/toolbelt/pull/293 was merged

sopel/modules/url.py
half-duplex: Good point, thanks.

half-duplex:
Some dns tests:
http://lo.4.sopel.sec.gd/
http://net.4.sopel.sec.gd/
http://net2.4.sopel.sec.gd/
http://lo.6.sopel.sec.gd/
http://lo4.6.sopel.sec.gd/
http://ula.6.sopel.sec.gd/
http://ll.6.sopel.sec.gd/

dgw:
Since @half-duplex reminded us this exists during another (unrelated) IRC discussion today, I came by to give it a look. Nothing jumps out at me, but I will defer final approval to @Exirel as the original reviewer who requested changes.

I'm excited at getting rid of `dnspython` as a dependency, though. 😁

dgw:
> I'm excited at getting rid of `dnspython` as a dependency, though. 😁

Even more so now that `dnspython` 2.7.0 has broken type-checks on `master` (see https://github.com/sopel-irc/sopel/pull/2594#issuecomment-2398040528). Python 3.8 isn't supported in the new version, so that CI job proceeded past the `lint` step to `test` before getting canceled when a sibling failed. I also confirmed myself, `make lint` succeeds on `master` with `dnspython` 2.6 and fails with 2.7.

@Exirel when you've some time to look at this again, it'd be much appreciated 😸

Exirel (approved):
Good riddance @ dnspython!

PR #2290, opened 2022-05-29
bot, loader, plugin, rules: add NOTICE message to rate limits

### LINKED ISSUE #830: Plugin API: Add optional message to module.rate()
If set, send `message` as a notice to the calling user if the command was canceled due to rate-limit. Ideally, it would support a placeholder such as `{time_left}` for the message to tell the user when they can use the command again.

I'd be happy to code it up and PR it; just want to establish that it'd have a chance of being merged first, before putting in that time.

~~Could also have a third parameter to toggle between using NOTICE and PRIVMSG.~~

Exirel:
### Description

Fix #830 by adding an optional message and three new decorators for specific rate limit messages (user, channel, and server).

### Checklist

### DIFF
--- a/sopel/bot.py
+++ b/sopel/bot.py
@@ -587,18 +587,34 @@ class Sopel(irc.AbstractBot):
         sopel: 'SopelWrapper',
         trigger: Trigger,
     ) -> None:
+        nick = trigger.nick
+        context = trigger.sender
+        is_channel = context and not context.is_nick()
+
         # rate limiting
         if not trigger.admin and not rule.is_unblockable():
-            if rule.is_rate_limited(trigger.nick):
+            if rule.is_user_rate_limited(nick):
+                message = rule.get_user_rate_message(nick)
+                if message:
+                    sopel.notice(message, destination=nick)
                 return
-            if not trigger.is_privmsg and rule.is_channel_rate_limited(trigger.sender):
+
+            if is_channel and rule.is_channel_rate_limited(context):
+                message = rule.get_channel_rate_message(nick, context)
+                if message:
+                    sopel.notice(message, destination=nick)
                 return
+
             if rule.is_global_rate_limited():
+                message = rule.get_global_rate_message(nick)
+                if message:
+                    sopel.notice(message, destination=nick)
                 return
 
         # channel config
-        if trigger.sender in self.config:
-            channel_config = self.config[trigger.sender]
+        if is_channel and context in self.config:
+            channel_config = self.config[context]
+            plugin_name = rule.get_plugin_name()
 
             # disable listed plugins completely on provided channel
             if 'disable_plugins' in channel_config:
@@ -606,13 +622,13 @@ class Sopel(irc.AbstractBot):
 
                 if '*' in disabled_plugins:
                     return
-                elif rule.get_plugin_name() in disabled_plugins:
+                elif plugin_name in disabled_plugins:
                     return
 
             # disable chosen methods from plugins
             if 'disable_commands' in channel_config:
                 disabled_commands = literal_eval(channel_config.disable_commands)
-                disabled_commands = disabled_commands.get(rule.get_plugin_name(), [])
+                disabled_commands = disabled_commands.get(plugin_name, [])
                 if rule.get_rule_label() in disabled_commands:
                     return
 
@@ -650,11 +666,11 @@ class Sopel(irc.AbstractBot):
         if not trigger.admin and not func.unblockable:
             if func in self._times[nick]:
                 usertimediff = current_time - self._times[nick][func]
-                if func.rate > 0 and usertimediff < func.rate:
+                if func.user_rate > 0 and usertimediff < func.user_rate:
                     LOGGER.info(
                         "%s prevented from using %s in %s due to user limit: %d < %d",
                         trigger.nick, func.__name__, trigger.sender, usertimediff,
-                        func.rate
+                        func.user_rate
                     )
                     return
             if func in self._times[self.nick]:
--- a/sopel/loader.py
+++ b/sopel/loader.py
@@ -48,9 +48,13 @@ def clean_callable(func, config):
     if is_limitable(func):
         # These attributes are a waste of memory on callables that don't pass
         # through Sopel's rate-limiting machinery
-        func.rate = getattr(func, 'rate', 0)
+        func.user_rate = getattr(func, 'user_rate', 0)
         func.channel_rate = getattr(func, 'channel_rate', 0)
         func.global_rate = getattr(func, 'global_rate', 0)
+        func.user_rate_message = getattr(func, 'user_rate_message', None)
+        func.channel_rate_message = getattr(func, 'channel_rate_message', None)
+        func.global_rate_message = getattr(func, 'global_rate_message', None)
+        func.default_rate_message = getattr(func, 'default_rate_message', None)
         func.unblockable = getattr(func, 'unblockable', False)
 
     if not is_triggerable(func) and not is_url_callback(func):
--- a/sopel/plugin.py
+++ b/sopel/plugin.py
@@ -40,6 +40,9 @@ __all__ = [
     'output_prefix',
     'priority',
     'rate',
+    'rate_user',
+    'rate_channel',
+    'rate_global',
     'require_account',
     'require_admin',
     'require_bot_privilege',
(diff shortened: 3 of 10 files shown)


sopel/bot.py
```diff
@@ -587,32 +587,47 @@ def call_rule(
         sopel: 'SopelWrapper',
         trigger: Trigger,
     ) -> None:
+        nick = trigger.nick
+        sender = trigger.sender
```
dgw, line 591: As an aside, I've come to dislike `trigger.sender` as an attribute name, because it sounds like that should be the `nick` who triggered the command. While you're creating this variable and updating usage of the attribute below, can this be called `channel` instead?

sopel/plugin.py
```diff
@@ -820,31 +823,168 @@ def add_attribute(function):
...
-                        the same channel, regardless of triggering user
-    :param int server: seconds between permitted calls of this function no
-                       matter who triggered it or where
+    :param user: seconds between permitted calls of this function by the same
+                 user
+    :param channel: seconds between permitted calls of this function in the
+                    same channel, regardless of triggering user
+    :param server: seconds between permitted calls of this function no matter
+                   who triggered it or where
+    :param message: optional keyword argument; default message send as notice
```
dgw, line 840:
```suggestion
    :param message: optional keyword argument; default message sent as NOTICE
```

sopel/plugin.py
```diff
@@ -820,31 +823,168 @@ def add_attribute(function):
...
+        Optional keyword argument ``message`` was added in Sopel 8.
+
+    .. note::
+
+        Users on the admin list in Sopel’s configuration are exempted from rate
+        limits.
+
+    .. seealso::
+
+        You can control each rate limit separatly, with their own custom
```
dgw, line 882:
```suggestion
        You can control each rate limit separately, with their own custom
```

sopel/plugin.py
```diff
@@ -820,31 +823,168 @@ def add_attribute(function):
...
+
+def rate_user(
+    rate: int,
+    message: typing.Optional[str] = None,
+) -> typing.Callable:
+    """Decorate a function to be rate-limited for a user.
+
+    :param rate: seconds between permitted calls of this function by the same
+                 user
+    :param message: optional; message send as notice when a user hits the limit
```
dgw, line 907:
```suggestion
    :param message: optional; message sent as NOTICE when a user hits the limit
```

sopel/plugin.py
```diff
@@ -820,31 +823,168 @@ def add_attribute(function):
...
+
+def rate_channel(
+    rate: int,
+    message: typing.Optional[str] = None,
+) -> typing.Callable:
+    """Decorate a function to be rate-limited for a channel.
+
+    :param rate: seconds between permitted calls of this function in the same
+                 channel, regardless of triggering user
+    :param message: optional; message send as notice when a user hits the limit
```
dgw, line 938:
```suggestion
    :param message: optional; message sent as NOTICE when a user hits the limit
```

sopel/plugin.py
```diff
@@ -820,31 +823,168 @@ def add_attribute(function):
...
+    will always take precedence::
+
+        @rate(10, 10, 10)
+        @rate_channel(5, 'You hit the channel rate limit for this function.')
+            # channel limit will be set to 5, other to 10
+            # will send a NOTICE only when a user hits the channel limit
+            # as other rate limit don't have any message set
+
+    If you don't provide a message, the default message set (if any) by
+    :func:`rate` will be used instead.
```
dgw, line 950: I have not applied this to every instance, as I'm not 100% sure the line-wrapping aligns with the standard, so I'm leaving it to you to tweak this if necessary. 😸

```suggestion
    If you don't provide a message, the default message set by :func:`rate`
    (if any) will be used instead.
```

sopel/plugin.py
```diff
@@ -820,31 +823,168 @@ def add_attribute(function):
...
+
+def rate_server(
+    rate: int,
+    message: typing.Optional[str] = None,
+) -> typing.Callable:
+    """Decorate a function to be rate-limited for a channel.
+
+    :param rate: seconds between permitted calls of this function no matter who
+                 triggered it or where
+    :param message: optional; message send as notice when a user hits the limit
```
dgw, line 969:
```suggestion
    :param message: optional; message sent as NOTICE when a user hits the limit
```

sopel/plugin.py
```diff
@@ -820,31 +823,168 @@ def add_attribute(function):
...
+        function.channel_rate_message = message
+        return function
+    return add_attribute
+
+
+def rate_server(
+    rate: int,
+    message: typing.Optional[str] = None,
+) -> typing.Callable:
+    """Decorate a function to be rate-limited for a channel.
```
dgw, line 965: Not channel. This is `rate_server`.

Also worth considering if this should be called `rate_global` to match the internal terminology. References to this type of limit in `plugins.rules` also appear to call this "global".

sopel/plugin.py
```diff
@@ -820,31 +823,168 @@ def add_attribute(function):
...
+    :param message: optional; message send as notice when a user hits the limit
+
+    This decorator can be used alone or with the :func:`rate` decorator, as it
+    will always take precedence::
+
+        @rate(10, 10, 10)
+        @rate_server(5, 'You hit the server rate limit for this function.')
+            # server limit will be set to 5, other to 10
+            # will send a NOTICE only when a user hits the servr limit
+            # as other rate limit don't have any message set
```
dgw, line 978:
```suggestion
            # as other rate limits don't have any message set
```

sopel/plugin.py
```diff
@@ -820,31 +823,168 @@ def add_attribute(function):
...
+    :param message: optional; message send as notice when a user hits the limit
+
+    This decorator can be used alone or with the :func:`rate` decorator, as it
+    will always take precedence::
+
+        @rate(10, 10, 10)
+        @rate_channel(5, 'You hit the channel rate limit for this function.')
+            # channel limit will be set to 5, other to 10
+            # will send a NOTICE only when a user hits the channel limit
+            # as other rate limit don't have any message set
```
dgw, line 947:
```suggestion
            # as other rate limits don't have any message set
```

sopel/plugin.py
```diff
@@ -820,31 +823,168 @@ def add_attribute(function):
...
+    :param message: optional; message send as notice when a user hits the limit
+
+    This decorator can be used alone or with the :func:`rate` decorator, as it
+    will always take precedence::
+
+        @rate(10, 10, 10)
+        @rate_user(20, 'You hit your rate limit for this function.')
+            # user limit will be set to 20, other to 10
+            # will send a NOTICE only when a user hits their own limit
+            # as other rate limit don't have any message set
```
dgw, line 916:
```suggestion
            # as other rate limits don't have any message set
```

sopel/plugins/rules.py
```diff
@@ -667,31 +667,52 @@ def is_unblockable(self) -> bool:
...
 
         :return: ``True`` when the rule reached the limit, ``False`` otherwise
-        :rtype: bool
+        """
+
+    @abc.abstractmethod
+    def get_user_rate_message(self, nick: tools.Identifier) -> Optional[str]:
+        """Give the message to send with a NOTICE to ``nick``.
+
+        :return: A formatted string or ``None`` if no message is set.
```
dgw, line 697:
```suggestion
        :return: A formatted string, or ``None`` if no message is set.
```

sopel/plugins/rules.py
```diff
@@ -667,31 +667,52 @@ def is_unblockable(self) -> bool:
...
+        """Give the message to send with a NOTICE to ``nick`` or ``channel``.
+
+        :return: A formatted string or ``None`` if no message is set.
+        """
+
+    @abc.abstractmethod
+    def get_global_rate_message(self) -> Optional[str]:
+        """Give the message to send with a NOTICE.
+
+        :return: A formatted string or ``None`` if no message is set.
```
dgw, line 715:
```suggestion
        :return: A formatted string, or ``None`` if no message is set.
```

sopel/plugins/rules.py
```diff
@@ -667,31 +667,52 @@ def is_unblockable(self) -> bool:
...
+
+    @abc.abstractmethod
+    def get_channel_rate_message(
+        self,
+        nick: tools.Identifier,
+        channel: tools.Identifier,
+    ) -> Optional[str]:
+        """Give the message to send with a NOTICE to ``nick`` or ``channel``.
+
+        :return: A formatted string or ``None`` if no message is set.
```
dgw, line 708:
```suggestion
        :return: A formatted string, or ``None`` if no message is set.
```

sopel/plugins/rules.py
```diff
@@ -783,6 +804,13 @@ def kwargs_from_callable(cls, handler):
             'rate_limit': getattr(handler, 'rate', 0),
             'channel_rate_limit': getattr(handler, 'channel_rate', 0),
             'global_rate_limit': getattr(handler, 'global_rate', 0),
+            'user_rate_message': getattr(handler, 'user_rate_message', None),
```
dgw, line 807: You know you want to, for consistency. 😏

```suggestion
            'user_rate_message': getattr(
                handler, 'user_rate_message', None),
```

sopel/plugins/rules.py
```diff
@@ -1068,6 +1104,22 @@ def is_global_rate_limited(self):
...
+    def get_channel_rate_message(self, nick, channel):
+        if self._channel_rate_message is not None:
+            return self._channel_rate_message.format(
+                nick=nick,
+                channel=channel,
+            )
+        return self._default_rate_message
+
+    def get_global_rate_message(self):
+        return self._global_rate_message or self._default_rate_message
```
dgw, line 1121: Since there's already some support for placeholders, what do you think of adding one for the time remaining until the rate limit expires? It's the main reason that _I_ wished for this feature to be built in, all those years ago.

Regardless of the above, these placeholders don't appear to be documented anywhere.

dgw (commented):
I haven't dived into the tests yet (there are a lot of them), but there might be enough to discuss already without them.

sopel/bot.py
Exirel: The thing is: it could be a nick or a channel. So on one hand I agree with you, I don't like "sender", but on the other `channel` is not always accurate either.

Something like `origin` or `source`? `maybe_channel_maybe_nick`? :grin:

sopel/plugin.py
Exirel: Yeah, I know, I didn't change the internal terminology because I'm not sure it's the right term or not, so I've only the decorator name for now. What do you prefer? Any opinion? I can change that to `rate_global` or change internals to `server`.

(the docstring is a bad copy/paste)

sopel/plugins/rules.py
Exirel: Yup I thought about it. I wanted to put the bulk of the code first for review, it would be easy to add anything after that.

sopel/plugin.py
dgw: I would not at all mind calling the decorator `rate_global` instead. Somehow `rate_server` makes it sound like the server (IRCd) will be calling commands all on its own… Doesn't make sense, but that's what it sounds like!

sopel/bot.py
dgw: `context`?

sopel/plugin.py
Exirel: Done!

sopel/bot.py
Exirel: Done!

sopel/bot.py
```diff
@@ -587,32 +587,48 @@ def call_rule(
         sopel: 'SopelWrapper',
         trigger: Trigger,
     ) -> None:
+        nick = trigger.nick
+        context = trigger.sender
+        is_channel = context and not trigger.is_privmsg
```
dgw, line 592: This interface looks weird when you know that `context` already knows if it's a channel (`and not context.is_nick()`). Is it worth suggesting a change? No. But it still made me stop and think.

sopel/bot.py
```diff
@@ -587,32 +587,48 @@ def call_rule(
...
+
         # rate limiting
         if not trigger.admin and not rule.is_unblockable():
-            if rule.is_rate_limited(trigger.nick):
+            if rule.is_rate_limited(nick):
```
dgw, line 596: Since we've documented the Rule internals as subject to change without warning, any thoughts on taking this opportunity to rename this method as `rule.is_user_rate_limited` so it's parallel to its `get_user_rate_message` counterpart?

I'm not sure yet if there would be any utility in a replacement `rule.is_rate_limited` that could check multiple types at once, but that would be the logical choice.

sopel/bot.py
```diff
@@ -587,32 +587,48 @@ def call_rule(
...
             if rule.is_global_rate_limited():
+                message = rule.get_global_rate_message()
+                if message:
+                    sopel.notice(message, destination=nick)
                 return
 
         # channel config
-        if trigger.sender in self.config:
-            channel_config = self.config[trigger.sender]
+        if context in self.config:
```
dgw, line 615: Should this be `if is_channel and context in self.config`? I realize it's already possible to make a "Channel configuration" for some specific nick because of how this works, but the question is whether that's a bug/needs changing. Triggering this section of code in PM from someone whose nick matches a plugin's section name, for instance, feels squicky.

sopel/loader.py
```diff
@@ -51,6 +51,10 @@ def clean_callable(func, config):
         func.rate = getattr(func, 'rate', 0)
```
dgw, line 51: Is it time to remove the `rate` attribute from callables and replace it with `func.user_rate`, for consistency?

Do we even have to consider the actual attributes of a plugin callable as part of the API? I hope not, because this is not something I want to go through a deprecation cycle for.

sopel/plugin.py
```diff
@@ -820,31 +823,168 @@ def add_attribute(function):
     return add_attribute
 
 
-def rate(user: int = 0, channel: int = 0, server: int = 0) -> typing.Callable:
+def rate(
+    user: int = 0,
+    channel: int = 0,
+    server: int = 0,
+    *, message: typing.Optional[str] = None,
```
dgw, line 830: Looks weird. Not that the alternatives (`*,` at the end of the previous line, or on its own line) look great either, but let me try this on for size:

```suggestion
    *,
    message: typing.Optional[str] = None,
```

(Nope, it still looks weird. Damn it.)

sopel/plugin.py
```diff
@@ -820,31 +823,168 @@ def add_attribute(function):
...
     not work properly.
+
+    .. versionchanged:: 8.0
+
+        Optional keyword argument ``message`` was added in Sopel 8.
+
+    .. note::
+
+        Users on the admin list in Sopel’s configuration are exempted from rate
+        limits.
```
dgw, line 878: Maybe include this note on all related decorators? Even those of us who work in this codebase on a regular basis keep forgetting about it. 😁

sopel/plugin.py
```diff
@@ -820,31 +823,168 @@ def add_attribute(function):
...
+    will always take precedence::
+
+        @rate(10, 10, 10)
+        @rate_user(20, 'You hit your rate limit for this function.')
+            # user limit will be set to 20, other to 10
+            # will send a NOTICE only when a user hits their own limit
+            # as other rate limits don't have any message set
+
+    If you don't provide a message, the default message set (if any) by
+    :func:`rate` will be used instead.
```
dgw, line 919: `rate_channel`'s format wins.

```suggestion
    If you don't provide a message, the default message set by :func:`rate`
    (if any) will be used instead.
```

sopel/plugins/rules.py
dgw: I also just noticed that the global and default versions don't support even a `nick` placeholder. 🤔

test/plugins/test_plugins_rules.py
```diff
@@ -1083,6 +1083,13 @@ def handler(wrapped, trigger):
     assert 'threaded' in kwargs
     assert 'output_prefix' in kwargs
     assert 'unblockable' in kwargs
+    assert 'rate_limit' in kwargs
```
dgw, line 1086: Along with what I said before about renaming the `rate` attr on callables, this kwarg could be changed to `user_rate_limit` too.

Here I am again, dropping a line note in some mostly-unrelated place because GitHub won't let me go put it [where it belongs](https://github.com/Exirel/sopel/blob/66bae33d60c35dd77460347529bf578f00961812/sopel/plugins/rules.py#L902) on code in `sopel/plugins/rules.py` that's too far from the diff.

dgw (changes requested):
Yes, I realize some of this could fall in "future PR" territory. One of us can make a tracking issue(s) once we agree on what's out of scope.

sopel/bot.py
Exirel: I totally agree. I'll go for the rename then.

sopel/bot.py
Exirel: Yeah I know, it's weird. I don't know how to feel about that. I remember that the reminder plugin didn't mind using "sender" even in PM. So I don't know yet.

sopel/plugin.py
Exirel: Ah ah I know right?

sopel/bot.py
Exirel: After double checking, I can use context instead of trigger here.

sopel/bot.py
Exirel: Fixed.

sopel/loader.py
Exirel: And it is done.

Exirel:
I've modified `func.rate` into `func.user_rate` (and many little changes like that).

Also now there are conflicts... I didn't look at them yet.

dgw (approved):
Merge conflicts are from a bit of CTCP stuff and some type-hint tweaks, AFAICT. Shouldn't be a huge deal to fix up. What's important is, this looks ready to merge so our brave `master` testers can try to break it in prod. 😁

Exirel:
> 1) this isn't squashed/rebased

This part: done. I didn't add anything yet, because I think it would be better for you if you wanted to check the diff between what you approved so far, and what will come next (docstrings, placeholders, etc.).

Exirel:
>  Since there's already some support for placeholders, what do you think of adding one for the time remaining until the rate limit expires? It's the main reason that I wished for this feature to be built in, all those years ago.
> 
> Regardless of the above, these placeholders don't appear to be documented anywhere.

I'm about to push a commit with just `nick` for user and global rate limits, `nick` and `channel` for channel rate limit.

For other placeholders, I need to think a bit more about it: can we assume the time left will be in seconds, and the rate limit in seconds as well? Like, if the rate limit is something like 60s/3600s, should we be smart about it? Should we provide a human readable version of that? (Using cool functions from `tools.time` for example?)

This leads me to think that, maybe, I could just commit what I have, and consider this PR "done": anyone could jump into the code and add the new placeholders as they wanted, with a separate/dedicated conversation about it, without all the other diff around.

sopel/plugin.py
```diff
@@ -830,31 +833,211 @@ def add_attribute(function):
...
     not work properly.
+
+    .. versionchanged:: 8.0
+
+        Optional keyword argument ``message`` was added in Sopel 8.
+
+    .. note::
+
+        Users on the admin list in Sopel’s configuration are exempted from rate
+        limits.
```
dgw, line 894: This Unicode single quote should just be a regular old ASCII `'`.

Yes I know it was already there in the old line you cut/copied, but we do like fixing weird stuff whenever we're already touching it. There are other occurrences elsewhere in this file, so I recommend a global find-and-replace in your IDE/editor.

If you're feeling ambitious you can do a quick search through the other files this touches to find any other instances, but I wouldn't expect any.

test/test_bot.py
```diff
@@ -661,8 +661,7 @@ def testrule(bot, trigger):
         plugin='testplugin',
         label='testrule',
         handler=testrule,
-        rate_limit=100,
-        threaded=False,
```
dgw, line 665: Why's this deleted? It came along recently, in a2a8ba0fb65fedfc39b7b7b9d639ba90a5a740df (from #2297), so it's probably important.

dgw (commented):
Looks like a small bit of collateral damage from the rebase happened; see line note.

We can certainly defer adding more kinds of placeholder for a later PR, but I really would like this patch to document the ones that it _does_ add. 🙂

test/test_bot.py
Exirel: Oh! It's because I messed up in #2297 at some point in my development process and it wasn't necessary at all, but I forgot to remove it. It's the equivalent of ` print('debug')` but very sneaky.

Exirel:
Squashed & rebased, as per @dgw's validation on IRC.

dgw (approved):
I should probably test this one last time on a real bot before merging, but that doesn't mean I can't approve it first.

PR #2292, opened 2022-05-29
db: use Text where length not limited

### LINKED ISSUE #2291: nick_values column isn't long enough
### Description
https://github.com/sopel-irc/sopel/blob/master/sopel/db.py#L66 states it should be 255 charecters. That doesn't even fit seen which is from core as that needs at least 512. This should be made higher.

### Reproduction steps
1. Run bot with mysql/MariaDB
2. Say something long3. 
3. See error4. 

### Expected behavior
Column is big enough

### Logs
None yet, I can post if really needed

### Environment
- Sopel `.version`: somewhere between 7 & 8
- Sopel installed via: source
- Python version: 3.9.2
- Operating system: Bullseye
- IRCd `/version`: Libera
- Relevant plugins: core, seen

### Notes
TEXT might be good, moaned on IRC. Will probably make TEXT.

RhinosF1:
Fixes #2291

### Description
What does this do? What problem does it solve? Does it resolve any GitHub issues?

### Checklist

### DIFF
--- a/sopel/db.py
+++ b/sopel/db.py
@@ -7,7 +7,7 @@ import os.path
 import traceback
 import typing
 
-from sqlalchemy import Column, create_engine, ForeignKey, Integer, String
+from sqlalchemy import Column, create_engine, ForeignKey, Integer, String, Text
 from sqlalchemy.engine.url import make_url, URL
 from sqlalchemy.exc import OperationalError, SQLAlchemyError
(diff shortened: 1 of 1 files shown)


RhinosF1:
This doesn't work when primary key is TEXT :(

dgw:
So don't try to make the keys TEXT? If your database _keys_ are over 255 bytes, then you need a therapist 😂

PR #2294, opened 2022-05-29
github: switch to yml templates

half-duplex:
### Description
Switches issue templates to yml so we have forms and markdown instead of just comments.

Are there ways to run Sopel besides pip install, ./setup.py install, or package manager install?
I also set the `needs-triage` label, which we don't currently have but might be helpful.

Examples:
[New issue](https://github.com/half-duplex/sopel/issues/new/choose)
[Result](https://github.com/half-duplex/sopel/issues/2)

### Checklist
- (No code changes) No issues are reported by `make qa` (runs `make quality` and `make test`)


dgw:
We do already have a "Needs Info" label, and a new "Needs Triage" label to go alongside it makes perfect sense to me. How hard is it to set a label containing a space from YAML templates?

PR #2297, opened 2022-06-05
rules: properly track rate limits of rules

### LINKED ISSUE #2295: Rate limiting is not updated until after the handler returns
### Description
A long-running plugin callable with a rate limit can be called again in the same rate-limit context while it is still running. Metrics are not updated [until after a Rule's handler completes](https://github.com/sopel-irc/sopel/blob/36b7c2f1064e7d734e50a66b24270cf7c8949426/sopel/plugins/rules.py#L1071-L1085), because the current metrics also need its exit code, but [rate limiting is checked before dispatching the callable](https://github.com/sopel-irc/sopel/blob/36b7c2f1064e7d734e50a66b24270cf7c8949426/sopel/bot.py#L584-L624).

### Reproduction steps
1. Build a long-running plugin callable ("dummy" way: `time.sleep(10)`)
2. Give the callable a command, let's say `.ratefail`
3. Apply `@plugin.rate(60)` or whatever test value you like
4. Try to call `.ratefail` twice from a non-admin user (admins bypass rate limits)
5. Actual result: `.ratefail` will run twice

### Expected behavior
`.ratefail` cannot be called twice because it is rate limited.

### Environment
- Sopel `.version`: 7, 8-dev, probably earlier
- Sopel installed via: any method
- Python version: any supported
- Operating system: any supported

### Notes
We discussed this on IRC today, and @Exirel already has a plan of attack. Therefore I've assigned the issue to him—but since we already thought there was an open issue about this when there actually wasn't, I thought it was a good idea to make one.

Exirel:
### Description

Fix #2295 by switching from a scalar object (tuple of datetime and return value) to the `RuleMetrics` object, which can tell if the rule is limited or not based on the internal state.

### Checklist

I had to deactivate the "admin" thing locally to test it live, and it works!

### DIFF
--- a/sopel/plugins/rules.py
+++ b/sopel/plugins/rules.py
@@ -24,13 +24,19 @@ import itertools
 import logging
 import re
 import threading
-from typing import Generator, Iterable, Optional, Type, TypeVar
+import typing
 from urllib.parse import urlparse
 
 
 from sopel import tools
 from sopel.config.core_section import (
     COMMAND_DEFAULT_HELP_PREFIX, COMMAND_DEFAULT_PREFIX, URL_DEFAULT_SCHEMES)
+
+
+if typing.TYPE_CHECKING:
+    from typing import Any, Dict, Generator, Iterable, Optional, Type
+
+    from sopel.tools.identifiers import Identifier
 
 
 __all__ = [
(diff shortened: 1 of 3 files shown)


dgw (approved):
This looks simple enough. Do you expect merging this to cause any conflicts in #2290?

Exirel:
Nope, or if any, only very minor and easy to fix conflicts.

dgw:
I :shipit: then. 🚀

PR #2299, opened 2022-06-08
cli: look harder for $PWD/cfg, don't automatically start wizard

half-duplex:
Github won't let me reopen #2093.
+`Bugfix` because it now prints the .cfg path instead of ".sopel/foo"

### Description
This fixes two annoyances:
- Sopel automatically adds `.cfg` when searching the config dir, but it doesn't when searching the current working directory
- If the config file is not found, it currently automatically starts the configuration wizard.

Combining these two, a user (me) will (does, regularly) `cd` to their non-default Sopel config dir, run `sopel -c mybot` expecting it to load mybot.cfg and run, but instead have Sopel create ~/.sopel/ and ~/.sopel/mybot.cfg and start the config wizard.

Searching `$PWD/{config}.{ext}` is straightforward, but not automatically running the wizard is probably subject to opinions. I changed it to the below output, but I would also be content with "Do you want to run the configuration wizard? y/n", or at least not creating the folder/file until the end of the wizard so the user has an opportunity to ^C without having to then `rm -rf ~/.sopel`

Amended "couldn't find config":
```
$ sopel
ConfigurationError: Unable to find the configuration file /home/user/.sopel/default.cfg
If you're just setting up Sopel, welcome! You can use `sopel configure` to get started easily.
```

### Checklist

### DIFF
--- a/sopel/cli/run.py
+++ b/sopel/cli/run.py
@@ -219,18 +219,17 @@ def get_configuration(options):
     .. seealso::
 
        The configuration file is loaded by
-       :func:`~sopel.cli.run.utils.load_settings` or created using the
-       configuration wizard.
+       :func:`~sopel.cli.run.utils.load_settings`
 
     """
     try:
         settings = utils.load_settings(options)
     except config.ConfigurationNotFound as error:
-        print(
-            "Welcome to Sopel!\n"
-            "I can't seem to find the configuration file, "
-            "so let's generate it!\n")
-        settings = utils.wizard(error.filename)
+        raise config.ConfigurationError(
+            "%s\n"
+            "If you're just setting up Sopel, welcome! "
+            "You can use `sopel configure` to get started easily." % error
+        )
 
     settings._is_daemonized = options.daemonize
     return settings
--- a/sopel/cli/utils.py
+++ b/sopel/cli/utils.py
@@ -198,47 +198,55 @@ def enumerate_configs(config_dir, extension='.cfg'):
             yield item
 
 
-def find_config(config_dir, name, extension='.cfg'):
+def find_config(config_dir: str, name: str, extension: str = '.cfg') -> str:
     """Build the absolute path for the given configuration file ``name``.
 
-    :param str config_dir: path to the configuration directory
-    :param str name: configuration file ``name``
-    :param str extension: configuration file's extension (default to ``.cfg``)
-    :return: the path of the configuration file, either in the current
-             directory or from the ``config_dir`` directory
+    :param config_dir: path to the configuration directory
+    :param name: configuration file ``name``
+    :param extension: configuration file's extension (defaults to ``.cfg``)
+    :return: the absolute path to the configuration file, either in the current
+             directory or in the ``config_dir`` directory
 
-    This function tries different locations:
+    This function appends the extension if absent before checking the following:
 
-    * the current directory
-    * the ``config_dir`` directory with the ``extension`` suffix
-    * the ``config_dir`` directory without a suffix
+    * If ``name`` is an absolute path, it is returned whether it exists or not
+    * If ``name`` exists in the ``config_dir``, the absolute path is returned
+    * If ``name`` exists in the current directory, its absolute path is returned
+    * Otherwise, the path to the nonexistent file within ``config_dir`` is returned
 
     Example::
 
         >>> from sopel.cli import utils
         >>> from sopel import config
+        >>> os.getcwd()
+        '/sopel'
         >>> os.listdir()
         ['local.cfg', 'extra.ini']
         >>> os.listdir(config.DEFAULT_HOMEDIR)
-        ['config.cfg', 'extra.ini', 'plugin.cfg', 'README']
+        ['config.cfg', 'extra.ini', 'logs', 'plugins']
         >>> utils.find_config(config.DEFAULT_HOMEDIR, 'local.cfg')
-        'local.cfg'
-        >>> utils.find_config(config.DEFAULT_HOMEDIR, 'local')
-        '/home/username/.sopel/local'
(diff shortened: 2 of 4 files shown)


sopel/cli/run.py
```diff
@@ -255,11 +255,13 @@ def get_configuration(options):
...
-            "I can't seem to find the configuration file, "
-            "so let's generate it!\n")
-        settings = utils.wizard(error.filename)
+        raise config.ConfigurationError(
+            str(error) + (
+                "\n"
+                "If you're just setting up Sopel, welcome! "
+                "You can use `sopel configure` to get started easily."
+            )
+        )
```
Exirel, line 264: No more extra parentheses with `%` (you can replace that with `.format()` if you prefer):

```suggestion
        raise config.ConfigurationError(
            "%s\n"
            "If you're just setting up Sopel, welcome! "
            "You can use `sopel configure` to get started easily." % error
        )
```

sopel/cli/utils.py
```diff
@@ -230,13 +230,10 @@ def find_config(config_dir, name, extension='.cfg'):
         '/home/username/.sopel/extra.ini'
 
     """
+    if not name.endswith(extension):
```
Exirel, line 233: You changed the behavior, but you didn't upgrade the docstring, which says:

```
>>> utils.find_config(config.DEFAULT_HOMEDIR, 'local')
'/home/username/.sopel/local'
```

This must be modified by this:

```
>>> utils.find_config(config.DEFAULT_HOMEDIR, 'local')
'/home/username/.sopel/local.cfg'
```

The test you modified is a proof of behavior change.

sopel/cli/run.py
```diff
@@ -255,11 +255,13 @@ def get_configuration(options):
     try:
         settings = utils.load_settings(options)
     except config.ConfigurationNotFound as error:
-        print(
-            "Welcome to Sopel!\n"
-            "I can't seem to find the configuration file, "
-            "so let's generate it!\n")
-        settings = utils.wizard(error.filename)
```
Exirel, line 262: The docstring still says this:

```rst
The configuration file is loaded by
:func:`~sopel.cli.run.utils.load_settings` or created using the
configuration wizard.
```

This need to be updated.

Exirel (changes requested):
I didn't check all the docstring, but I believe it would be better to check more of them. This is a clear behavior change, one that I agree with, but I really don't like that docstrings and/or documentation have not been updated accordingly. This is user facing, so it is of the upmost importance.

sopel/cli/run.py
half-duplex: I'll just make it an f-string :smile_cat:

sopel/cli/utils.py
half-duplex: The old docstring was also incorrect :upside_down_face: The code would find ./extra.ini before $conf/extra.ini, and the local one would be abspath.

The whole searching pwd thing is kinda weird too... I can guess why we do it, but it feels fucky for anything but sopel .cfgs - should we make it only search pwd when `search_pwd=True` or something, so plugins will only get stuff within the homedir the main sopel config sets?

dgw: > but it feels fucky for anything but sopel .cfgs

Fortunately that's all this function is supposed to be used for. This is `sopel.cli.utils`, not a module we expect plugin authors ever to touch.

sopel/cli/utils.py
```diff
@@ -197,47 +197,54 @@ def enumerate_configs(config_dir, extension='.cfg'):
...
+    :param name: configuration file ``name``
+    :param extension: configuration file's extension (defaults to ``.cfg``)
+    :return: the absolute path to the configuration file, either in the current
+             directory or in the ``config_dir`` directory
 
-    * the current directory
-    * the ``config_dir`` directory with the ``extension`` suffix
-    * the ``config_dir`` directory without a suffix
+    This function appends the extension if absent before checking the following:
+    * If ``name`` is an absolute path, it is returned whether it exists or not
```
Exirel, line 210:
```suggestion
    This function appends the extension if absent before checking the following:

    * If ``name`` is an absolute path, it is returned whether it exists or not
```

Exirel (changes requested):
Just a nitpick. Looks good to me otherwise.

Exirel:
I added the label breaking change because it changes the behavior of an existing feature that was kept as is between Sopel 6.9 and 7.1. I think it's an acceptable change that also need a fair warning to our users in the release note, without needing a whole deprecation cycle.

sopel/cli/utils.py
```diff
@@ -198,47 +198,55 @@ def enumerate_configs(config_dir, extension='.cfg'):
...
-        'local.cfg'
-        >>> utils.find_config(config.DEFAULT_HOMEDIR, 'local')
-        '/home/username/.sopel/local'
+        '/sopel/local.cfg'
         >>> utils.find_config(config.DEFAULT_HOMEDIR, 'config')
         '/home/username/.sopel/config.cfg'
         >>> utils.find_config(config.DEFAULT_HOMEDIR, 'extra', '.ini')
         '/home/username/.sopel/extra.ini'
 
+    .. versionchanged:: 8.0
```
dgw, line 234: **Iff** there are other amendments to this patch, I suggest adding a blank line here for consistency with other docstrings in the project, but it isn't worth running CI again just for this.

```suggestion
    .. versionchanged:: 8.0

```

Exirel (approved):
Same nitpick as dgw regarding the blank line needed after the rst directive. I'm not worried about missing something, the basics are covered nicely with docstrings and tests, as far as I can tell.

dgw:
Through the magic of `git merge --no-ff --no-commit`, I will take care of the whitespace without needing another force-push.

PR #2301, opened 2022-06-09
remove checkstyle.sh and ci_build.sh

half-duplex:
### Description
`checkstyle.sh` is now just a bunch of pointless window-dressing around running `flake8`, so let's get rid of it.

Can I also kill ci_build.sh? Making it resemble the current CI flow would basically be `pip install && make qa`

### Checklist


half-duplex:
Also removed ci_build re: irc conversation.

dgw (approved):
Kinda curious what @HumorBaby would have to say about the Git hooks, but they'll still work. (I just wonder why the hook script doesn't use `make` targets.)

Exirel:
It's possible that the git hook scripts are not in the same path as the `Makefile` file? I don't know.

PR #2303, opened 2022-06-09
cli: remove ssl cert warning

half-duplex:
### Description
I missed this in #2278 :neutral_face:
Since `ca_certs=None` is now valid, the warning is nonsense. If the setting is truly invalid, Sopel will quit with an exception:
```
ValueError: Invalid value for core.ca_certs: Value must be an existing or creatable file.
```
or
```
ssl.SSLError: [X509: NO_CERTIFICATE_OR_CRL_FOUND] no certificate or crl found (_ssl.c:4123)
```
(Though after the latter, it'll still attempt to reconnect after shutting down??)

### Checklist

### DIFF
--- a/sopel/cli/run.py
+++ b/sopel/cli/run.py
@@ -53,10 +53,6 @@ def run(settings, pid_file, daemon=False):
     # Also show the location of the config file used to load settings
     print("\nLoaded config file: {}".format(settings.filename))
 
-    if not settings.core.ca_certs:
-        tools.stderr(
-            'Could not open CA certificates file. SSL will not work properly!')
-
     # Define empty variable `p` for bot
     p = None
     while True:


dgw:
> (Though after the latter, it'll still attempt to reconnect after shutting down??)

Some weird lifecycle stuff happening, maybe. That should be at least [caught](https://github.com/sopel-irc/sopel/blob/e38942ee19768d63af053ddeb0e2cb520955f3f7/sopel/irc/backends.py#L309-L311), but attrs like `hasquit` are still a bit wonky. Good for another patch if you fall into that rabbit hole.

dgw (approved):
I believe #2306 touches some stuff that will fix trying to reconnect after SSL failure, so there's no need to block this on it.

half-duplex:
(Also, this doesn't affect the result, just removes the nonsense message preceding it)

dgw:
(Should it be relabeled as a Tweak or Housekeeping instead of Bugfix?)

half-duplex:
I figure telling the user something outright incorrect counts as a bug. Not quitting is more or less unrelated.

dgw:
Exirel agreeing with me on IRC was enough to push me into "this isn't a bug" territory. (More of a record-keeping thing than any kind of defense. Just didn't like leaving the thread dangling like that after the label chance.)

PR #2306, opened 2022-06-11
irc, memories: fix handling QUITs, and prevent infinite failed-connection retries

Exirel:
### Description

This started with our instance of Sopel running from source with the bot becoming unresponsive after a user quit the server: there was a bug in how `SopelIdentifierMemory` deals with `None in memory`. This caused an unexpected exception at the wrong place, which put the bot in a weird unstable state from which it doesn't recover.

I also fixed the case where an SSL connection error would block the bot in an infinite restart loop.

I'm still not sure the message handling is safe, but at least *this particular cause* is fixed.

### Checklist

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -23,7 +23,9 @@ exclude_lines =
     if 0:
     if False:
     if __name__ == .__main__.:
-    
+    if typing.TYPE_CHECKING:
+    if TYPE_CHECKING:
+
 show_missing = True
 
 [html]
--- a/sopel/plugins/rules.py
+++ b/sopel/plugins/rules.py
@@ -475,7 +475,7 @@ class RuleMetrics:
         time_limit: datetime.datetime,
     ) -> bool:
         """Determine if the rule hits the time limit."""
-        if not self.started_at and not self.ended_at:
+        if not self.started_at:
             # not even started, so not limited
             return False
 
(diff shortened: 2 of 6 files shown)


.coveragerc
```diff
@@ -23,7 +23,9 @@ exclude_lines =
     if 0:
     if False:
     if __name__ == .__main__.:
-    
+    if typing.TYPE_CHECKING:
+    if TYPE_CHECKING:
+
```
dgw, line 28: What is happening with the whitespace here? Shouldn't this change appear as a simple +2/-0 diff?

sopel/plugins/rules.py
```diff
@@ -475,7 +475,7 @@ def is_limited(
         time_limit: datetime.datetime,
     ) -> bool:
         """Determine if the rule hits the time limit."""
-        if not self.started_at and not self.ended_at:
+        if not self.started_at:
```
dgw, line 478: Both of us missed this in #2297, huh?

dgw (approved):
Through the magic of `git checkout master -- sopel/tools/memories.py && make test` I have verified that indeed, the new test coverage does validate a fix. 🚀

.coveragerc
Exirel: I've no idea.

sopel/plugins/rules.py
Exirel: Technically, if `self.started_at` is `None`, then `self.ended_at` should be `None` too. However, this is an unnecessary check that made mypy grumpy (for good reason).

dgw: my*grum*py 😉 got it

.coveragerc
dgw: I figured it out. There was whitespace on the empty line before, which GitHub isn't showing even though I have "Ignore whitespace" turned off in diffs. This being +3/-1 is a good thing.

PR #2307, opened 2022-06-11
url: fix various type error

Exirel:
### Description

Making mypy happy with `url.py`.

### Checklist

### DIFF
--- a/sopel/modules/url.py
+++ b/sopel/modules/url.py
@@ -323,8 +323,15 @@ def title_auto(bot: SopelWrapper, trigger: Trigger):
 
 
 def process_urls(
-    bot: SopelWrapper, trigger: Trigger, urls: List[str], requested: bool = False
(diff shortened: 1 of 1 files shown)


sopel/modules/url.py
```diff
@@ -360,6 +367,11 @@ def process_urls(
         # FIXME: This does nothing when an attacker knows how to host a 302
         # FIXME: This whole concept has a TOCTOU issue
         if not bot.config.url.enable_private_resolution:
+            if not parsed_url.hostname:
```
half-duplex, line 370: I'm guessing this is an unrelated bugfix?

half-duplex (approved):

sopel/modules/url.py
Exirel: Nope, `hostname` is `Optional[str]`, because `file:///path` will have `None` for the hostname, and yes this is a valid URL. However, `ip` requires `str`, so mypy legitimately complains about it.

PR #2310, opened 2022-06-17
tools: fix `get_sendable_message()` with both multibyte chars & whitespace

dgw:
### Description
Fixes a bug reported on IRC by deadyeti, where text containing both emoji and spaces (tweets, in their case) was not split correctly in `tools.get_sendable_message()`, resulting in the second message containing only one word.

I also added test cases (some as regression insurance) and tweaked the function documentation as long as I was touching it.

h/t to @half-duplex for jumping straight into the code and finding the root cause while I was still working on a test case to reproduce the bug. 😅

### Checklist

### RFC
Is this important enough to prompt a maintenance release? We haven't been planning to cut any more package versions off the 7.1.x series, in favor of focusing everyone's time and attention on finishing 8.0.

@sopel-irc/rockstars please weigh in if you have Thoughts.

### DIFF
--- a/sopel/tools/__init__.py
+++ b/sopel/tools/__init__.py
@@ -111,7 +111,7 @@ def get_sendable_message(text, max_length=400):
             unicode_max_length = unicode_max_length - 1
         else:
             # Split at the last best space found
-            excess = text[last_space:]
+            excess = text[last_space:] + excess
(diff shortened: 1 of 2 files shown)


Exirel (approved):
Yeah OK, good one.

PR #2313, opened 2022-06-27
plugins.handlers: override get_version() for EntryPointPlugin

dgw:
### Description
The `_module` of an entry point plugin is unlikely to have its own `__version__`. What's important is the version of the package from which that entry point was loaded.

### Checklist

### Notes
I haven't written a test for this (yet?). Am not even sure if it should call the superclass version (`PyModulePlugin.get_version()`) somewhere as a fallback. Doing that doesn't seem very useful, but I'm open to others' thoughts.

### DIFF
--- a/sopel/plugins/handlers.py
+++ b/sopel/plugins/handlers.py
@@ -591,9 +591,9 @@ class EntryPointPlugin(PyModulePlugin):
         :return: the plugin's version string
         :rtype: Optional[str]
         """
-        version: Optional[str] = None
+        version: Optional[str] = super().get_version()
 
-        if hasattr(self._module, "__package__"):
+        if version is None and hasattr(self._module, "__package__"):
             try:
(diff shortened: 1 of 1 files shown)


dgw:
What do you know? The answer was stupid-simple. Still haven't answered the question about calling the superclass's handler, but at least the broken test works now. In real usage, I doubt we'll ever see an entrypoint plugin that doesn't come from a package (as the one in our test suite does), so I haven't tried to test the output. (Should just be "unknown" like a regular module with no `__version__`.)

dgw:
I answered my own `super()` question. Really ready for review now.

Exirel (approved):
It's OK, I think it should work fine, and if it doesn't there is enough try/except for me to feel confident about this.

I'll add a mental note to write tests for this later.

dgw:
Apparently I forgot to squash this before merging. RIP "Only Checkmarks in the History" goals.

PR #2315, opened 2022-07-12
wikipedia: use say() instead of reply() when reporting a URL error

SnoopJeDi:
### Description

This PR removes user highlights `bot.reply()` from the `wikipedia` plugin, in favor of the non-highlighting `bot.say()`

Note: this PR is motivated by a personal opinion about netiquette; a user sharing a Wikipedia URL may not even know the Sopel instance exists, and the bot highlighting them is the kind of thing I consider a bit of an aggravation. Feel free to close this PR on that basis :sweat_smile: 

### Existing behavior
```
16:33 <SnoopJ> https://en.wikipedia.org/wiki/Diffraction_spike#/media/File:Comparison_strut_diffraction_spikes.svg
16:33 <terribot> SnoopJ: Error fetching section "/media/File:Comparison_strut_diffraction_spikes.svg" for page "Diffraction spike".
```

### Behavior with this patch
```
16:34 <SnoopJ> https://en.wikipedia.org/wiki/Diffraction_spike#/media/File:Comparison_strut_diffraction_spikes.svg
16:34 <terribot> [wikipedia] Error fetching section "/media/File:Comparison_strut_diffraction_spikes.svg" for page "Diffraction spike".
```

### Checklist

### DIFF
--- a/sopel/modules/wikipedia.py
+++ b/sopel/modules/wikipedia.py
@@ -173,7 +173,7 @@ def mw_search(server, query, num):
     return None
 
 
-def say_snippet(bot, trigger, server, query, show_url=True):
+def say_snippet(bot, trigger, server, query, show_url=True, commanded=False):
     page_name = query.replace('_', ' ')
     query = quote(query.replace(' ', '_'))
     url = 'https://{}/wiki/{}'.format(server, query)
@@ -188,8 +188,11 @@ def say_snippet(bot, trigger, server, query, show_url=True):
         # see https://github.com/sopel-irc/sopel/issues/2259
         snippet = re.sub(r"\s+", " ", snippet)
     except KeyError:
-        if show_url:
-            bot.reply("Error fetching snippet for \"{}\".".format(page_name))
+        msg = 'Error fetching snippet for "{}".'.format(page_name)
+        if commanded:
+            bot.reply(msg)
+        else:
+            bot.say(msg)
         return
 
     msg = '{} | "{}'.format(page_name, snippet)
@@ -223,7 +226,7 @@ def say_section(bot, trigger, server, query, section):
 
     snippet = mw_section(server, query, section)
     if not snippet:
-        bot.reply("Error fetching section \"{}\" for page \"{}\".".format(section, page_name))
+        bot.say('Error fetching section "{}" for page "{}".'.format(section, page_name))
         return
 
     msg = '{} - {} | "{}"'.format(page_name, section.replace('_', ' '), snippet)
@@ -309,7 +312,7 @@ def wikipedia(bot, trigger):
         return plugin.NOLIMIT
     else:
         query = query[0]
-    say_snippet(bot, trigger, server, query)
+    say_snippet(bot, trigger, server, query, commanded=True)
 
 
 @plugin.command('wplang')


dgw:
I think IRC discussion yesterday established that this is desirable for URL-triggered calls only. A command that results in an error should still `reply()`.

See the [xkcd](https://github.com/sopel-irc/sopel/blob/365cad7d128b31b97c463445b1853de82c6c3f84/sopel/modules/xkcd.py#L132) and [reddit](https://github.com/sopel-irc/sopel/blob/9279220a348e7d78a8662439242441d64b49c6a9/sopel/modules/reddit.py#L335) plugins for examples of other core plugins that respond differently depending on trigger type.

sopel/modules/wikipedia.py
```diff
@@ -189,7 +189,11 @@ def say_snippet(bot, trigger, server, query, show_url=True):
         snippet = re.sub(r"\s+", " ", snippet)
     except KeyError:
         if show_url:
-            bot.reply("Error fetching snippet for \"{}\".".format(page_name))
+            msg = 'Error fetching snippet for "{}".'.format(page_name)
```
SnoopJeDi, line 192: I was going to replace some of these simpler `str.format()` calls with f-strings (which should be fine since Sopel is 3.7+ now?) but I notice that the library doesn't use them anywhere else.

```suggestion
            msg = f'Error fetching snippet for "{page_name}".'
```

sopel/modules/wikipedia.py
```diff
@@ -223,7 +227,7 @@ def say_section(bot, trigger, server, query, section):
 
     snippet = mw_section(server, query, section)
     if not snippet:
-        bot.reply("Error fetching section \"{}\" for page \"{}\".".format(section, page_name))
+        bot.say('Error fetching section "{}" for page "{}".'.format(section, page_name))
```
SnoopJeDi, line 230: see https://github.com/sopel-irc/sopel/pull/2315/files#r920725745

```suggestion
        bot.say(f'Error fetching section "{section}" for page "{page_name}".')
```

sopel/modules/wikipedia.py
SnoopJeDi: Clarified on IRC: this is a stylistic choice!

sopel/modules/wikipedia.py
```diff
@@ -189,7 +189,11 @@ def say_snippet(bot, trigger, server, query, show_url=True):
         snippet = re.sub(r"\s+", " ", snippet)
     except KeyError:
         if show_url:
-            bot.reply("Error fetching snippet for \"{}\".".format(page_name))
+            msg = 'Error fetching snippet for "{}".'.format(page_name)
+            if commanded:
+                bot.reply(msg)
+            else:
+                bot.say(msg)
         return
```
dgw, line 197: This logic doesn't entirely make sense. While there's no code path to call `say_snippet()` with `..., show_url=False, commanded=True` _right now_, there might be in the future.

Given the addition of `commanded` as an explicit parameter, there's no reason any more to abuse the `show_url` argument's value for inferring whether to output an error or stay silent.

dgw:
This is good to squash down, IMO. Note our commit message convention, which would go something like this for this PR:

`wikipedia: use say() instead of reply() when reporting a URL error`

dgw (approved):
No one told you I prefer _not_ to actually rebase onto `master` during squash if possible, because it makes verifying that the PR contains no extra changes easier, but this patch is small enough that it's nbd.

PR #2317, opened 2022-07-13
implement alembic to handle database migrations

RustyBower:
### Description
Implement Alembic to handle current and future database migrations

### Checklist

### DIFF
--- a/alembic.ini
+++ b/alembic.ini
@@ -1,0 +1,104 @@
+# A generic, single database configuration.
+
+[alembic]
+# path to migration scripts
+script_location = alembic
+
+# template used to generate migration file names; The default value is %%(rev)s_%%(slug)s
+# Uncomment the line below if you want the files to be prepended with date and time
+# see https://alembic.sqlalchemy.org/en/latest/tutorial.html#editing-the-ini-file
+# for all available tokens
+# file_template = %%(year)d_%%(month).2d_%%(day).2d_%%(hour).2d%%(minute).2d-%%(rev)s_%%(slug)s
+
+# sys.path path, will be prepended to sys.path if present.
+# defaults to the current working directory.
+prepend_sys_path = .
+
+# timezone to use when rendering the date within the migration file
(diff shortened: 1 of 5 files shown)


RustyBower:
I do need to figure out how to automatically run Alembic for new versions of Sopel 🤔

dgw:
That'd be cool. I'm not clear on how this works with packaged installs vs. being installed from Git, because the version appears based on commit SHAs, but also haven't read this whole patch yet (nor read Alembic docs).

Maybe later when I'm not out and on a phone. 😁

Exirel:
As said on IRC, I'm sorry I've to say no, because I've been there, I've done the work on previous project so I know what it means. However when we want migrations in Sopel:

* it needs to be a sopel command line, allowing to select a different configuration file
* it needs to allow us to generate new migrations that can be packaged
* it needs to allow Sopel's owner to run migrations that was packaged
* it needs to allow plugin's author to generate new migrations for their plugin (in that case, maybe only packaged version?)
* it needs to collect plugin's migrations to run them alongside Sopel's own migrations

Basically, we don't need the default behavior of Alembic, which is to run for a project you 100% control, we need a system closer to what Django's ORM support: generate migrations for your project, package it, then collect all migrations from the project and its installed plugin to run them. And trust me, I know it's not easy, I had to done it once a few years ago and it wasn't fun at all.

As a side note, the migration needs to have a complete downgrade handler, this one is incomplete.

RustyBower:
There's definitely a way to package this within the application itself to migrate on boot / startup, since this is functionality that Flask-Migrate has, so it must be possible.

Exirel:
I know there is a way: I've done it. I also know this PR is not the way.

RustyBower:
I'm open to other suggestions

dgw:
This seems to have well and truly stalled. Since it didn't seem to be the right approach anyway, let's take any future discussion back to #2293. I'll copy over @Exirel's list of requirements.

PR #2320, opened 2022-07-14
irc: configure anti-looping system

### LINKED ISSUE #1618: Allow the disabling of repetition detection
It would be useful for bot owners to be able to disable the repetition detection feature ("...").

We use Sopel primarily for devops and moderation, with modules which will often return repetitive answers (i.e. checking IPs or e-mails against blocklists), and thus far have had to fork Sopel to remove the "..." responses.

Other rate limits throughout Sopel are configurable (for instance, `@rate`); it would make sense to bring this in line with them, be it through a config option or a per-callable option to disable the feature.

Additionally, this would be very helpful for debugging modules.

Exirel:
### Description

Tin, fix #1618.

The feature was surprisingly easy to implement, this is a draft because I want to make sure I've the right names before writing more documentation.

### Checklist

### DIFF
--- a/docs/source/configuration.rst
+++ b/docs/source/configuration.rst
@@ -333,8 +333,8 @@ policy.
 
 .. versionadded:: 7.0
 
-    Additional configuration options: ``flood_burst_lines``, ``flood_empty_wait``,
-    and ``flood_refill_rate``.
+    Additional configuration options: ``flood_burst_lines``,
+    ``flood_empty_wait``, and ``flood_refill_rate``.
 
 .. versionadded:: 7.1
 
@@ -349,6 +349,79 @@ policy.
     ``@dgw`` said once about Sopel's flood protection logic:
 
         *"It's some arcane magic from AT LEAST a decade ago."*
+
+Loop prevention
+---------------
+
+In order to prevent the bot from entering a loop (for example when there is
+another bot in the same channel, or if a user spams a command), it'll try to
+see if the next message to send is repeating too often in a short time period.
+If that happens, the bot will send ``...`` a few times before remaining silent::
+
+    <bot> I repeat myself!
+    <bot> I repeat myself!
+    <bot> I repeat myself!
+    <bot> I repeat myself!
+    <bot> I repeat myself!
+    # wanted to say: "I repeat myself"
+    <bot> ...
+    # wanted to say: "I repeat myself"
+    <bot> ...
+    # wanted to say: "I repeat myself"
+    <bot> ...
+    # silence, wanted to say: "..." instead of "I repeat myself"
+
+This doesn't affect non-repeating messages, and if enough time has passed
+between now and the last message sent, the loop prevention won't be triggered.
+
+This behavior can be configured with:
+
+* :attr:`~CoreSection.antiloop_threshold`: the number of repeating messages
+  before triggering the loop prevention.
+* :attr:`~CoreSection.antiloop_silent_after`: how many times the bot will send
+  the repeat text until it remains silent.
+* :attr:`~CoreSection.antiloop_window`: how much time (in seconds) since the
+  last message must pass before ignoring the loop prevention.
+* :attr:`~CoreSection.antiloop_repeat_text`: the text used to replace repeating
+  messages (default to ``...``).
+
+For example this configuration::
+
+    [core]
+    antiloop_threshold = 2
+    antiloop_silent_after = 1
+    antiloop_window = 60
+    antiloop_repeat_text = Ditto.
+
+will activate the loop prevention feature if there are at least 2 messages
+in the last 60 seconds, **and** exactly 2 of those messages are the same.
+After sending ``...`` *once* (a third message), the bot will remain silent.
+
+This doesn't affect other messages, i.e. messages that don't repeat::
+
+    <bot> I repeat myself!
+    <bot> No I don't!
+    <bot> I can talk.
+    <bot> I repeat myself!
+    <bot> No I don't!
+    # wanted to say: "I repeat myself"
+    <bot> Ditto.
+    # silence, wanted to say: "Ditto." instead of "No I don't!"
+    <bot> This message is unique.
+
+You can **deactivate** the loop prevention by setting
+:attr:`~CoreSection.antiloop_threshold` to 0.
+
+.. versionadded:: 8.0
+
+    The loop prevention feature wasn't configurable before Sopel 8.0. The
+    new configuration options are: ``antiloop_threshold``,
+    ``antiloop_silent_after``, and ``antiloop_window``.
+
+.. note::
+
+    Since Sopel remembers only the last ten messages, it will use the minimum
+    value between ``antiloop_threshold`` and ten.
 
 Perform commands on connect
 ---------------------------
--- a/sopel/config/core_section.py
+++ b/sopel/config/core_section.py
@@ -133,6 +133,92 @@ class CoreSection(StaticSection):
     :func:`~sopel.plugin.nickname_command`.
     """
 
+    antiloop_repeat_text = ValidatedAttribute(
+        'antiloop_repeat_text', default='...')
+    """The replacement text sent when detecting a repeated message.
+
+    :default: ``...``
+
+    This is equivalent to the default value:
+
+    .. code-block:: ini
+
+        antiloop_repeat_text = ...
+
+    .. seealso::
+
+        The :ref:`Loop Prevention` chapter to learn what each antiloop-related
+        setting does.
+
+    .. versionadded:: 8.0
+    """
+
+    antiloop_silent_after = ValidatedAttribute(
+        'antiloop_silent_after', int, default=3)
+    """How many times the anti-looping message will be sent before stopping.
+
+    :default: ``3``
+
+    This is equivalent to the default value:
+
+    .. code-block:: ini
+
+        antiloop_silent_after = 3
+
+    .. seealso::
+
+        The :ref:`Loop Prevention` chapter to learn what each antiloop-related
+        setting does.
+
+    .. versionadded:: 8.0
+    """
+
+    antiloop_threshold = ValidatedAttribute(
+        'antiloop_threshold', int, default=5)
+    """How many times a message can be repeated without anti-looping action.
+
+    :default: ``5``
+
+    This is equivalent to the default value:
+
+    .. code-block:: ini
+
+        antiloop_threshold = 5
+
+    You can deactivate the anti-looping feature (not recommended) by setting
+    this to ``0``:
+
+    .. code-block:: ini
+
+        antiloop_threshold = 0
+
+    .. seealso::
+
+        The :ref:`Loop Prevention` chapter to learn what each antiloop-related
+        setting does.
+
+    .. versionadded:: 8.0
+    """
+
+    antiloop_window = ValidatedAttribute('antiloop_window', int, default=120)
+    """The time period (in seconds) checked when detecting repeated messages.
+
+    :default: ``120``
+
+    This is equivalent to the default value:
+
+    .. code-block:: ini
+
+        antiloop_window = 120
+
+    .. seealso::
+
+        The :ref:`Loop Prevention` chapter to learn what each antiloop-related
+        setting does.
+
+    .. versionadded:: 8.0
+    """
+
     auth_method = ChoiceAttribute('auth_method', choices=[
         'nickserv', 'authserv', 'Q', 'sasl', 'server', 'userserv'])
     """Simple method to authenticate with the server.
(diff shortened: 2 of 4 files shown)


sopel/config/core_section.py
```diff
@@ -133,6 +133,58 @@ class CoreSection(StaticSection):
...
+    .. code-block:: ini
+
+        anti_looping_repeat = 3
+
+    .. versionadded:: 8.0
+    """
+
+    anti_looping_timeframe = ValidatedAttribute(
+        'anti_looping_timeframe', int, default=120)
+    """The timeframe used to activate anti-looping.
```
dgw, line 175: Needs to mention a unit (seconds) in the docs, when you flesh them out.

dgw (commented):
I keep forgetting that people ask about that, even though we _do_ have an issue about it. 😅 Thanks for just sitting down and writing the patch. Few general thoughts:

- How about `antiloop_` as a prefix for these options? Though I'm all ears for something that fits a bit better with `flood_`.
  - As ideas for new names:
    - `anti_looping` -> `antiloop_threshold`
    - `anti_looping_repeat` -> `antiloop_silent_after`
    - `anti_looping_timeframe` -> `antiloop_window`
- With the other options already set, why not add a way to override the `...` message? Maybe someone would like Sopel to say "Ditto", for example.

Exirel:
And I even added tests.

docs/source/configuration.rst
```diff
@@ -350,6 +350,74 @@ policy.
 
         *"It's some arcane magic from AT LEAST a decade ago."*
 
+Loop prevention
+---------------
+
+In order to prevent the bot from entering a loop (for example when there is
+another bot in the same channel, or if a user spams a command), it'll try to
+see if the next message to send is repeating too often in the last eight
+messages in the last few minutes. If that happens, the bot will send ``...``
```
dgw, line 359: Here's a thought: Why both a time _and_ message-count threshold? With a configurable `antiloop_window` and `antiloop_threshold`, I can't think of any reason to hard-code looking at only the last 8 messages.

docs/source/configuration.rst
```diff
@@ -350,6 +350,74 @@ policy.
...
+between now and the last message sent, the loop prevention won't be triggered.
+
+This behavior can be configured with:
+
+* :attr:`~CoreSection.antiloop_threshold`: the number of repeating messages
+  before triggering the loop prevention.
+* :attr:`~CoreSection.antiloop_silent_after`: how many times the bot will send
+  the repeat text until it remains silent.
+* :attr:`~CoreSection.antiloop_window`: how much times (in seconds) since the
+  last message must pass before ignoring the loop prevention.
```
dgw, line 385:
```suggestion
* :attr:`~CoreSection.antiloop_window`: how much time (in seconds) since the
  last message must pass before ignoring the loop prevention.
```

First nemesi**S** in a long time. 😮

docs/source/configuration.rst
```diff
@@ -350,6 +350,74 @@ policy.
...
+For example this configuration::
+
+    [core]
+    antiloop_threshold = 2
+    antiloop_silent_after = 1
+    antiloop_window = 60
+    antiloop_repeat_text = Ditto.
+
+will activate the loop prevention feature if there is at least one message
+in the last 60s, from the last 8 messages, 2 are already the same. In that case
```
dgw, line 398: Tie-in with the above, about using 8 as an arbitrary cutoff for checking repetition.

dgw (commented):
Why didn't I notice the hard-coded lookback number (8 lines) before? No idea. Must have looked at the code on my phone, which always makes it easy to get lost. Small screen problems.

> And I even added tests.

Blasphemy! Ban this man! /s

Exirel:
Albeit not perfect, the time window has it uses as it prevents *some* false positive. However, it's not really well done, because it only check if the last message is within that time window, so you could have one message 5s away, and the previous 7 a day old, and it'll *still* think there is a loop.

However, I don't feel like fixing that in this PR, I would like to keep the configuration change and the behavior change as separated as possible.

> Why didn't I notice the hard-coded lookback number (8 lines) before?

On the other hand, Sopel remembers 10 messages per channel/user anyway. So I can remove that limit and just look at the last 10 messages directly.

Exirel:
And so I changed the 8 to 10 and changed the doc accordingly. I added a warning in the doc for good measure, because setting the threshold to more than 10 will deactivate the loop prevention system and that might not be obvious to a user.

dgw:
Is there a reason not to clamp the value to max 10 with a custom parse (and/or serialize) function? Disabling the flood prevention is documented explicitly (and improved on by one of my recent PRs).

sopel/irc/__init__.py
```diff
@@ -830,18 +835,22 @@ def say(
...
+                        return
 
             self.backend.send_privmsg(recipient, text)
-            recipient_stack['flood_left'] = max(0, recipient_stack['flood_left'] - 1)
+
+            # update recipient meta-data
+            flood_left = recipient_stack['flood_left'] - 1
+            recipient_stack['flood_left'] = max(0, flood_left)
             recipient_stack['messages'].append((time.time(), safe(text)))
             recipient_stack['messages'] = recipient_stack['messages'][-10:]
```
dgw, line 855: This actually annoys me, upon looking at the whole file in context. I know a "stack" is different from a FIFO queue, but that's what really should be used for the `messages` buffer (specifically, `collections.deque(maxlen=10)`). Maybe I'll take this branch as a base and try to convert it, instead of making Yet Another Tracking Issue.

dgw (commented):
One more implementation thing about the _number_ of messages considered, while I'm awaiting Exi's thoughts on clamping the `antiloop_threshold` value to max 10.

Exirel:
Given the status of the review, I squashed the commits while adding the `min(10, 'antiloop_threshold')`. I replaced the warning by the note, and that's all I did in here.

docs/source/configuration.rst
```diff
@@ -350,6 +350,79 @@ policy.
...
         *"It's some arcane magic from AT LEAST a decade ago."*
 
+Loop prevention
+---------------
+
+In order to prevent the bot from entering a loop (for example when there is
+another bot in the same channel, or if a user spams a command), it'll try to
+see if the next message to send is repeating too often in the last ten messages
+in the last few minutes. If that happens, the bot will send ``...`` a few times
+before remaining silent::
```
dgw, line 360:
```suggestion
see if the next message to send is repeating too often in a short time period.
If that happens, the bot will send ``...`` a few times before remaining silent::
```

I have been trying to think of a solution to the awkward "in the last ten messages in the last few minutes", and I think the actual solution is to just not get into those details yet. The settings to control this, described later, cover it for users who are curious without confusing someone who's just skimming the first paragraph to see what the heck this feature is.

docs/source/configuration.rst
```diff
@@ -350,6 +350,79 @@ policy.
...
+    [core]
+    antiloop_threshold = 2
+    antiloop_silent_after = 1
+    antiloop_window = 60
+    antiloop_repeat_text = Ditto.
+
+will activate the loop prevention feature if there is at least one message in
+the last 60s, from the last 10 messages, 2 are already the same. In that case
+the bot will send ``...``, but only *once*. After that, the bot will remain
+silent. This doesn't affect other messages, i.e. messages that don't repeat::
```
dgw, line 400: I thought about this one for quite a while, ngl.

```suggestion
will activate the loop prevention feature if there are at least 2 messages
in the last 60 seconds, **and** exactly 2 of those messages are the same.
After sending ``...`` *once* (a third message), the bot will remain silent.

This doesn't affect other messages, i.e. messages that don't repeat::
```

Here again, the 10-message maximum history doesn't matter and can only confuse the reader with too much detail. The `.. note::` below about the maximum value of `antiloop_threshold` covers it enough, IMO.

Can a highly active Sopel instance exhibit behavior that contradicts this simplified explanation of how loop prevention works? Of course it can. Running more than ten different commands in one minute, all with different output, and then repeating one that has been pushed off the top of the history list will make it appear that loop prevention "doesn't work", but that's not a typical case.

(I think it usually goes without saying, but I'll say it anyway: If my reasoning doesn't make sense, _please do_ tell me instead of just accepting the suggestion anyway, so I can suggest another way of clarifying what's being said here without losing as many details.)

sopel/config/core_section.py
```diff
@@ -133,6 +133,92 @@ class CoreSection(StaticSection):
     :func:`~sopel.plugin.nickname_command`.
     """
 
+    antiloop_repeat_text = ValidatedAttribute(
+        'antiloop_repeat_text', default='...')
+    """The replacement text when detecting a repeating message.
```
dgw, line 138:
```suggestion
    """The replacement text sent when detecting a repeated message.
```

sopel/config/core_section.py
```diff
@@ -133,6 +133,92 @@ class CoreSection(StaticSection):
...
+    .. seealso::
+
+        The :ref:`Loop Prevention` chapter to learn what each antiloop-related
+        setting does.
+
+    .. versionadded:: 8.0
+    """
+
+    antiloop_window = ValidatedAttribute('antiloop_window', int, default=120)
+    """The time window (in seconds) used to activate anti-looping.
```
dgw, line 204:
```suggestion
    """The time period (in seconds) checked when detecting repeated messages.
```

I _think_ this is better, but am not convinced it's as clear as it _could_ be.

sopel/irc/__init__.py
```diff
@@ -840,18 +845,22 @@ def say(
                     time.sleep(sleep_time)
 
             # Loop detection
-            messages = [m[1] for m in recipient_stack['messages'][-8:]]
+            if antiloop_threshold > 0 and elapsed < antiloop_window:
+                messages = [m[1] for m in recipient_stack['messages'][-10:]]
```
dgw, line 849: I realize it would obsolete #2327 before it even makes it out of draft status (and make more work for myself in the documentation that I just reworded above), but what if Sopel maintained a time-based rather than count-based history of messages it has sent?

As I said in one of those rephrasing line notes, the `antiloop_window` becomes useless if the bot has sent too many messages to a given channel, because the 10-slot history may be far too small for the volume of output being generated.

sopel/irc/__init__.py
```diff
@@ -840,18 +845,22 @@ def say(
...
+                    if messages.count(text) >= antiloop_silent_after:
+                        # If we've already said that N times, discard message
+                        return
 
             self.backend.send_privmsg(recipient, text)
-            recipient_stack['flood_left'] = max(0, recipient_stack['flood_left'] - 1)
+
+            # update recipient meta-data
+            flood_left = recipient_stack['flood_left'] - 1
+            recipient_stack['flood_left'] = max(0, flood_left)
```
dgw, line 863: Changing the one-liner to two lines feels like a crusade without a cause, but it's not like the code is _wrong_ now.

dgw (commented):
> I squashed the commits while adding the `min(10, 'antiloop_threshold')`. I replaced the warning by the note, and that's all I did in here.

You also rebased while doing that, though. I can't click "force-pushed" in the event history to get the diff between the previous state and the current state without wading through clutter pulled from upstream. :/

So I re-reviewed everything, and came up with more line notes than I probably would have if I'd only had to look at the _new_ changes. Serves you right. 🤪

docs/source/configuration.rst
Exirel: It's a removal then!

docs/source/configuration.rst
Exirel: No you are absolutely right: it makes sense to *me* because I'm knee deep into that ~~garbage~~ code, however I realize it's not going to help the reader. Good catch!

(thanks for taking the time to check on that) (your reasoning makes a lot of sense to me)

sopel/irc/__init__.py
Exirel: This will be for another time. I think the list should contain up to N messages, N being configurable, as long as it is up to the last T seconds, for a maximum size of B bytes.

This is, I think, for another good PR or two.

sopel/irc/__init__.py
Exirel: I... don't even know why I did that.

sopel/irc/__init__.py
dgw: Let me add a note to that other PR, then, and resolve this thread.

sopel/irc/__init__.py
Exirel: Oh, I know: because 80 chars and all that. The rewrap was quite ugly.

Exirel:
And I applied the changes. No squash, no rebase yet. You tell me when it's all OK for you.

dgw (approved):
As far as I can see, this is good to go (squash). If I missed anything by reviewing on my phone, so be it. 🙃

Exirel:
Squashed.

PR #2321, opened 2022-07-14
coretasks: join channels on MOTD

### LINKED ISSUE #2289: UHNAMES Race Condition
### Description
When Sopel joins a server that supports UHNAMES but not the userhost-in-names CAP, it sends `PROTOCTL UHNAMES` to accomplish the same goal.

In my testing, this results in:
```
>> CAP LS 302
>> NICK Sopel
>> USER ...
<< 001 ...
>> JOIN #test
<< 005 [...] UHNAMES
>> PROTOCTL UHNAMES
<< JOIN #test
<< 353 Sopel = #test :@Op Sopel
```

Followed by an exception (expected 2 to unpack, got 1) on [coretasks.py:524](https://github.com/sopel-irc/sopel/blame/36b7c2f1064e7d734e50a66b24270cf7c8949426/sopel/coretasks.py#L524)

Sending `PROTOCTL UHNAMES` *before* the `JOIN` seems to work.

The same problem _may_ exist with the CAP, if not in practice, than in theory.

Previously: #2102

### Reproduction steps
1. Find a server with UHNAMES but no userhost-in-names
2. Configure Sopel to connect to it and join a channel
3. Start sopel
4. Note exception

### Expected behavior
Joins without exception

### Environment
- Sopel `.version`: 36b7c2f1064e7d734e50a66b24270cf7c8949426
- Sopel installed via: pip
- Python version: 3.9.2
- Operating system: Debian 10
- IRCd `/version`: InspIRCd 2

### Notes
@dgw broke it

Exirel:
### Description

Tin, should fix #2289 for good but I didn't test that section in particular.

This was way too easy. The tests didn't even fail, everything worked on the first try. I don't really complain, I'm just... yeah, OK, all good then.

I've tested it with the Libera servers (they use solanum-1.0-dev), and with a friend's server using UnrealIRCd-5.2.4. Both are fairly modern IRC servers, so it is possible that this breaks with older server that don't follow the MOTD specification.

### Checklist

### DIFF
--- a/sopel/coretasks.py
+++ b/sopel/coretasks.py
@@ -64,6 +64,7 @@ def setup(bot):
     The setup phase is used to activate the throttle feature to prevent a flood
     of JOIN commands when there are too many channels to join.
     """
+    bot.memory['retry_join'] = SopelMemory()
     bot.memory['join_events_queue'] = collections.deque()
 
     # Manage JOIN flood protection
@@ -82,6 +83,7 @@ def setup(bot):
 
 def shutdown(bot):
     """Clean up coretasks-related values in the bot's memory."""
+    bot.memory['retry_join'] = SopelMemory()
     try:
         bot.memory['join_events_queue'].clear()
     except KeyError:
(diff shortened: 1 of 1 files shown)


half-duplex:
This fixes the specific issues I was observing. I haven't poked into the state graph - is this still a race, just one we more reliably win? Or do we know that we'll never get an ENDOFMOTD/NOMOTD until after the UHNAMES has been received?

dgw:
[Horse docs says](https://modern.ircdocs.horse/#connection-registration) that:

> Upon successful completion of the registration process, the server MUST send, in this order, the [RPL_WELCOME](https://modern.ircdocs.horse/#rplwelcome-001) (001), [RPL_YOURHOST](https://modern.ircdocs.horse/#rplyourhost-002) (002), [RPL_CREATED](https://modern.ircdocs.horse/#rplcreated-003) (003), [RPL_MYINFO](https://modern.ircdocs.horse/#rplmyinfo-004) (004), and at least one [RPL_ISUPPORT](https://modern.ircdocs.horse/#rplisupport-005) (005) numeric to the client.

The UHNAMES flag, if advertised, will appear in a 005 line, at which point Sopel will immediately send the PROTOCTL to enable it.

Some other stuff "SHOULD" or "MAY" happen before the server again "MUST" respond as if the client sent it the MOTD command, which will trigger the start of channel joins as of this patch. By that time, since `coretasks` stuff tends to use `plugin.thread(False)` to guard against processing protocol events in parallel, PROTOCTL UHNAMES should have taken effect.

Is it _technically_ still a race? Yeah, I guess. But by the time Sopel _processes_ everything the server has sent and starts asking to join channels, a race should be practically impossible because disabling threaded handling forces everything the server sends after registration to be processed in sequence.

Exirel:
So, is that an approval?

dgw:
Ask @half-duplex, as the original reporter of #2289.

half-duplex (approved):
Even if there's still potentially a race, between the try/catch, this, and that bad results don't actually matter, I'm fine calling it resolved.

PR #2322, opened 2022-07-14
quality: add flake8-quotes

Exirel:
### Description

Tin. This was a nightmare. Please never again. Good luck reviewers.

### Checklist

### DIFF
--- a/sopel/modules/reddit.py
+++ b/sopel/modules/reddit.py
@@ -88,9 +88,9 @@ def get_is_cakeday(entrytime):
 @plugin.output_prefix(PLUGIN_OUTPUT_PREFIX)
 def image_info(bot, trigger, match):
     url = match.group(0)
-    preview = match.group("subdomain") == "preview"
+    preview = match.group('subdomain') == 'preview'
     if preview:
-        url = "https://i.redd.it/{}".format(match.group("image"))
+        url = 'https://i.redd.it/{}'.format(match.group('image'))
     results = list(
         bot.memory['reddit_praw']
         .subreddit('all')
@@ -127,11 +127,11 @@ def post_or_comment_info(bot, trigger, match):
     match = match or trigger
     groups = match.groupdict()
 
-    if groups.get("comment"):
-        say_comment_info(bot, trigger, groups["comment"])
+    if groups.get('comment'):
+        say_comment_info(bot, trigger, groups['comment'])
         return
 
-    say_post_info(bot, trigger, groups["submission"])
+    say_post_info(bot, trigger, groups['submission'])
 
 
 @plugin.url(gallery_url)
@@ -258,14 +258,14 @@ def subreddit_info(bot, trigger, match, commanded=False):
     if match_lower in ['all', 'popular']:
         message = '{link}{nsfw} | {public_description}'
         nsfw = ' ' + bold(color('[Possible NSFW]', colors.ORANGE))
-        link = "https://reddit.com/r/" + match_lower
+        link = 'https://reddit.com/r/' + match_lower
         public_description = ''
         if match_lower == 'all':
-            public_description = ('Today\'s top content from hundreds of '
+            public_description = ("Today's top content from hundreds of "
                                   'thousands of Reddit communities.')
         elif match_lower == 'popular':
             public_description = ('The top trending content from some of '
-                                  'Reddit\'s most popular communities')
+                                  "Reddit\'s most popular communities")
         message = message.format(
             link=link, nsfw=nsfw, public_description=public_description)
         bot.say(message)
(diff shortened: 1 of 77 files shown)


Exirel:
I re-read the diff, just to be extra-sure, and I didn't see any mistake. I followed the preferences of @dgw written on #1765 about this flake8 plugin (which is: to use the default behavior).

I might want to ask for a fast track merge on this one, or this will increase the number of conflicts and new flake8 errors later on, since it touches almost everything.

sopel/modules/reddit.py
```diff
@@ -258,14 +258,14 @@ def subreddit_info(bot, trigger, match, commanded=False):
...
+        link = 'https://reddit.com/r/' + match_lower
         public_description = ''
         if match_lower == 'all':
-            public_description = ('Today\'s top content from hundreds of '
+            public_description = ("Today's top content from hundreds of "
                                   'thousands of Reddit communities.')
         elif match_lower == 'popular':
             public_description = ('The top trending content from some of '
-                                  'Reddit\'s most popular communities')
+                                  "Reddit\'s most popular communities")
```
half-duplex, line 268:
```suggestion
                                  "Reddit's most popular communities")
```
Does this flake plugin not catch these?

sopel/modules/translate.py
```diff
@@ -70,7 +70,7 @@ def translate(text, in_lang='auto', out_lang='en'):
         data = json.loads(result)
     except ValueError:
         LOGGER.error(
-            'Error parsing JSON response from translate API (%s to %s: "%s")',
+            "Error parsing JSON response from translate API (%s to %s: '%s')",
```
half-duplex, line 73: Why?

sopel/modules/safety.py
```diff
@@ -65,41 +65,41 @@ def configure(settings: Config):
...
-        "protection.\nOtherwise, only the configured hosts list will be used.",
+        'Optionally, enter a VirusTotal API key to improve malicious URL '
+        'protection.\nOtherwise, only the configured hosts list will be used.',
     )
     settings.safety.configure_setting(
-        "domain_blocklist_url",
-        "Optionally, provide the URL for a hosts-file formatted domain "
+        'domain_blocklist_url',
+        'Optionally, provide the URL for a hosts-file formatted domain '
         "blocklist to use instead of StevenBlack's.",
```
half-duplex, line 89: Same gross

half-duplex (changes requested):
I'll say again that I disagree with the premise, I much prefer double quotes over typing a string then noticing it has an apostrophe and needing to replace the singles with doubles. Apostrophes are a markedly more common case than nested quoting.

```
sopel#flake8-quotes$ git grep '".*'"'"'.*"' | wc -l  # How many "...'..."
213
sopel#flake8-quotes$ git grep "'"'.*".*'"'" | wc -l  # How many '..."...'
126
```

Not that the world should cater to C, but for anyone who's used a language like it, anything but double quotes on a string is something to trip over constantly.

That said, besides the actionable comments, this does what it says on the tin.

sopel/modules/translate.py
Exirel: Oh, that might have been either an honest mistake (I did some find & replace regex), or there is something else. I think it's probably a mistake.

dgw:
> I'll say again that I disagree with the premise, I much prefer double quotes over typing a string then noticing it has an apostrophe and needing to replace the singles with doubles. Apostrophes are a markedly more common case than nested quoting.

What gets me about your figures is actually how common `'..."...'` is. I expected it to be rarer. I guess we use a lot of that pattern for logging and/or command-usage messages.

In reality, I don't mind double quotes in most cases. If I really didn't like them, there are several @half-duplex PRs I would have blocked. 😝 (I mostly remember calling h-d out on changing quotes in lines that didn't otherwise need to be touched.)

Looking back at the text @Exirel cited from me, I think that was an incomplete expression of preference—for which I owe everyone who's looked at the list an apology. Where double quotes bug me is in subscripts (`some_array['key']`), some function args (one-word strings), and other code-y places. For logs, bot output, etc., double quotes are great.

This PR may have nudged me away from the idea of using this plugin after all, since it doesn't appear possible to configure it such that my exact preferences would work. The only distinction it makes is between `inline` and `multiline` quotes, and that just doesn't match my (weird) desire for (questionable) consistency.

Unless I'm missing something, we can't have this plugin enforce _all_ of these with one configuration (examples could be more comprehensive, but I ran out of ideas):

```py
# dict keys and similar with singles
some_dict['keyname']

# output string with doubles, but a short parameter with singles
bot.say("Some {} message.".format('useless'))

# short parameter with singles
helper_func(varname, 'truth_check', 30)

# continuation with doubles (actually the plugin probably *could* do this if `double` is the default)
long_helper(
    "Something wicked this way comes. "
    "(It's a witch.)"
)
```

I just wish this realization had come from experimenting with the cleanup _myself_ and trying to make this plugin do what I want it to do, rather than taking @Exirel's time for that. 😢

sopel/modules/reddit.py
dgw: I don't think it's designed to catch this. zheller/flake8-quotes#98

Just another reason I'm disappointed in how this turned out, and wish I'd experimented with it before adding to the wishlist so Exirel didn't waste time implementing. 😦

Exirel:
Yeah that makes sense. I think we should close this for now, we might want to give it another try later in the future, when we don't have 30 other issues to fix before releasing Sopel 8 then. I think it was worth the try to show what it means, in term of diff and efforts.

Exirel:
As per IRC discussion: we close this one for now!

PR #2327, opened 2022-07-20
irc: use deque for "stack" of recent messages

dgw:
### Description
Tin. Reading #2320 made me wonder why we used a list with manual pruning (`l = l[-10:]` after appending), and I couldn't think of a good reason. Updated tests made me pretty confident I could just swap to a deque with `maxlen` and get pruning for free.

### Checklist
  - Caveat: Some type-checking stuff gets flagged by flake8, but those should be sorted out already on the actual `master` branch (they reference code that wasn't changed in this patch).
  - (thanks to Exirel's updated test suite)

### Notes
This is a draft because it's built on top of #2320. I will rebase and mark ready for review when that's squashed down.

### DIFF
--- a/sopel/irc/__init__.py
+++ b/sopel/irc/__init__.py
@@ -25,6 +25,7 @@ who should worry about :class:`sopel.bot.Sopel` only.
 from __future__ import annotations
 
 import abc
+from collections import deque
 from datetime import datetime
 import logging
 import os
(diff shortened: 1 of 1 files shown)


dgw:
Thanks for breaking CI, Coveralls. Thankfully it doesn't actually matter yet, because this is only a draft.

dgw:
Coveralls is fixed. Still keeping as draft until the base PR is completed.

dgw:
Further discussion on #2320 says we should keep thinking about this, perhaps converting the "stack" to prune messages not just by a dumb count, but by time range and/or maximum size in bytes. This can be considered Replaced if one of us gets the motivation to do the fancy way pre-8.0, but should be merged if we don't.

SnoopJeDi (approved):
Nice little improvement

Exirel (approved):
Perfection.

Issue #2120, opened 2021-06-27
Cleanup/refactoring for 8.0


dgw:
_Originally posted by @half-duplex in https://github.com/sopel-irc/sopel/issues/2062#issuecomment-856886914_

- [x] sopel/formatting.py TODO (#2122)
- [x] conftest.py 2x "Python 3.3" comments (#2123)
- [x] sopel/plugins/rules.py has a 2.7 TODO (#2124)
- ~~[ ] sopel/irc/backends.py SSLAsynchatBackend.handle_connect~~ NAK; replaced by asyncio backend

dgw:
Got PRs for the first three. I'd have to study more about that `SSLAsynchatBackend` one, if it's even worth doing before @Exirel starts on that planned rewrite using `asyncio` instead of `asynchat`.

Exirel:
Not sure it's worth it indeed.

On the other hand, I'm still pondering if using asyncio vs using a threading approach wouldn't be better anyway, since Sopel *does not* take advantage of asyncio in any way (there is no coroutine/async/await, and it won't be possible to use them any time soon).

dgw:
Whether threading or asyncio, it's probably best to refactor Soon™. Looks like the current asyncore/asynchat backend is going to trigger `DeprecationWarning` in Python 3.10: python/cpython#26904

Honestly we should be surprised (and grateful) that it took this long (almost 5 years since 3.6.0 final).

Exirel:
All done from this list.

dgw:
> All done from this list.

Is the backend one "done" by virtue of that class being replaced? 😁

Exirel:
Absolutely. :grin: 
I don't know if there are other items to consider, so I didn't close the issue.

dgw:
Consider? Yes. I found some stuff to triage by searching the whole repo for "todo".

Py2 and cleanup/follow-through categories are the ones I'd consider important to either complete or update comments with a revised plan to handle in 8.1/9. Some of these items may require not-insignificant refactoring, which is why I included "update comments with a revised plan" as an option.

Future planning category would be nice to think about before we cut a new major version, but aren't _required_.

Misc. is obviously the least important. I threw things in there when I didn't necessarily understand what's being asked, or think it'll take under 15 mins to address.

<details>
<summary>Py2-is-dead stuff</summary>

[...]

PR #2330, opened 2022-07-23
currency: fiat_provider config option takes precedence

Exirel:
### Description

Tin, part of #2120.

### Checklist

### DIFF
--- a/sopel/modules/currency.py
+++ b/sopel/modules/currency.py
@@ -41,9 +41,11 @@ rates_updated = 0.0
 
 
 class CurrencySection(types.StaticSection):
-    fiat_provider = types.ChoiceAttribute('fiat_provider',
-                                          list(FIAT_PROVIDERS.keys()),
-                                          default='exchangerate.host')
+    fiat_provider = types.ChoiceAttribute(
+        'fiat_provider',
+        list(FIAT_PROVIDERS.keys()),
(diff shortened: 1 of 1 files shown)


sopel/modules/currency.py
```diff
@@ -215,28 +216,25 @@ def update_rates(bot):
     rates_crypto = response.json()
 
     # Update fiat rates
-    if bot.config.currency.fixer_io_key is not None:
+    settings = bot.settings.currency
```
dgw, line 219: I have mixed feelings about making such aliases. To someone unfamiliar with the structure of a Sopel instance, reading core plugin code to learn from existing examples, this is easy to miss (and then later wonder where the hell `settings` comes from).

In this particular case, converting to use this alias also pollutes the diff 😝

dgw (approved):
I'm nitpicking. There's nothing actually "wrong" with the patch; it just took a bit more mental energy to review than it would have otherwise due to extraneous code style changes. :P

sopel/modules/currency.py
Exirel: Yeah, I'll change that to something easier to read, even for me this is not optimal.

Exirel:
And done.

dgw (approved):
Nitpick resolved, I see. Squash away 🚀

PR #2332, opened 2022-07-23
help: replace with dependency on `sopel-help`

### LINKED ISSUE #1212: help: using a pastebin should be optional

Exirel:
### Description

Fix #1212 by proxy.

Make the built-in `help` plugin obsolete with warning in its `setup()` hook and all rules have been removed (rules & commands), and installing Sopel now requires `sopel-help`.

Since it is possible to install a package without its dependencies, I kept `help.py` so users can have a warning. It isn't much, but it might help someone.

This is based of #2342 for Python 3.10.

Thank you for your service, help, you will be remembered.

### Checklist


dgw:
Are the internals that `help`/`sopel-help` use well-enough documented that someone could easily make a custom equivalent? (Not talking about adding e.g. paste backends to `sopel-help`, but an actual full-on replacement.)

Otherwise, I'm 👎 on replacing an essential built-in plugin like this with a no-op.  Moving to package it separately (so it can be updated independently of core) is fine, but core should require it.

Exirel:
> Are the internals that `help`/`sopel-help` use well-enough documented that someone could easily make a custom equivalent?

Not yet! This is in the work with Sopel 8, by replacing the hard-coded dict and stuffs like that by the Rule Managers, which could be the origin of a new chapter in "advanced tips & tricks" in the documentation.

> core should require it.

Alright then! Waiting on #2328 first, then I'll add it to the list of install-requires. 😁

sopel/modules/help.py
```diff
@@ -1,350 +1,18 @@
...
-@plugin.priority('low')
-def help2(bot, trigger):
-    response = (
-        "Hi, I'm a bot. Say {1}commands to me in private for a list "
-        "of my commands, or see https://sopel.chat for more "
-        "general details. My owner is {0}."
-        .format(bot.config.core.owner, bot.config.core.help_prefix))
-    bot.reply(response)
+    LOGGER.warning(
+        'Sopel help built-in module is obsolete. '
```
dgw, line 16:
```suggestion
        'Sopel's built-in help plugin is obsolete. '
```

How many times have we talked about why "module" is confusing? 😝

dgw (commented):
You know, the irony here is that I was ready to merge this despite the use of "module" because it's unlikely that most users would ever see it. But because you rebased it onto another PR that isn't merged yet, I _have_ to wait, and so do you. 😝

(If you wait until #2342 goes in before amending this one, the UI will be less confusing. The extra commits will disappear after you force-push, but _only if_ they've been merged.)

dgw:
Oh, I just had another (possibly stupid) idea: Could `help.py` be moved to some other place called "shims" or similar, so we have a place for this and other plugins that are moved out of core but still want to keep around with warnings like this for a bit?

Obviously I know it _can_, but is the idea _good_? 😛

dgw:
Wait is over; #2342 has been merged. Amend away, and soon ye shall receive another merge. 😁

Exirel:
Done. I had to fix the message, and decided to keep a single `\'` because I could have that or use double quotes and use two `\"` for the last line.

dgw (approved):
> and decided to keep a single `\'` because I could have that or use double quotes and use two `\"` for the last line.

Maybe you forgot that the below totally works 😁 but it's totally a choice between two almost equally annoying alternatives (mixed quotes vs. escapes). The third option would be using double quotes and wrapping the command in something else, maybe backticks.

Anyway, I'm already over it.

```pycon
>>> print(
...     "Sopel's built-in help plugin is obsolete. "
...     'Install sopel-help as the official help plugin for Sopel.\n'
...     'You can install sopel-help with "pip install sopel-help".')
Sopel's built-in help plugin is obsolete. Install sopel-help as the official help plugin for Sopel.
You can install sopel-help with "pip install sopel-help".
```

Exirel:
I did not. I remembered that you didn't like that syntax on my attempt at using flake8-quotes, mixing `"` and `'` in multi-lines.

dgw:
Well, I don't. But I also don't like escapes. But this is also almost never going to be seen by anyone, until its eventual deletion in 9.0 or whatever, so again: I'm over it. 😁 😹

PR #2333, opened 2022-07-25
Add FreeBSD service

DtxdF:
### Description
Service support for FreeBSD.

### Checklist

### DIFF
--- a/contrib/FreeBSD/README
+++ b/contrib/FreeBSD/README
@@ -1,0 +1,46 @@
+This folder contains sopel and sopel-default.cfg designed to be distributed by third parties such as FreeBSD.
+
+`sopel-default.cfg` is a default configuration file for sopel. The -default prefix is mandatory and will be used as a 'profile'. The profiles may be changed with the variable `sopel_profiles` in `/etc/rc.conf`. For example:
+
+```sh
+sysrc sopel_profiles="profile1 profile2 profile3"
+```
+
+The configuration files should be `${sopel_confdir}/${sopel_prefix}profile1.cfg`, `${sopel_confdir}/${sopel_prefix}profile2.cfg` and `${sopel_confdir}/${sopel_prefix}profile3.cfg`. The variable `sopel_confdir` is the default configuration directory (by default `/usr/local/etc`). `sopel_prefix`is the prefix for all files (by default `sopel-`).
+
+The service must be installed in `/usr/local/etc/rc.d`:
+
+```sh
+cp sopel /usr/local/etc/rc.d
+chmod +x /usr/local/etc/rc.d/sopel
+```
+
+The default configuration file must be installed in ${sopel_confdir}. For example:
+
+```sh
+cp sopel-default.cfg /usr/local/etc
+```
+
+If you want to run sopel at startup, the variable `sopel_enable` must be set to `YES`:
+
+```sh
+sysrc sopel_enable="YES"
+```
+
+If you want to change the python version, the variable `sopel_interpreter` (by default `/usr/local/bin/python3.9`) must be changed.
+
+Default values:
+
+```sh
+$ egrep -E '^: \$\{sopel_.+:=.+}' sopel
+: ${sopel_enable:="NO"}
+: ${sopel_piddir:="/var/run/sopel"}
+: ${sopel_confdir:=/usr/local/etc}
+: ${sopel_flags:=--config-dir "${sopel_confdir}"}
+: ${sopel_program:="/usr/local/bin/sopel"}
+: ${sopel_user:=${name}}
+: ${sopel_interpreter:=/usr/local/bin/python3.9}
+: ${sopel_profiles:=default}
+: ${sopel_prefix:=sopel-}
+: ${sopel_output:=/dev/null}
+```
--- a/contrib/FreeBSD/sopel
+++ b/contrib/FreeBSD/sopel
@@ -1,0 +1,120 @@
+#!/bin/sh
+
+# PROVIDE: sopel
+# REQUIRE: FILESYSTEM NETWORKING usr
+
+. /etc/rc.subr
+
+name="sopel"
+desc="Simple, easy-to-use, open-source IRC utility bot, written in Python"
+rcvar="${name}_enable"
+start_cmd="sopel_start"
+stop_cmd="sopel_stop"
+restart_cmd="sopel_restart"
+status_cmd="sopel_status"
+configure_cmd="sopel_configure"
+extra_commands="configure"
+
+load_rc_config "${name}"
+
+: ${sopel_enable:="NO"}
+: ${sopel_piddir:="/var/run/sopel"}
+: ${sopel_confdir:=/usr/local/etc}
+: ${sopel_flags:=--config-dir "${sopel_confdir}"}
+: ${sopel_program:="/usr/local/bin/sopel"}
+: ${sopel_user:=${name}}
+: ${sopel_interpreter:=/usr/local/bin/python3.9}
+: ${sopel_profiles:=default}
+: ${sopel_prefix:=sopel-}
+: ${sopel_output:=/dev/null}
+
+sopel_start()
+{
+	local profile
+
+	profile="$1"; shift
+
+	echo "Starting sopel profile '${profile}'." && sleep 1
+        /usr/sbin/daemon \
+		-o "${sopel_output}" \
+		-t "${desc}" \
+		-u "${sopel_user}" \
+		${sopel_interpreter} \
+		${sopel_program} start ${sopel_flags} \
+			-c "${sopel_prefix}${profile}" $@
+}
+
+sopel_stop()
+{
+	local pid pidfile profile
+	
+	profile="$1"; shift
+
+	pidfile="${sopel_piddir}/${sopel_prefix}${sopel_prefix}${profile}.pid"
+	if ! [ -f "${pidfile}" ]; then
+		return 1
+	fi
+
+	pid=`cat ${pidfile}`
+
+	echo "Stopping sopel profile '${profile}'."
+        /usr/sbin/daemon \
+		-o "${sopel_output}" \
+		${sopel_interpreter} \
+		${sopel_program} stop ${sopel_flags} \
+			-c "${sopel_prefix}${profile}" $@
+
+	wait_for_pids $pid
+}
+
+sopel_restart()
+{
+	local profile
+	
+	profile="$1"; shift
+
+        run_rc_command stop "${profile}" $@ &&
+        sleep 1 &&
+        run_rc_command start "${profile}" $@
+}
+
+sopel_status()
+{
+	local profile pid
+
+	profile="$1"; shift
+
+	pid=`check_pidfile \
+		"${sopel_piddir}/${sopel_prefix}${sopel_prefix}${profile}.pid" \
+		"${sopel_program}" \
+		"${sopel_interpreter}"`
+
+	if [ -n "${pid}" ]; then
+		echo "Sopel profile '${profile}' is running as pid ${pid}."
+	else
+		echo "Sopel profile '${profile}' is not running."
+	fi
+}
+
+sopel_configure()
+{
+	local profile
+
+	profile="$1"; shift
+
+	echo "Configuring profile '${profile}'..."
+
+	${sopel_interpreter} \
+        ${sopel_program} configure ${sopel_flags} \
+		-c "${sopel_confdir}/${sopel_prefix}${profile}" $@
+}
+
+cmd="$1"; shift
+for profile in $sopel_profiles; do
+	if ! [ -f "${sopel_confdir}/${sopel_prefix}${profile}.cfg" ]; then
+		echo "Sopel profile '${profile}' does not exist."
+		continue
+	fi
+
+	run_rc_command "${cmd}" "${profile}" $@
+done
--- a/contrib/FreeBSD/sopel-default.cfg
+++ b/contrib/FreeBSD/sopel-default.cfg
@@ -1,0 +1,26 @@
+#
+# IMPORTANT NOTES!
+# You must delete the not_configured line in order for the bot to work,
+# otherwise it will refuse to start.
+#
+# You must create homedir. Sopel should create the others directories:
+#  mkdir -p $homedir
+# The service file has a variable called 'sopel_user'. This user name
+# must exist. After creating the user, the owner of the logdir, 
+# pid_dir and homedir directories must be changed. By default,
+# the value is 'sopel':
+#  chown sopel:sopel $logdir
+#  chown sopel:sopel $pid_dir
+#  chown sopel:sopel $homedir
+# 
+[core]
+nick=sopel
+not_configured=True
+host=irc.libera.chat
+port=6697
+use_ssl=True
+verify_ssl=True
+owner=
+logdir=/var/log/sopel
+pid_dir=/var/run/sopel
+homedir=/var/db/sopel


contrib/FreeBSD/README
```diff
@@ -0,0 +1,46 @@
...
+: ${sopel_piddir:="/var/run/sopel"}
+: ${sopel_confdir:=/usr/local/etc}
+: ${sopel_flags:=--config-dir "${sopel_confdir}"}
+: ${sopel_program:="/usr/local/bin/sopel"}
+: ${sopel_user:=${name}}
+: ${sopel_interpreter:=/usr/local/bin/python3.9}
+: ${sopel_profiles:=default}
+: ${sopel_prefix:=sopel-}
+: ${sopel_output:=/dev/null}
+```
```
half-duplex, line 46: Missing trailing newline

contrib/FreeBSD/README
```diff
@@ -0,0 +1,46 @@
+This folder contains sopel and sopel-default.cfg designed to be distributed by third parties such as FreeBSD.
+
+`sopel-default.cfg` is a default configuration file for sopel. The -default prefix is mandatory and will be used as a 'profile'. The profiles may be changed with the variable `sopel_profiles` in `/etc/rc.conf`. For example:
+
+```sh
```
half-duplex, line 5: If using markdown formatting, this file should be named *.md

contrib/FreeBSD/README
```diff
@@ -0,0 +1,46 @@
...
+sysrc sopel_profiles="profile1 profile2 profile3"
+```
+
+The configuration files should be `${sopel_confdir}/${sopel_prefix}profile1.cfg`, `${sopel_confdir}/${sopel_prefix}profile2.cfg` and `${sopel_confdir}/${sopel_prefix}profile3.cfg`. The variable `sopel_confdir` is the default configuration directory (by default `/usr/local/etc`). `sopel_prefix`is the prefix for all files (by default `sopel-`).
```
half-duplex, line 9: Is editing these a routine enough use case to warrant documenting it this way? I'd prefer to avoid making people reading docs parse variable syntax. Also, GitHub handles long lines *horribly*, I'd suggest sticking to 80-120 chars.
Phrasing:
```suggestion
The configuration files should be `${sopel_confdir}/${sopel_prefix}profile1.cfg`, `${sopel_confdir}/${sopel_prefix}profile2.cfg` and `${sopel_confdir}/${sopel_prefix}profile3.cfg`. The variable `sopel_confdir` is the configuration directory (by default `/usr/local/etc`). `sopel_prefix`is the prefix for all files (by default `sopel-`).
```

contrib/FreeBSD/sopel
```diff
@@ -0,0 +1,120 @@
...
+}
+
+sopel_restart()
+{
+	local profile
+	
+	profile="$1"; shift
+
+        run_rc_command stop "${profile}" $@ &&
+        sleep 1 &&
```
half-duplex, line 77: Necessary?

contrib/FreeBSD/sopel-default.cfg
```diff
@@ -0,0 +1,26 @@
+#
```
half-duplex, line 1: Would it be worth suggesting use of `sopel configure`, either through the service manager or directly?

contrib/FreeBSD/README
DtxdF: I added the trailing newline

contrib/FreeBSD/README
```diff
@@ -0,0 +1,46 @@
...
+
+The service must be installed in `/usr/local/etc/rc.d`:
+
+```sh
+cp sopel /usr/local/etc/rc.d
```
half-duplex, line 14: Consider naming the `sopel` script `sopel.sh` or `sopel-freebsd.sh` for clarity, and stripping that off here?

half-duplex (commented):
Thanks for the PR!

It's not clear who the target is for the readme, but I think answering that question might let in be more concise and less duplicative of the contents of the script.

DtxdF:
> Thanks for the PR!
> 
> It's not clear who the target is for the readme, but I think answering that question might let it be more concise and less duplicative of the contents of the script.

It is for FreeBSD users. The README explains how to set variables for sopel using the FreeBSD tools and explains how to get more variables to configure sopel. My intention is to show how to configure sopel as another FreeBSD service through `/etc/rc.conf`.

contrib/FreeBSD/README
DtxdF: I will change the extension

contrib/FreeBSD/README
DtxdF: I could change the content of the README to make it easier to understand without the need to parse the variables. I will keep between 80-120 chars

contrib/FreeBSD/sopel
DtxdF: The `sleep 1`?

contrib/FreeBSD/sopel-default.cfg
DtxdF: The service file has a wrapper to `sopel configure`. I think using a wrapper like `service sopel configure` is more convenient because constantly the user is running `service sopel <start|stop|status|restart|configure>`.

contrib/FreeBSD/README
DtxdF: I name the service file as such, without extension, because FreeBSD init does not need the .sh extension, and other services do not use that extension

contrib/FreeBSD/sopel
half-duplex: Yeah - will `stop` ever return before sopel has finished stopping? If not, is there a reason for the sleep?

contrib/FreeBSD/sopel-default.cfg
half-duplex: Right, if that's available, it would be good to mention it - either in this file or in the readme.

contrib/FreeBSD/README
half-duplex: Right, I'm not suggesting naming the service `sopel.sh`, just naming the script one of the above in the repo, and having the command here be `cp sopel.sh /usr/local/etc/rc.d/sopel` to install it without the extension. Not a blocker if you don't like it.

contrib/FreeBSD/README.md
```diff
@@ -0,0 +1,58 @@
...
+cp sopel-default.cfg /usr/local/etc
+```
+
+If you want to run sopel at startup, the variable `sopel_enable` must be set to `YES`:
+
+```sh
+sysrc sopel_enable="YES"
+```
+
+If you want to change the python version, the variable `sopel_interpreter` must be changed. For example:
```
half-duplex, line 38: Necessary for end user docs? How common is this need?

contrib/FreeBSD/README.md
```diff
@@ -0,0 +1,58 @@
...
+
+The profiles may be changed with the variable `sopel_profiles` in `/etc/rc.conf`. For example:
+
+```sh
+sysrc sopel_profiles="profile1 profile2 profile3"
+```
+
+By default, the configuration directory is `/usr/local/etc`.
+
+The `sopel-profile1.cfg`, `sopel-profile2.cfg` and `sopel-profile3.cfg` files must be stored in the configuration directory.
```
half-duplex, line 15:
```suggestion
For the above example, the configuration files `sopel-profile1.cfg`, `sopel-profile2.cfg`, and `sopel-profile3.cfg` must be present there.
```
Conditional on the example - otherwise I can see someone reading it as mandatory.

half-duplex (commented):
If the readme is intended for end users (rather than packagers), I think even fewer of the options need to be specifically pointed out. If someone has unusual needs (in particular, changing sopel_prefix or sopel_interpreter) they probably understand how to do that just by the sopel_profiles instruction and the list of defaults.

The average end user will be reading documentation with one goal in mind, making it work. They're looking for "pip install sopel, then copy this file here, that file there, and run these two commands. For these common changes do [...], or for less common changes apply that pattern to one of [variable list]", and the doc should be structured and written to support them in that endeavor.

contrib/FreeBSD/sopel
DtxdF: Yes, some IRC servers may block an IP with two or more simultaneous connections to prevent flooding. The message `Session limit exceeded` should be printed in the IRC client. I will use `sleep 1` because `daemon(8)` will run sopel in the background without waiting. I know sopel has a flag called --fork, but the daemon is convenient in context because the output is redirected to /dev/null (sopel already prints the output to `/var/log/sopel`, so there is no need to have two outputs).

contrib/FreeBSD/sopel-default.cfg
DtxdF: I will suggest that in the README

contrib/FreeBSD/README
DtxdF: No problem, I will change the name

contrib/FreeBSD/README.md
```diff
@@ -0,0 +1,58 @@
...
+
+sopel- is mandatory and can be changed with the `sopel_prefix` variable in `/etc/rc.conf` as another FreeBSD service.
```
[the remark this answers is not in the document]
DtxdF: I will remove the `sopel_prefix`, it is intended for packagers, but I suggest that `sopel_interpreter` should be kept because, on FreeBSD, there is no `/usr/local/bin/python` link to `/usr/local/bin/python3.8|3.9|x.xx`, so it is necessary to reconfigure `sopel_interpreter` when python has a new version on FreeBSD. The `sopel_interpreter` variable is not only used to run sopel, it is also used to check if sopel is running with the `check_pidfile` function.

contrib/FreeBSD/README.md
half-duplex: https://github.com/sopel-irc/sopel/pull/2333#discussion_r930317048

contrib/FreeBSD/sopel
half-duplex: Fair, but I think that belongs in the start function - and it looks like there already is one there, making this one unnecessary, unless I'm misunderstanding `run_rc_command` or something.

DtxdF: In the case of `sopel_stop`, there is no `sleep` because the user does not want to wait. In the case of `sopel_start` there is a `sleep` because `sopel_restart` has one. To avoid the above problem, there are 2 seconds to wait, the `sleep` of `sopel_start`, and the `sleep` of `sopel_restart`. I can place `sleep 2` in `sopel_start`, but it may be annoying for the user to wait 2 seconds for each startup profile. The `sleep 1` is unnoticeable.

contrib/FreeBSD/sopel-default.cfg
```diff
@@ -0,0 +1,26 @@
...
+# You must create homedir. Sopel should create the others directories:
+#  mkdir -p $homedir
+# The service file has a variable called 'sopel_user'. This user name
+# must exist. After creating the user, the owner of the logdir, 
+# pid_dir and homedir directories must be changed. By default,
+# the value is 'sopel':
+#  chown sopel:sopel $logdir
+#  chown sopel:sopel $pid_dir
+#  chown sopel:sopel $homedir
```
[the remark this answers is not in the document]
DtxdF: I will put it in the README, and leave a suggestion about the `sopel configure`.

dgw:
@half-duplex When this is satisfactory, are you up to walking OP through the process to squash & amend commit message to meet project standards?

DtxdF:
> @half-duplex When this is satisfactory, are you up to walking OP through the process to squash & amend commit message to meet project standards?

Sorry, rule 5 in CONTRIBUTING.md. I will comply with your rules.

dgw:
@DtxdF Don't worry about it just yet. We allow time to squash/tweak the history after PR approval. 🙂

contrib/FreeBSD/sopel
half-duplex: I'm just not clear on why starting multiple bots would need a 1 second delay, but *re*starting would need a 2 second delay. If only 1 second is necessary, the additional delay in restart can be removed. If 2 seconds is always necessary, the start delay should be 2 and the additional delay in restart can still be removed.

contrib/FreeBSD/sopel-default.cfg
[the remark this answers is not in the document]
DtxdF: I think that configuration file should stay there. It is explicit for the service requirements, especially with `pid_dir`, `logdir`, `homedir`. It is also explicit with the required name.

contrib/FreeBSD/sopel
DtxdF: I understand, I will move the delay from `restart` to `stop`.

2 seconds delay is necessary for fast connections or network delays. It makes more sense to move the delay from `restart` to `stop`.

I understand it is a bit confusing, but the 2 second delay is necessary, and the addition of the same functions makes sense to solve the above problem.

Do we agree with this change?

half-duplex: - Why do stops need to be staggered?
- Why do restarts need a larger delay than starts?

DtxdF: If I move the delay from `restart` to `stop`, the restart is simply:

```sh
run_rc_command stop "${profile}" $@ &&
run_rc_command start "${profile}" $@
```

The delay is set to `stop`, there is no need to set the delay to `restart` because there is already one.

Why is it necessary to put the delay in `start` and `stop`? Simple: to solve the above problem, it is necessary to delay to 2 seconds. No more, no less.

The delay in `start` is not perceptible. The delay in `stop` is not perceptible.

Yes, i could put the delay in `restart`, but the above problem may occur if the user runs `service sopel stop && service sopel start` instead of `service sopel restart`. I know the user prefers the latter case, but it is possible for it to happen.

half-duplex: Sorry, it's not really clear from that - Why do restarts need a larger delay (1s stop + 1s start) than starts (1s)?

DtxdF: Due to a slow connection or a network problem that delays a packet before disconnecting it.

half-duplex: Normally Sopel sends a `QUIT` message and waits before exiting. Unless this `stop` doesn't wait for sopel to exit, or you know of routine cases I don't, you should be able to assume that if the process exits you're all set.

https://github.com/sopel-irc/sopel/blob/71380df85212747a6c02d8c37c9ffb7836001278/sopel/irc/__init__.py#L635

DtxdF: As I mentioned in another comment, the service uses daemon(8) which does not wait for the program to run.

DtxdF: Do you agree with the change I mentioned?

half-duplex: Starting two:
echo;sleep;fork;echo;sleep;fork;echo;sleep;fork
Restarting two:
stop;wait_for_pids;sleep;echo;sleep;fork;stop;wait_for_pids;sleep;echo;sleep;fork

So for each, to start it simply waits 1 second and then forks, for a total delay between instances of one second.
To stop, it's sending a signal, waiting for Sopel to exit, waiting one second, then in `start` it's waiting another second, then forking the new one. The total delay here is 2 seconds between each instance.

I still don't see why there needs to be twice the delay when restarting vs starting.

dgw:
Did this one fall by the wayside? I would defer to @half-duplex's expertise here, so just looking for the 👍 that it's ready and a squash of the intermediate commits + messages meeting guidelines.

PR #2341, opened 2022-08-13
bot, coretasks, irc, plugin, plugins: new capability negotiation system

### LINKED ISSUE #972: Sopel blindly requests unadvertised CAPs
Hi,

I noticed Sopel requests `multi-prefix`, even if it is not in the `CAP LS` response.
This is technically valid according to the spec, but you have code in `coretask.py` that looks supposed to filter capabilities in such cases. I guess its purpose is to avoid getting a `CAP NAK` if a single capability in the list is not available.

raw log file:

```
>>1450516159.3153656    CAP LS 302
>>1450516159.3155537    NICK Sopel
>>1450516159.315693 USER sopel +iw Sopel :Sopel: http://sopel.chat
<<1450516159.3162677    CAP * LS :
>>1450516159.31694  CAP REQ multi-prefix
>>1450516159.3170621    CAP END
```

Version: 047a99ed2d3f76c7d54e5bacaac1c1d38d1b743c (Git master at the time I am writing this)

### LINKED ISSUE #974: Stuck if configured to use SASL but server NAKs the capability
Config file:

```
[core]
nick = Sopel
host = 0.0.0.0
use_ssl = false
port = 39154
owner = me
channels = 
sasl_username = jilles
sasl_password = sesame
sasl_mechanism = PLAIN
auth_method = sasl
```

Raw logs:

```
>>1450542318.5877385    CAP LS 302
>>1450542318.5879178    NICK Sopel
>>1450542318.5880473    USER sopel +iw Sopel :Sopel: http://sopel.chat
<<1450542318.588655 CAP * LS :
>>1450542318.588997 CAP REQ multi-prefix
>>1450542318.589132 CAP REQ sasl
<<1450542318.6277814    CAP * NAK :multi-prefix
<<1450542318.6675858    
<<1450542318.6681972    CAP * NAK :sasl
>>1450542560.3821402    PING 0.0.0.0
```

Version: 047a99ed2d3f76c7d54e5bacaac1c1d38d1b743c (Git master at the time I am writing this)

Exirel:
### Description

Fix  #972 and should fix  #974 as well.

I rewrote the entire capability negotiation feature.
I haven't tested it live properly, and I want to add so much more tests, this will have to wait.

Don't review the docstring yet or you are going to have a bad time, as I'm going to rebase/squash this multiple times until I'm happy with the result.

This PR exists so anyone can take a chance to test it live on their own setup.

### Checklist


Exirel:
So, it's far from it, but it's getting there:

* I've added way more tests, both for sasl support and cap request en general
* I've find a way to make the system more or less backward compatible, with some caveat, with `bot.cap_req()` (which is a nightmare)
* I still need to write some good documentation on how to use all of that, probably in the "advanced tips and tricks" because cap negotiation is *not* for the faint of heart

We are reaching slowly, but surely, the 60% coverage rate!

sopel/plugins/capabilities.py
```diff
@@ -0,0 +1,312 @@
...
+            }
+            text = ' '.join(cap_req)
+            if cap_req_enables <= capabilities:
+                self._registered[cap_req] = {
+                    plugin_name: (handler_info[0], False)
+                    for plugin_name, handler_info in plugin_requests.items()
+                }
+                self._requested.add(cap_req)
+                LOGGER.debug('Capability negotiation request: "%s".', text)
+                bot.write(('CAP', 'REQ'), text=text)
```
dgw, line 154: I would like to propose handling `-unadvertised-cap` by just omitting that from the REQ, instead of skipping the request.

A CAP that _can't_ be enabled because the server doesn't support it should yield the same server behavior as explicitly disabling it on a server that _does_ support it.

(Also, `cap_req_enables` is a confusing var name. How about something like `absolute_caps`, after the `abs()` built-in function stripping the sign off a number?)

sopel/plugins/capabilities.py
```diff
@@ -0,0 +1,312 @@
...
+            else:
+                LOGGER.debug(
+                    'Unable to negotiate capability request: "%s".', text,
+                )
```
dgw, line 158: Am I missing where the callback for a request that is never sent gets notified that the capability(/ies) asked for were essentially NAK'd due to one or more not being available?

I'm not certain it's really more efficient to proactively avoid sending REQs that will obviously be NAK'd like this. Wouldn't it be easier to send the request anyway and let the server send a NAK that would propagate through the callbacks as normal?

sopel/plugins/capabilities.py
```diff
@@ -0,0 +1,312 @@
...
+        denied instead.
+
+        If the capability wasn't requested, the result will be ``None``.
+        """
+        # nothing to deny
+        if cap_req not in self._requested:
+            LOGGER.debug('Received CAP NAK for an unknwon CAP REQ.')
+            return None
+
+        # update denied: callbacks may fail, server NAK ninetheless
```
dgw, line 255:
```suggestion
        # update denied: callbacks may fail, server NAK nonetheless
```

sopel/plugins/capabilities.py
```diff
@@ -0,0 +1,312 @@
...
+    from sopel.bot import Sopel, SopelWrapper
+    from sopel.plugin import capability, CapabilityNegotiation
+
+
+LOGGER = logging.getLogger(__name__)
+
+
+class Manager:
+    """Manager of Capability Requests."""
+    def __init__(self):
```
dgw, line 29: For something like this, that's designed to be instantiated as part of another object (a bot, in this case), is there anything besides the possible added difficulty of unit tests that makes you not want to make the parent object provide itself as a constructor argument, to avoid needing `bot` as method args later?

dgw (commented):
OK, I'm going to stop this round before I get absolutely lost in `test/`. There's _a lot_ of new stuff in there, but as long as the suite passes there's no reason to spend time looking in there yet.

Some big "what if" and "why not" questions from me. I'll probably be excoriated for even thinking some of them—and with good reason 😁—but I hope that overall, this collection of feedback will be useful in polishing the approach here.

sopel/plugins/handlers.py
```diff
@@ -348,7 +356,19 @@ def has_setup(self):
...
+            for module_attribute in vars(self._module).values()
+            if isinstance(module_attribute, plugin_decorators.capability)
+        ]
+
+    def register(self, bot: Sopel) -> None:
+        # capabilities are directly registered
+        for cap_request in self.get_capability_requests():
+            bot.cap_requests.register(self.name, cap_request)
+
+        # plugin callables go through ``bot.add_plugin``
```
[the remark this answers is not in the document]
Exirel: Busted: this is absolutely a not-so-secret way to migrate from the old-style `bot.add_plugin` method to a version where the Plugin handler exposes properly its plugin callables (rules, commands, jobs, URLs, and now CAP requests).

sopel/coretasks.py
```diff
@@ -55,10 +60,105 @@
...
-batched_caps = {}
+def _auth_capability(
+    cap_req: Tuple[str, ...], bot: SopelWrapper, acknowledged: bool,
+) -> plugin.CapabilityNegotiation:
+    if acknowledged:
+        return plugin.CapabilityNegotiation.DONE
+
+    name = ', '.join(cap_req)
+    owner_account = bot.settings.core.owner_account
+    admin_accounts = bot.settings.core.admin_accounts
```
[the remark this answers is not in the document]
Exirel: Readability! So I don't have to use `bot.settings.core.*` everywhere.

sopel/irc/capabilities.py
```diff
@@ -0,0 +1,198 @@
...
+        assert trigger.args[1] == 'LS'
+
+        # extracting capabilities
+        for available_capability in trigger.split():
+            name, *params = available_capability.split('=', maxsplit=1)
+            self._available[name] = params[0] if params else None
+
+        # return if multiline or not
+        multiline = trigger.args[2] == '*'
+        return not multiline
```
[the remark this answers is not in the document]
Exirel: Yes, and:

* this is the LS command, which has a proper protocol for multi-line
* I don't want to support multi-line CAP requests, plugin should know better than doing that

There is nothing in Sopel that makes it easy to handle multi-line CAP REQ/CAP ACK. And I can live with that limitation.

sopel/bot.py
```diff
@@ -877,6 +878,119 @@ def _update_running_triggers(self, running_triggers: list) -> None:
...
+
+        self._cap_requests_manager.request_available(
+            self, available_capabilities)
+
+        return bool(self._cap_requests_manager.requested)
+
+    def resume_capability_negotiation(
+        self, cap_req: Tuple[str, ...],
+        plugin_name: str,
+    ) -> None:
```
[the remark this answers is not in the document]
Exirel: **I KNOW RIGHT???**

It's on my roadmap, but it's far in the future (think: Sopel 11 or 12 maybe).

sopel/coretasks.py
```diff
@@ -893,175 +1001,144 @@ def track_quit(bot, trigger):
...
+    bot: SopelWrapper,
+    cap_req: Tuple[str, ...],
+    results: List[Tuple[bool, Optional[plugin.CapabilityNegotiation]]],
+    was_completed: bool,
+) -> None:
+    if any(
+        callback_result[1] == plugin.CapabilityNegotiation.ERROR
+        for callback_result in results
+    ):
+        # error: a plugin needs something and the bot cannot function properly
```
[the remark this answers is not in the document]
Exirel: The problem here is that now, the plugin *is* loaded, and its callables are active. If a plugin cannot properly handle CAP REQ, and either crash or explicitly returns `CapabilityNegotiation.ERROR`, I think it is best for the bot to stop here and let the bot owner figure out what is what.

Beside, this should be a very rare scenario. Remember that most plugins just don't use CAP REQ. We are working with edge cases here.

sopel/plugins/capabilities.py
Exirel: > besides the possible added difficulty of unit tests

Not "possible": very real and very annoying. And yes there is! Beside the difficulty (more like: complexity, because state and stuff), it is already very annoying to deal with the bot's interface alone, especially when split between the `AbstractBot` base class (that handles only IRC-related stuffs) and the `Sopel` class (that handles the plugins & various setups).

Splitting that section into its own object makes it much simpler to work with, since it's simpler to reason on a very specific subset of features. This also prevents a bunch of name conflict. This is, by the way, the same reason we have a `bot.rules` object that proxy all the plugin callables management for the bot.

sopel/plugins/capabilities.py
Exirel: Well, that's the point of #972 isn't it?

Exirel:
> Some big "what if" and "why not" questions from me. I'll probably be excoriated for even thinking some of them—and with good reason 😁—but I hope that overall, this collection of feedback will be useful in polishing the approach here.

I took the time to read everything, and I think I tried to reply to most of the important things.

I think most of the questions are about a very specific issue with the protocol's specification: what does it mean to validate a request? Is a request that disable something not available is valid, or not valid?

### single cap request vs multi cap request

It is very easy to deal with single-cap request: it either works, or it doesn't. The problems come only when there are multi-cap requests:

* `cap1 cap2`
* `cap1 -cap2`
* `-cap1 -cap2`

From reading the spec, it is entirely possible for a server to NAK the request `cap1 cap2 cap3`, while ACK three independent requests (one for each cap). Does it make sense? Not really, but that doesn't change the fact that these are *not* the same.

That's also why I don't want Sopel to send one single `CAP REQ` with all the capability at once, because that's not how the bot uses these capabilities, and it makes no sense to ACK/NAK all of them as a whole.

### callback anti-pattern

While working on that, I discovered some sort of anti-pattern: registering a handler for a CAP REQ that changes the state of the bot (or its memory), or load something. This is, except for one exception, a bad idea.

The proper way to handle when a capability is enabled or disabled is to use the method `bot.capabilities.is_enabled()` whenever it is needed, instead of storing that information in `bot.memory` (or a side-effect). It looks cool, but it is actually more a problem than anything else.

It is especially true for capability that are **not available**, so the bot should not request them, and so it won't received either an ACK nor a NAK. In that case, the callback **won't** be called.

The good reasons I've seen so far are:

* log the state of the capability, and possibly emit a warning for the bot's owner
* engage with the IRC protocol, such as SASL authentication, that must be performed while negotiating capabilities (this is related to #974)

I think we should agree on that, or find a consensus about that, and then document it (probably in an "advanced" chapter of the documentation).

sopel/plugins/handlers.py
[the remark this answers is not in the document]
dgw: Maybe annotate `bot.add_plugin` with a TODO to that effect (that it's intended to be replaced)?

I think it's great for the plugin handler classes to take care of collecting the relevant callables, eventually. Just important to remember that `bot.add_plugin` is _technically_ part of a public API, and therefore it needs to be marked as deprecated (or otherwise flagged as "will change") well in advance.

sopel/irc/capabilities.py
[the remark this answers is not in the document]
dgw: The thing about following specs is, you follow the _whole_ spec and leave nothing out, or you can't say you're compliant. You and I might be just fine with this limitation, but anything that the spec allows SHOULD be handled. (See what I did there? 😁)

sopel/bot.py
```diff
@@ -877,6 +878,119 @@ def _update_running_triggers(self, running_triggers: list) -> None:
...
+            :class:`sopel.plugin.capability` for a replacement.
+
+        .. warning::
+
+            This method must not be used. This will emulate the old behavior
+            by adding a :class:`sopel.plugin.capability` with a wrapper around
+            ``success_callback`` and ``failure_callback``, however the behavior
+            is not the same as before. The callback won't be called if the
+            request is never made, i.e. never REQ, ACK, or NAK, as the new
+            system doesn't try to emulate the server's response.
```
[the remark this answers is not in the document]
dgw: Somehow I doubt that the existing method is used by very many third parties. Of course I would prefer to leave it alone, working as-is in 7.x, marked as deprecated for 8.0 and removed in 9.0—but it's been long enough that I don't remember how much Other Stuff was required to make it work (that has been deleted or changed by this PR).

sopel/plugins/capabilities.py
dgw: Yeah, I could have asked the same question about other stuff that's been added/refactored in the last few versions, but I guess I didn't think of it.

Passing `bot` around all the time is _kind of_ annoying in the same way as always having to restate `plugin_name` (as said in that other thread here), but not as bad because it's something you're always just going to _have_ and not a string that has to be kept up-to-date if the name of something changes. (Well, OK, it _could_ change if someone's being silly with their args, but you know what I meant.)

Just to be clear, what I had in mind was (consider this as pseudo-code):

```py
# in the bot's init:
self._capability_manager = capabilities.Manager(bot=self)

# in the Manager's init
self._bot = bot

# possibly also as part of the Manager
@property
def bot(self):
    return self._bot

# then use the attr/property in Manager methods,
# instead of passing bot as an arg
```

Especially the last part of your reply sounds like you were reacting to a suggestion that this Manager be moved into the bot/Sopel class somehow, which is definitely _not_ what I meant to suggest! 😅 The modularization is good!

sopel/plugins/capabilities.py
dgw: Tested with Libera via telnet (1337 h4x0r skills activated) and indeed, `CAP REQ :account-notify -cap2` (obviously Libera doesn't advertise anything named `cap2`) results in a NAK.

But here's the thing: If a server doesn't even know about a given CAP, it's not going to implement the functionality that CAP would enable, whether or not it's asked to do so. Sending a server that doesn't support `cap2`, `CAP REQ :cap1 cap3`, _should_ yield the same behavior in the end as sending a server that _does_ support it `CAP REQ :cap1 -cap2 cap3`. In both cases (assuming a CAP ACK response), the `cap1` and `cap3` behavior will be active, and the `cap2` behavior will be inactive, just as the `-` in the request demands.

This is the sort of thing that could benefit from asking the IRCv3 folks, though. I'm not up for a conversation on that today, personally, but one of us really should dig into this a bit more before making a decision.

sopel/plugins/capabilities.py
dgw: Ultimately I don't really care whether the REQ actually gets sent. I remember #972 (and that was one reason this rewrite is even happening).

I just didn't see any other code path that would give the requester feedback when a request is simply not sent due to one or more of the capabilities in it not being advertised, so suggested the easy way out (send the REQ anyway and let the NAK propagate naturally through the paths I did see).

sopel/coretasks.py
```diff
@@ -55,10 +60,105 @@
...
+CAP_SERVER_TIME = plugin.capability('server-time')
+CAP_USERHOST_IN_NAMES = plugin.capability('userhost-in-names')
+CAP_MESSAGE_TAGS = plugin.capability('message-tags')
+CAP_ACCOUNT_NOTIFY = plugin.capability(
+    'account-notify', handler=_auth_capability)
+CAP_EXTENDED_JOIN = plugin.capability(
+    'extended-join', handler=_auth_capability)
+CAP_ACCOUNT_TAG = plugin.capability(
+    'account-tag', handler=_auth_capability)
+CAP_SASL = plugin.capability('sasl', handler=_sasl_capability)
```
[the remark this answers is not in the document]
dgw: Clearly I sat on this for a while. 😁 Didn't want to rush into saying any more on the topic, as it's clearly a _very big_ part of how this will look from a plugin author's perspective.

And in that time, I landed on _what_ bugs me about how it looks now: The top level of the plugin's module gets cluttered.

So the second way you listed to expose properties (an object/class or `Protocol`) sounds much better to me. Heck, if we land on something that can be adapted to a "version 2" of the `StaticSection` idiom for settings, even more better.

I think a main sticking point for us would be introducing another reserved name like `setup`, `shutdown`, and `configure`, right? Even if it's a reserved class name, not a function name. Leaving aside that it's _probably_ fine to add a reserved name in a major version, let me run with the alternatives for a bit. Hope you don't mind a bit of brainstorming-out-loud.

### Option 1: Class registered at `setup()`

- In brief: Like `config.types.StaticSection`, but for capability requests.
- Usage model: Plugin defines a `PluginNameCAPs` class (or whatever the convention ends up being), then registers it in `setup()` like how `PluginNameSection` classes get registered for settings now. Class members could look like the top-level module attributes do now in this revision of `coretasks`, i.e. each member/attr is a `plugin.capability`.
- **Pros:**
  - Familiar idiom, to anyone who's written even a moderately complex Sopel plugin
  - Creates only one name at the module top level
  - Can be named anything the author wants
  - Multiple instances possible, if someone wants to go totally nuts
    - For example, register the plugin's settings first and then conditionally add CAP requests based on the loaded settings
- **Cons:**
  - I (at least) want to move _away_ from that exact model of how config sections are registered in `setup()`
    - Not sure yet if the CAP version of this approach would require passing the plugin's name again, but that's something I/we don't like about `bot.config.add_section()`
  - I had included "naming attrs is hard" (but more verbose/technical) here, but then realized that the attr names literally don't matter as long as the attrs are self-describing, i.e. `plugin.capability` objects

### Option 2: Abstract class to extend in plugin

- In brief: Also like `config.types.StaticSection`, but no need to register in `setup()`.
- Usage model: Plugin defines `PluginNameCAPs(plugin.Capabilities)` (whatever the names are) and the loader/plugin-handler auto-collects all subclasses of the core-defined type. Members of this class are the same as Option 1.
- **Pros:**
  - Also creates only one name at the module top level
  - Also can be named anything
  - Also multiple instances possible
    - Less useful here because the `bot.settings` won't be available for conditional enabling/disabling of individual requests
  - Could be adapted to plugin config sections pretty easily (`config.types.AutoSection` in Sopel 9.0, anyone?)
    - The plugin loader/handler already knows the plugin's name, so this can eliminate needing to repeat it when calling `bot.config.add_section()`
- **Cons:**
  - …I'm sure they exist, but this fixes the cons I listed for Option 1 and I couldn't think of any specific to Option 2.

### Conclusion… for now?

I didn't write up an option for a reserved name because it would be basically the same as Option 2 but with a predefined name the plugin author has to use instead of extending a specific class/interface.

Now I wait and see what (if anything) survives The Exirel Gauntlet, i.e. how many of these ideas are Bad™ based on your much greater past experience in software engineering. 🙃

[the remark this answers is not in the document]
Exirel: Since I spent times away from writing code for a while, I had time to think about this. I still want the same thing, i.e. declaration of plugin's features at import time, not by running `setup`. And we still have the same constraint on CAP: requesting cap1 and cap2 separately is *not* the same thing as requesting them together.

This means that using a single class would be quite annoying: you would have to do something like that:

```python
class PluginCapRequests():
    CAP_REQUESTS = (
        ('cap1',),
        ('cap2', 'cap3'),
    )

    def cap_is_ack(self, cap_req):
        # check what is in cap_req so you know which CAP_REQUESTS is ack
```

All it does is moving complexity into the handling custom code. Another option would be to do something like:

```python
class PluginCapRequests():
    @plugin.capability('cap1')
    def cap1_is_ack(self, ...):
        ...

    @plugin.capability('cap2', 'cap3')
    def cap2_3_is_ack(self, ...):
        ...
```

Which is exactly like having a bunch a module-level attribute, but now you *must* have a CAP handler to decorate.

Last but not least: how many plugins are going to need that? Like, I get that you don't like the view, but there is only one plugin that need that: `coretasks`. This clutter is ours only, for a feature that is so specific, so advanced, I don't see how the average plugin author will get confused with it.

So, even if I considered changing to something slightly different for a moment, I would rather stick with what I suggest in this PR for now.

sopel/irc/capabilities.py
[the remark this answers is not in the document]
Exirel: Can you show me an example of that multi-line? Because the spec is a bit light on what a multi-line CAP REQ/ACK is.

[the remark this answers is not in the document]
Exirel: Hey, I found a [recommendation in the spec](https://ircv3.net/specs/extensions/capability-negotiation.html#the-cap-req-subcommand):

> Clients SHOULD ensure that their list of requested capabilities is not too long to be replied to with a single ACK or NAK message. If a REQ’s final parameter gets sufficiently large (approaching the 510 byte limit), clients SHOULD instead send multiple REQ subcommands.

So, Sopel is the Client. Sopel SHOULD instead send multiple REQ subcommands. So when I say: this is a limitation of Sopel, i.e. the Client, I think it's acceptable and in accordance with the specification.

sopel/plugins/capabilities.py
Exirel: I'm really against this, for two reasons:

* This is clearly a edge case, and for CAP, I'm all for accepting limitation on how Sopel interpret what is requested from plugins. If someone wants to remove a Capability while requesting others at the same time, they have the option to split their request, or to accept that their request will be sent "as is", meaning will be accepted or denied based on the specification.
* This is **really* hard to keep track internally. The current tracking is already quite complex because of the protocol, adding yet another level of complexity like "this is my request but you have to map this to other request that contains a specific edge case" is just... hugh... no. I really don't want to do that.

This system is really complex, and I don't want to make it even more complex by having developers getting confused over what is actually send or not to the server: what you see is what you get. The capability is not advertised by the server? The request should not be sent. Otherwise, it can be sent, as it is, and it may or may not fail *all at once*, as specified.

sopel/plugins/capabilities.py
Exirel: I understand that. In my opinion, Plugins should react to ACK and NAK only if it is relevant to the negotiation phase. If it isn't, the rules/commands/event handlers should look at the bot's state directly, and not some derived value from the ACK/NAK handler.

What I mean is that (and I may need to change the doc to explain that further/properly) is that instead of doing this:

```python
@plugin.capability('cap1')
def cap1_handler(cap_req, bot, acknowledged):
    if acknowledged:
        bot.memory['my_plugin_cap1'] = True
    else:
        bot.memory['my_plugin_cap1'] = False

@plugin.rule('something'):
def rule_something(bot, trigger):
    if bot.memory['my_plugin_cap1']:
        # do something
```

They should do this:

```python
CAP_CAP1 = plugin.capability('cap1')

@plugin.rule('something'):
def rule_something(bot, trigger):
    if bot.capabilities.is_enabled('cap1'):
        # do something
```

The difference is on how the bot's state is verified: in the first version, a proxy variable is used to know the state (the `bot.memory` flag), and a proxy needs to be synchronized. In the second version, the current bot's state is always used, there is no need for any sort of synchronization, so less risk of errors.

Exirel:
> That's a core function, of course, and so it needs to be supported somehow.

And it is! So we are good on that front.

I've re-read all the comments, and now I think I'm confident with my current design. I have some improvements in mind, such as:

* allowing a callback for non-advertised CAP
* allowing conditional CAP REQ, i.e. "don't actually send this CAP REQ if config/env is not OK", to prevent requesting SASL when the configuration doesn't need it, for example
* an advanced documentation chapter for CAP REQ handling

Given your blessing to make this as provisional, I may suggest to just go with it for now, and see how we can improve that further. For instance, even thought I'm against, working around the "this CAP is not available so -CAP should not be a problem" thing is something that we may want to work on. We could also change the way to declare CAP REQ in a plugin code.

I'll try to fix all the spelling & docstring issues today.

sopel/plugins/capabilities.py
Exirel: Ha! I see. The thing is, in most case we are not interested in the instance of `Sopel` itself, but in a `SopelWrapper` instance, which is contextual and not available when instantiating the `Manager`. So for now, I won't change that.

sopel/coretasks.py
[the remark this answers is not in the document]
Exirel: It is doable, yes, and yet I'm reluctant to allow a bot to run with a broken plugin that soon in the launch process: unlike many other features, capability negotiation can have a big impact on the server & client's behaviors. As a result, I think there are bigger risks here for security.

For instance, if something mess up with the SASL authentication, or any kind of authentication, a bot owner may not realize that its bot instance is not secure at all, because the plugin that was supposed to handle that with CAP REQ is not loaded at all. It's unlikely to be a problem, but the risk is here.

In that case, I don't think Sopel should unload the plugin and carry on, I think it should tell the bot owner that something is wrong with their setup, and they must act on it.

sopel/plugins/capabilities.py
Exirel: Quick note on how to track CAP REQ with ACK/NAK: it is all done with simple string manipulation, there is no way to mark a CAP REQ as "this is request nu°7814" and link that number to a callback handler internally. We have to check if a tuple of strings matches anything in memory: having to keep track of modified CAP REQ is just too much.

docs/source/plugin/advanced.rst
```diff
@@ -64,3 +64,115 @@ a QUIT/PART event, and they will be available after a JOIN event.
 
 Note that you don't specifically need to use ``@plugin.thread(False)``, but
 it is still recommended to prevent any race condition.
+
+
+Capability Negociation
```
dgw, line 69:
```suggestion
Capability negotiation
```

Case change because it seems that we haven't used Title Case for this level of heading elsewhere in the page.

docs/source/plugin/advanced.rst
```diff
@@ -64,3 +64,115 @@ a QUIT/PART event, and they will be available after a JOIN event.
...
+======================
+
+Capability negociation is a feature of IRCv3 that allows a server to advertise
```
dgw, line 72:
```suggestion
Capability negotiation is a feature of IRCv3 that allows a server to advertise
```

docs/source/plugin/advanced.rst
```diff
@@ -64,3 +64,115 @@ a QUIT/PART event, and they will be available after a JOIN event.
...
+a list of optional capabilities, and allows its clients to request such
+capabilities. You can see that as feature flags, activated by the client.
+
+Capability negociation takes place after:
```
dgw, line 76:
```suggestion
Capability negotiation takes place after:
```

docs/source/plugin/advanced.rst
```diff
@@ -64,3 +64,115 @@ a QUIT/PART event, and they will be available after a JOIN event.
...
+
+* connecting to the IRC server
+* client's identification (``USER`` and ``NICK``)
+
+And before:
+
+* the ``RPL_WELCOME`` event (001)
+* ``ISUPPORT`` messages
+* client's authentification (except for SASL, which occurs in the capability
+  negociation phase)
```
dgw, line 86:
```suggestion
* client's authentication (except for SASL, which occurs in the capability
  negotiation phase)
```

'authentification' _is_ a word (despite my browser's spell-checker flagging it), but its meaning is not what's meant here.

docs/source/plugin/advanced.rst
```diff
@@ -64,3 +64,115 @@ a QUIT/PART event, and they will be available after a JOIN event.
...
+
+.. warning::
+
+    This is a very advanced feature, and plugin authors should understand how
+    capability negociation works before using it. Even if Sopel tries to make
```
dgw, line 91:
```suggestion
    capability negotiation works before using it. Even if Sopel tries to make
```

docs/source/plugin/advanced.rst
```diff
@@ -64,3 +64,115 @@ a QUIT/PART event, and they will be available after a JOIN event.
...
+    limitations and possible caveats.
+
+Declaring requests: the ``capability`` decorator
+------------------------------------------------
+
+In :mod:`sopel.plugin` there is an advanced :class:``capability`` decorator: it
+is a class that acts like a decorator, and can be used with or without a
+callback to decorate::
+
+    # plugin file
```
dgw, line 102:
```suggestion
    """Sample plugin file"""

```

Module-level docstring is probably clearer than a comment that looks attached to the import. (Non-blocking suggestion.)

docs/source/plugin/advanced.rst
```diff
@@ -64,3 +64,115 @@ a QUIT/PART event, and they will be available after a JOIN event.
...
+* client's authentification (except for SASL, which occurs in the capability
+  negociation phase)
+
+.. warning::
+
+    This is a very advanced feature, and plugin authors should understand how
+    capability negociation works before using it. Even if Sopel tries to make
+    it as simple as possible, plugin authors should be aware of the known
+    limitations and possible caveats.
+
```
dgw, line 94:
```suggestion


```

docs/source/plugin/advanced.rst
```diff
@@ -64,3 +64,115 @@ a QUIT/PART event, and they will be available after a JOIN event.
...
+.. autoclass:: sopel.plugin.CapabilityNegotiation
+   :members:
+
+
+Working with capabilities
+-------------------------
+
+A plugin that requires capabilities, or that can enhance its features with
+capabilities, should rely on the bot's
+:attr:`capability manager <sopel.irc.AbstractBot.capabilities>`::
```
dgw, line 126: Not strictly related to this PR, but contextually relevant: Did we ever figure out how to make x-refs like this appear in not-monospace? When the link text is overridden to something that's not a code object, it looks odd.

docs/source/plugin/advanced.rst
```diff
@@ -64,3 +64,115 @@ a QUIT/PART event, and they will be available after a JOIN event.
...
+        if bot.capabilities.is_enabled('cap1'):
+            # be fancy with enhanced capabilities
+        else:
+            # stick to the basics
+
+The manager's :meth:`~sopel.irc.capabilities.Capabilities.is_enabled` in
+particular is the most interesting, as it allows a plugin to always know if a
+capability is available or not.
+
+Ending negociations
```
dgw, line 139:
```suggestion


Ending negotiations
```

docs/source/plugin/advanced.rst
```diff
@@ -64,3 +64,115 @@ a QUIT/PART event, and they will be available after a JOIN event.
...
+-------------------
+
+Sopel automatically sends a ``CAP END`` message when all requests are handled.
+However in some case, a plugin author may need to delay the end of the
+negociation to perform an action that must be done first.
```
dgw, line 144:
```suggestion
However in some cases, a plugin author may need to delay the end of CAP
negotiation to perform an action that must be done first.
```

docs/source/plugin/advanced.rst
```diff
@@ -64,3 +64,115 @@ a QUIT/PART event, and they will be available after a JOIN event.
...
+
+In that case, plugin author must return
+:attr:`~sopel.plugin.CapabilityNegotiation.CONTINUE` in their callback.
```
dgw, line 147:
```suggestion
In that case, a plugin must return
:attr:`~sopel.plugin.CapabilityNegotiation.CONTINUE` in its callback.
```

docs/source/plugin/advanced.rst
```diff
@@ -64,3 +64,115 @@ a QUIT/PART event, and they will be available after a JOIN event.
...
+
+This is the case for the SASL authentification, as seen in the ``coretasks``
```
dgw, line 149:
```suggestion
This is the case for the SASL authentication, as seen in the ``coretasks``
```

Same as before, real word with a different meaning than intended.

docs/source/plugin/advanced.rst
```diff
@@ -64,3 +64,115 @@ a QUIT/PART event, and they will be available after a JOIN event.
...
+    @plugin.unblockable
+    @plugin.priority('medium')
+    def sasl_success(bot: SopelWrapper, trigger: Trigger):
+        """Resume capability negotiation on successful SASL auth."""
+        LOGGER.info("Successful SASL Auth.")
+        bot.resume_capability_negotiation(CAP_SASL.cap_req, 'coretasks')
+
+.. important::
+
+    Note that these plugin callable should always use ``@plugin.thread(False)``
```
dgw, line 177:
```suggestion
    Plugin callables that modify the bot's capability negotiation state should always use ``@plugin.thread(False)``
```

No rewrap because I'm lazy and the PR already has a bunch of stuff that needs squashing down anyway.

sopel/irc/capabilities.py
```diff
@@ -1,4 +1,21 @@
-"""Management of IRC server capabilities."""
+"""Tracking of IRC's capability negotiation.
```
dgw, line 1:
```suggestion
"""Management of IRC capability negotiation.
```

Splitting the difference between old and new. 😁

sopel/irc/capabilities.py
```diff
@@ -18,21 +35,70 @@
...
 class CapabilityInfo(NamedTuple):
-    """Capability metadata."""
+    """Capability metadata.
+
+    This contains the details of a capability: if :attr:`available`,
+    :attr:`enabled`, and its :attr:`params` (if advertised).
+
+    .. note::
+
+        You can get a capability info through
```
dgw, line 45:
```suggestion
        You can get a capability's info through
```

sopel/irc/capabilities.py
```diff
@@ -18,21 +35,70 @@
...
+        :meth:`Capabilities.get_capability_info`.
+
+    """
     name: str
+    """Name of the capability.
+
+    The name of a capability is the name as it appears in the ``CAP LS``
+    subcommand such as ``multi-prefix`` or ``sasl``.
```
dgw, line 53:
```suggestion
    subcommand, such as ``multi-prefix`` or ``sasl``.
```

sopel/irc/capabilities.py
```diff
@@ -18,21 +35,70 @@
...
+    """
     params: Optional[str]
+    """Advertised parameters for this capability.
+
+    When a server support ``CAP`` version 302, capabilities can have
```
dgw, line 58:
```suggestion
    When a server supports ``CAP`` version 302, capabilities can have
```

sopel/irc/capabilities.py
```diff
@@ -18,21 +35,70 @@
...
+    parameters. The format and the meaning of the parameters depend on the
+    capability itself.
+
+    For example, the ``sasl`` capability can provide the list of SASL
+    mechanisms the server support, such as ``PLAIN,EXTERNAL``.
```
dgw, line 63:
```suggestion
    mechanisms the server supports, such as ``PLAIN,EXTERNAL``.
```

sopel/irc/capabilities.py
```diff
@@ -18,21 +35,70 @@
...
     def __init__(self) -> None:
         self._available: Dict[str, Optional[str]] = {}
         self._enabled: Set[str] = set()
 
     def get_capability_info(self, name: str) -> CapabilityInfo:
-        """Retrieve meta-data about a capability."""
+        """Retrieve metadata about a capability.
+
+        The returned :class:`CapabilityInfo` will tell if the capability is
+        advertised by the server, and if so its parameters and if it is
```
dgw, line 96:
```suggestion
        advertised by the server; and if so, its parameters and whether it is
```

sopel/irc/capabilities.py
```diff
@@ -43,12 +109,22 @@ def get_capability_info(self, name: str) -> CapabilityInfo:
 
     @property
     def available(self) -> Dict[str, Optional[str]]:
-        """Dict of available server capabilities."""
+        """Dict of available server capabilities.
+
+        Each key is the name of a capability advertised by the server, and each
+        key is the parameters as advertised (if any) for this capability.
```
dgw, line 115:
```suggestion
        value is the parameters as advertised (if any) for this capability.
```

I assume this is not a special dict in which each item has two keys and no values 😉

sopel/plugins/capabilities.py
```diff
@@ -1,4 +1,15 @@
...
+.. versionadded:: 8.0
+
+.. important::
+
+    This is all relatively new. Its usage and documentation is for Sopel core
+    development and advanced developers. It is subject to rapid changes
+    between versions without much (or any) warning.
+
+    Do **not** build your plugin based on what is here, you do **not** need to.
+"""
```
dgw, line 12:
```suggestion
"""

```

sopel/plugins/capabilities.py
```diff
@@ -25,7 +36,29 @@
 
 
 class Manager:
-    """Manager of Capability Requests."""
+    """Manager of plugin's capability requests.
```
dgw, line 39:
```suggestion
    """Manager of plugins' capability requests.
```

The manager manages _all_ plugins' requests, and not one manager instance for each plugin, right?

(Can also just be "plugin capability requests" with no possessive at all.)

sopel/plugins/capabilities.py
```diff
@@ -25,7 +36,29 @@
...
+
+    Whenever a plugin declares a capability request (through
+    :class:`sopel.plugin.capability`), the bot will register the request with
+    this manager.
+
+    The bot is responsible to call the manager at the appropriate time, and the
+    manager will store requests' state and handle acknowledgement, by follwing
```
dgw, line 46:
```suggestion
    manager will store requests' state and handle acknowledgement, by following
```

sopel/plugins/capabilities.py
```diff
@@ -25,7 +36,29 @@
...
+    this workflow:
+
+    1. the bot register the plugin's requests with the :meth:`register` method
```
dgw, line 49:
```suggestion
    1. the bot will register plugins' requests with the :meth:`register` method
```

sopel/plugins/capabilities.py
```diff
@@ -41,22 +74,65 @@ def __init__(self):
...
-        """Set of registered capabilities."""
+        """Set of registered capability requests.
+
+        Each element is a capability request as a tuple of capability names::
+
+            >>> manager.registered
+            {('cap1',), ('cap2', 'cap3')}
+
+        A registered requests is a request wanted by a plugin. The request may
+        or may not be requested, acknowledged or denied.
```
dgw, line 85:
```suggestion
        A registered request is a request wanted by a plugin. The request may
        or may not be requested, acknowledged, or denied.
```

sopel/plugins/capabilities.py
```diff
@@ -129,11 +228,40 @@ def request_available(
...
-        Requests made are stored as requested, other are ignored.
+        Requests made are stored as requested; others are ignored::
+
+            >>> # manager.register => cap1 and cap2, not cap3
+            >>> manger.request_available(bot, ('cap1', 'cap3'))
+            >>> manger.is_requested(('cap1',))
+            True
+            >>> manger.is_requested(('cap2',))
+            False
+            >>> manger.is_requested(('cap3',))
```
dgw, line 243:
```suggestion
            >>> manager.request_available(bot, ('cap1', 'cap3'))
            >>> manager.is_requested(('cap1',))
            True
            >>> manager.is_requested(('cap2',))
            False
            >>> manager.is_requested(('cap3',))
```

No eating involved here!

(This joke works for both the English and French definitions of "manger"!)

sopel/plugins/capabilities.py
```diff
@@ -129,11 +228,40 @@ def request_available(
...
+            capability only if it is advertised first. As a result, if a
+            request is never made, its callback will never be called, because
+            there won't be any ACK/NAK message from the server for them.
+
+            Plugin author should not use the request's callback to activate or
+            deactivate features, and instead check the bot's capabilities every
+            time they need to.
+
+            The only exception to that is when the plugin needs to perform an
+            operation while negociating the capabilities (such as SASL auth).
```
dgw, line 263:
```suggestion
            As a result, Sopel will send a request to enable or disable
            a capability only if it is advertised first. If a request is
            never made, its callback will never be called, because there
            won't be any related ACK/NAK message from the server.

            Plugin authors should not use the request's callback to activate or
            deactivate features, and instead check the bot's capabilities every
            time they need to.

            The only exception to that is when the plugin needs to perform an
            operation while negotiating the capabilities (such as SASL auth).
```

(Rewrap is sloppy, but good enough.)

sopel/plugins/capabilities.py
```diff
@@ -174,6 +302,21 @@ def resume(
...
         If the capability request cannot be found for that plugin, the result
         value remains the same (it stay uncomplete or complete)
+
+        .. important::
+
+            When a request's callback returns
+            :attr:`~sopel.plugin.CapabilityNegotiation.CONTINUE`, this method
+            must be called later (once the plugin has finished its job) or the
+            bot will never send the ``CAP END`` command, and the bot will hang
+            forever.
```
dgw, line 312:
```suggestion
            bot will never send the ``CAP END`` command and hang forever.
```

dgw (commented):
Scary line note count, but it's mostly typo fixes.

You were right when you said on IRC that this was waiting on me taking another look. I think for my part, when looking for things that needed attention in the PR list, I have been skipping anything that's still marked as a draft. It's probably time to Ready this one after this set of comments, isn't it?

docs/source/plugin/advanced.rst
Exirel: Oh. I think there is a trick, I'll see what I can do to make it less odd. The minimum I can do is to replace this by `bot.capabilities` to make it less odd.

sopel/plugins/capabilities.py
Exirel: I'm honestly lost with the usage of possessive or lack of possessive to express slightly different things. The nuances are too subtle for me to detect and I can't get around.

Exirel:
Everything that was typo-related has been fixed. Let me know if you want me to squash this before doing another review, other than that, it's ready on my end.

sopel/irc/capabilities.py
```diff
@@ -0,0 +1,274 @@
...
+        assert trigger.args[1] == 'ACK'
+
+        # extracting capabilities
+        ack_capabilities = tuple(sorted(trigger.split()))
+
+        # update enabled capabilities
+        for ack_capability in ack_capabilities:
+            if ack_capability.startswith('-'):
+                # ACK a disable request
+                cap_reversed = ack_capability.strip('-')
```
SnoopJ, line 186: Nitpick: as I understand it, it's possible for a particularly antisocial capability name to have a trailing dash, i.e. `nuisance-cap-`, which would be stripped away here. I'm not sure if a capability is allowed to have a leading dash, though? If so, we could slice here, but I think `lstrip()` probably satisfies my nitpick.

```suggestion
                cap_reversed = ack_capability.lstrip('-')
```

sopel/plugins/capabilities.py
```diff
@@ -0,0 +1,455 @@
...
+        :param request: a set of capabilities that form a capability request
+                        together; this can be any sortable iterable
+        """
+        return tuple(sorted(request)) in self._acknowledged
+
+    def is_denied(self, request: Iterable[str]) -> bool:
+        """Tell if a capability request is denied.
+
+        :param request: a set of capabilities that form a capability request
+                        together; this can be any sortable iterable
```
SnoopJ, line 187: All (finite) iterables are compatible with `sorted()`

```suggestion
    def is_registered(self, request: Iterable[str]) -> bool:
        """Tell if a capability request is registered.

        :param request: a set of capabilities that form a capability request
                        together; this can be any iterable
        """
        return tuple(sorted(request)) in self._registered

    def is_requested(self, request: Iterable[str]) -> bool:
        """Tell if a capability request is requested.

        :param request: a set of capabilities that form a capability request
                        together; this can be any iterable
        """
        return tuple(sorted(request)) in self._requested

    def is_acknowledged(self, request: Iterable[str]) -> bool:
        """Tell if a capability request is acknowledged.

        :param request: a set of capabilities that form a capability request
                        together; this can be any iterable
        """
        return tuple(sorted(request)) in self._acknowledged

    def is_denied(self, request: Iterable[str]) -> bool:
        """Tell if a capability request is denied.

        :param request: a set of capabilities that form a capability request
                        together; this can be any iterable
```

sopel/plugins/capabilities.py
```diff
@@ -0,0 +1,455 @@
...
+
+            >>> manager.is_requested(('cap1',))
+            False
+
+        See :meth:`request_available` to automatically request capabilities
+        advertised by the server.
+        """
+        self._registered.setdefault(request.cap_req, {})[plugin_name] = (
+            request, False,
+        )
```
SnoopJ, line 219: Nitpick: I would prefer this to be two statements with an additional name

```suggestion
        plugin_caps = self._registered.setdefault(request.cap_req, {})
        plugin_caps[plugin_name] = (
            request, False,
        )
```

SnoopJ (commented):
Submitting my second-opinion now, without having gone through the test suite with a fine-toothed comb.

Based on the number of my remarks that ended up being nitpicks and docs, I say this looks good to me. Some parts are a little tricky to follow but overall I 'get' it.

sopel/irc/capabilities.py
dgw: Ooh, that's a _very_ edge case. Is this worth bringing up in the #ircv3 channel or issue tracker? I went looking in [the spec](https://ircv3.net/specs/extensions/capability-negotiation) for a formal grammar or allowable charset, but found none; only that capability names MUST be treated opaque identifiers, and that names using lowercase-with-hyphens-between-words is merely a SHOULD.

sopel/plugin.py
```diff
@@ -71,6 +89,207 @@
...
+
+    def __call__(
+        self,
+        handler: Callable[
+            [Tuple[str, ...], SopelWrapper, bool],
+            CapabilityNegotiation,
+        ],
+    ) -> capability:
+        """Register a capability negotiation callback."""
+        self._handler = handler
```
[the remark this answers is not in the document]
Exirel: Oh, good catch!

sopel/plugins/handlers.py
```diff
@@ -348,12 +356,26 @@ def has_setup(self):
         """
         return hasattr(self._module, 'setup')
 
-    def register(self, bot):
+    def get_capability_requests(self) -> List[plugin_decorators.capability]:
+        return [
+            module_attribute
+            for module_attribute in vars(self._module).values()
+            if isinstance(module_attribute, plugin_decorators.capability)
+        ]
```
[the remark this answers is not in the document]
Exirel: You are right. The `capability` class is not made to be that flexible.

sopel/bot.py
```diff
@@ -395,6 +395,8 @@ def reload_plugins(self) -> None:
             LOGGER.info("Reloaded %s plugin %s from %s",
                         meta['type'], name, meta['source'])
 
+    # TODO: deprecate both add_plugin and remove_plugin
```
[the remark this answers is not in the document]
Exirel: See #2425 now!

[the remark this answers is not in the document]
SnoopJ: Awesome, thanks 😁
```suggestion
    # TODO: deprecate both add_plugin and remove_plugin; see #2425
```

sopel/irc/capabilities.py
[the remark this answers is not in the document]
Exirel: Re-reading this: the result is that Sopel should not even allow to send a CAP request that is too big!

sopel/plugin.py
```diff
@@ -120,6 +120,13 @@ class CapabilityNegotiation(enum.Enum):
     """
 
 
+if TYPE_CHECKING:
+    CapabilityHandler = Callable[
+        [Tuple[str, ...], SopelWrapper, bool],
+        CapabilityNegotiation,
+    ]
```
SnoopJ, line 127: Nitpick: Is the `TYPE_CHECKING` guard necessary? I ask because it seems plausible that this could be used by consumers who want to generate their handlers, although this is absolutely a fringe edge case.

SnoopJ (approved):
I have no nits left to pick, this LGTM.

sopel/irc/capabilities.py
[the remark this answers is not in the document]
dgw: Does it check for the case of a REQ that's too long? I would expect that to happen either when the request is made (`plugins.capabilities.Manager.register()`) or when the request is sent (`plugins.capabilities.Manager.request_available()`), but don't see any code for it. Will add a line note on `request_available()` to make it easier to find.

sopel/plugins/capabilities.py
```diff
@@ -0,0 +1,456 @@
...
+            }
+            text = ' '.join(cap_req)
+            if cap_req_enables <= capabilities:
+                self._registered[cap_req] = {
+                    plugin_name: (handler_info[0], False)
+                    for plugin_name, handler_info in plugin_requests.items()
+                }
+                self._requested.add(cap_req)
+                LOGGER.debug('Capability negotiation request: "%s".', text)
+                bot.write(('CAP', 'REQ'), text=text)
```
dgw, line 284: [A discussion above indicates](https://github.com/sopel-irc/sopel/pull/2341/files#r1141195054) that this method should be responsible for ensuring that the resulting `CAP REQ` subcommand isn't too long to be acknowledged in a single line, per a SHOULD in the spec:

> Hey, I found a [recommendation in the spec](https://ircv3.net/specs/extensions/capability-negotiation.html#the-cap-req-subcommand):
> 
> > Clients SHOULD ensure that their list of requested capabilities is not too long to be replied to with a single ACK or NAK message. If a REQ’s final parameter gets sufficiently large (approaching the 510 byte limit), clients SHOULD instead send multiple REQ subcommands.
> 
> So, Sopel is the Client. Sopel SHOULD instead send multiple REQ subcommands.

dgw (commented):
There are a couple of nits that can be handled during squash, but we should make a decision about one apparently unresolved thread that led into discussion about CAP REQs that are too long to be acknowledged in one line. It's fine to say that's for a later patch, but in that case let's have a follow-up issue about it before merging. 😸

sopel/irc/capabilities.py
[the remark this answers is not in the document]
Exirel: It doesn't, that why I didn't resolve it, and then I forgot to deal with it... :upside_down_face:

sopel/plugin.py
Exirel: Yes it is, sadly: ` SopelWrapper` cannot be imported in `sopel.plugin`, because `sopel.bot` defines `SopelWrapper` and imports `sopel.plugin` (for a single constant that I really need to move out...).

So, yeah, to prevent circular import, I have to. At least for now.

sopel/plugins/capabilities.py
Exirel: Done!

Exirel:
Oof. Done with the third? (I didn't count) round. I think it's the third, or maybe the fourth...

Anyway, I added the limitation on the `CAP REQ` size, I think it's a very unlikely scenario, however Sopel deals with it now, and there are tests to prove it.

Permission to rebase & squash requested! :grin:

sopel/plugins/capabilities.py
```diff
@@ -194,6 +194,8 @@ def register(self, plugin_name: str, request: capability) -> None:
         """Register a capability ``request`` for ``plugin_name``.
 
         :param request: the capability request to register for later
+        :raise RuntimeError: when the capability request is too long for a
+                             single ``CAP REQ`` and ``CAP * ACK``.
```
dgw, line 198:
```suggestion
                             single ``CAP REQ`` and ``CAP * ACK``
```

Style nit: existing `:raise:` declarations in other files have no trailing period

sopel/plugins/capabilities.py
```diff
@@ -213,7 +215,26 @@ def register(self, plugin_name: str, request: capability) -> None:
...
         """
+        cap_req = ' '.join(request.cap_req)
+        if len(cap_req.encode('utf-8')) > 500:
+            # "CAP * ACK " is 10 bytes, leaving 500 bytes for the capabilities.
+            # Sopel cannot allow multi-line requests, as it won't know how to
+            # deal properly with multi-line ACK.
+            # The spec says a client SHOULD send multiple request, however
+            # the spec also says that a server will ACK or NAK a whole request
+            # at once, so technically, multiple REQ are not the same as a
+            # single REQ.
```
dgw, line 235:
```suggestion
            # The spec says a client SHOULD send multiple requests; however
            # the spec also says that a server will ACK or NAK a whole request
            # at once. So technically, multiple REQs are not the same as a
            # single REQ.
```

sopel/plugin.py
```diff
@@ -207,6 +207,13 @@ def cap_example_incompatible_cap(cap_req, bot, ack):
         If you want to handle a ``CAP`` message without requesting the
         capability, you should use the :func:`event` decorator instead.
 
+    .. warning::
+
+        The list of ``cap_req`` is limited in size to prevent the bot from
+        separating the ``CAP REQ`` in multiple lines as the bot does not know
+        how to call back the capability handler upon receving the multi-line
```
dgw, line 214:
```suggestion
        how to call back the capability handler upon receiving the multi-line
```

sopel/plugins/capabilities.py
```diff
@@ -0,0 +1,477 @@
...
+.. versionadded:: 8.0
+
+.. important::
+
+    This is all relatively new. Its usage and documentation is for Sopel core
+    development and advanced developers. It is subject to rapid changes
+    between versions without much (or any) warning.
+
+    Do **not** build your plugin based on what is here, you do **not** need to.
+"""
```
dgw, line 12:
```suggestion
    Do **not** build your plugin based on what is here, you do **not** need to.

"""
```

This is our usual style when indented blocks occur at the end of a docstring.

sopel/plugins/capabilities.py
```diff
@@ -0,0 +1,477 @@
...
+       ``CAP LS`` subcommand
+    3. upon receiving the list, it'll use the :meth:`request_available` method
+       to let the manager send the required ``CAP REQ`` messages.
+    4. when the server ``ACK`` a request, the bot will call the
+       :meth:`acknowledge` method; when the server ``NAK`` a request, the bot
+       will call the :meth:`deny` method
+    5. once all requests are handled (either directly or after calling the
+       :meth:`resume` method), the bot will send the ``CAP END`` message to end
+       capability negotiation
+
```
dgw, line 61: Does this blank line affect rendering? It seems like it shouldn't be necessary.

```suggestion
```

dgw (commented):
Just style & typo nits. Any style mishaps I've missed at this point will be allowed to live to see the day when a relevant `flake8` plugin is added and flags them for correction, if ever.

I'm still worried a bit about the possibility of (as @SnoopJ called it) "a particularly antisocial capability" named with a leading hyphen that would be stripped in `plugins.capabilities.Manager.request_available()`, but that's a concern that needs to be run by the IRCv3 folks. If it's a possibility, a small patch later can try to account for it. No need to hold this any longer on account of such a wildly unlikely case.

sopel/plugins/capabilities.py
Exirel: It doesn't. I think I just like the extra space after a list.

Exirel:
> such a wildly unlikely case

I have honestly never think that it would be possible. There is nothing that technically prevent that from happening for standard capabilities, however they should start as a draft first, so people will see the problem quickly.

As for [vendor capabilities](https://ircv3.net/specs/extensions/capability-negotiation.html#vendor-specific-capabilities):

> These names are prefixed by a valid DNS domain name.

And a valid DNS domain name **MUST NOT** start with an hyphen, and in case there is a non-ASCII character in there:

> In cases where the prefix contains non-ASCII characters, punycode MUST be used, e.g. xn--e1afmkfd.org/foo.

So I think the spec covers the problem by not allowing anyone to do that in practice.

Also, PR is ready again!

dgw (approved):
Go-go gadget SQUASHER!

sopel/irc/capabilities.py
```diff
@@ -0,0 +1,274 @@
...
+        This method behaves as a plugin callable with its ``bot`` and
+        ``trigger`` arguments, with the precise goals to handle ``CAP ACK``
+        command only, registering enabled or disabled capabilities and tracking
+        acknowledgement.
+
+        Then it returns the list of acknowledged capability requests.
+        """
+        # dev checking
+        assert trigger.event == 'CAP'
+        assert trigger.args[1] == 'ACK'
```
dgw, line 177: Was this supposed to be left in the "production" branch that's ready to merge?

I know some of my past protocol work has shipped `assert` in production code for a few versions, so I can't throw stones at this from my glass house, but just checking.

sopel/irc/capabilities.py
```diff
@@ -0,0 +1,274 @@
...
+
+        This method behaves as a plugin callable with its ``bot`` and
+        ``trigger`` arguments, with the precise goals to handle ``CAP NAK``
+        command only.
+
+        Then it returns the list of denied capability requests.
+        """
+        # dev checking
+        assert trigger.event == 'CAP'
+        assert trigger.args[1] == 'NAK'
```
dgw, line 211: Same question/comment as above in `handle_ack`.

sopel/irc/capabilities.py
```diff
@@ -0,0 +1,274 @@
...
+        This method behaves as a plugin callable with its ``bot`` and
+        ``trigger`` arguments, with the precise goals to handle ``CAP NEW``
+        command only.
+
+        It registers which new capabilities are available, then returns this
+        list.
+        """
+        # dev checking
+        assert trigger.event == 'CAP'
+        assert trigger.args[1] == 'NEW'
```
dgw, line 235: Same Q as in `handle_ack`.

sopel/irc/capabilities.py
```diff
@@ -0,0 +1,274 @@
...
+        This method behaves as a plugin callable with its ``bot`` and
+        ``trigger`` arguments, with the precise goals to handle ``CAP DEL``
+        command only.
+
+        It registers which capabilities are not available and not enabled
+        anymore, then returns this list.
+        """
+        # dev checking
+        assert trigger.event == 'CAP'
+        assert trigger.args[1] == 'DEL'
```
dgw, line 262: Same Q as in `handle_ack`.

sopel/plugins/capabilities.py
```diff
@@ -0,0 +1,477 @@
...
+        :return: a 2-value tuple with (was completed, is completed)
+
+        The capability request, for that plugin, will be marked as done,
+        and the result will be about the capability negotiation process:
+
+        * was it already completed before the resume?
+        * is it completed now?
+
+        If the capability request cannot be found for that plugin, the result
+        value remains the same (it stay uncomplete or complete)
```
dgw, line 327:
```suggestion
        value remains the same (it stays incomplete or complete)
```

Rats. Missed this one.

sopel/plugins/capabilities.py
```diff
@@ -0,0 +1,477 @@
...
+        :param bot: the bot instance used to send capability requests
+        :param available_capabilities: available capabilities
+
+        This sends ``CAP REQ`` commands for requests that can be made, i.e.
+        all the requested capabilities (with or without prefix) must be
+        available for Sopel to send the request.
+
+        Requests made are stored as requested; others are ignored::
+
+            >>> # manager.register => cap1 and cap2, not cap3
```
dgw, line 260: I just realized I don't understand this example. What is this line of the repr supposed to represent? Since the `register()` method above doesn't return anything, I think that's where I'm getting lost.

sopel/plugin.py
```diff
@@ -71,6 +89,221 @@
...
+        The list of ``cap_req`` is limited in size to prevent the bot from
+        separating the ``CAP REQ`` in multiple lines as the bot does not know
+        how to call back the capability handler upon receiving the multi-line
+        ``ACK * REQ``.
+
+    .. seealso::
+
+        The IRCv3 specification on `Client Capability Negotiation`__.
+
+    .. __: https://ircv3.net/specs/extensions/capability-negotiation.html
```
dgw, line 221:
```suggestion
    .. __: https://ircv3.net/specs/extensions/capability-negotiation
```

The site never generates URLs ending in `.html`. (I should probably propose that they add redirects to the "pretty" versions… Shouldn't be too hard; IIRC they use Netlify, like us.)

docs/source/plugin/advanced.rst
```diff
@@ -64,3 +64,134 @@ a QUIT/PART event, and they will be available after a JOIN event.
...
+  negotiation phase)
+
+.. warning::
+
+    This is a very advanced feature, and plugin authors should understand how
+    capability negotiation works before using it. Even if Sopel tries to make
+    it as simple as possible, plugin authors should be aware of the known
+    limitations and possible caveats.
+
+.. __: https://ircv3.net/specs/extensions/capability-negotiation.html
```
dgw, line 95:
```suggestion
.. __: https://ircv3.net/specs/extensions/capability-negotiation
```

Ditto what I said on the other occurrence.

sopel/irc/__init__.py
```diff
@@ -52,8 +50,8 @@
 from sopel import tools, trigger
```
dgw, line 50:
```suggestion
from sopel import tools, trigger
from sopel.lifecycle import deprecated
```

Goes with newly deprecated properties in the `AbstractBot`.

dgw (changes requested):
Unfortunately for @Exirel I went through the whole thing\* again and found a few more nitpicks—but more importantly, I found an example in one docstring that seems to need clarification.

\* — Except for tests. I didn't reread most of the `test/` files.

sopel/irc/capabilities.py
Exirel: Yes, assert are fine here, I used them to help fellow developers: if they use them in the wrong context, they will trigger `AssertionError` while working on the code. It is very unlikely that someone will use these outside of Sopel's core.

sopel/plugins/capabilities.py
Exirel: That will make more sense.

```suggestion
            >>> manager.register('example', 'cap1')
            >>> manager.register('example', 'cap2')
```

dgw (approved):
Yes, the revised example makes more sense and we've cleared up everything else too. 🥳

sopel/irc/capabilities.py
dgw: To tie this up once and for all, I proposed ircv3/ircv3-specifications#519, which was accepted. We needn't worry about a leading hyphen!

PR #2346, opened 2022-08-28
irc: `LINELEN` ISUPPORT token support/use

dgw:
Take advantage of the `LINELEN` token in 005 to send longer messages than the RFC limit of 512 bytes, when possible.

Can't say I've ever seen a network actually advertise this token, but it's relatively easy to add and test so why not support it.

### Checklist

### Notes
This was previously flagged as obsolete, but InspIRCd still uses it (refs: ircdocs/irc-defs#82, https://github.com/inspircd/inspircd/blob/e147b461173c58eb443eee0ac56f7d3c3c4b2c3a/src/server.cpp#L218).

Docs link in the code points to a nonexistent section at time of writing. See ircdocs/modern-irc#189 for the patch that will add it.

### DIFF
--- a/sopel/irc/__init__.py
+++ b/sopel/irc/__init__.py
@@ -184,6 +184,11 @@ class AbstractBot(abc.ABC):
             and can be used with :func:`sopel.tools.get_sendable_message`.
 
         """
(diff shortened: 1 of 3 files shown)


Exirel (approved):
Looks straightforward to me.

PR #2352, opened 2022-10-04
privileges, url: configurable minimum channel access requirement for `.urlban` and friends

dgw:
### Description
Tin, more or less.

Added `__all__` to `sopel.privileges` for ease of fetching the names of valid privilege constants programmatically.

`.urlban`/`.urlallow` and friends are now only available to bot owner/admins and channel OPs by default, with the option to change the channel access level required via config setting.

### Checklist

### Notes
Naming Things Is Hard as always. Suggestions welcome if anyone's got a better idea for the new option name in `UrlSection`.

Decided not to add the new option to `configure()`, at least for now. Doing so seemed like overkill.

### DIFF
--- a/sopel/modules/url.py
+++ b/sopel/modules/url.py
@@ -20,7 +20,7 @@ import dns.resolver
 import requests
 from urllib3.exceptions import LocationValueError  # type: ignore[import]
 
-from sopel import plugin, tools
+from sopel import plugin, privileges, tools
 from sopel.config import types
 from sopel.tools import web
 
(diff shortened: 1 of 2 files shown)


half-duplex (approved):
Why is `sopel.privileges` a module, anyway? If it was an [`enum.Flag`](https://docs.python.org/3/library/enum.html#flag) class, optionally with `X=enum.auto()`, we would get a lot of this stuff free...
Besides that, lgtm.

dgw:
> Why is `sopel.privileges` a module, anyway?

Because of reorganizing in #2179. These constants used to live inside `sopel.plugin`, which was even worse.

> If it was an [`enum.Flag`](https://docs.python.org/3/library/enum.html#flag) class, optionally with `X=enum.auto()`, we would get a lot of this stuff free...

You're not wrong. No package version has ever been released with `sopel.privileges`, so if you want to throw together a follow-up to #2179 that does something sensible with an `enum.Flag` class (while staying backward compatible with the aliases in `sopel.plugin` for the time being), I'd be all for it. If it's replaced by something better, we could just pretend `sopel.privileges`-the-module never happened.

half-duplex:
Such flexibility! Releases suck, -dev is comfy, we should stay here forever.

Exirel:
> Why is `sopel.privileges` a module, anyway? If it was an [`enum.Flag`](https://docs.python.org/3/library/enum.html#flag) class, optionally with `X=enum.auto()`, we would get a lot of this stuff free... Besides that, lgtm.

We would still need to find a place for this `enum` class. Fixing circular import is way more important than the small free stuffs that `enum` would give us in that case. My opinion here is that there is not a *lot* of free stuff, only *few* free stuffs.

And it would not be worth it.

PR #2356, opened 2022-10-08
send_notice: sanitize `dest` argument before sending it on wire

semarie:
all arguments should be inspected for newlines to avoid IRC injection.
the `text` argument is already checked (in `prepare_command()`).

### Description
The PR adds a sanitization (remove possible newlines) on `dest` argument. It is a security measure to avoid injection of arbirtrary IRC command via `dest` parameter.

### Checklist

### DIFF
--- a/sopel/irc/abstract_backends.py
+++ b/sopel/irc/abstract_backends.py
@@ -245,4 +245,4 @@ class AbstractIRCBackend(abc.ABC):
         :param str dest: nickname or channel name
         :param str text: the text to send
         """
-        self.send_command('NOTICE', dest, text=text)
+        self.send_command('NOTICE', safe(dest), text=text)


half-duplex:
Good catch! Feel like writing a regression test?

dgw:
The only question I had after looking at this was "why aren't the `args` _all_ sanitized in `prepare_command()`? Why only `text`?

Pretty sure @half-duplex had the same reaction, based on the IRC logs I saw from earlier today.

dgw:
You might have noticed that this sits unreviewed and unmerged, and it's because I keep coming back to the idea that actually (ackshully?), `send_command()` should be responsible for sanitizing its own inputs, rather than requiring all of its callers to remember to pass `safe(dest)`.

@semarie do you have any interest in performing such a refactor yourself?

semarie:
@dgw I will not have the time to do to. I reviewed a bit the sopel code as a bot based on it joined an IRC chan I am using. So it was more a play than a real interest in the code base.

Regarding the fact to centralize sanitization inside `send_command()`, I think it is a right move.

dgw:
@semarie I appreciate the quick reply, and we all appreciate that you pointed out this problem in the first place. I will keep this PR open as a reminder, but I think we're going to replace it with saner sanitization. It so happens that @half-duplex _loves_ patching silly security-adjacent stuff like this, and this idea already got a "yay" on IRC. 🙂

dgw:
Thanks to @half-duplex for the refactor follow-up in #2368

Issue #2331, opened 2022-07-23
trigger: inconsistent `sender` for other general messages like QUIT


dgw:
This is a follow-up to #2217, which reports an exception case that is technically fixed, but the underlying cause (weird behavior with certain message types) still exists.

Code referenced below sets an intermediate variable to `None`, then populates it based on the arguments to whatever command Sopel received _unless_ that command was QUIT. This leaves some edge cases hanging about, like handling of AWAY on servers with `away-notify` turned on—another command that's just generally sent to any clients that share a channel with the acting user, but isn't associated with any channel or private-message buffer.

https://github.com/sopel-irc/sopel/blob/71380df85212747a6c02d8c37c9ffb7836001278/sopel/trigger.py#L213-L225

Making this consistent would be best. `trigger.sender` being `None` is safer now that #2217 was fixed in #2306, and it's definitely _not_ ideal for the sender to be set to `Identifier("the away message")` on received AWAY commands.

dgw:
Going to say #2359 resolves this. If we find other cases, we'll reopen/follow-up.

PR #2359, opened 2022-10-13
trigger: Make `sender` property more consistent

dgw:
### Description
Short sleep does not make me a great comprehensive reader of specs, but commands that are 1) sent from the server to a client and 2) include a <target> (channel/nick), are actually quite rare. The IRC protocol is mostly numeric replies, and very few event types actually refer to a user (nick) or channel.

The idea here, though, is that unless I forgot about an event type that actually is tied to a channel or query (PM) buffer, it is ONLY the new list of `COMMANDS_WITH_CONTEXT` that should have the `sender` set to any value other than `None`.

### Checklist

### Notes
Undecided yet if this is enough to call #2331 closed. I remember a discussion about transitioning to a new property called `trigger.context` or similar, for clarity, but can't find it on the relevant issue. Maybe it was an IRC conversation.

'Breaking Change' label because this could be a [spacebar heating](https://xkcd.com/1172/) incident if anyone relies on the old (buggy) behavior for some weird plugin thing and it needs to be mentioned in the changelog for 8.0.

While updating/adding tests I also found an actual bug in how `text` gets set (it contradicts Sopel's documentation). TODO comments related. I will work on that next.

### DIFF
--- a/sopel/trigger.py
+++ b/sopel/trigger.py
@@ -38,6 +38,27 @@ __all__ = [
 IdentifierFactory = Callable[[str], identifiers.Identifier]
 
 
+COMMANDS_WITH_CONTEXT = [
+    'INVITE',
+    'JOIN',
+    'KICK',
+    'MODE',
+    'NOTICE',
+    'PART',
+    'PRIVMSG',
+    'TOPIC',
+]
+"""List of commands with a :attr:`trigger.sender<Trigger.sender>`.
+
+Most IRC messages a plugin will want to handle (channel messages and PMs) are
+associated with a context, exposed in the Trigger's ``sender`` property.
+However, not *all* commands can be directly associated to a channel or nick.
+
+For IRC message types other than those listed here, the ``trigger``\'s
+``sender`` property will be ``None``.
+"""
+
+
 class PreTrigger:
     """A parsed raw message from the server.
 
@@ -90,6 +111,14 @@ class PreTrigger:
     .. py:attribute:: sender
 
         Channel name or query where this message was received.
+
+        .. warning::
+
+            The ``sender`` Will be ``None`` for commands that have no implicit
+            channel or private-message context, for example AWAY or QUIT.
+
+            The :attr:`COMMANDS_WITH_CONTEXT` attribute lists IRC commands for
+            which ``sender`` can be relied upon.
 
     .. py:attribute:: tags
 
@@ -210,11 +239,11 @@ class PreTrigger:
         nick, self.user, self.host = components_match.groups()
         self.nick: identifiers.Identifier = self.make_identifier(nick)
 
-        # If we have arguments, the first one is the sender
-        # Unless it's a QUIT event
+        # If we have arguments, the first one is *usually* the sender,
+        # most numerics and certain general events (e.g. QUIT) excepted
         target: Optional[identifiers.Identifier] = None
 
-        if self.args and self.event != 'QUIT':
+        if self.args and self.event in COMMANDS_WITH_CONTEXT:
             target = self.make_identifier(self.args[0])
 
             # Unless we're messaging the bot directly, in which case that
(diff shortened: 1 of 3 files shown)


Exirel (approved):
I'm impressed. It's actually quite good: logic is good, code is sane, there are tests! All good to me.

sopel/trigger.py
```diff
@@ -38,6 +38,27 @@
...
+COMMANDS_WITH_CONTEXT = [
+    'INVITE',
+    'JOIN',
+    'KICK',
+    'MODE',
+    'NOTICE',
+    'PART',
+    'PRIVMSG',
+    'TOPIC',
+]
```
SnoopJ, line 50: ## PERFORMANCE WANKERY WARNING

Is order important here? If not, a `set()` (or `frozenset()`?) might be a small improvement that accelerates membership tests, which are I imagine going to be the typical operations with this.

Because the perf is in the pudding, here's what I think is a representative microbenchmark for `'PART'`:

```17:36 [snoopjedi@denton ~]
$ python3.9 -m timeit -n 1000000 -s "COMMANDS_WITH_CONTEXT = [ 'INVITE', 'JOIN', 'KICK', 'MODE', 'NOTICE', 'PART', 'PRIVMSG', 'TOPIC', ]" "'PART' in COMMANDS_WITH_CONTEXT"
1000000 loops, best of 5: 122 nsec per loop
17:37 [snoopjedi@denton ~]
$ python3.9 -m timeit -n 1000000 -s "COMMANDS_WITH_CONTEXT = { 'INVITE', 'JOIN', 'KICK', 'MODE', 'NOTICE', 'PART', 'PRIVMSG', 'TOPIC', }" "'PART' in COMMANDS_WITH_CONTEXT"
1000000 loops, best of 5: 30.4 nsec per loop
```

Tests for members that aren't present are fairly comparable to this, although early hits in the list do seem to eke out a small victory, so if `PRIVMSG` (or some other member) is the source of _most_ triggers, maybe it makes sense to keep this a list and put that member at the front of it.

test/test_trigger.py
```diff
@@ -80,6 +80,56 @@ def test_quit_pretrigger(nick):
...
 
+def test_quit_pretrigger_empty(nick):
+    """Make sure empty quit message is also fine"""
+    line = ':Foo!foo@example.com QUIT'
+    pretrigger = PreTrigger(nick, line)
+    assert pretrigger.tags == {}
+    assert pretrigger.hostmask == 'Foo!foo@example.com'
+    assert pretrigger.line == line
+    assert pretrigger.args == []
+    assert pretrigger.text == 'QUIT'  # TODO BUG: the command is not part of `pretrigger.args`
```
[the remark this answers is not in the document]
dgw: Nope, I discovered it while working on this PR and created a follow-up immediately: #2360.

SnoopJ (approved):
LGTM, nice one

sopel/trigger.py
dgw: Want to add tuple to that microbenchmark?

SnoopJ: Good idea, it's comparable to the `list` spelling (which makes sense, it's basically the same linear scan):

```
17:54 [snoopjedi@denton ~/scratch]
$ python3.9 -m timeit -n 1000000 -s "COMMANDS_WITH_CONTEXT = ( 'INVITE', 'JOIN', 'KICK', 'MODE', 'NOTICE', 'PART', 'PRIVMSG', 'TOPIC', )" "'QUIT' in COMMANDS_WITH_CONTEXT"
1000000 loops, best of 5: 171 nsec per loop
```

test/test_trigger.py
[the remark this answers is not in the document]
SnoopJ: I'm a big fan of leaving an explicit issue reference with a comment like this, but as long as it's tracked, it's not a big deal.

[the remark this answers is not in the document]
dgw: I would be, too, except in cases like this where such an issue reference doesn't yet exist. Creating an issue instead of just making the +1-1 patch immediately seemed silly. (OK, the actual patch is +5-4, but that's because of updating comments + tests.)

sopel/trigger.py
dgw: A set really is the way to go, then. You really can't beat a good old hashtable backend.

PR #2360, opened 2022-10-13
trigger: make `text` empty if `args` is empty, per docs

dgw:
### Description
An IRC command/event such as 'AWAY' with no message, or 'PING', will have an empty `args` list.

Documentation for the (Pre)Trigger `text` attribute says that it will contain the last item in `args` if the line does not contain ':', but it previously—erroneously—contained the command name in such cases.

### Checklist

### Notes
Starting this as a draft because it's built on top of #2359. Whether I separate the histories or not, I'll have to faff with _something_ after merging one before the other can be merged. Would rather have the simplest rebase imaginable (which this strategy will get me) than a rebase that requires the interactive mode.

Same as the parent PR, 'Breaking Change' label because fixing this after so long could cause a [spacebar heating](https://xkcd.com/1172/) incident.

### DIFF
--- a/sopel/trigger.py
+++ b/sopel/trigger.py
@@ -228,7 +228,8 @@ class PreTrigger:
             self.args.append(self.text)
         else:
             self.args = line.split(' ')
-            self.text = self.args[-1]
+            # see `text` attr documentation above, and #2360
+            self.text = self.args[-1] if len(self.args) > 1 else ''
 
         self.event = self.args[0]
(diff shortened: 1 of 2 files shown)


Exirel (approved):

dgw:
(I just pushed to the wrong branch, but that's OK. It'll be something I can force-push after #2359 gets merged, to make GH realize what is actually part of the PR.)

Issue #971, opened 2015-12-13
Add remaining IRCv3 standards (as of 3.2)


embolalia:
Most of this is in PR #961, or already merged.

* [ ] `account-notify`
* [ ] `away-notify`
* [ ] `extended-join`
* [ ] `tls`
* [ ] `CAP`
* [ ] `Metadata`
* [ ] `Monitor`
* [ ] `account-tag`
* [ ] `batch`
* [ ] `cap-notify`
* [ ] `chghost`
* [ ] `echo-message`
* [ ] `invite-notify`
* [ ] `sasl`
* [ ] `server-time`
* [ ] `user-host-in-names`

dgw:
Some thoughts after looking into a few of the remaining unchecked entries:

- `setname`: fairly simple, and I have a patch (with test) ready for 8.1
- `msgid`: no specific handling for this should be necessary now that `message-tags` were enabled in #2089
  - I guess `Trigger` could gain an `id` attribute/property that returns `Trigger.tags['msgid']` (or `None`, if message IDs aren't in use). Seems silly to wrap a property around a tag, but that's exactly what `Trigger.account` is. 🤷‍♂️
- `invite-notify`: even simpler than `setname`, and likewise I have a patch ready for 8.1
  - There's nothing for Sopel to actually do with these events, though. We can log them at INFO level like joins/parts/kicks/quits, but a plugin that wants to do something with `INVITE` events could just as easily request the CAP itself instead of relying on core to do that.

dgw:
Checked off `invite-notify` with a PR containing my previously-mentioned patch, and fixes for `admin` that I hadn't thought of last year.

Added a note to SASL v3.2, since I think we already implemented handling for the mechanism list and such for 8.0 (but didn't dig into the PR history to verify).

Kufat:
Reply and channel-context have been ratified:
https://github.com/ircv3/ircv3-specifications/pull/535
https://github.com/ircv3/ircv3-specifications/pull/596

dgw:
@Kufat Thanks for the reminder.

Tripped on GitHub's "create sub-issue" not working the way we'd like, but I think everything is sorted out now with the two related issues (one pre-existing, one new).

PR #2362, opened 2022-10-23
coretasks: implement scram-sha-256

half-duplex:
### Description
This implements the SASL SCRAM-SHA-256 authentication mechanism.

I've tested it using irc.alphachat.net:6697 - PM for creds or BYO.

It adds a dependency on `scramp` (~29kb), which could be made optional if we're picky.

It adds tests for both PLAIN and SCRAM-SHA-256, which could probably be done better. Neither tests nor implementation try to catch every edge case - any big ones we care a lot about?

cc @dgw for ircv3 feature tracking, see also #971

### Checklist


Neustradamus:
Linked to:
- https://github.com/scram-sasl/info/issues/1

dgw:
> It adds a dependency on `scramp` (~29kb), which could be made optional if we're picky.

Avoiding/tryong to eliminate dependencies in the current era of Sopel (can't speak for past maintainers) is more about future-proofing than install-size concerns. I worry whether Package X will still be releasing new versions in five years, and tend to prefer projects that are maintained by multiple users and live under a GitHub organization rather than someone's personal account. Sole-maintainer stuff has a nasty habit of going stale without warning when the author loses interest.

dgw:
Wasn't saying anything about this specific library, haha, just dependencies in general. Sure, I've seen ProgVal around a lot, and would probably have picked that one had I written this patch, but it's already written with the other. The important part is that there's another library we could probably switch to if needed.

sopel/config/core_section.py
```diff
@@ -1181,12 +1181,13 @@ def homedir(self):
 
     :default: ``PLAIN``
 
-    ``EXTERNAL`` is also supported, e.g. for using :attr:`client_cert_file` to
-    authenticate via CertFP.
+    ``EXTERNAL`` is supported, e.g. for using :attr:`client_cert_file` to
+    authenticate via CertFP, and ``SCRAM-SHA-256`` for challenge-response
+    authentication.
```
dgw, line 1186: Can we turn this into something more like:

```suggestion
    Supported mechanisms are:

        * ``PLAIN`` — for basic password-based SASL; the default
        * ``EXTERNAL`` — for using :attr:`client_cert_file` to
          authenticate via CertFP
        * ``SCRAM-SHA-256`` — for challenge-response authentication
```

(Have not spent any time trying to confirm line lengths or exact wrapping/indents in the GitHub comment-box.)

sopel/coretasks.py
```diff
@@ -31,6 +31,9 @@
 import re
 import time
 
+from scramp import ScramClient, ScramException
+from scramp.core import ClientStage
```
dgw, line 35: Despite what I just said on your UTF8ONLY PR, I don't mind these `from` imports because the imported names are mostly self-evident. They aren't generic in the way `raises` is.

The only thing I might have done differently is `from scramp.core import ClientStage as ScramClientStage`.

dgw (commented):
I'm doing the nitpick-docs thing again, huh?

sopel/config/core_section.py
half-duplex: Much more clear, thanks.

sopel/config/core_section.py
```diff
@@ -1181,12 +1181,15 @@ def homedir(self):
 
     :default: ``PLAIN``
 
-    ``EXTERNAL`` is also supported, e.g. for using :attr:`client_cert_file` to
-    authenticate via CertFP.
+    Supported mechanisms are:
+        * ``PLAIN``, to authenticate by sending a plaintext password
+        * ``EXTERNAL``, to authenticate using a TLS client certificate
+          (see :attr:`client_cert_file`)
+        * ``SCRAM-SHA-256``, for password-based challenge-response authentication
```
dgw, line 1188: Indenting so much makes Sphinx generate a `<dl>`, with the `<ul>` and children nested inside a `<dd>`:

<details>
<summary>Dev Tools HTML view</summary>

image

</details>

That's probably not what we want.

```suggestion
  * ``PLAIN``, to authenticate by sending a plaintext password
  * ``EXTERNAL``, to authenticate using a TLS client certificate
    (see :attr:`client_cert_file`)
  * ``SCRAM-SHA-256``, for password-based challenge-response authentication
```

half-duplex: With an added bonus of that 81-char line now being 77

dgw (dismissed):
I'm happy, but let's give @Exirel a chance to weigh in.

progval:
There are two uncaught exceptions from scramp when the server returns either garbage or an invalid signature instead of the last response. In the former case, it also does not send `AUTHENTICATE *` to abort the authentication, so it hangs on a pingpong loop

Respectively:

```
1666768709.393 C: CAP REQ sasl
1666768709.393 S: CAP Sopel ACK :sasl
1666768709.441 C: AUTHENTICATE SCRAM-SHA-256
1666768709.441 S: AUTHENTICATE +
[2022-10-26 09:18:29,489] sopel.coretasks      INFO     - Sending SASL SCRAM client first
1666768709.490 C: AUTHENTICATE biwsbj1qaWxsZXMscj1jOGMxNGUxZTRjOTc0Mzk5ODdiZmZlNjJjNGI5YjVmZA==
1666768709.547 S: AUTHENTICATE :cj1jOGMxNGUxZTRjOTc0Mzk5ODdiZmZlNjJjNGI5YjVmZDhkNjFiNjJiZDc2MDQ5NzQ4MzFmMTRkYjgyMTIxZGIzLHM9WlRJMVkyTmhOMk0zWmpSaU5EQTRZMkkxTVdSaFlUZG1OVFpoTnpVNVkyRT0saT00MDk2
[2022-10-26 09:18:29,623] sopel.coretasks      INFO     - Sending SASL SCRAM client final
1666768709.624 C: AUTHENTICATE Yz1iaXdzLHI9YzhjMTRlMWU0Yzk3NDM5OTg3YmZmZTYyYzRiOWI1ZmQ4ZDYxYjYyYmQ3NjA0OTc0ODMxZjE0ZGI4MjEyMWRiMyxwPVl4UWtZUmJGVkM3QjZRRERVejhOeklYcGlEczJQbWdRMEZUcUc1NTdVK289
1666768709.624 S: AUTHENTICATE :AAAA
[2022-10-26 09:18:29,668] sopel.bot            ERROR    - Unexpected KeyError ('v') from  at 2022-10-26 07:18:29.668634. Message was: AAAA
  File "/home/dev-irc/.local/lib/python3.9/site-packages/sopel/bot.py", line 648, in call_rule
    rule.execute(sopel, trigger)
  File "/home/dev-irc/.local/lib/python3.9/site-packages/sopel/plugins/rules.py", line 1203, in execute
  File "/home/dev-irc/.local/lib/python3.9/site-packages/sopel/coretasks.py", line 1189, in auth_proceed
    bot._scram_client.set_server_final(server_final)
  File "/home/dev-irc/.local/lib/python3.9/site-packages/scramp/core.py", line 269, in set_server_final
    _set_server_final(message, self.server_signature)
  File "/home/dev-irc/.local/lib/python3.9/site-packages/scramp/core.py", line 562, in _set_server_final
    if server_signature != msg["v"]:
KeyError: 'v'
1666768711.920 C: PING 0.0.0.0
```

and

```
1666768717.764 C: CAP REQ sasl
1666768717.764 S: CAP Sopel ACK :sasl
1666768717.821 C: AUTHENTICATE SCRAM-SHA-256
1666768717.821 S: AUTHENTICATE +
[2022-10-26 09:18:37,864] sopel.coretasks      INFO     - Sending SASL SCRAM client first
1666768717.865 C: AUTHENTICATE biwsbj1qaWxsZXMscj05YjJkYWZjOTRlNzM0NTNlOGMxNDM2MjFlNDEwNzRiMw==
1666768717.915 S: AUTHENTICATE :cj05YjJkYWZjOTRlNzM0NTNlOGMxNDM2MjFlNDEwNzRiMzA1NGE5MzU3NTFjMjQ2NmQ4YjM1MjIwZTAxMTEzZTM4LHM9WVdJellqVmlNemN6WW1FME5ESmxZamd6WWpnMk56TmtPR014WXpRd09UWT0saT00MDk2
[2022-10-26 09:18:37,993] sopel.coretasks      INFO     - Sending SASL SCRAM client final
1666768717.994 C: AUTHENTICATE Yz1iaXdzLHI9OWIyZGFmYzk0ZTczNDUzZThjMTQzNjIxZTQxMDc0YjMwNTRhOTM1NzUxYzI0NjZkOGIzNTIyMGUwMTExM2UzOCxwPUNhQXFCZlladjJWSFZzU3U2cUdnZHo3M0dUODNuZFZ4NkJDejNaNTdkMFk9
1666768717.994 S: AUTHENTICATE :dj1ubU1mM1FIV2NKUWk5cE1ndHFLU0tQclZueUk2c3FOTzZJN3BFLzBveUdjPQ==
[2022-10-26 09:18:38,040] sopel.coretasks      ERROR    - SASL SCRAM failed: ScramException("The server signature doesn't match.")
[2022-10-26 09:18:38,040] sopel.bot            ERROR    - Unexpected ScramException (The server signature doesn't match. other-error) from  at 2022-10-26 07:18:38.040819. Message was: dj1ubU1mM1FIV2NKUWk5cE1ndHFLU0tQclZueUk2c3FOTzZJN3BFLzBveUdjPQ==
  File "/home/dev-irc/.local/lib/python3.9/site-packages/sopel/bot.py", line 648, in call_rule
    rule.execute(sopel, trigger)
  File "/home/dev-irc/.local/lib/python3.9/site-packages/sopel/plugins/rules.py", line 1203, in execute
  File "/home/dev-irc/.local/lib/python3.9/site-packages/sopel/coretasks.py", line 1193, in auth_proceed
    raise e
  File "/home/dev-irc/.local/lib/python3.9/site-packages/sopel/coretasks.py", line 1189, in auth_proceed
    bot._scram_client.set_server_final(server_final)
  File "/home/dev-irc/.local/lib/python3.9/site-packages/scramp/core.py", line 269, in set_server_final
    _set_server_final(message, self.server_signature)
  File "/home/dev-irc/.local/lib/python3.9/site-packages/scramp/core.py", line 563, in _set_server_final
    raise ScramException(
scramp.core.ScramException: The server signature doesn't match. other-error
1666768718.042 C: AUTHENTICATE *
```

half-duplex:
Re-raising the exception (example 2) was intentional, but I suppose is unnecessary. Good catch on the scramp KeyError.
Also added try/catch for bad server_first.

tlocke:
Hi all, maintainer of Scramp here 👋 With the latest release of Scramp 1.4.4 we now check that every message is well formed and raise a `ScramException` if not. So in the example above:

```
  File "/home/dev-irc/.local/lib/python3.9/site-packages/sopel/coretasks.py", line 1189, in auth_proceed
    bot._scram_client.set_server_final(server_final)
  File "/home/dev-irc/.local/lib/python3.9/site-packages/scramp/core.py", line 269, in set_server_final
    _set_server_final(message, self.server_signature)
  File "/home/dev-irc/.local/lib/python3.9/site-packages/scramp/core.py", line 562, in _set_server_final
    if server_signature != msg["v"]:
KeyError: 'v'
```
A `ScramException` with a descriptive message is raised instead. Thanks to @half-duplex for the patch :thumbsup: Let me know if you come across any other problems.

dgw:
@progval @tlocke Have your concerns all been addressed?

I'm going through some of our older open PRs and this one looks like it should be either ready or very nearly so.

progval:
yes

tlocke:
Actually I just got involved because I maintain the scramp library. If any problems come up with it, just let me know.

dgw:
@tlocke Thanks. I saw a traceback example and assumed you were advising a fix for something still in the patch, or modifying the version range.

@Exirel It might not be feasible to wait for #2341 as that's still in draft state, vs. this being practically ready to merge. Unless you have time to finish that off before mal finds the time to finalize this one. It's a race! 🏁

sopel/coretasks.py
```diff
@@ -106,6 +112,12 @@ def _handle_sasl_capability(
             'cannot authenticate with SASL.',
         )
         return plugin.CapabilityNegotiation.ERROR
+    elif auth_target not in ["PLAIN", "EXTERNAL", "SCRAM-SHA-256"]:
+        LOGGER.error(
+            'Configured SASL method %r is not supported by Sopel.',
+            auth_target,
+        )
+        return plugin.CapabilityNegotiation.ERROR
```
Exirel, line 120: In my opinion there are better places to check this:

* few lines down in the file we check that the configured SASL method is supported by the IRC server
* the configuration option can have validation
* maybe the setup function of coretasks could check for that

sopel/coretasks.py
```diff
@@ -94,8 +98,10 @@ def _handle_sasl_capability(
     # Manage CAP REQ :sasl
     auth_method = bot.settings.core.auth_method
     server_auth_method = bot.settings.core.server_auth_method
+    auth_target = bot.settings.core.auth_target
     is_required = 'sasl' in (auth_method, server_auth_method)
 
+    LOGGER.critical("FOO. BAR!")
```
Exirel, line 104: Leftover to be removed.

```suggestion
```

sopel/coretasks.py
```diff
@@ -94,8 +98,10 @@ def _handle_sasl_capability(
     # Manage CAP REQ :sasl
     auth_method = bot.settings.core.auth_method
     server_auth_method = bot.settings.core.server_auth_method
+    auth_target = bot.settings.core.auth_target
```
Exirel, line 101: If only `server_auth_method` is set, then you need to check `server_auth_sasl_mech`, instead of `auth_target`.

Exirel (changes requested):
The implementation of the SCRAM workflow looks good. However, when it comes to the appropriate checks and validation, I think more need to be done.

About the automated tests, I'm glad there are some, however the SASL related tests need to go into `test/coretasks/test_coretasks_sasl.py`. Maybe a part of it is a leftover from the rebase, which I assume was quite difficult from what I can remember of my own modifications.

While you are updating the doc, I think the example is wrong, as it uses `server_auth_target` instead of `server_auth_sasl_mech`, that need to be checked.

test/test_coretasks.py
```diff
@@ -502,6 +509,210 @@ def test_handle_rpl_myinfo(mockbot):
...
+    mockbot.settings.core.auth_method = "sasl"
+    mockbot.settings.core.auth_target = "SCRAM-MD4"
+    mockbot.on_message("CAP * LS :sasl")
+    mockbot.on_message("CAP TestBot ACK :sasl")
+    assert mockbot.backend.message_sent == rawlist(
+        "CAP REQ :sasl",
+        "CAP END",
+    )
+    #with pytest.raises(Exception):
+    mockbot.on_message("AUTHENTICATE +")
```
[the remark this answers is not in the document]
half-duplex: Forgot to finish making the test actually test the thing. Still not sure if it's worth testing.

sopel/coretasks.py
half-duplex: Fixed. I wouldn't mind making these config options less confusing at some point...

sopel/coretasks.py
```diff
@@ -1246,7 +1258,43 @@ def auth_proceed(bot, trigger):
             bot.write(('AUTHENTICATE', '*'))
             return
 
-    # TODO: Implement SCRAM challenges
+    elif mech == "SCRAM-SHA-256":
+        if trigger.args[0] == "+":
+            bot._scram_client = ScramClient([mech], sasl_username, sasl_password)
```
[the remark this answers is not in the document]
half-duplex: Do we have a line on what belongs in bot.memory vs an attribute? To me, bot.memory is meant for users to touch, so the scram client belongs as a _attribute. On the other hand, I'd rather leave the scramp dependency in coretasks.py only...

sopel/coretasks.py
half-duplex: Do we care about failing the config if the server doesn't support SASL at all? If yes, we should also error on SASL configured but no server support. If no, I've made the check consistent with server mech mismatch.

sopel/coretasks.py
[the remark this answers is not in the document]
SnoopJ: I don't think there's any clearly-defined line, but the core does use `bot.memory` for some stateful things (`'join_events_queue'` in particular) so I tend to agree that this would be better stored in memory.

I could see a case for having an attribute on the bot that contains a generalized authentication client of some sort, but I can see and agree with the objection to the definition given here, especially since the attribute isn't guaranteed to exist.

sopel/coretasks.py
```diff
@@ -1250,7 +1264,43 @@ def auth_proceed(bot, trigger):
...
+            try:
+                server_final = base64.b64decode(trigger.args[0]).decode("utf-8")
+                bot._scram_client.set_server_final(server_final)
+            except (BinasciiError, KeyError, ScramException) as e:
+                LOGGER.error("SASL SCRAM server_final failed: %r", e)
+                bot.write(("AUTHENTICATE", "*"))
+                return
+            LOGGER.info("SASL SCRAM succeeded")
+            bot.write(("AUTHENTICATE", "+"))
+            bot._scram_client = None
```
SnoopJ, line 1302: Looking at this, I find myself wondering how much of these 30 lines could be pushed into `ScramClient`, so that the `elif` clauses could be folded away behind something roughly like `else: bot._scram_client.proceed()`

I'm not sure if it's really worth pushing across the class boundary, but maybe it would still make sense to define a separate method on the bot.

sopel/coretasks.py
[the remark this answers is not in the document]
Exirel: Yeah, at some point, maybe have a fully extendable authentication system within Sopel so plugin can actually hook into it properly... but at this time, this isn't it.

The bot offer a memory object exactly for this type of purpose: plugin specific piece of in-memory data. In this case, this is even a throwaway piece of data, because once the scram challenge is complete, you could delete it from memory.

test/coretasks/test_coretasks_sasl.py
```diff
@@ -344,3 +353,201 @@ def test_sasl_nak(botfactory: BotFactory, tmpconfig) -> None:
...
+    mockbot.on_message("CAP TestBot ACK :sasl")
+    assert mockbot.backend.message_sent == rawlist(
+        "CAP REQ :sasl",
+        "AUTHENTICATE SCRAM-SHA-256",
+    )
+    mockbot.on_message("AUTHENTICATE +")
+
+    scram_server.set_client_first(
+        b64decode(mockbot.backend.message_sent[-1].split(b" ")[-1]).decode("utf-8")
+    )
```
Exirel, line 421: The `[-1]` here is an assumption that I don't like. I'd rather see that there is, indeed, only the expected number of messages. And if possible, to check the content of said messages (it might not be possible, I don't know if SCRAM has a random part that we can't control in any way here).

test/coretasks/test_coretasks_sasl.py
```diff
@@ -344,3 +353,201 @@ def test_sasl_nak(botfactory: BotFactory, tmpconfig) -> None:
...
+
+    mockbot.on_message(
+        "900 TestBot test!test@test TestBot :You are now logged in as TestBot"
+    )
+    mockbot.on_message("903 TestBot :SASL authentication succeeded")
+    assert (
+        len(mockbot.backend.message_sent) == 6
+        and mockbot.backend.message_sent[-1] == rawlist("CAP END")[0]
+    )
```
Exirel, line 445: Ditto here about having only one assert here to check multiple condition. I really don't think it's good.

Exirel, line 445: Ditto here about having only one assert here to check multiple condition. I really don't think it's good.

test/coretasks/test_coretasks_sasl.py
```diff
@@ -344,3 +353,201 @@ def test_sasl_nak(botfactory: BotFactory, tmpconfig) -> None:
...
+
+
+def test_sasl_scram_sha_256_nonsense_server_first(mockbot: Sopel):
+    """Verify the bot handles a nonsense SCRAM-SHA-256 server_first correctly."""
```
Exirel, line 449: Nit: wording.

```suggestion
def test_sasl_scram_sha_256_invalid_server_first(mockbot: Sopel):
    """Verify the bot handles an invalid SCRAM-SHA-256 server_first correctly."""
```

test/coretasks/test_coretasks_sasl.py
```diff
@@ -344,3 +353,201 @@ def test_sasl_nak(botfactory: BotFactory, tmpconfig) -> None:
...
+    )
+    mockbot.on_message("AUTHENTICATE " + b64encode(b"junk").decode("utf-8"))
+    assert (
+        len(mockbot.backend.message_sent) == 4
+        and mockbot.backend.message_sent[-1] == rawlist("AUTHENTICATE *")[0]
+    )
+
+
+def test_sasl_scram_sha_256_nonsense_server_final(mockbot: Sopel):
+    """Verify the bot handles a nonsense SCRAM-SHA-256 server_final correctly."""
```
Exirel, line 475: Nit: wording.

```suggestion
def test_sasl_scram_sha_256_invalid_server_final(mockbot: Sopel):
    """Verify the bot handles an invalid SCRAM-SHA-256 server_final correctly."""
```

test/coretasks/test_coretasks_sasl.py
```diff
@@ -344,3 +353,201 @@ def test_sasl_nak(botfactory: BotFactory, tmpconfig) -> None:
...
+    mockbot.on_message("AUTHENTICATE " + b64encode(b"junk").decode("utf-8"))
+    assert (
+        len(mockbot.backend.message_sent) == 5
+        and mockbot.backend.message_sent[-1] == rawlist("AUTHENTICATE *")[0]
+    )
+
+
+def test_sasl_scram_sha_256_error_server_first(mockbot: Sopel):
+    """Verify the bot handles an error SCRAM-SHA-256 server_first correctly."""
+
```
Exirel, line 509:
```suggestion
```

test/test_coretasks.py
```diff
@@ -17,6 +17,7 @@
 [core]
 owner = Uowner
 nick = TestBot
+auth_password = hunter2
```
Exirel, line 20: Why is that?

Exirel (changes requested):
Same as before: the SASL mechanism implementation looks sensible. However on the test side, I've several things I would like to see modified. A couple of nitpicks on wording (I really don't like using the word "nonsense" as it isn't precise).

Also, I noted a few changes that I don't think should be in here for now: the should be discussed in their own issue/separate PR.

sopel/coretasks.py
```diff
@@ -118,9 +121,18 @@ def _handle_sasl_capability(
     cap_info = bot.capabilities.get_capability_info('sasl')
     cap_params = cap_info.params
 
-    available_mechs = cap_params.split(',') if cap_params else []
+    server_mechs = cap_params.split(',') if cap_params else []
 
-    if available_mechs and mech not in available_mechs:
+    sopel_mechs = ["PLAIN", "EXTERNAL", "SCRAM-SHA-256"]
+    if mech not in sopel_mechs:
```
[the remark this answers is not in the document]
Neustradamus: I think that the order needs to be: "SCRAM-SHA-256", "EXTERNAL", "PLAIN"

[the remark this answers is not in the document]
half-duplex: Why, @Neustradamus?

test/coretasks/test_coretasks_sasl.py
```diff
@@ -344,3 +353,201 @@ def test_sasl_nak(botfactory: BotFactory, tmpconfig) -> None:
         'CAP END',
         'QUIT :Error negotiating capabilities.',
     )
+
+
+def test_sasl_bad_method(mockbot: Sopel, caplog: pytest.LogCaptureFixture):
```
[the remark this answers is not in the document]
half-duplex: The exception raised is caught by Sopel and turned into a log message - I'm not sure how I would catch the exception.

test/test_coretasks.py
half-duplex: Missed when moving tests to the other file

Neustradamus:
Any progress on it?

half-duplex:
@Neustradamus Can you help me understand the situations where scram is beneficial? I stopped tending the PR because I realized that as far as I can tell TLS reduces its benefit to "malicious IRCd can't see your plaintext password during auth (but can during signup/change, and can always impersonate you)"

I think this (PR? ircv3 spec?) is also missing channel binding to prevent an active MITM from proxying the scram auth to another service/network.

test/coretasks/test_coretasks_sasl.py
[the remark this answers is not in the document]
half-duplex: I think the thought was that if Sopel doesn't understand a mech the user may have installed a plugin to handle it so we shouldn't bail too aggressively

PR #2364, opened 2022-10-24
Replace dev-requirements.txt with a [dev] extra

### LINKED ISSUE #2363: Move dev requirements into an extra
### Requested Feature

Right now, setting up a new development environment for Sopel requires installing `dev-requirements.txt` as well as the package itself.

It would be marginally more convenient to define a `[dev]` extra for the project and move the development requirements there, so that a user can do `pip install --editable .[dev]` to install the library _and_ development requirements.

### Problems Solved

Slightly simpler/friendlier contributor UX. I have personally made the mistake of forgetting about `dev-requirements.txt` several times

### Alternatives

_No response_

### Notes

_No response_

SnoopJ:
### Description

This PR removes `dev-requirements.txt` and moves the optional development dependencies it defines into a new `[dev]` extra. Closes #2363 

### Checklist


dgw (approved):
I will happily merge this once squashed. Maybe the final commit message is something like `install: use [dev] extra instead of dev-requirements.txt`?

(We like our easily scannable module/component prefixes, if you couldn't tell. 😛)

SnoopJ:
> I will happily merge this once squashed. Maybe the final commit message is something like `install: use [dev] extra instead of dev-requirements.txt`?
> 
> (We like our easily scannable module/component prefixes, if you couldn't tell. stuck_out_tongue)

Yea, I just didn't know which one made the most sense here, will use `install` in the squash.

Can you drop the best reference on the `flake8` issue this uncovered?

dgw:
> Can you drop the best reference on the `flake8` issue this uncovered?

Uh, that's probably PyCQA/flake8#1004 though there are numerous threads from consumers trying to solve the incompatibility later (post-release) on both flake8's and importlib_metadata's repos.

dgw:
Ultimately reverted in #2392, due to workflow issues we can't solve with pip. The "eager" upgrade strategy doesn't work the way we thought, so it became difficult to keep dev-deps up to date.

PR #2365, opened 2022-10-25
backends: try only utf8 when UTF8ONLY

half-duplex:
### Description
When UTF8ONLY is active, raise a decoding error if the message fails to decode instead of attempting other encodings.

As far as I can tell, Sopel never sends non-utf8 data unless someone's poking at the backend directly, so no other changes should be necessary to consider UTF8ONLY supported. (Actually, even this doesn't seem mandatory, just sensible)

### Checklist

### DIFF
--- a/sopel/irc/abstract_backends.py
+++ b/sopel/irc/abstract_backends.py
@@ -67,16 +67,15 @@ class AbstractIRCBackend(abc.ABC):
         # We can't trust clients to pass valid Unicode.
         try:
             data = str(line, encoding='utf-8')
-        except UnicodeDecodeError:
+        except UnicodeDecodeError as e:
+            # ...unless the server announces UTF8ONLY
+            if "UTF8ONLY" in self.isupport:
+                raise e
             # not Unicode; let's try CP-1252
             try:
                 data = str(line, encoding='cp1252')
             except UnicodeDecodeError:
-                # Okay, let's try ISO 8859-1
-                try:
-                    data = str(line, encoding='iso8859-1')
-                except UnicodeDecodeError:
-                    raise ValueError('Unable to decode data from server.')
+                raise ValueError('Unable to decode data from server.')
 
         return data
 
--- a/test/irc/test_irc_abstract_backends.py
+++ b/test/irc/test_irc_abstract_backends.py
@@ -1,7 +1,9 @@
 """Tests for core ``sopel.irc.backends``"""
 from __future__ import annotations
 
+import pytest
 
+from sopel.irc.isupport import ISupport
 from sopel.tests.mocks import MockIRCBackend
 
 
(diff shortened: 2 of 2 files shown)


half-duplex:
Was trying to add tests and... Windows-1252 is a superset of ISO/IEC 8859-1. I'm removing the latter.

sopel/irc/abstract_backends.py
```diff
@@ -67,16 +67,15 @@ def decode_line(self, line: bytes) -> str:
         # We can't trust clients to pass valid Unicode.
         try:
             data = str(line, encoding='utf-8')
-        except UnicodeDecodeError:
+        except UnicodeDecodeError as e:
+            # ...unless the server announces UTF8ONLY
+            if "UTF8ONLY" in self.isupport:
+                raise e
```
dgw, line 73: Wow, this is the perfect place for `e.add_note("UTF8ONLY server sent a line that could not be decoded as UTF-8")`, but that requires Python 3.11 🙁 (or a nasty monkeypatch, I guess)

dgw (commented):
No test case with invalid input that will trigger the final `raise ValueError`?

test/irc/test_irc_abstract_backends.py
```diff
@@ -298,3 +300,31 @@ def test_send_notice_safe():
...
+    assert backend.decode_line(test.encode("utf-8")) == test
+
+
+def test_decode_line_utf8only():
+    bot = BotCollector()
+    backend = MockIRCBackend(bot)
+    backend.isupport = ISupport(UTF8ONLY=None)
+
+    test = "PRIVMSG #sopel :Hello, Martín!"
+    with raises(UnicodeDecodeError):
```
dgw, line 320: I tend to prefer `with pytest.raises` because it's clearer. Others (Exi) might disagree, but I like to keep the namespace of imported stuff. See-also: My preference for e.g. `from sopel import plugin` and then `@plugin.command()`.

(Forgot to add this _before_ submitting that review just now, sorry.)

sopel/irc/abstract_backends.py
half-duplex: Yeah. Not a bad place for a regular LOGGER.warning() though.

half-duplex: Never mind, there's no LOGGER in there... meh

test/irc/test_irc_abstract_backends.py
half-duplex: Yep, fair. It's also tests so being thrifty with includes isn't exactly priority zero.

half-duplex:
Added, but more for coverage than because it matters =p

dgw (approved):
Cool. Assuming CI passes and all that (why wouldn't it?), it's squash time.

PR #2366, opened 2022-10-27
formatting: fix `color('some text', 0)`

dgw:
### Description
A literal `0` (nominally the color white, in mIRC convention) evaluates as `False`, which was being skipped by the `not` and making the function return its input text unchanged.

Regression check added to existing `test_color()` case.

### Checklist

### Notes
No further 7.x releases are planned, but this fix probably _should_ be included if another patch version happens.

This took about 10x as long as it should have because I ran into `setuptools>=64.0` breaking how Sopel does editable installs and had to figure out that problem before I could actually test the patch. That problem needs its own issue/PR, but I still need to do a little more research before I can open a useful one of either.

Kudos to @xnaas for hacking some extra color schemes into a forked sopel-rainbow plugin, which revealed this bug.

### DIFF
--- a/test/test_formatting.py
+++ b/test/test_formatting.py
@@ -29,6 +29,12 @@ def test_color():
     assert color(text, 0) == '\x0300' + text + '\x03'  # ensure that color 0 isn't treated as boolean false
     pytest.raises(ValueError, color, text, 100)
     pytest.raises(ValueError, color, text, 'INVALID')
+    pytest.raises(ValueError, color, text, '')
+    # test background color as well
+    pytest.raises(ValueError, color, text, '', '')
+    pytest.raises(ValueError, color, text, None, '')
+    pytest.raises(ValueError, color, text, colors.PINK, '')
+    pytest.raises(ValueError, color, text, '', colors.PINK)
 
 
 def test_hex_color():
--- a/test/test_formatting.py
+++ b/test/test_formatting.py
@@ -24,6 +24,9 @@ def test_color():
     assert color(text) == text
     assert color(text, colors.PINK) == '\x0313' + text + '\x03'
     assert color(text, colors.PINK, colors.TEAL) == '\x0313,10' + text + '\x03'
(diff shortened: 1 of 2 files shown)


test/test_formatting.py
```diff
@@ -24,6 +24,9 @@ def test_color():
     assert color(text) == text
     assert color(text, colors.PINK) == '\x0313' + text + '\x03'
     assert color(text, colors.PINK, colors.TEAL) == '\x0313,10' + text + '\x03'
+    assert color(text, '3') == '\x0303' + text + '\x03'  # ensure single-digit str is zero-padded
+    assert color(text, 3) == '\x0303' + text + '\x03'  # ensure ints are zero-padded
+    assert color(text, 0) == '\x0300' + text + '\x03'  # ensure that color 0 isn't treated as boolean false
     pytest.raises(ValueError, color, text, 100)
     pytest.raises(ValueError, color, text, 'INVALID')
```
Exirel, line 31: One more group of tests is required here: check what happen with an empty string.

```suggestion
    pytest.raises(ValueError, color, text, 100)
    pytest.raises(ValueError, color, text, 'INVALID')
    pytest.raises(ValueError, color, text, '')
    # test background color as well
    pytest.raises(ValueError, color, text, '', '')
    pytest.raises(ValueError, color, text, None, '')
    pytest.raises(ValueError, color, text, colors.PINK, '')
    pytest.raises(ValueError, color, text, '', colors.PINK)
```

Exirel (changes requested):

test/test_formatting.py
dgw: Good point.

Relatedly, I thought about adding type hints to this stuff as long as I was touching it, but couldn't find a sensible way to indicate "this takes an int _in a certain value range_, or a str _but not an empty one_". If there is none and we're meant to just annotate the type and let `ValueError` etc. handle the acceptable values within that type, then sure, type hinting this module could be a future PR of its own.

Exirel (approved):

PR #2367, opened 2022-10-27
build: pin setuptools while we figure out PEP-0660 editable install

dgw:
### Description
In setuptools 64.0, editable installs were changed. As of 65.x, I was unable to install sopel as an editable distribution without getting ImportErrors even when trying to do something as simple as run `pytest --version` (which shouldn't load any Sopel code).

I *believe* this is enough to work around the issue for now, given that pyproject.toml packages are supposed to use build isolation and the used setuptools version is separate from whatever's installed in the user's environment. At least my brief testing bore this out, with ~=63.0 requested by the TOML file and 65.something installed in pip.

### Checklist

### Notes
No milestone; this is entirely for us developers, not tied to any release. (It's a build-time dependency, not required at runtime.)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,5 +1,5 @@
 [build-system]
-requires = ["setuptools>=63.0", "wheel"]
+requires = ["setuptools~=63.0", "wheel"]
 build-backend = "setuptools.build_meta"
 
 [tool.setuptools]


Exirel (approved):
This is sad. Maybe it's time to switch to another backend.

PR #2370, opened 2022-11-02
trigger, irc: use STATUSMSG to handle status specific messages in channels

### LINKED ISSUE #1886: STATUSMSG not handled in `Identifier.is_nick()`

Exirel:
### Description

Fix #1886.

In this PR, I modified `PreTrigger` to parse the `sender` to check if the prefix was a `STATUSMSG` prefix, i.e. a message intended to a specific privilege level in the channel (such as ops, voice, etc.). Once this is done, the `PreTrigger` remove that prefix from the `sender` to create an instance of `Identifier`. On the bot's side, every time a `PreTrigger` is instantiated, `bot.isupport.get('STATUSMSG')` is provided.

I added some tests to show how that works and ensure it works as intended without breaking anything, so we should be all good.

Two notes to consider for the future:

* we may want to change the list of channel prefix to something more restrictive, or not
* we may want to set a default value for the `STATUSMSG` value, or have something a bit smarter about that

Both to be considered later, not in this PR.

### Checklist

### DIFF
--- a/sopel/irc/__init__.py
+++ b/sopel/irc/__init__.py
@@ -322,6 +322,7 @@ class AbstractBot(abc.ABC):
             message,
             url_schemes=self.settings.core.auto_url_schemes,
             identifier_factory=self.make_identifier,
+            statusmsg_prefixes=self.isupport.get('STATUSMSG'),
         )
         if all(
             cap not in self.enabled_capabilities
@@ -371,6 +372,7 @@ class AbstractBot(abc.ABC):
                 ":{0}!{1}@{2} {3}".format(self.nick, self.user, host, raw),
                 url_schemes=self.settings.core.auto_url_schemes,
                 identifier_factory=self.make_identifier,
+                statusmsg_prefixes=self.isupport.get('STATUSMSG'),
             )
             self.dispatch(pretrigger)
 
--- a/sopel/trigger.py
+++ b/sopel/trigger.py
@@ -169,6 +169,7 @@ class PreTrigger:
         line: str,
         url_schemes: Optional[Sequence] = None,
         identifier_factory: IdentifierFactory = identifiers.Identifier,
+        statusmsg_prefixes: Tuple[str, ...] = tuple(),
     ):
         self.make_identifier = identifier_factory
         line = line.strip('\r\n')
@@ -243,9 +244,13 @@ class PreTrigger:
         # If we have arguments, the first one is *usually* the sender,
         # most numerics and certain general events (e.g. QUIT) excepted
         target: Optional[identifiers.Identifier] = None
+        status_prefix: Optional[str] = None
 
         if self.args and self.event in COMMANDS_WITH_CONTEXT:
-            target = self.make_identifier(self.args[0])
+            raw_target = self.args[0]
+            if statusmsg_prefixes and raw_target[0] in statusmsg_prefixes:
+                status_prefix, raw_target = raw_target[0], raw_target[1:]
+            target = self.make_identifier(raw_target)
 
             # Unless we're messaging the bot directly, in which case that
             # second arg will be our bot's name.
@@ -253,6 +258,7 @@ class PreTrigger:
                 target = self.nick
 
         self.sender = target
+        self.status_prefix = status_prefix
 
         # Parse CTCP
         if self.event == 'PRIVMSG' or self.event == 'NOTICE':
@@ -331,6 +337,16 @@ class Trigger(str):
         The :attr:`COMMANDS_WITH_CONTEXT` attribute lists IRC commands for
         which ``sender`` can be relied upon.
 
+    """
+    status_prefix = property(lambda self: self._pretrigger.status_prefix)
+    """The prefix used for the :attr:`sender` for status-specific messages.
+
+    :type: Optional[str]
+
+    This will be ``None`` by default. If a message is sent to a channel, it may
+    have a status prefix for status-specific messages. In that case, the prefix
+    is removed from the channel's identifier (see :attr:`sender`) and saved to
+    this attribute.
     """
     time = property(lambda self: self._pretrigger.time)
     """When the message was received.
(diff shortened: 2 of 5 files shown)


sopel/irc/__init__.py
```diff
@@ -309,6 +309,7 @@ def on_message(self, message: str) -> None:
             message,
             url_schemes=self.settings.core.auto_url_schemes,
             identifier_factory=self.make_identifier,
+            statusmsg_prefixes=self._isupport.get('STATUSMSG'),
```
dgw, line 312: Why the `_` version?

```suggestion
            statusmsg_prefixes=self.isupport.get('STATUSMSG'),
```

sopel/irc/__init__.py
```diff
@@ -358,6 +359,7 @@ def on_message_sent(self, raw: str) -> None:
                 ":{0}!{1}@{2} {3}".format(self.nick, self.user, host, raw),
                 url_schemes=self.settings.core.auto_url_schemes,
                 identifier_factory=self.make_identifier,
+                statusmsg_prefixes=self._isupport.get('STATUSMSG'),
```
dgw, line 362: Same question as above. Previously existing uses of ISUPPORT data in this file use the property, not the attr directly.

```suggestion
                statusmsg_prefixes=self.isupport.get('STATUSMSG'),
```

sopel/trigger.py
```diff
@@ -58,6 +58,9 @@
 ``sender`` property will be ``None``.
 """
 
+DEFAULT_STATUSMSG = ('@',)
```
dgw, line 61: Clients ["SHOULD NOT"](https://modern.ircdocs.horse/#statusmsg-parameter) try to use this feature unless a list of accepted prefixes is advertised, and servers "MUST" specify a value for the 005 token. The default should probably be an empty tuple.

Two older references ([draft-hardy-irc-isupport-00](https://datatracker.ietf.org/doc/html/draft-hardy-irc-isupport-00#section-4.18) and [draft-brocklesby-irc-isupport-01](https://datatracker.ietf.org/doc/html/draft-brocklesby-irc-isupport-01#section-3.11)) reiterate that the 005 token's value is required.

sopel/trigger.py
```diff
@@ -331,6 +343,15 @@ class Trigger(str):
...
 
+    """
+    status_prefix = property(lambda self: self._pretrigger.status_prefix)
+    """The prefix used for the :attr:`sender` for status-specific messages.
+
+    :type: str | None
+
+    This will be ``None`` by default. If a message is sent to a channel, it may
+    have a status prefix for status-specific messages. In that case, the prefix
+    is removed from the channel's identifier and set to this attribute.
```
dgw, line 354:
```suggestion
    is removed from the channel's identifier and saved to this attribute.
```

(Also, cross-ref `identifier` to the class?)

sopel/trigger.py
```diff
@@ -331,6 +343,15 @@ class Trigger(str):
         The :attr:`COMMANDS_WITH_CONTEXT` attribute lists IRC commands for
         which ``sender`` can be relied upon.
 
+    """
+    status_prefix = property(lambda self: self._pretrigger.status_prefix)
+    """The prefix used for the :attr:`sender` for status-specific messages.
+
+    :type: str | None
```
dgw, line 350: I'm too lazy to test it (and it's 01:30 here anyway), but would this work?

```suggestion
    :type: Optional[str]
```

dgw (changes requested):
I don't think this version of the patch complies with the STATUSMSG spec. See line notes.

Don't worry, there are also some of the usual doc/comment nitpicks. 😜

sopel/trigger.py
Exirel: Oh, nice catch! I was too focus on that and forgot that there should not be any default. Will fix that.

sopel/irc/__init__.py
Exirel: Prevent a call to a property which create one more call than necessary (object -> property -> _ version).

sopel/irc/__init__.py
Exirel: I wonder if the other usage are there because it was made to be backward compatible? I honestly can't tell, it's been too long.

sopel/trigger.py
Exirel: Both work.

sopel/trigger.py
Exirel: Fixed, with added tests.

sopel/trigger.py
Exirel: I tried to do `Channel.name` and it was ugly and not really helpful. So I went for `:attr:sender` which is more relevant to the context.

dgw (approved):
I think it's squash time. Might be spaghetti squash, but time will tell us that. 🙃

Exirel:
Squashed!

PR #2371, opened 2022-11-02
config: strip quotes from FilenameAttribute

### LINKED ISSUE #2263: config: FilenameAttribute doesn't handle quotes well
### Description
If the value of a `FilenameAttribute` set in the config file is quoted, it silently creates a mess.

### Reproduction steps
1. Set `core.logdir = "/path/to/logs"`
2. Run Sopel
3. Sopel quietly creates a silly directory tree inside `$HOMEDIR`:
```
$ tree ~/.sopel
/home/dgw/.sopel
├── "
│   └── path
│       └── to
│           └── logs"
│               ├── default.error.log
│               ├── default.exceptions.log
│               ├── default.raw.log
│               └── default.sopel.log
...
```

### Expected behavior
Even though quotes aren't required in Sopel's config file, they should be sensibly handled.

### Environment
- Sopel `.version`: current `master`, aca10553da2a996aaaa4051741fae6fb7e6c4940
- Sopel installed via: `-e .` @ above commit
- Python version: 3.8, 3.9
- Operating system: Ubuntu 20.04 (WSL)

### Notes
This is a leftover todo from #1959, where I updated the docs but didn't make any code changes related to this.

Exirel:
### Description

Fix #2263. That was way easier than I expected.

### Checklist

### DIFF
--- a/sopel/config/types.py
+++ b/sopel/config/types.py
@@ -697,6 +697,11 @@ class FilenameAttribute(BaseValidated):
         if not value:
             return self.parse(value)
 
+        if value.startswith('"') and value.endswith('"'):
+            value = value.strip('"')
+        elif value.startswith("'") and value.endswith("'"):
+            value = value.strip("'")
+
         result = os.path.expanduser(value)
         if not os.path.isabs(result):
             if not self.relative:
--- a/test/test_config.py
+++ b/test/test_config.py
@@ -227,6 +227,67 @@ def test_fileattribute_given_file_when_dir(fakeconfig):
         fakeconfig.fake.ad_fileattr = testfile
 
 
+def test_fileattribute_relative_with_quotes(fakeconfig):
+    testfile = os.path.join(fakeconfig.core.homedir, 'test.tmp')
+
+    # double quotes
+    fakeconfig.fake.rf_fileattr = '"test.tmp"'
+    assert fakeconfig.fake.rf_fileattr == testfile
+
+    # single quotes
+    fakeconfig.fake.rf_fileattr = "'test.tmp'"
+    assert fakeconfig.fake.rf_fileattr == testfile
+
+
+def test_fileattribute_relative_with_nested_double_quotes(fakeconfig):
+    testfile = os.path.join(fakeconfig.core.homedir, '"test.tmp"')
+
+    # double quotes nested inside single quotes
+    fakeconfig.fake.rf_fileattr = "'\"test.tmp\"'"
+    assert fakeconfig.fake.rf_fileattr == testfile
+
+
+def test_fileattribute_relative_with_nested_single_quotes(fakeconfig):
+    testfile = os.path.join(fakeconfig.core.homedir, "'test.tmp'")
+
+    # single quotes nested inside double quotes
+    fakeconfig.fake.rf_fileattr = '"\'test.tmp\'"'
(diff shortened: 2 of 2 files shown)


sopel/config/types.py
```diff
@@ -696,6 +696,9 @@ def _parse(self, value, settings, section):
         if not value:
             return self.parse(value)
 
+        if value.startswith('"') or value.endswith('"'):
+            value = value.strip('"')
```
dgw, line 700: I would understand using this `if` for the `and` case, but why for `or`?

dgw (commented):
One possibly dumb question in a line note, and a general question here: Why not also for single-quotes?

sopel/config/types.py
Exirel: You're right, there is no need for the `or`. As for the `and`... now I wonder if it is what I wanted to write, or if it is unnecessary.

Do we want to consider `"some_filename.txt` a file *with* a `"` or not? I guess not, probably? That would be a breaking change, but one we can accept for Sopel 8. People shouldn't have quotes in their filename anyway.

dgw: I, uh… Hmm.

Common sense says "of course that should have the `"` removed!" But then I remember that if the `"` appears anywhere else in the basename, it won't be removed, and that annoys me. So maybe the suggestion below is what it should be (added handling of `'` too while I was at it):

```suggestion
        if value.startswith('"') and value.endswith('"'):
            value = value.strip('"')

        if value.startswith("'") and value.endswith("'"):
            value = value.strip("'")
```

Who are we to tell users they _can't_ set a `FilenameAttribute` to `"Who's On First?".txt` if they _really_ want it? It's perfectly legal (on common Linux distros, at least).

half-duplex: Can we please just switch to TOML where all strings are quoted and escaping stupid filenames is the user's problem?
I agree with `and` until then.

dgw: > Can we please just switch to TOML

See: all the IRC discussions about that, lol

sopel/config/types.py
```diff
@@ -696,9 +696,12 @@ def _parse(self, value, settings, section):
         if not value:
             return self.parse(value)
 
-        if value.startswith('"') or value.endswith('"'):
+        if value.startswith('"') and value.endswith('"'):
             value = value.strip('"')
 
+        if value.startswith("'") and value.endswith("'"):
+            value = value.strip("'")
```
dgw, line 703: My first thought looking at this structure was "what if the name has multiple nested levels of quotes?"

And I thought about it a little more. This will strip `"'filename.txt'"` correctly, but `'"filename.txt"'` will only be stripped down to `"filename.txt"`.

My gut says we should either only support the first layer of quoting, or repeatedly strip both types of quote until nothing changes. I'm in favor of the first option because it's very simple to do. 😁

```suggestion
        if value.startswith('"') and value.endswith('"'):
            value = value.strip('"')
        elif value.startswith("'") and value.endswith("'"):
            value = value.strip("'")
```

Exirel: That's fair.

Exirel:
Fixed and squashed (as per discussion on IRC).

dgw (approved):
New tests for nested quotes are 👌

PR #2375, opened 2022-11-04
bot, irc, coretasks: more robust `bot.connection_registered` value

dgw:
### Description
The `connection_registered` attribute itself is now a property, which first checks to see if the bot's `backend` exists (is not `None`) and then whether the `backend.is_connected()`, before finally seeing if the internal `_connection_registered` flag has been set.

Also added a few places where the internal flag should be reset back to its initial value of `False`, such as if the bot is about to quit or the backend's `on_close()` handler was called.

All of this is to try and make `bot.connection_registered` reliable for plugins like `remind`, which have scheduled tasks that might need to send data to IRC and need a reliable way to check whether the connection is ready to accept such data before operating.

### Checklist
  - Note that this required adding a bit more setup to two `test_coretasks` units.

### DIFF
--- a/sopel/bot.py
+++ b/sopel/bot.py
@@ -976,6 +976,8 @@ class Sopel(irc.AbstractBot):
     def _shutdown(self) -> None:
         """Internal bot shutdown method."""
         LOGGER.info("Shutting down")
+        # Proactively tell plugins (at least the ones that bother to check)
+        self._connection_registered = False
         # Stop Job Scheduler
         LOGGER.info("Stopping the Job Scheduler.")
         self._scheduler.stop()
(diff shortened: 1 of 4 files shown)


Exirel (approved):
That will do.

I wonder if [a Threading.Event](https://docs.python.org/3/library/threading.html#threading.Event) could be used for `_connection_registered` instead of a boolean, but that's a nitpick and not necessary right now.

dgw:
> [Threading.Event](https://docs.python.org/3/library/threading.html#threading.Event) could be used for `_connection_registered` instead of a boolean

This is a neat idea and worth looking into, largely because of the `.wait()` method that would make available.

PR #2379, opened 2022-11-19
tools.time: versionadded annotations for great justice

dgw:
### Description
Was working on something in sopel-github and got annoyed that I had to check `git blame` to verify that the functions I wanted to use would be available in that plugin's minimum required Sopel version.

`willie.tools.time` came into existence in the 5.3 release, so I used that as the module's `versionadded`, and treated the namespace change from `willie` to `sopel` as a `versionchanged` event.

If people are nice to me (read: this gets reviewed in something less than a month 😝), I might go through the rest of `sopel.tools` and add any missing version annotations in its other submodules, too. This is the sort of thing I'm always on the lookout for when working on other stuff, because if the documentation annoys _me_ (who's been around the project for almost a decade), it will _almost certainly_ annoy noobs.

### Relevant preview link

I touched only `sopel/tools/time.py`, so of the whole preview deploy available from Netlify, only https://deploy-preview-2379--sopel-unstable-docs.netlify.app/package/tools/time.html is relevant for review.

### Checklist
  - (It doesn't touch code. I test-built the docs.)

### DIFF
--- a/sopel/tools/time.py
+++ b/sopel/tools/time.py
@@ -1,4 +1,11 @@
-"""Tools for getting and displaying the time."""
+"""Tools for getting and displaying the time.
+
+.. versionadded:: 5.3
+.. versionchanged:: 6.0
+
+    Moved from ``willie`` namespace to ``sopel`` namespace for project rename.
+
+"""
 from __future__ import annotations
 
 import datetime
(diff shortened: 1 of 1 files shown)


Exirel (approved):
I like this change. I think we could have a spec for our docstrings that says "add the versionadded/changed at the bottom but before ..note and such". This is more or less what we did so far and I think it works for us.

dgw:
> versionadded/changed at the bottom but before ..note and such

Seems like that's what the Python docs do, in general, too. Which is probably why I used that style. 😁

PR #2382, opened 2022-12-02
CodeQL minor nitpick pass

dgw:
### Description
For #2381 I made sure we wouldn't introduce any Error-level alerts, because those would have given the `master` branch a ❌ check status. Anything below that level (that wasn't a false-positive or a test case) was mostly left alone.

This PR's branch name, `useless-code`, is inspired by the alert label that caught my eye first. At time of opening that label also applies to most of the alerts that will be fixed by this patch.

### Checklist
  - I managed to break my flake8 somehow, but these tweaks are so minor I did not waste time trying to fix it just for this PR

### DIFF
--- a/sopel/config/types.py
+++ b/sopel/config/types.py
@@ -101,8 +101,10 @@ class StaticSection:
         attribute = getattr(self.__class__, name)
         if default is NO_DEFAULT and not attribute.is_secret:
             try:
+                # get current value of this setting to use as prompt default
                 default = getattr(self, name)
             except AttributeError:
+                # there is no current value; that's OK
                 pass
             except ValueError:
                 print('The configured value for this option was invalid.')
--- a/sopel/formatting.py
+++ b/sopel/formatting.py
@@ -23,6 +23,9 @@ __all__ = [
     'CONTROL_STRIKETHROUGH',
     'CONTROL_MONOSPACE',
     'CONTROL_REVERSE',
+    # convenience lists
+    'CONTROL_FORMATTING',
+    'CONTROL_NON_PRINTING',
     # utility functions
     'color',
     'hex_color',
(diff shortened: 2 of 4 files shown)


SnoopJ (approved):
Changes look fine to me. The commit message for 0b50255 should probably reference the CodeQL rule that it's fixing but you may plan to squash those references away anyhow.

I would like to formally declare my annoyance with CodeQL's `py/import-and-import-from` rule which looks more like someone's personal aesthetics than a serious concern about maintenance (what's the alternative supposed to be, `import foo; # other imports; something = foo.something` ?!)

dgw:
> The commit message for [0b50255](https://github.com/sopel-irc/sopel/commit/0b50255af3ac7a00278d906156423f3f1a2131eb) should probably reference the CodeQL rule that it's fixing

Yeah, I forgot that one, whoops. Fixed.

> I would like to formally declare my annoyance with CodeQL's `py/import-and-import-from` rule which looks more like someone's personal aesthetics than a serious concern about maintenance (what's the alternative supposed to be, `import foo; # other imports; something = foo.something` ?!)

The alternative is supposed to be just using `foo.something` with the namespace, in your example case.

I personally agree with the idea that things should be imported _either_ under a module namespace _or_ individually, but not both. As a contrived demo, I think it would be silly to `from sopel import plugin` and then also `from sopel.plugin import commands`, and I think it _would_ lead to potential confusion later.

But I also don't like importing individual module members in most cases anyway. I've found that the best way to have future-proof imports is always importing a (sub)module, and using that module's members as members (`.` notation). Then you 1) always know where something came from, because the module it's from is always included at the site of use; and 2) you can use other features of the module without modifying imports.

_Because of (1) above, I also tend to avoid using `import ____ as`, except to avoid name collisions._

Exirel (approved):

PR #2385, opened 2022-12-10
tools: deprecations & reorganization (feat. `sopel.cli`)

dgw:
### Description
Started looking at `sopel.tools` as a follow-up to #2379 and noticed some more cleanup that could be done. Specifically:

* `OutputRedirect` class is used by nothing, appears to be a relic of the 3.x-era logging system, and should be deprecated.
* `check_pid()` belongs under the `cli` namespace, because that's the only code that needs it. And plugins shouldn't be touching PIDs anyway.
* `stderr()` also ideally should never be used by plugins. If they really really want to print directly to a terminal instead of using a logger, they have access to `import sys; print('some bs', file=sys.stderr)` on their own.
* `get_input()` is already deprecated, but I had to replace uses of it to eliminate circular imports (because I'm perhaps foolishly keeping the moved functions available in `tools` until 8.1).

### Checklist

### Notes
I was going to do this stuff a bit more atomically, moving things from `tools` to `cli` in a series of several commits, but that approach ran into circular-import errors if I tried to only move one function at a time. So I'm sorry, but I couldn't find a reasonable way to break it down into smaller commits. You'll just have to review the whole thing at once. 🙏

### DIFF
--- a/sopel/cli/config.py
+++ b/sopel/cli/config.py
@@ -4,7 +4,6 @@ from __future__ import annotations
 import argparse
 import os
 
-from sopel import tools
 from . import utils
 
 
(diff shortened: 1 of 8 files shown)


SnoopJ (approved):
Pretty straightforward reorganization, LGTM :+1:

Exirel (approved):
Let's deprecate everything. 😈

Issue #2373, opened 2022-11-04
Backend is not initialized during plugin setup


dgw:
### Problem analysis
When the `AbstractBot` that forms the foundation of a Sopel instance is initialized, it sets its `backend` to `None`:

https://github.com/sopel-irc/sopel/blob/53c5543d15a73aaba904ee5093212a575b3b6cf9/sopel/irc/__init__.py#L75-L76

This attribute isn't populated until the bot's `run()` method gets called:

https://github.com/sopel-irc/sopel/blob/53c5543d15a73aaba904ee5093212a575b3b6cf9/sopel/irc/__init__.py#L254-L263

Calling `run()` initiates the connection to the IRC network, so all plugins (especially `coretasks`) must be loaded by this point. But this means that _before_ `run()` is called, there is no `backend`, and therefore plugins that check the state of the IRC connection "the nice way" (by checking `bot.backend.is_connected()`) will raise an `AttributeError` if that check runs during the setup phase, while the `backend` is still `None`.

One such plugin is `remind`, which starts a scheduled task that dutifully checks (as of #2351) whether the IRC connection is ready yet:

https://github.com/sopel-irc/sopel/blob/53c5543d15a73aaba904ee5093212a575b3b6cf9/sopel/modules/remind.py#L163-L165

Per the documentation for `@plugin.interval`, that scheduled task _may_ be called before the bot is connected to IRC, and sometimes it is. This previously (before #2351) led to a `RuntimeError` because the plugin tried to `say()` messages before the `backend` was initialized, and those messages were subsequently lost. Now that the plugin is more well behaved, it causes the above-mentioned `AttributeError` instead under the same circumstances—but at least now it no longer silently eats the offending reminders.

### Proposed solution
I would like to separate initialization of the `AbstractBot.backend` attribute from running the backend, so the backend will be available to query even if a scheduled task happens to run before the network connection is initiated.

The reason I'm putting this up for discussion first is that it would be a redesign of the interface. The `host` and `port` used by the backend are currently passed in through a long chain of method calls:

  - [`cli.run.command_start(opts)`](https://github.com/sopel-irc/sopel/blob/53c5543d15a73aaba904ee5093212a575b3b6cf9/sopel/cli/run.py#L483) (when `sopel` runs with [the [default] `start` action](https://github.com/sopel-irc/sopel/blob/53c5543d15a73aaba904ee5093212a575b3b6cf9/sopel/cli/run.py#L476))
[...]

PR #2394, opened 2023-01-04
irc, backends: dummy non-connected backend for convenience

dgw:
### Description
Part of a strategy for addressing #2373, we want to have a backend that works like a normal backend but isn't connected yet, for use instead of a `NoneType` object (which obviously won't have any of the methods a backend would).

The idea is to make it so calls to things like `backend.is_connected()` still work even pre-connection, so it's easier for core & plugin code to be responsible and check the connection status before trying to send things over the IRC connection.

### Checklist
  - Will revisit this when the PR is actually final (starting as draft)
  - Ditto above

### DIFF
--- a/sopel/irc/__init__.py
+++ b/sopel/irc/__init__.py
@@ -45,7 +45,7 @@ from typing import (
 
 from sopel import tools, trigger
 from sopel.tools import identifiers
-from .backends import AsyncioBackend
+from .backends import AsyncioBackend, NotConnectedBackend
 from .isupport import ISupport
 from .utils import CapReq, safe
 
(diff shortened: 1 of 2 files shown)


Exirel:
As per our IRC discussion:

> Honestly, this looks pretty good to me.
> I think I'd reconsider "NotConnected", and maybe say "NotInitialized", because that's what it is in the end.
> And maybe add some more docstring here and there.
> And last but not least, I'd add a note in the "advanced.rst" chapter in the doc for Plugins to say "Hey, between setup and connection attempt, all you have is a fake backend that will raise exception and return False if you ask if it is connected.".
> Take this opportunity to document the lifecycle of Sopel's backend and the IRC connection.
> Which... could be done in a future PR, related to the refactoring of said "how to tell if the bot is connected, if the bot started capability negotiation, if the bot finished negotiation, if the bot got RPL_WELCOME, if the bot started joining channels, etc."
> That's a lot of events to track actually!

Not everything need or should be done in this PR. Probably just the new name and a bit more docstring. Everything else (Sopel connection/IRC initialization events and lifecycle) should probably be in another PR.

dgw:
I am going to ready this PR, even though I haven't had a chance to run any manual tests against it. Really want feedback on the docstring improvements; the dummy backend itself is pretty straightforward.

Exirel (approved):
Since this respect the kind of type safety I'd like more in Sopel, I think we can merge that, and see later if we want to be pedantic about names and stuff.

PR #2383, opened 2022-12-03
Track `User`'s realname

dgw:
### Description
I saw a very angry person talking about Sopel somewhere (intentionally vague). Their complaint _du jour_ was that Sopel doesn't expose `User`s' realnames.

While I chose not to engage them in conversation (some users, you're better off without, as goes the conventional business wisdom), the gap in our API is a valid point and it's not difficult to fix.

### Checklist
- [x] I have read [CONTRIBUTING.md](https://github.com/sopel-irc/sopel/blob/master/CONTRIBUTING.md)
- [x] I can and do license this contribution under the EFLv2
- [ ] No issues are reported by `make qa` (runs `make quality` and `make test`)
  - My flake8 is still busted 🙁 and the test suite doesn't cover WHO/WHOX handling yet anyway ☹️
- [x] I have tested the functionality of the things this change touches
  - Automated testing no, but I manually verified (via `sopel-ipython` console) that realnames were correctly populated from both WHO (on Rizon) and WHOX (on Libera) replies

### Notes
I needed to get back to something else and didn't have time to work on tests right now, but still wanted to get the code itself out there for y'all wonderful people to look at. It'd be entirely valid to refuse approval until the test suite covers these functions, and I'd absolutely not take it personally. 😉

PR #2396, opened 2023-01-10
coretasks: send WHO to get realname for never-seen users on JOIN

### LINKED ISSUE #2387: coretasks: no realname for new JOINs on `away-notify` servers
As far as I can tell, the only place the realname of a newly joined user might be fetched is in the periodic WHO requests Sopel sends, which are currently skipped if the server supports `away-notify`:

https://github.com/sopel-irc/sopel/blob/9965ccea7b9aaa566e2402c1bb1c12c3ff31bfc1/sopel/coretasks.py#L806-L808

My half-baked ideas for this (not a ton of free time to think/research today) are:

- Skip the above check and always `_send_who()` for channels periodically
- Immediately `WHO` the exact nick after a user JOINs a channel and hasn't been seen before

I will edit the list if I come up with more ideas when actually working on this.

Follow-up to: #2383

dgw:
### Description
This should make it so that, even on networks with `away-notify` support (where Sopel disables its regular WHO polling), users who were not in a given channel when Sopel joined will still have a `realname` value set in relatively short order.

The WHO command can take a nickname, a channel name, or a pattern mask. Sopel previously made use of only the channel name type. I refactored the `_send_who()` helper a bit so it can take either a channel or nick, and tweaked the JOIN handler to send a WHO for any user Sopel didn't already know about (wasn't present in `bot.users` at the time of the JOIN event).

Patches #2383; resolves #2387

### Checklist
  - Lazily letting GHA do this for me
  - Hopped onto Rizon (which has `away-notify`) with my testbot, used `sopel-ipython` to inspect the `bot.users` list, then joined the test channel again with a webchat client and verified that its realname was populated through the ipython console. Before this patch, the new client's realname was empty when following this same procedure.

### DIFF
--- a/sopel/coretasks.py
+++ b/sopel/coretasks.py
@@ -782,7 +782,7 @@ def _remove_from_channel(bot, nick, channel):
                 bot.users.pop(nick, None)
 
 
-def _send_who(bot, channel):
+def _send_who(bot, mask):
     if 'WHOX' in bot.isupport:
         # WHOX syntax, see http://faerion.sourceforge.net/doc/irc/whox.var
         # Needed for accounts in WHO replies. The `WHOX_QUERYTYPE` parameter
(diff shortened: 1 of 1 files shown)


sopel/coretasks.py
```diff
@@ -839,16 +840,19 @@ def track_join(bot, trigger):
     to know more about said user (privileges, modes, etc.).
     """
     channel = trigger.sender
+    new_channel = channel not in bot.channels
+    self_join = trigger.nick == bot.nick
+    new_user = trigger.nick not in bot.users
```
SnoopJ, line 845: nice improvement to legibility here :grin:

SnoopJ (approved):
Changes look good to me

Exirel (approved):
I like that. The little change for `new_channel`, `self_join`, etc. is especially nice.

dgw:
Merged & pulled to Sopel bot at Libera (which has `away-notify`) for testing.

PR #2404, opened 2023-01-26
cli.run: remove obsolete `--quiet` arg

dgw:
This argument was already nonfunctional in 7.x anyway, due to #1678 and related logging changes. Seems like no one was using it (since we never received a report/complaint that it was broken).

Instead of this, users who need Sopel to run without printing to the terminal can use the usual shell trick: `command >> /dev/null 2>&1`

### Checklist

### Notes
Labeled as breaking change because it needs to be mentioned in the changelog, even though the actual breakage occurred in 7.0.


dgw:
Fixed. Forgot to modify tests before.

SnoopJ (approved):
Change looks good to me. The related tests look a little weird (especially all that `hasattr`) like they're effectively testing `argparse`, but such are the joys of long maintenance tails I guess.

dgw:
> The related tests look a little weird (especially all that `hasattr`) like they're effectively testing `argparse`

It's testing that someone remembered to add the expected arguments to the parser for each subcommand—or rather, that no one accidentally deletes one of the expected arguments later. @Exirel is _very_ thorough, at least usually. 😁

Exirel:
> Change looks good to me. The related tests look a little weird (especially all that `hasattr`) like they're effectively testing `argparse`, but such are the joys of long maintenance tails I guess.

They are testing we 1. declare an option that will create an attribute of the right name, and then 2. that the value is what was actually expected, with the goal to ensure we don't break our cli's interface.

If a dev remove an argument, or change the name of the destination, that dev will very quickly know that something they might have not tested could be broken. It's a lot of "maybe", I know, yet it's a lot of very cheap maybe so why not.

PR #2411, opened 2023-02-13
py, build: drop `py` plugin from core & require external package

dgw:
### Description
https://github.com/sopel-irc/sopel/blob/4c3e45eec8dec05a0a2f604804da1e8639049b8c/sopel/modules/py.py#L20

This Oblique service URL will stop working at some point. End of support for Python 2.7 on Google App Engine, in January 2024, approaches. After that point we'll be unable to update that deployment. https://github.com/sopel-irc/oblique is already archived, because we're not intending to update it (the package contains _a lot_ more than just the `py` service).

Dropping this plugin from core now will let us make the transition to a new backend easier, without having to fit it into Sopel's release cadence. Marked the requirement with a TODO for 8.1, when it should become an optional manually-installed plugin like the others that have been packageized (ipython, spellcheck, etc.).

While there is a `pylambda` private repo with a potential replacement, it makes more sense to just drop this from the first-party features. Among users active in the #sopel channel, this plugin only seems to come up when someone's bot's users notice it in the command listing and start trying to break it.

### Checklist

### Notes
Plugin moved to https://github.com/sopel-irc/sopel-py


dgw:
Damn it, `pip`, I _want_ prerelease versions to be allowed. Will have to tweak the new package, republish, and rerun checks. 😩

Exirel (approved):
I'll be so happy when we reach Sopel 8.1's release day.

Issue #2413, opened 2023-02-19
Provide `urllib.parse.ParseResult` to plugins with URL triggers


SnoopJ:
### Requested Feature

Sopel should take better advantage of `urllib.parse` when responding to URLs and passing them off to plugins for handling. See discussion in #2412.

What I have in mind is something along the lines of the following, and hopefully this is in the direction that @Exirel and @dgw were thinking.

Suppose the `Trigger` provided to the handler had new behavior available if the handler call was caused by a URL, `Trigger.parse_url() -> urllib.parse.ParseResult`. In this hypothetical, the code in the `wikipedia` plugin that caused us all to start thinking about this might look something like:

```python
@plugin.url(r'https?:\/\/([a-z]+(?:\.m)?\.wikipedia\.org)\/wiki\/((?!File\:)[^ #]+)#?([^ ]*)')
@plugin.output_prefix(PLUGIN_OUTPUT_PREFIX)
def mw_info(bot, trigger):
    """Retrieves and outputs a snippet of the linked page."""                                                                                                         
    url = trigger.parse_url()
    article = url.path.lstrip("/wiki")
    section = url.fragment

    do_the_rest(article, section)
```

But this is something I put together in less than an hour and definitely needs the eyes of the experts on it. In particular, I'm not sure how I feel about the method call, and don't know what's missing from this interface for messages that might contain multiple URLs or other things I'm not accounting for. I opted for a method here because it defers the work until the user actually wants the parsed result, but it's entirely possible that this needs to be implemented in some other manner.

### Problems Solved

https://xkcd.com/1313/

dgw:
You can implement it as a property, still deferring the work until the handler code asks for it (and caching the result once the property is accessed once, if that seems useful enough) but without making it a method call from plugin authors' perspective.

> don't know what's missing from this interface for messages that might contain multiple URLs

Nothing is missing: Each URL is a separate trigger for `@url` callables, like each occurrence of the pattern in a line for `@find` rules.

Thing is, `@url` was intended to have some special behavior before. `@url` decorated callables can have a third parameter called `match` which once upon a time contained only data about the URL match (without the rest of the line). That's been deprecated, and it's now functionally the same as `trigger`.

[...]

PR #2414, opened 2023-02-19
wikipedia: support query strings in Wikipedia URLs

### LINKED ISSUE #2412: wikipedia: query parameters cause URL parse failure
### Description

Wikipedia URLs that contain `?query=params` can confuse Sopel, leading to a `KeyError`:

```
10:31 <+SnoopJ> Sopel, do you beef it on this URL? https://en.wikipedia.org/wiki/Gauss%27s_law?useskin=vector#Differential_form
10:31 <+Sopel> Unexpected KeyError ('parse') from SnoopJ at 2023-02-17 15:31:51.723938. Message was: https://en.wikipedia.org/wiki/Gauss%27s_law?useskin=vector#Differential_form
```

### Reproduction steps

1. Send a Wikipedia URL with an embedded query parameter to a Sopel instance with the `wikipedia` plugin active

### Expected behavior

The same article-retrieval behavior as a Wikipedia URL without query parameters:

```
10:51 <+SnoopJ> Sopel, https://en.wikipedia.org/wiki/Gauss%27s_law#Differential_form
10:51 <+Sopel> [wikipedia] Gauss's law - Differential form | "By the divergence theorem, Gauss's law can alternatively be written in the differential form: ∇ ⋅ E = 
ρ ε 0 ε r {\displaystyle \nabla \cdot \mathbf {E} ={\frac {\rho }{\varepsilon _{0}\varepsilon _{r}}}} where ∇ · E is the divergence of the electric field, ε0 is the vacuum permittivity, ε r {\displaystyle \varepsilon _{r}} is the relative permittivity, and ρ is the volume charge density […]"
```

### Relevant logs

_No response_

### Notes

Exception details from the error log:

```
[2023-02-17 11:16:12,662] sopel.bot            ERROR    - Unexpected error ('parse') from SnoopJ at 2023-02-17 16:16:12.662537. Message was: https://en.wikipedia.org/wiki/Gauss%27s_law?useskin=vector#Differential_form                                                                                                                   
  File "/home/snoopjedi/.sopel/venv/lib/python3.9/site-packages/sopel/bot.py", line 763, in call_rule                                                                 
    rule.execute(sopel, trigger)                                                                                                                                      
  File "/home/snoopjedi/.sopel/venv/lib/python3.9/site-packages/sopel/plugins/rules.py", line 1057, in execute                                                        
  File "/home/snoopjedi/.sopel/venv/lib/python3.9/site-packages/sopel/plugins/rules.py", line 1608, in execute_handler                                                
    return handler(bot, trigger, match=trigger)                                                                                                                       
  File "/home/snoopjedi/.sopel/venv/lib/python3.9/site-packages/sopel/modules/wikipedia.py", line 300, in mw_info                                                     
    say_section(bot, trigger, match.group(1), unquote(match.group(2)), unquote(match.group(3)))                                                                       
  File "/home/snoopjedi/.sopel/venv/lib/python3.9/site-packages/sopel/modules/wikipedia.py", line 243, in say_section                                                 
  File "/home/snoopjedi/.sopel/venv/lib/python3.9/site-packages/sopel/modules/wikipedia.py", line 264, in mw_section                                                  
    for entry in sections['parse']['sections']:                                                                                                                       
KeyError: 'parse'
```

### Sopel version

7.1.9 (also happens on `master`)

### Installation method

`pip install`

### Python version

3.9.9

### Operating system

Ubuntu 20.04

### IRCd

_No response_

### Relevant plugins

wikipedia

SnoopJ:
### Description
This PR fixes #2412 by discarding the query string from the name of the article to be searched.

### Checklist

### DIFF
--- a/sopel/modules/wikipedia.py
+++ b/sopel/modules/wikipedia.py
@@ -10,12 +10,12 @@ from __future__ import annotations
 from html.parser import HTMLParser
 import logging
 import re
+from urllib.parse import quote, unquote, urlparse
 
 from requests import get
 
 from sopel import plugin
 from sopel.config import types
-from sopel.tools.web import quote, unquote
 
 
 LOGGER = logging.getLogger(__name__)
@@ -321,26 +321,27 @@ def mw_image_description(server, image):
     return desc
 
 
-# Matches a wikipedia page (excluding spaces and #, but not /File: links), with a separate optional field for the section
-@plugin.url(r'https?:\/\/([a-z]+(?:\.m)?\.wikipedia\.org)\/wiki\/((?!File\:)[^ #]+)#?([^ ]*)')
+# Matches a Wikipedia link (excluding /File: pages)
+@plugin.url(r'https?:\/\/([a-z]+(?:\.m)?\.wikipedia\.org)\/wiki\/((?!File\:)[^ ]+)')
 @plugin.output_prefix(PLUGIN_OUTPUT_PREFIX)
 def mw_info(bot, trigger, match=None):
     """Retrieves and outputs a snippet of the linked page."""
     server = match.group(1)
-    query = unquote(match.group(2))
-    section = unquote(match.group(3))
+    page_info = urlparse(match.group(2))
+    article = unquote(page_info.path)
+    section = unquote(page_info.fragment)
 
     if section:
         if section.startswith('cite_note-'):  # Don't bother trying to retrieve a snippet when cite-note is linked
-            say_snippet(bot, trigger, server, query, show_url=False)
+            say_snippet(bot, trigger, server, article, show_url=False)
         elif section.startswith('/media'):
             # gh2316: media fragments are usually images; try to get an image description
             image = section[7:]  # strip '/media' prefix in pre-3.9 friendly way
             say_image_description(bot, trigger, server, image)
         else:
-            say_section(bot, trigger, server, query, section)
+            say_section(bot, trigger, server, article, section)
     else:
-        say_snippet(bot, trigger, server, query, show_url=False)
+        say_snippet(bot, trigger, server, article, show_url=False)
 
 
 @plugin.command('wikipedia', 'wp')


sopel/modules/wikipedia.py
```diff
@@ -327,7 +328,13 @@ def mw_image_description(server, image):
 def mw_info(bot, trigger, match=None):
     """Retrieves and outputs a snippet of the linked page."""
     server = match.group(1)
-    query = unquote(match.group(2))
+
+    # NOTE: Discard the query string if there is one, avoids the error in gh-2412 https://github.com/sopel-irc/sopel/issues/2412
+    # TODO: replace this usage when/if gh-2413 is resolved https://github.com/sopel-irc/sopel/issues/2413
```
dgw, line 333: If you put a direct number reference, it's linked in GH code view, btw.

```suggestion
    # NOTE: Discard the query string if there is one; see #2412
    # TODO: replace this usage when/if #2413 is resolved
```

(I also don't know if this TODO is useful, considering how uncertain it is that anything will be implemented based on #2413; and if it is, we'll certainly take a pass through the plugins that are still in core to find uses of `urlparse` or any straggling regex-based "parsing" of links that might still be left by then.)

SnoopJ: Hmm, fair enough, I'll drop it

dgw:
I am foolish for not looking more closely at this earlier. Why not use the parsed URL for the fragment, also? Are you trying to avoid touching too much? If you change the URL pattern to `https?:\/\/([a-z]+(?:\.m)?\.wikipedia\.org)\/wiki\/(?!File\:)[^ ]+` it should be possible to offload fetching all the bits to `urllib.parse` instead of the regex.

(Not a review comment because I'm just trying to understand the intended scope; Exi's comment you referenced was about converting all of the link-matching logic to use `urllib.parse`, stripping the regex down to _only_ what's necessary to make sure what's been given is a Wikipedia article of some kind.)

sopel/modules/wikipedia.py
```diff
@@ -323,18 +323,14 @@ def mw_image_description(server, image):
 
 
 # Matches a wikipedia page (excluding spaces and #, but not /File: links), with a separate optional field for the section
-@plugin.url(r'https?:\/\/([a-z]+(?:\.m)?\.wikipedia\.org)\/wiki\/((?!File\:)[^ #]+)#?([^ ]*)')
+@plugin.url(r'https?:\/\/([a-z]+(?:\.m)?\.wikipedia\.org)\/wiki\/((?!File\:)[^ ]*)')
```
dgw, line 326: Probably don't want to match just `/wiki/`?

```suggestion
@plugin.url(r'https?:\/\/([a-z]+(?:\.m)?\.wikipedia\.org)\/wiki\/((?!File\:)[^ ]+)')
```

sopel/modules/wikipedia.py
```diff
@@ -323,18 +323,14 @@ def mw_image_description(server, image):
 
 
 # Matches a wikipedia page (excluding spaces and #, but not /File: links), with a separate optional field for the section
```
dgw, line 325:
```suggestion
# Matches a Wikipedia link (excluding /File: pages)
```

dgw (commented):
Since you're updating comments anyway...

sopel/modules/wikipedia.py
```diff
@@ -10,6 +10,7 @@
 from html.parser import HTMLParser
 import logging
 import re
+import urllib.parse
```
Exirel, line 13: D doesn't care, and I do too much. 😁 

```suggestion
from urllib.parse import urlparse
```

sopel/modules/wikipedia.py
```diff
@@ -321,26 +322,27 @@ def mw_image_description(server, image):
...
-@plugin.url(r'https?:\/\/([a-z]+(?:\.m)?\.wikipedia\.org)\/wiki\/((?!File\:)[^ #]+)#?([^ ]*)')
+# Matches a Wikipedia link (excluding /File: pages)
+@plugin.url(r'https?:\/\/([a-z]+(?:\.m)?\.wikipedia\.org)\/wiki\/((?!File\:)[^ ]+)')
 @plugin.output_prefix(PLUGIN_OUTPUT_PREFIX)
 def mw_info(bot, trigger, match=None):
     """Retrieves and outputs a snippet of the linked page."""
     server = match.group(1)
-    query = unquote(match.group(2))
-    section = unquote(match.group(3))
+    page_info = urllib.parse.urlparse(match.group(2))
```
Exirel, line 331: To go with the other change.

```suggestion
    page_info = urlparse(match.group(2))
```

sopel/modules/wikipedia.py
Exirel: Or, even:

```suggestion
from urllib.parse import urlparse, unquote
```

And then remove the import `from sopel.tools.web import unquote` which is really a useless tool.

Exirel (changes requested):
I think it's the right moment to push a little further the cleanup of the plugin and remove the use of `sopel.tools.web.unquote` when it's literally just a shortcut for the stdlib `urllib.parse.unquote`.

Edit: actually, same for `sopel.tools.web.quote`. Both are useless now.

Exirel (approved):

Exirel:
Alright, all that is left is a squash, since @dgw approved and I checked the changes. I don't think there is enough to require more than one commit, and I don't mind if you lose the co-author from my suggestions (I don't assume you know, but you might, all that to say it's up to you).

dgw (approved):
I would have put changing the existing imports from `sopel.tools.web` as out of scope for this, but it doesn't ultimately matter. The `quote` and `unquote` wrappers there have comments that we should consider deprecating them, anyway.

PR #2416, opened 2023-02-28
reload: remove `.update` command

dgw:
### Description
Sopel's not designed to run directly from a Git working directory any more, so let's not advertise this command that relies upon it.

### Checklist
  - It was really easy to test, because this only removes stuff.


dgw:
_(How dumb do you have to be to click "close" instead of "merge"? At least as dumb as me, I guess.)_

Issue #2255, opened 2022-02-13
Logging to IRC creates errors after QUIT


dgw:
### Description
The `IrcLoggingHandler` trips its `handleError()` method with log output immediately after sending a QUIT to the server.

### Reproduction steps
1. Enable outputting logs to an IRC channel with `logging_channel = #somechannel`
2. Set Sopel to flood its IRC logging handler with `logging_channel_level = DEBUG`
3. Start the bot
4. Quit the bot with e.g. <kbd>Ctrl</kbd>+<kbd>C</kbd>

The WARNING ("sending QUIT to server") and replying ERROR ("ERROR received from server: Closing Link") never make it to the IRC channel. The subsequent INFO ("Closing connection") generates an exception in the logger itself (`BrokenPipeError: [Errno32] Broken pipe`). Full logs below.

### Expected behavior
Ideally the WARNING about sending a QUIT command should appear on IRC before the actual QUIT is sent/processed, but that's not the deal-breaker here. Neither is the ERROR not making it there; obviously, the bot has already quit (the server is acknowledging it), so the logger can't send anything more to IRC.

What bugs me is the `BrokenPipeError`. I expect the `IrcLoggingHandler` to check and realize that it doesn't have a connection, and then not even try to send anything to IRC after the QUIT is processed.

### Logs

<details>
<summary>Long traceback</summary>

```
^C[2022-02-12 17:30:13,217] sopel.bot            WARNING  - Got quit signal, sending QUIT to server.
[2022-02-12 17:30:13,248] sopel.irc            ERROR    - ERROR received from server: Closing Link: 68-nn-nn-nn.fttp.usinternet.com ()
[2022-02-12 17:30:13,249] sopel.irc.backends   INFO     - Closing connection
--- Logging error ---
Traceback (most recent call last):
  File "/mnt/c/Users/dgw/github/sopel/sopel/logger.py", line 37, in emit
    self._bot.say(msg, self._channel)
  File "/mnt/c/Users/dgw/github/sopel/sopel/irc/__init__.py", line 760, in say
    self.backend.send_privmsg(recipient, text)
  File "/mnt/c/Users/dgw/github/sopel/sopel/irc/abstract_backends.py", line 201, in send_privmsg
    self.send_command('PRIVMSG', safe(dest), text=text)
  File "/mnt/c/Users/dgw/github/sopel/sopel/irc/abstract_backends.py", line 69, in send_command
    self.irc_send(raw_command.encode('utf-8'))
[...]

PR #2419, opened 2023-03-13
logging: disallow `DEBUG` channel logs & stop inheriting console log level

dgw:
### Description
Setting `logging_channel_level = DEBUG` sends so much to IRC that the bot got stuck waiting for flood protection to release.

The only sensible solution is to disallow setting channel logs to the `DEBUG` level, and along the way I've also made it so that channel logging always defaults to `WARNING` if not configured, instead of inheriting the console's `logging_level` setting.

### Checklist

### Notes
Indirectly related to #2255, both because I discovered that bug while working on this one last year _and_ because I devised this solution while testing to see if that issue could be closed after #2410.

### DIFF
--- a/docs/source/configuration.rst
+++ b/docs/source/configuration.rst
@@ -825,6 +825,12 @@ Example of configuration to log errors only in the ``##bot_logs`` channel::
     logging_channel_level = ERROR
     logging_channel_format = %(message)s
 
+.. note::
+
+   ``DEBUG`` is not supported for ``logging_channel_level``. The bot would
+   either flood itself off the network or spend most of its time waiting for
+   flood protection delays (thus becoming unusable).
+
 .. versionadded:: 7.0
 
    Configuration options ``logging_channel_level``, ``logging_channel_format``
--- a/docs/source/configuration.rst
+++ b/docs/source/configuration.rst
@@ -807,9 +807,10 @@ Log to a Channel
 ----------------
 
 It is possible to send logs to an IRC channel, by configuring
-:attr:`~CoreSection.logging_channel`. By default, it uses the same log level,
-format, and date-format parameters as console logs. This can be overridden
-with these settings:
+:attr:`~CoreSection.logging_channel`. By default, it logs ``WARNING`` or higher
+messages, using the same date and log-line format parameters as the console
+logging output. The following settings can be used to customize output to the
+logging channel:
 
 * ``format`` with :attr:`~CoreSection.logging_channel_format`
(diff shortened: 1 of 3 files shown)


Exirel (approved):

dgw:
> we could even prevent people from using the INFO level

I didn't do that because "Channel joined:" log lines are emitted as INFO, and that seems like a useful thing to have in a log channel. [Bucket](https://github.com/zigdon/xkcd-Bucket), for example, also emits messages to its _control channel_ when joining/parting a channel.

Regarding reporting channel joins/parts to IRC, we can treat that as a separate feature to be considered for 8.1. [Bucket's join/part reporting](https://github.com/zigdon/xkcd-Bucket/blob/47698fd0e92d1011f2472779be855c349d736b90/bucket.pl#L635) is targeted at admin-requested actions, and we could extend our `admin` plugin with that functionality independent of channel logs.

About INFO logs: I think we still have work to do regarding which log level to use for various actions. As an example, see [the logging for a kicked user](https://github.com/sopel-irc/sopel/blob/4efc48ed48fc29611bfe8d9fc28edb15816ecb56/sopel/coretasks.py#L758), which is currently INFO but probably more appropriate for DEBUG.

Those two points aside, I _think_ allowing INFO but defaulting to WARNING is appropriate for now. WARNING is low-volume enough that it should be fine on just about any bot, and it is safe for owners of low-to-medium traffic Sopel instances to drop it down to INFO if they want to see more of what's going on for a specific reason. I removed DEBUG because it breaks Sopel immediately upon joining IRC, even if it's only in one channel and no events have happened since startup—i.e. there's _absolutely no chance_ that it would be useful.

Does all of this reasoning make sense? If I've missed considering something, please do bug me about it before I merge!

Exirel:
> Does all of this reasoning make sense?

Yes! With a small caveat: this applies if we keep sending technical logs to channels. I'd argue that a better feature (not to be done in this PR of course, and probably a new feature for 8.1/9.x) would be to completely separate these two types of logs:

* technical logs, i.e. the code did something for technical reason
* business logs, i.e. the code did something to support a feature (such as JOIN, INVITE)

The first one should probably not go to the channel logger unless it's a WARNING or an ERROR, and the second type could go as low as INFO without much issue. After all, we already separate application logs from raw IRC logs, so that would make sense to me to do as well for other types of logs.

dgw:
> a better feature (not to be done in this PR of course, and probably a new feature for 8.1/9.x)

Definitely. We need to think about how it should work. 😅

This should go on a tickler list. If there's no issue about it by merge time, I'll bang out a quick one.

PR #2420, opened 2023-03-13
safety: logging level tweaks

dgw:
### Description
* failure to delete old malwaredomains file is upgraded to a warning: fixing the problem requires manual intervention
* domain list update is now all debug, all the time, except if there is an error (which generates a warning so the bot owner can see if it's repeatedly failed)
* cache cleanup being skipped is also debug now; it didn't make sense for "cleanup already running" to be louder than "starting cleanup"

### Checklist

### Notes
I'm looking specifically at `info()` logging calls in light of #2419 and very well might run through all the places where the most common messages I see in logs of the bots I run/admin just to double-check that we're using the INFO level sensibly.

### DIFF
--- a/sopel/modules/safety.py
+++ b/sopel/modules/safety.py
@@ -120,6 +120,7 @@ def setup(bot: Sopel):
             # Python on Windows throws an exception if the file is in use
             LOGGER.warning('Could not delete %s: %s', old_file, str(err))
 
+    LOGGER.info('Ensuring unsafe domain list is up-to-date (safety plugin setup)')
     update_local_cache(bot, init=True)
 
 
--- a/sopel/modules/safety.py
+++ b/sopel/modules/safety.py
@@ -118,7 +118,7 @@ def setup(bot: Sopel):
         except Exception as err:
             # for lack of a more specific error type...
             # Python on Windows throws an exception if the file is in use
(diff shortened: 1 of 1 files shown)


sopel/modules/safety.py
```diff
@@ -141,12 +141,12 @@ def download_domain_list(bot: Sopel, path: str) -> bool:
     url = bot.settings.safety.domain_blocklist_url
     if url is None or not url.startswith("http"):
         url = "https://raw.githubusercontent.com/StevenBlack/hosts/master/hosts"
-    LOGGER.info("Downloading unsafe domain list from %s", url)
+    LOGGER.debug("Downloading unsafe domain list from %s", url)
```
Exirel, line 144: I'd keep that one as info, as it is a useful information at startup. That, or I'd add an INFO log in the setup hook function.

Exirel (changes requested):
I approve most of these change, I have one small request that I think make sense, with an alternate solution if needed.

sopel/modules/safety.py
dgw: I chose to add this in the setup function, leaving this more-detailed one for debug only.

Exirel (approved):

Issue #2423, opened 2023-03-17
Idea: Lazy channel-restriction decorator


dgw:
### Requested Feature

New decorator (tentative name: `@plugin.allowed_channels`) that takes one or more functions as arguments which will be used at runtime to retrieve the list of channels, similar to how `@plugin.rule_lazy` and friends work.

The decorator's purpose is to allow restricting in which channels certain commands can be used, without using "a bunch of ifs" (as went the tongue-in-cheek complaint in #2082).

### Problems Solved

It is relatively common that bot owners may want to add "spammy" games or other content to their bots that are not appropriate for general use in all channels where the bot is joined. While creating a second config just for such games is an option, in practice bot owners tend to make bespoke copies of plugin code instead with the channel restrictions hard-coded.

In #2082, it was requested to add an optional argument to `@plugin.require_chanmsg` that would allow restricting the decorated callable to a specific channel(s) named at import time. This isn't an ideal solution, because it still requires hard-coding allowed channels into the plugin file.

However, decorators like `@plugin.rule_lazy` and its friends _are_ configurable at runtime via their reliance on helper functions. In the `rule_lazy`-and-friends case, such helper functions are given access to the bot's `settings` so they can generate a list of text patterns to match based on the bot's config. This approach can be applied to generating a list of allowed channels, too.

This solves the problem both the _de facto_ and previously-suggested approaches have regarding configurability.

### Alternatives

The main alternative is a custom decorator as detailed in the comments of #2082, or the aforementioned "bunch of ifs" with hard-coded channel names. An upcoming pull request will propose documentation of the custom decorator approach as part of our Advanced Tips & Tricks page for plugin devs.

### Notes

A sister decorator called `@plugin.restricted_channels` that performs the opposite function would be logical, but that _does_ fall squarely under the existing per-channel command blocking feature, especially once it becomes easier to configure, so it would (IMO) have a shorter useful lifespan as part of Sopel's plugin API.

**However, I also think this kind of thing is ultimately best done by _the bot_ rather than _the plugin_. I'm opening this issue to gather Proper Feedback™ on what could be an alternative to a previously suggested idea.** (#2082)

It's obviously _more useful to more bot owners_ if they have the freedom to allow/block channels at will for any command (or entire plugin) regardless of whether a given plugin's developer has implemented this decorator. Therefore, **even if the only result of this issue is its closure as "not worth it", followed by renewed discussion on #2358, I'll still be happy.**

Exirel:
[...]

Issue #2082, opened 2021-06-01
Expand @plugin.require_chanmsg to specify channels


xnaas:
### The problem
Limiting commands to specific channels requires some or many if statements such as:

```py
if trigger.sender == "#nsfw":
```

### The solution
Chatlog:
```
<cottongin> <xnaas> Is there a better way to  have an entire plugin confined to one channel than putting an if in every command? [...] Hmm...I guess I see how it can be done with dgw's duel.py, for example https://github.com/xnaas/sopel-duel/blob/b864c7f91d862d6e035dd759bb34228482de4804/duel.py#L20-L31
14:11:57 <cottongin> looking real briefly at that, that'd maybe be a nice improvement for 8.0, add in an optional argument to `@plugin.require_chanmsg(chans=[])`
<cottongin> idk how difficult that'd be to implement
* xnaas salivates
<xnaas> 🤤
<+dgw> Probably not very hard
<+dgw> Make a feature request I guess
* xnaas opens a feature request issue?
<xnaas> 👏
<cottongin> 👍
<cottongin> it makes sense to me
```

Particular solution proposed by @cottongin is to have something like `@plugin.require_chanmsg(chans=[])` where you could specify channels a command is limited to.

### Alternatives
`if` statements for days.

### Notes
Sopel is da best. :)

xnaas:
@Exirel said "no, get good." 🤪

cottongin:
To elaborate (via `IRC`):

```
15:22:49 <+Exirel> Not going to expand require_chanmsg, but I know how to deal with that.
15:23:08 <+Exirel> It's just Python, btw.
...
18:02:26 <+Exirel> Haven't tested it, but you get the gist of it. https://www.irccloud.com/pastebin/O9zQUoiu/snippet_require_channel_from_settings.py
18:02:47 <+Exirel> cottongin: the problem is: how do you dynamically define the list of channels to allow?
18:03:00 <+Exirel> You can't do that at plugin import time, you have to do it at runtime only.
...
18:04:04 <+Exirel> But yeah, just as mal said: you can write your own decorator. I'd suggest to call it first (i.e. the closer to your function).
...
18:06:26 <+Exirel> But yeah, there is a rework that I've in mind, that will replace the `require_chanmsg` by something that simply add a function to be called at runtime.
18:06:36 <cottongin> 👍
18:06:39 <+Exirel> Adding as many function as one wish into a list, and the bot can call it.
[...]

PR #2424, opened 2023-03-17
docs: add channel restriction decorator examples to Advanced Tips page

dgw:
### Description
This documents the way to restrict commands to certain channels, _with_ the option to configure the list of allowed channels using the bot's config file instead of relying on hard-coded names, _without_ using "a bunch of ifs".

Follow-up to #2082 that doesn't require adding a technically limited, forever-half-baked API feature.

### Checklist

### Feedback requests
First, the examples using a setting currently fail closed: if the `allowed_channels` list is empty, the decorated callable is not allowed to be used in _any_ channel. Should this be changed to allow plugins that adopt such a decorator directly from our examples to be used _anywhere, unless_ at least one channel is listed?

Second, does this need more variations to show the possibilities? I got tired of writing the examples for today so I stopped at two ("pull from bot.settings" and "use arg if given, pull from bot.settings if not"), but _at least_ the following cases have no demonstration:

- "use arg only"
- "use arg, but take an iterable instead of a string"
- "use iterable arg if given, pull from bot.settings if not"

I was going to leave those as an exercise to the reader because we really want to encourage using config rather than code for things like this, but am open to arguments in favor of demonstrating anyway.

### Future
I considered that we could make a "lazy" decorator like those for rule/find/search/url that accepts a function to call, but it is _far_ too late in the cycle for 8.0 for that. See #2423 for a much longer discussion of that idea.


Exirel:
I think I'd describe only the "Simple channel restriction decorator" version, and add a note "from this, implementing more complicated rules is left as an exercise to the reader".

dgw:
Think I got this to a good spot, with that feedback. Thanks @Exirel.

docs/source/plugin/advanced.rst
```diff
@@ -15,6 +15,59 @@ If something is not in here, feel free to ask about it on our IRC channel, or
...
+.. important::
+
+    When using this example in your own plugin code, remember to change
+    ``myplugin`` to a section name appropriate for your plugin. It is also a
+    good idea to rename the ``MyPluginSection`` class accordingly.
+
+.. note::
+
+    This kind of decorator could be written in many different ways; this
+    example is just the simplest. Implementation of more complex approaches
```
Exirel, line 67: I'm always reluctant to use words like "simplest" or "easiest". In that case, I'd say "is the most common usage". This would reflect what people ask on IRC and what we tell them to do, so that count as "most common" IMHO.

Exirel (changes requested):

docs/source/plugin/advanced.rst
dgw: Pushed some more edits, including to fix the "old habits die hard" above w/`bot.config`.

Exirel (approved):
🚢 📜 🖊️

docs/source/plugin/advanced.rst
```diff
@@ -15,6 +15,60 @@ If something is not in here, feel free to ask about it on our IRC channel, or
...
+Restricting commands to certain channels
+========================================
+
+Allowing games, for example, to be run only in specific channels is a
+relatively common request, but a difficult feature to support directly in
+Sopel's plugin API. Fortunately it is fairly trivial to build a custom
+decorator function that handles this in a configurable way.
+
+Here is a sample plugin that defines such a custom decorator, along with the
+scaffolding needed to pull a list of channels from the bot's settings::
```
SnoopJ, line 27: Nit: a nudge to the anaomy document would be helpful for a reader who may not be familiar but ends up finding this example because they went looking for it

```suggestion
scaffolding needed to pull a list of channels from the :ref:`bot's settings <plugin-anatomy-config>`:
```

SnoopJ (approved):
I left a nit, but LGTM anyway

docs/source/plugin/advanced.rst
Exirel: This will probably need to be split in two lines with that suggestion. :grin:

dgw: Yeah.

```suggestion
scaffolding needed to pull a list of channels from the bot's
:ref:`settings <plugin-anatomy-config>`::
```

(I _think_ the `::` is required still, but I need to test-render in Sphinx before pushing a correction commit.)

dgw: Have tested, played with wording/which parts of the sentence to link, and arrived at a happy place.

Issue #2195, opened 2021-10-18
isupport: hexadecimal escape sequences


dgw:
Recent discussion (ircv3/ircv3-ideas#81, ircdocs/modern-irc#137) led me to https://datatracker.ietf.org/doc/html/draft-brocklesby-irc-isupport-03#page-5 — which says that escapes like `\x20` should be valid in ISUPPORT tokens. Sopel might want to support these.

Added test case for `test/irc/test_irc_isupport.py::test_parse_parameter` in parameter list `VALID_PARSE_VALUE` would be:
`('NETWORK=Libera\\x20Chat', ('NETWORK', 'Libera Chat')),` (fails on current master)

PR #2429, opened 2023-03-22
irc: unescape hexadecimal sequences in ISUPPORT parameter values

SnoopJ:
### Description

This changeset handles hexadecimal escape sequences in `ISUPPORT` parameter values, implementing #2195

### Checklist

### DIFF
--- a/sopel/irc/isupport.py
+++ b/sopel/irc/isupport.py
@@ -131,10 +131,37 @@ ISUPPORT_PARSERS = {
 }
 
 
+def _unescape_param(param):
+    """Handle escape sequences in ISUPPORT parameter values.
+
+    Sopel follows the recommendation of `modern.ircdocs.horse`__ which only
+    recognizes the escape sequences ``\\x20, \\x5C, \\x3D``. All other such
+    escape sequences will be passed through unaltered.
+
+    .. __: https://modern.ircdocs.horse/#rplisupport-005
+    """
+    HEX_PATTERN = r"\\x([0-9a-fA-F]{2})"
+
+    def _unescape(match):
+        num = match.group(1).upper()
+        if num == "20":
+            result = " "
+        elif num == "5C":
+            result = "\\"
+        elif num == "3D":
+            result = "="
+        else:
+            result = match.group(0)
+
+        return result
+
+    return re.sub(HEX_PATTERN, _unescape, param)
+
+
 def parse_parameter(arg):
     items = arg.split('=', 1)
     if len(items) == 2:
-        key, value = items
+        key, value = items[0], _unescape_param(items[1])
     else:
         key, value = items[0], None
 
--- a/test/irc/test_irc_isupport.py
+++ b/test/irc/test_irc_isupport.py
@@ -271,6 +271,10 @@ VALID_PARSE_VALUE = (
     ('STATUSMSG=ABCD', ('STATUSMSG', ('A', 'B', 'C', 'D'))),
     ('TOPICLEN=5', ('TOPICLEN', 5)),
     ('USERLEN=5', ('USERLEN', 5)),
+    ('NETWORK=Libera\\x20Chat', ('NETWORK', 'Libera Chat')),
+    ('NETWORK=Libera\\x5CChat', ('NETWORK', 'Libera\\Chat')),
+    ('NETWORK=Libera\\x3DChat', ('NETWORK', 'Libera=Chat')),
+    ('NETWORK=Libera\\x0aChat', ('NETWORK', 'Libera\\x0aChat')),
 )
 
 


SnoopJ:
The [specification](https://datatracker.ietf.org/doc/html/draft-brocklesby-irc-isupport-03#page-5) says also that:

>    If the server has not advertised a CHARSET parameter, it MUST not use
>   such sequences with a value outside those permitted by the above ABNF
>   grammar, with the exception of "\x20";

which I believe is not currently checked by Sopel. Not sure if we want to worry about that, but if so, it should probably be part of this changeset.

dgw:
Refreshing to have such a small patch to review after _you know what_ 😅 (and thank you @SnoopJ for helping with that one).

> The [specification](https://datatracker.ietf.org/doc/html/draft-brocklesby-irc-isupport-03#page-5) says also that:
> 
> > If the server has not advertised a CHARSET parameter, it MUST not use
> > such sequences with a value outside those permitted by the above ABNF
> > grammar, with the exception of "\x20";
> 
> which I believe is not currently checked by Sopel. Not sure if we want to worry about that, but if so, it should probably be part of this changeset.

I wouldn't worry about servers violating the MUST; we can "[be liberal in what you accept](https://en.wikipedia.org/wiki/Robustness_principle)" and decode the escapes anyway, assuming UTF-8 in the absence of a specified CHARSET.

You could add a test case where something that looks like an escape but isn't valid (what you called "pathological counter-examples" on IRC) gets handled. More importantly, the following behavior should be validated in tests as well:

> Characters in multibyte encodings such as UTF-8 should be
> sent as a series of \x sequences.

If we're gonna do this, it had better not _mojibake_ the encoded values 😉

SnoopJ:
After discussion with @Exirel, it's come to light that the [Modern IRC](https://modern.ircdocs.horse/#rplisupport-005) specification and the [IETF specification](https://datatracker.ietf.org/doc/html/draft-brocklesby-irc-isupport-03) have slightly different opinions about escape sequences, with Modern IRC being more strict, limiting escapes to only `\x20, \x5C, \x3D`.

I've updated this PR to follow the Modern IRC perspective, which is not only simpler to implement, but avoids tangling with the complication of the antiquated/removed `CHARSET` field, and the ambiguity of bytes sent in a multi-byte encoding.

Exirel:
Note from IRC: we have a different understanding of the spec, and we agree that it is a bit confusing. Following these two things led me to believe that we don't need to un-escape everything and only a subset of characters:

* Latest version of the escaping function in [inspircd](https://github.com/inspircd/inspircd/blob/2b810b412aeee2d9c79d97fa54a80e317981a6d8/src/server.cpp)
* [Modern IRC event 005 specification](https://modern.ircdocs.horse/#rplisupport-005)

As a result, we discussed it with @SnoopJ and we came to the conclusion that we are going to un-escape only the few bytes that need to.

Edit: @SnoopJ beat me to it!

sopel/irc/isupport.py
```diff
@@ -131,10 +131,38 @@ def _parse_prefix(value):
 }
 
 
+def _hex_unescape(param):
+    """
+    Unescape hexadecimal escape sequences (i.e. ``'\\xHH'``)
```
Exirel, line 136:
```suggestion
    """Unescape hexadecimal escape sequences (i.e. ``'\\xHH'``)
```

sopel/irc/isupport.py
```diff
@@ -131,10 +131,38 @@ def _parse_prefix(value):
...
+    def _unescape(match):
+        num = match.group(1).upper()
+        if num == "20":
+            result = " "
+        elif num == "5C":
+            result = "\\"
+        elif num == "3D":
+            result = "="
+        else:
+            raise ValueError("Encountered invalid escape \\x{num}".format(num=num))
```
Exirel, line 155: This shouldn't raise an exception: for instance, [inspircd will happily escape](https://github.com/inspircd/inspircd/blob/2b810b412aeee2d9c79d97fa54a80e317981a6d8/src/server.cpp#L181-L200) `\n` and replace it with `"\x0a"`, which is going to raise an exception here.

As a result, I suggest to do the following:

```suggestion
            result = match.group(0)
```

test/irc/test_irc_isupport.py
```diff
@@ -281,6 +284,11 @@ def test_parse_parameter(arg, expected):
     assert (key, value) == expected
 
 
+def test_parse_parameter_invalid_escape():
+    with pytest.raises(ValueError):
+        isupport.parse_parameter('NETWORK=\\xc5\\x81ibera\\x20Chat')
+
+
```
Exirel, line 291: As a result you will need to remove that:

```suggestion
```

test/irc/test_irc_isupport.py
```diff
@@ -271,6 +271,9 @@ def test_isupport_targmax_undefined():
     ('STATUSMSG=ABCD', ('STATUSMSG', ('A', 'B', 'C', 'D'))),
     ('TOPICLEN=5', ('TOPICLEN', 5)),
     ('USERLEN=5', ('USERLEN', 5)),
+    ('NETWORK=Libera\\x20Chat', ('NETWORK', 'Libera Chat')),
+    ('NETWORK=Libera\\x5CChat', ('NETWORK', 'Libera\\Chat')),
+    ('NETWORK=Libera\\x3DChat', ('NETWORK', 'Libera=Chat')),
```
Exirel, line 276:
```suggestion
    ('NETWORK=Libera\\x3DChat', ('NETWORK', 'Libera=Chat')),
    ('NETWORK=Libera\\x0aChat', ('NETWORK', 'Libera\\x0aChat')),
```

Exirel (changes requested):

Exirel (approved):

sopel/irc/isupport.py
```diff
@@ -131,10 +131,37 @@ def _parse_prefix(value):
 }
 
 
+def _hex_unescape(param):
+    """Unescape hexadecimal escape sequences (i.e. ``'\\xHH'``)
```
RustyBower, line 135: Change this to e.g. not i.e.

https://www.grammarly.com/blog/know-your-latin-i-e-vs-e-g/

RustyBower (changes requested):

sopel/irc/isupport.py
half-duplex: I think "i.e." is appropriate, as it only escapes things in the form `\xHH`

RustyBower: then the comment before should be more specific. "escape sequences" is too general

half-duplex: Sure, but "hexadecimal escape sequences" as written is sufficiently specific imo :p

SnoopJ: The grammar here matches my intent, but the name/doc can be a little more explicit, change made in fa45f08

RustyBower (approved):

sopel/irc/isupport.py
```diff
@@ -131,10 +131,38 @@ def _parse_prefix(value):
 }
 
 
+def _unescape_param(param):
+    """
+    Handle escape sequences in ISUPPORT parameter values
```
dgw, line 136:
```suggestion
    """Handle escape sequences in ISUPPORT parameter values.
```

dgw (commented):
Everyone else already did the good nitpicks, so this was all that was left for me to find. 🤣

SnoopJ:
Squashed and rebased with a summary in the commit message. Another ~one~8.0 bites the dust?

dgw (approved):
:shipit: :shipit: :shipit: :rocket:

PR #2433, opened 2023-03-25
url: fix premature loop termination, docstring cleanup & slight refactoring

### LINKED ISSUE #2230: url: documented as calling other plugins when redirects match, but doesn't
Once upon a time, the `url` plugin's `title_auto` feature would check if a link _redirected to_ a URL that should be handled by another plugin, but that appears to no longer be the case since [the conversion to use `requests`](https://github.com/sopel-irc/sopel/commit/ccda12e3f5440ee87edc2689faccca1f3f027f4d). Its documentation and that for the `process_urls` helper (below) still say this is the case, but the behavior is gone.

https://github.com/sopel-irc/sopel/blob/d36a19dad548068e34357d63761258cdd5e5efc0/sopel/modules/url.py#L290-L295
https://github.com/sopel-irc/sopel/blob/d36a19dad548068e34357d63761258cdd5e5efc0/sopel/modules/url.py#L322-L329

I would argue that this is a useful feature, and we should bring it back. For example, our first-party GitHub plugin could get free handling of repo/issue/PR/commit links shortened through git.io if this was still the case. _Any_ shortened URL at present, actually, is only ever going to get the generic title-fetching behavior.

While it's certainly possible to create Yet Another Sopel Plugin that resolves short URLs, it would be fairly annoying to implement—and it wouldn't block the built-in title handling, either.

_Note, I don't care if this is part of `url`, but it makes more sense to put it there than into the bot's core URL handling. We don't necessarily want the core bot making HTTP requests every time it sees a link. Some of the thoughts in this issue were previously expressed in #sopel@libera on 2021-11-30._

SnoopJ:
### Description

This changeset resolves #2230 by dropping the incorrect information from the relevant docstrings, and also does some light refactoring of `title_auto(), process_urls()` for the sake of future maintenance. Also included is a fix for a bug discovered while writing this PR, where a URL matching a registered callback causes following URLs in the message to be ignored.

### Checklist
  - `1426 passed, 8 xfailed, 1 warning in 52.07s`

### DIFF
--- a/sopel/modules/url.py
+++ b/sopel/modules/url.py
@@ -13,7 +13,7 @@ from __future__ import annotations
 from ipaddress import ip_address
 import logging
 import re
-from typing import Generator, List, Optional, Tuple, TYPE_CHECKING
+from typing import Generator, List, NamedTuple, Optional, TYPE_CHECKING
 from urllib.parse import urlparse
 
 import dns.resolver
@@ -289,10 +289,10 @@ def title_command(bot: SopelWrapper, trigger: Trigger):
         # needs to be a list so len() can be checked later
         urls = list(web.search_urls(trigger))
 
-    for url, title, domain, tinyurl, dispatched in process_urls(
+    for url, title, domain, tinyurl, ignored in process_urls(
         bot, trigger, urls, requested=True
     ):
-        if dispatched:
+        if ignored:
             result_count += 1
             continue
         message = "%s | %s" % (title, domain)
@@ -317,8 +317,13 @@ def title_command(bot: SopelWrapper, trigger: Trigger):
 def title_auto(bot: SopelWrapper, trigger: Trigger):
     """
     Automatically show titles for URLs. For shortened URLs/redirects, find
-    where the URL redirects to and show the title for that (or call a function
-    from another plugin to give more information).
+    where the URL redirects to and show the title for that.
+
+    .. note::
+
+        URLs that match (before redirection) any other registered callbacks
+        will *not* have their titles shown.
+
     """
     # Enabled or disabled by feature flag
     if not bot.settings.url.enable_auto_title:
@@ -342,8 +347,8 @@ def title_auto(bot: SopelWrapper, trigger: Trigger):
             continue
         urls.append(url)
 
-    for url, title, domain, tinyurl, dispatched in process_urls(bot, trigger, urls):
-        if not dispatched:
+    for url, title, domain, tinyurl, ignored in process_urls(bot, trigger, urls):
+        if not ignored:
             message = '%s | %s' % (title, domain)
             if tinyurl:
                 message += ' ( %s )' % tinyurl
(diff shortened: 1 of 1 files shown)


Exirel (changes requested):
Small nitpick here and there.

sopel/modules/url.py
```diff
@@ -353,16 +356,20 @@ def title_auto(bot: SopelWrapper, trigger: Trigger):
...
     bot: SopelWrapper,
     trigger: Trigger,
     urls: List[str],
     requested: bool = False,
-) -> Generator[
-    Tuple[str, Optional[str], Optional[str], Optional[str], bool],
-    None,
-    None,
-]:
+) -> Iterable[UrlInfo]:
```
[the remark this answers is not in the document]
SnoopJ: That's reasonable. I usually prefer `Iterable[T]` over `Generator[T, None, None]` in cases where the sent/return value isn't being used, but my legibility preference is a weak one and I'm happy to revert here.

sopel/modules/url.py
```diff
@@ -353,16 +357,30 @@ def title_auto(bot: SopelWrapper, trigger: Trigger):
         bot.memory["last_seen_url"][trigger.sender] = url
 
 
+class UrlInfo(NamedTuple):
```
dgw, line 360: Do we have a compelling reason not to call this `URLInfo`? We already have e.g. `sopel.plugins.rules.URLCallback` as precedent for not forcing strict CamelCase on acronyms/initialisms.

dgw (changes requested):
~~Some~~ Mostly nitpicks. When you've addressed these, go ahead and re-request review from both me and Exi.

sopel/modules/url.py
Exirel: Oh yeah, `URLInfo` would be better.

sopel/modules/url.py
```diff
@@ -353,16 +358,30 @@ def title_auto(bot: SopelWrapper, trigger: Trigger):
...
+    url: str
+
+    title: Optional[str]
+    """The title associated with ``url``, if appropriate."""
+
+    hostname: Optional[str]
+    """The hostname associated with ``url``, if appropriate."""
+
+    tinyurl: Optional[str]
+    """A shortened form of ``url``, if appropriate."""
```
dgw, line 373: This is just following the rest of the code's name for the returned value, and _at present_ it is still an actual `tinyurl` from the TinyURL service. I just wonder how long that's going to last, since the API endpoint this plugin uses appears not to be advertised anywhere on tinyurl.com (that I could find).

But if the TinyURL backend stops working, we can rename this and every other related variable when that time comes. No need to bother with it now.

dgw (approved):
Just a side note, "doom and gloom" for the future of one `url` plugin subfeature that I doubt gets much use (but we don't collect usage metrics, so who knows?)

Exirel (approved):
I haven't tested this myself, however it looks sane to me.

SnoopJ:
Re: resting, I showed off on IRC that with this patch, `[url]` does not swallow URLs that come after one handled by some other plugin:

```
13:06 <+SnoopJ> https://github.com/sopel-irc/sopel/ https://arxiv.org/abs/2303.12132
13:06 <+testibot> [url] [2303.12132] Fundamentals of Generative Large Language Models and Perspectives in Cyber-Defense | arxiv.org
13:06 <+testibot> [GitHub] sopel-irc/sopel - :robot::speech_balloon: An easy-to-use and highly extensible IRC Bot framework. Formerly Willie. | 99.5% Python 0.4% Shell 0.1% Makefile | Last Push: 2023-03-28 - 15:09:05UTC | Stargazers: 924 | Watchers: 65 | Forks: 414 | Network: 414 | Open Issues: 95
13:06 <+Sopel> [GitHub] sopel-irc/sopel - :robot::speech_balloon: An easy-to-use and highly extensible IRC Bot framework. Formerly Willie. | 99.5% Python 0.4% Shell 0.1% Makefile | Last Push: 2023-03-28 - 15:09:05UTC | Stargazers: 924 | Watchers: 65 | Forks: 414 | Network: 414 | Open Issues: 95
```

PR #2434, opened 2023-03-26
Add more placeholders to rate-limit messages, move template interpolation from `Rule` to `Sopel`

### LINKED ISSUE #2325: More placeholders in rate-limit messages
### Requested Feature

We discussed a bit in #2290 what other things might be useful to have as placeholders in the optional message sent when a command is blocked by rate limiting. Some ideas from that, plus a bit of reflection just now:

- `{time_left}` — a `tools.time.seconds_to_human()` of the time left until the rate limit expires
- `{seconds_left}` — the raw version of `{time_left}`
- `{limit}` — what the actual rate limit is, probably in `seconds_to_human()` format
- `{limit_type}` — especially for the default message, useful to differentiate why a command was blocked
- `{command}` — the name of the command that was rate-limited
  - Potentially tricky because [almost any kind of callable](https://github.com/sopel-irc/sopel/blob/3a5ba8ec47c97e4de247473cec4c1badb546ae48/sopel/loader.py#L114-L128) can be rate-limited.
- `{plugin}` — the name of the plugin in which the limited feature lives

Names subject to change, as always, if better ideas come up.

### Problems Solved

The time-related placeholders are good for games and such, where players can only perform a given action every _n_ seconds, and would want to know how long until they can use it again.

Command and plugin names could be useful if the message needs to include some usage info, and would reduce code duplication (and later, l10n workload) by allowing a given plugin the use of just one templatized rate-limit string in place of many.

### Alternatives

There are no alternatives to the remaining-time placeholders. Plugin code can't otherwise insert those values. (Same for the existing nick and channel placeholders defined in the current code. The values are only available at runtime to the dispatcher.)

The plugin author obviously knows what the rate limit is, in simple cases, but if multiple types of rate limit are in effect on a single callable a hard-coded value in the message isn't always going to work.

### Notes

_No response_

SnoopJ:
### Description

Fix #2325.

This changeset includes:

  - [x] plumbing changes in data flow about rate-limiting between `Sopel` and `Rule`
  - [x] new template fields for rate limit messages
    - unless somebody has suggestions for more fields, I'm considering this complete

---

### Checklist

### DIFF
--- a/sopel/bot.py
+++ b/sopel/bot.py
@@ -592,7 +592,11 @@ class Sopel(irc.AbstractBot):
                 except plugins.exceptions.PluginError as err:
                     LOGGER.error("Cannot register URL callback: %s", err)
 
-    def rate_limit_info(self, rule: plugin_rules.Rule, trigger: Trigger) -> Tuple[bool, Optional[str]]:
+    def rate_limit_info(
+        self,
+        rule: plugin_rules.Rule,
+        trigger: Trigger,
+    ) -> Tuple[bool, Optional[str]]:
         if trigger.admin or rule.is_unblockable():
             return False, None
 
@@ -609,16 +613,22 @@ class Sopel(irc.AbstractBot):
             template = rule.user_rate_template
             rate_limit_type = "user"
             rate_limit = rule.user_rate_limit
+            metrics = user_metrics
         elif is_channel and channel_metrics.is_limited(at_time - rule.channel_rate_limit):
             template = rule.channel_rate_template
             rate_limit_type = "channel"
             rate_limit = rule.channel_rate_limit
+            metrics = channel_metrics
         elif global_metrics.is_limited(at_time - rule.global_rate_limit):
             template = rule.global_rate_template
             rate_limit_type = "global"
             rate_limit = rule.global_rate_limit
+            metrics = global_metrics
         else:
             return False, None
+
+        next_time = metrics.last_time + rate_limit
+        time_left = next_time - at_time
 
         if template:
             message = template.format(
@@ -627,8 +637,10 @@ class Sopel(irc.AbstractBot):
                 sender=trigger.sender,
                 plugin=rule.get_plugin_name(),
                 label=rule.get_rule_label(),
+                time_left=time_left,
+                time_left_sec=time_left.total_seconds(),
                 rate_limit=rate_limit,
-                rate_limit_sec=rate_limit_sec,
+                rate_limit_sec=rate_limit.total_seconds(),
                 rate_limit_type=rate_limit_type,
             )
         else:
--- a/sopel/bot.py
+++ b/sopel/bot.py
@@ -630,6 +630,8 @@ class Sopel(irc.AbstractBot):
         next_time = metrics.last_time + rate_limit
         time_left = next_time - at_time
 
+        message: Optional[str] = None
+
         if template:
             message = template.format(
                 nick=trigger.nick,
@@ -643,8 +645,6 @@ class Sopel(irc.AbstractBot):
                 rate_limit_sec=rate_limit.total_seconds(),
                 rate_limit_type=rate_limit_type,
             )
-        else:
-            message = None
 
         return True, message
 
--- a/sopel/bot.py
+++ b/sopel/bot.py
@@ -599,16 +599,17 @@ class Sopel(irc.AbstractBot):
         is_channel = trigger.sender and not trigger.sender.is_nick()
         channel = trigger.sender if is_channel else None
 
-        # TODO: these functions call now() and probably shouldn't
-        if rule.is_user_rate_limited(trigger.nick):
+        at_time = trigger.time
+
+        if rule.is_user_rate_limited(nick=trigger.nick, at_time=at_time):
             template = rule.user_rate_template
             rate_limit_type = "user"
             rate_limit_sec = rule._user_rate_limit
-        elif is_channel and rule.is_channel_rate_limited(channel=channel):
+        elif is_channel and rule.is_channel_rate_limited(channel=channel, at_time=at_time):
             template = rule.channel_rate_template
             rate_limit_type = "channel"
             rate_limit_sec = rule._channel_rate_limit
-        elif rule.is_global_rate_limited():
+        elif rule.is_global_rate_limited(at_time=at_time):
             template = rule.global_rate_template
             rate_limit_type = "global"
             rate_limit_sec = rule._global_rate_limit
--- a/sopel/bot.py
+++ b/sopel/bot.py
@@ -8,7 +8,7 @@
 from __future__ import annotations
 
 from ast import literal_eval
-from datetime import datetime
+import datetime
 import inspect
 import itertools
 import logging
@@ -601,15 +601,19 @@ class Sopel(irc.AbstractBot):
 
         at_time = trigger.time
 
-        if rule.is_user_rate_limited(nick=trigger.nick, at_time=at_time):
+        user_metrics = rule.get_user_metrics(trigger.nick)
+        channel_metrics = rule.get_channel_metrics(channel)
+        global_metrics = rule.get_global_metrics()
+
+        if user_metrics.is_limited(at_time - rule.user_rate_limit):
             template = rule.user_rate_template
             rate_limit_type = "user"
             rate_limit = rule.user_rate_limit
-        elif is_channel and rule.is_channel_rate_limited(channel=channel, at_time=at_time):
+        elif is_channel and channel_metrics.is_limited(at_time - rule.channel_rate_limit):
             template = rule.channel_rate_template
             rate_limit_type = "channel"
             rate_limit = rule.channel_rate_limit
-        elif rule.is_global_rate_limited(at_time=at_time):
+        elif global_metrics.is_limited(at_time - rule.global_rate_limit):
             template = rule.global_rate_template
             rate_limit_type = "global"
             rate_limit = rule.global_rate_limit
@@ -1029,7 +1033,7 @@ class Sopel(irc.AbstractBot):
 
         if trigger:
             message = '{} from {} at {}. Message was: {}'.format(
-                message, trigger.nick, str(datetime.utcnow()), trigger.group(0)
+                message, trigger.nick, str(datetime.datetime.utcnow()), trigger.group(0)
             )
 
         LOGGER.exception(message)
--- a/sopel/bot.py
+++ b/sopel/bot.py
@@ -592,6 +592,44 @@ class Sopel(irc.AbstractBot):
                 except plugins.exceptions.PluginError as err:
                     LOGGER.error("Cannot register URL callback: %s", err)
 
+    def rate_limit_info(self, rule: plugin_rules.Rule, trigger: Trigger) -> Tuple[bool, Optional[str]]:
+        if trigger.admin or rule.is_unblockable():
+            return False, None
+
+        is_channel = trigger.sender and not trigger.sender.is_nick()
+        channel = trigger.sender if is_channel else None
+
+        # TODO: these functions call now() and probably shouldn't
+        if rule.is_user_rate_limited(trigger.nick):
+            template = rule.user_rate_template
+            rate_limit_type = "user"
+            rate_limit_sec = rule._user_rate_limit
+        elif is_channel and rule.is_channel_rate_limited(channel=channel):
+            template = rule.channel_rate_template
+            rate_limit_type = "channel"
+            rate_limit_sec = rule._channel_rate_limit
+        elif rule.is_global_rate_limited():
+            template = rule.global_rate_template
+            rate_limit_type = "global"
+            rate_limit_sec = rule._global_rate_limit
+        else:
+            return False, None
+
+        if template:
+            message = template.format(
+                nick=trigger.nick,
+                channel=channel or 'private message',
+                sender=trigger.sender,
+                plugin=rule.get_plugin_name(),
+                label=rule.get_rule_label(),
+                rate_limit_sec=rate_limit_sec,
+                rate_limit_type=rate_limit_type,
+            )
+        else:
+            message = None
+
+        return True, message
+
     # message dispatch
 
     def call_rule(
@@ -604,25 +642,11 @@ class Sopel(irc.AbstractBot):
         context = trigger.sender
         is_channel = context and not context.is_nick()
 
-        # rate limiting
-        if not trigger.admin and not rule.is_unblockable():
-            if rule.is_user_rate_limited(nick):
-                message = rule.get_user_rate_message(nick)
-                if message:
-                    sopel.notice(message, destination=nick)
-                return
-
-            if is_channel and rule.is_channel_rate_limited(context):
-                message = rule.get_channel_rate_message(nick, context)
-                if message:
-                    sopel.notice(message, destination=nick)
-                return
-
-            if rule.is_global_rate_limited():
-                message = rule.get_global_rate_message(nick)
-                if message:
-                    sopel.notice(message, destination=nick)
-                return
+        limited, limit_msg = self.rate_limit_info(rule, trigger)
+        if limit_msg:
+            sopel.notice(limit_msg, destination=nick)
+        if limited:
+            return
 
         # channel config
         if is_channel and context in self.config:
--- a/sopel/bot.py
+++ b/sopel/bot.py
@@ -604,15 +604,15 @@ class Sopel(irc.AbstractBot):
         if rule.is_user_rate_limited(nick=trigger.nick, at_time=at_time):
             template = rule.user_rate_template
             rate_limit_type = "user"
-            rate_limit_sec = rule._user_rate_limit
+            rate_limit = rule.user_rate_limit
         elif is_channel and rule.is_channel_rate_limited(channel=channel, at_time=at_time):
             template = rule.channel_rate_template
             rate_limit_type = "channel"
-            rate_limit_sec = rule._channel_rate_limit
+            rate_limit = rule.channel_rate_limit
         elif rule.is_global_rate_limited(at_time=at_time):
             template = rule.global_rate_template
             rate_limit_type = "global"
-            rate_limit_sec = rule._global_rate_limit
+            rate_limit = rule.global_rate_limit
         else:
             return False, None
 
@@ -623,6 +623,7 @@ class Sopel(irc.AbstractBot):
                 sender=trigger.sender,
                 plugin=rule.get_plugin_name(),
                 label=rule.get_rule_label(),
+                rate_limit=rate_limit,
                 rate_limit_sec=rate_limit_sec,
                 rate_limit_type=rate_limit_type,
             )
--- a/sopel/plugin.py
+++ b/sopel/plugin.py
@@ -1112,10 +1112,27 @@ def rate(
         @rate(10, 10, 10, message='Hit the rate limit for this function.')
             # will send a NOTICE
 
-    The message can contain a placeholder for ``nick``: the message will be
-    formatted with the nick that hit rate limit::
+    The message can contain placeholders which will be filled in:
 
-        @rate(10, 10, 2, message='Sorry {nick}, you hit the rate limit!')
+    * ``nick``: the nick that hit the rate limit
+    * ``channel``: the channel in which the rate limit was hit (will be
+      ``'private-message'`` for private messages)
+    * ``sender``: the sender (nick or channel) of the message which hit
+      the rate limit
+    * ``plugin``: the name of the plugin that hit the rate limit
+    * ``label``: the label of the plugin handler that hit the rate limit
+    * ``time_left``: the time remaining before the rate limit expires, as
+      a string
+    * ``time_left_sec``: the time remaining before the rate limit expires,
+      expressed in number of seconds
+    * ``rate_limit``: the rate limit, as a string
+    * ``rate_limit_sec``: the rate limit, expressed in number of seconds
+    * ``rate_limit_type``: the type of rate limit that was hit (one of
+      ``user, group, global``)
+
+    For example::
+
+        @rate(10, 10, 2, message='Sorry {nick}, you hit the {rate_limit_type} rate limit!')
 
     Rate-limited functions that use scheduled future commands should import
     :class:`threading.Timer` instead of :mod:`sched`, or rate limiting will
@@ -1168,10 +1185,9 @@ def rate_user(
             # will send a NOTICE only when a user hits their own limit
             # as other rate limits don't have any message set
 
-    The message can contain a placeholder for ``nick``: the message will be
-    formatted with the nick that hit rate limit::
+    The message can contain the same placeholders supported by :func:`rate`::
 
-        @rate_user(5, 'Sorry {nick}, you hit your 5s limit!')
+        @rate_user(5, 'Sorry {nick}, you hit your {rate_limit_sec}s limit!')
 
     If you don't provide a message, the default message set by :func:`rate`
     (if any) will be used instead.
@@ -1213,13 +1229,11 @@ def rate_channel(
     If you don't provide a message, the default message set by :func:`rate`
     (if any) will be used instead.
 
-    The message can contain placeholders for ``nick`` and ``channel``: the
-    message will be formatted with the nick that hit rate limit for said
-    ``channel``::
+    The message can contain the same placeholders supported by :func:`rate`::
 
         @rate_channel(
             5,
-            'Sorry {nick}, you hit the 5s limit for the {channel} channel!',
+            'Sorry {nick}, you hit the {rate_limit_sec}s limit for the {channel} channel!',
         )
 
     .. versionadded:: 8.0
@@ -1248,7 +1262,9 @@ def rate_global(
     :param message: optional; message sent as NOTICE when a user hits the limit
 
     This decorator can be used alone or with the :func:`rate` decorator, as it
-    will always take precedence::
+    will always take precedence.
+
+    For example::
 
         @rate(10, 10, 10)
         @rate_global(5, 'You hit the global rate limit for this function.')
@@ -1259,8 +1275,7 @@ def rate_global(
     If you don't provide a message, the default message set by :func:`rate`
     (if any) will be used instead.
 
-    The message can contain a placeholder for ``nick``: the message will be
-    formatted with the nick that hit rate limit::
+    The message can contain the same placeholders supported by :func:`rate`::
 
         @rate_global(5, 'Sorry {nick}, you hit the 5s limit!')
 
--- a/sopel/plugins/rules.py
+++ b/sopel/plugins/rules.py
@@ -735,55 +735,50 @@ class AbstractRule(abc.ABC):
     def is_user_rate_limited(self, nick: Identifier) -> bool:
         """Tell when the rule reached the ``nick``'s rate limit.
 
-        :return: ``True`` when the rule reached the limit, ``False`` otherwise
+        :return: ``True`` when the rule reached the limit, ``False`` otherwise.
         """
 
     @abc.abstractmethod
     def is_channel_rate_limited(self, channel: Identifier) -> bool:
         """Tell when the rule reached the ``channel``'s rate limit.
 
-        :return: ``True`` when the rule reached the limit, ``False`` otherwise
+        :return: ``True`` when the rule reached the limit, ``False`` otherwise.
         """
 
     @abc.abstractmethod
     def is_global_rate_limited(self) -> bool:
         """Tell when the rule reached the global rate limit.
 
-        :return: ``True`` when the rule reached the limit, ``False`` otherwise
+        :return: ``True`` when the rule reached the limit, ``False`` otherwise.
         """
 
+    @property
     @abc.abstractmethod
-    def get_user_rate_message(self, nick: Identifier) -> Optional[str]:
-        """Give the message to send with a NOTICE to ``nick``.
+    def user_rate_template(self) -> Optional[str]:
+        """Give the message template to send with a NOTICE to ``nick``.
 
-        :param nick: the nick that is rate limited
         :return: A formatted string, or ``None`` if no message is set.
 
-        This method is called by the bot when a trigger hits the user rate
+        This property is accessed by the bot when a trigger hits the user rate
         limit (i.e. for the specificed ``nick``).
         """
 
+    @property
     @abc.abstractmethod
-    def get_channel_rate_message(
-        self,
-        nick: Identifier,
-        channel: Identifier,
-    ) -> Optional[str]:
-        """Give the message to send with a NOTICE to ``nick``.
+    def channel_rate_template(self) -> Optional[str]:
+        """Give the message template to send with a NOTICE to ``nick``.
 
-        :param nick: the nick that reached the channel's rate limit
-        :param channel: the channel that is rate limited
         :return: A formatted string, or ``None`` if no message is set.
 
         This method is called by the bot when a trigger hits the channel rate
         limit (i.e. for the specificed ``channel``).
         """
 
+    @property
     @abc.abstractmethod
-    def get_global_rate_message(self, nick: Identifier) -> Optional[str]:
+    def global_rate_template(self) -> Optional[str]:
         """Give the message to send with a NOTICE to ``nick``.
 
-        :param nick: the nick that reached the global rate limit
         :return: A formatted string, or ``None`` if no message is set.
 
         This method is called by the bot when a trigger hits the global rate
@@ -1158,36 +1153,20 @@ class Rule(AbstractRule):
         rate_limit = datetime.timedelta(seconds=self._global_rate_limit)
         return self._metrics_global.is_limited(now - rate_limit)
 
-    def get_user_rate_message(self, nick):
+    @property
+    def user_rate_template(self):
         template = self._user_rate_message or self._default_rate_message
+        return template
 
-        if not template:
-            return None
-
-        return template.format(
-            nick=nick,
-        )
-
-    def get_channel_rate_message(self, nick, channel):
+    @property
+    def channel_rate_template(self):
         template = self._channel_rate_message or self._default_rate_message
+        return template
 
-        if not template:
-            return None
-
-        return template.format(
-            nick=nick,
-            channel=channel,
-        )
-
-    def get_global_rate_message(self, nick):
+    @property
+    def global_rate_template(self):
         template = self._global_rate_message or self._default_rate_message
-
-        if not template:
-            return None
-
-        return template.format(
-            nick=nick,
-        )
+        return template
 
     def execute(self, bot, trigger):
         if not self._handler:
--- a/sopel/plugins/rules.py
+++ b/sopel/plugins/rules.py
@@ -732,7 +732,11 @@ class AbstractRule(abc.ABC):
         """
 
     @abc.abstractmethod
-    def is_user_rate_limited(self, nick: Identifier, at_time: Optional[datetime.datetime] = None) -> bool:
+    def is_user_rate_limited(
+        self,
+        nick: Identifier,
+        at_time: Optional[datetime.datetime] = None,
+    ) -> bool:
         """Tell when the rule reached the ``nick``'s rate limit.
 
         :param nick: the nick associated with this check
@@ -742,7 +746,11 @@ class AbstractRule(abc.ABC):
         """
 
     @abc.abstractmethod
-    def is_channel_rate_limited(self, channel: Identifier, at_time: Optional[datetime.datetime] = None) -> bool:
+    def is_channel_rate_limited(
+        self,
+        channel: Identifier,
+        at_time: Optional[datetime.datetime] = None,
+    ) -> bool:
         """Tell when the rule reached the ``channel``'s rate limit.
 
         :param channel: the channel associated with this check
@@ -752,7 +760,10 @@ class AbstractRule(abc.ABC):
         """
 
     @abc.abstractmethod
-    def is_global_rate_limited(self, at_time: Optional[datetime.datetime] = None) -> bool:
+    def is_global_rate_limited(
+        self,
+        at_time: Optional[datetime.datetime] = None,
+    ) -> bool:
         """Tell when the rule reached the global rate limit.
 
         :param at_time: optional aware datetime for the rate limit check;
@@ -1144,6 +1155,15 @@ class Rule(AbstractRule):
     def is_unblockable(self):
         return self._unblockable
 
+    def get_user_metrics(self, nick: Identifier) -> RuleMetrics:
+        return self._metrics_nick.get(nick, RuleMetrics())
+
+    def get_channel_metrics(self, channel: Identifier) -> RuleMetrics:
+        return self._metrics_sender.get(channel, RuleMetrics())
+
+    def get_global_metrics(self) -> RuleMetrics:
+        return self._metrics_global
+
     @property
     def user_rate_limit(self) -> datetime.timedelta:
         return datetime.timedelta(seconds=self._user_rate_limit)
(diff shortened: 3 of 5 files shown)


sopel/plugins/rules.py
```diff
@@ -1158,36 +1166,63 @@ def is_global_rate_limited(self):
...
+        if trigger.admin or self.is_unblockable():
+            return RateLimitCheckInfo()
+
+        # TODO: these functions call now() and probably shouldn't
+        if self.is_user_rate_limited(trigger.nick):
+            template = self.get_user_rate_message(trigger.nick)
+        elif self.is_channel_rate_limited(trigger.sender):
+            template = self.get_channel_rate_message(trigger.nick, trigger.sender)
+        elif self.is_global_rate_limited():
+            template = self.get_global_rate_message(trigger.nick)
```
SnoopJ, line 1179: If the getters are not public API that we want to preserve, their logic can be moved here

```suggestion
        if self.is_user_rate_limited(trigger.nick):
            template = self._user_rate_message or self._default_rate_message
        elif self.is_channel_rate_limited(trigger.sender):
            template = self._channel_rate_message or self._default_rate_message
        elif self.is_global_rate_limited():
            template = self._global_rate_message or self._default_rate_message
```

SnoopJ: Similarly, the business of the `is_*_rate_limited()` methods might find a home in this function as well, since they are only being called internally to `Rule` now

sopel/plugins/rules.py
```diff
@@ -814,6 +816,12 @@ def execute(self, bot, trigger):
         """
 
 
+class RateLimitCheckInfo(NamedTuple):
+    """Information from checking if a rate limit has been enforced"""
```
dgw, line 820:
```suggestion
    """Information from checking if a rate limit has been enforced."""
```

Nitpick early, nitpick often. 🙃

sopel/plugins/rules.py
dgw: None of `Rule`'s interface is really "public API" in the sense that we encourage or expect plugin developers to use it. One look at the `.. important::` at the top of the docs page that covers `Rule` & friends will tell you that. ("Do **not** build your plugin based on what is here, you do **not** need to.")

No need to plan around deprecation cycles for this stuff. No one outside of Sopel core is supposed to be using it, and if they are we can't be held responsible for their stuff breaking. _They_ made the choice to use clearly marked internal machinery. 🤷‍♂️

sopel/plugins/rules.py
```diff
@@ -1158,36 +1166,63 @@ def is_global_rate_limited(self):
...
+        #     * time_left_sec
+        #     * rate_limit
+        #     * rate_limit_type
+        #     * command
+        #     * plugin
+        #     * ???
+        if template:
+            message = template.format(
+                nick=trigger.nick,
+                # TODO: add is_channel() to trigger?
```
dgw, line 1194: We haven't done this because it seems pretty redundant and pointless to have a method whose body can be implemented as `return !self.other_method()`.

sopel/plugins/rules.py
```diff
@@ -1158,36 +1166,63 @@ def is_global_rate_limited(self):
...
     def get_global_rate_message(self, nick):
         template = self._global_rate_message or self._default_rate_message
 
         if not template:
             return None
 
-        return template.format(
-            nick=nick,
-        )
+        return template
```
dgw, line 1225: I was surprised there was no flake8 warning about unused method parameters on any of this. Then I remembered that #1765 exists, with a task item to add `flake8-unused-arguments` to the dev requirements someday.

Surely you just forgot to remove the now-unused positional arguments?

dgw (commented):
Since I wanted to respond to the thread about public API concerns, thought I'd give the rest a quick skim too.

Exirel (changes requested):
I've been thinking a lot about rate limiting (yes, since yesterday evening when I said I went to bed, apparently that kept me awake for an hour or so...). I understand why you want to change the existing methods, yet I want to go another way: add more methods!

For instance, most of the logic is in the `RuleMetrics` class: once you have it, you can ask if it is limited, and it contains most of the required information to format the rule's template for rate limiting.

Consider this:

```
@property
def user_rate_limit(self) -> datetime.timedelta:
    return  datetime.timedelta(seconds=self._*_rate_limit)

@property
def user_rate_template(self) -> Optional[str]:
    return self._user_rate_message or self._default_rate_message

def get_user_metrics(self, nick: Identifier) -> RuleMetrics:
    return self._metrics_nick.setdefault(nick, RuleMetrics())
```

Then you can have something like that:

```
triggered_at = datetime.datetime.utcnow()
user_rate_limit = rule.user_rate_limit  
channel_rate_limit = rule.channel_rate_limit
global_rate_limit = rule.global_rate_limit
user_metrics = rule.get_user_metrics(trigger.nick)
channel_metrics = rule.get_channel_metrics(trigger.sender)
global_metrics = rule.global_metrics

metrics, template = None, None
if user_metrics.is_limited(triggered_at - user_rate_limit):
    template = rule.user_rate_template
    metrics = user_metrics
elif channel_metrics.is_limited(triggered_at - user_rate_limit):
    template = rule.channel_rate_template
    metrics = channel_metrics
elif global_metrics.is_limited(triggered_at - global_rate_limit):
    template = rule.global_rate_template
    metrics  = global_metrics

message = None
if metrics and template:
    message = template.format(...)

return RateLimitCheckInfo(limited=metrics is not None, metrics, triggered_at, message)
```

In the end, you still get an object that can tell you if 1. it was rate limited, 2. the data and metrics for that, and 3. the message to send if any! With some magic `@property` you can even prevent yourself from having to pass the `limited`, as it can be derived from the existence of a `metrics` object.

sopel/plugins/rules.py
```diff
@@ -1158,36 +1166,63 @@ def is_global_rate_limited(self):
         rate_limit = datetime.timedelta(seconds=self._global_rate_limit)
         return self._metrics_global.is_limited(now - rate_limit)
 
+    def rate_limit_info(self, trigger: Trigger) -> RateLimitCheckInfo:
```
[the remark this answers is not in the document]
SnoopJ: Agree as discussed on IRC. Moved this function back across the interface boundary in 3a7063b

sopel/plugins/rules.py
SnoopJ: Yea that's just oversight on my part as things churned around in this draft. These getters are removed in the latest revision and the `format()` lives on the `Sopel` side now.

sopel/plugins/rules.py
SnoopJ: Fair enough, dropped the TODO, which I only committed by mistake anyway, it was there to remind me to ask about it :sweat_smile:

sopel/plugins/rules.py
SnoopJ: For the sake of posterity, the `get_*` family are removed in 3a7063b, but the `is_*` family are staying around as possible future public API. I'm sure this PR will make some kind of changes to them, but they aren't going anywhere, so this thread is now resolved.

SnoopJ:
This PR is functionally complete now, but still needs a pass for testing and general quality (docstrings etc.)

Additionally, using `at_time = trigger.time` in `Sopel.rate_limit_info()` (which might need a better name, can anybody think of one?) seems to break some of the tests for `call_rule()`:

```
___________________________________________________________________________ test_call_rule ___________________________________________________________________________
test/test_bot.py:643: in test_call_rule
    assert mockbot.backend.message_sent == rawlist(
E   AssertionError: assert [b'PRIVMSG #channel :hi\r\n'] == [b'PRIVMSG #c...nnel :hi\r\n']
E     Right contains one more item: b'PRIVMSG #channel :hi\r\n'
E     Use -v to get more diff
__________________________________________________________ test_call_rule_rate_limited_channel_with_message __________________________________________________________
test/test_bot.py:855: in test_call_rule_rate_limited_channel_with_message
    assert mockbot.backend.message_sent == rawlist(
E   AssertionError: A NOTICE should appear here.
E   assert [b'PRIVMSG #channel :hi\r\n'] == [b'PRIVMSG #c...e limit.\r\n']
E     Right contains one more item: b'NOTICE Test :You reached the channel rate limit.\r\n'
E     Use -v to get more diff
__________________________________________________________ test_call_rule_rate_limited_global_with_message ___________________________________________________________
test/test_bot.py:963: in test_call_rule_rate_limited_global_with_message
    assert mockbot.backend.message_sent == rawlist(
E   AssertionError: A NOTICE should appear here.
E   assert [b'PRIVMSG #channel :hi\r\n'] == [b'PRIVMSG #c...e limit.\r\n']
E     Right contains one more item: b'NOTICE Test :You reached the server rate limit.\r\n'
E     Use -v to get more diff
====================================================================== short test summary info =======================================================================
FAILED test/test_bot.py::test_call_rule - AssertionError: assert [b'PRIVMSG #channel :hi\r\n'] == [b'PRIVMSG #c...nnel :hi\r\n']
FAILED test/test_bot.py::test_call_rule_rate_limited_channel_with_message - AssertionError: A NOTICE should appear here.
FAILED test/test_bot.py::test_call_rule_rate_limited_global_with_message - AssertionError: A NOTICE should appear here.
==================================================================== 3 failed, 47 passed in 2.60s ====================================================================
```

Looks like `test_call_rule` is getting rate-limited where no limit is expected, and a `NOTICE` is not being sent for `test_call_rule_rate_limited_channel_with_message` and `test_call_rule_rate_limited_global_with_message`.

I suspect this is caused by some implicit dependence of the tests on the passage of time between the construction of a `Trigger` associated with the relevant even and calling of `Sopel.call_rule()`. It's not a very long difference, I observe it to be `O(1 ms)`, but something is amiss there.

dgw:
Haven't poked at _all_ of the failing-with-`at_time = trigger.time` tests, but I threw `pdb` into `Sopel.rate_limit_info()` and tried out `test_call_rule_rate_limited_channel_with_message`:

```
test/test_bot.py::test_call_rule_rate_limited_channel_with_message
>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>> PDB set_trace (IO-capturing turned off) >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
> /home/dgw/github/sopel/sopel/bot.py(597)rate_limit_info()
-> if trigger.admin or rule.is_unblockable():
(Pdb) c

>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>> PDB continue (IO-capturing resumed) >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>

>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>> PDB set_trace (IO-capturing turned off) >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
> /home/dgw/github/sopel/sopel/bot.py(597)rate_limit_info()
-> if trigger.admin or rule.is_unblockable():
(Pdb) n
> /home/dgw/github/sopel/sopel/bot.py(600)rate_limit_info()
-> is_channel = trigger.sender and not trigger.sender.is_nick()
(Pdb) n
> /home/dgw/github/sopel/sopel/bot.py(601)rate_limit_info()
-> channel = trigger.sender if is_channel else None
(Pdb) n
> /home/dgw/github/sopel/sopel/bot.py(606)rate_limit_info()
-> at_time = trigger.time
(Pdb) n
> /home/dgw/github/sopel/sopel/bot.py(608)rate_limit_info()
-> user_metrics = rule.get_user_metrics(trigger.nick)
(Pdb) print(at_time)
2023-03-27 07:51:05.409947+00:00
(Pdb) print(rule.channel_rate_limit)
0:01:40
(Pdb) n
> /home/dgw/github/sopel/sopel/bot.py(609)rate_limit_info()
-> channel_metrics = rule.get_channel_metrics(channel)
(Pdb) n
> /home/dgw/github/sopel/sopel/bot.py(610)rate_limit_info()
-> global_metrics = rule.get_global_metrics()
(Pdb) n
> /home/dgw/github/sopel/sopel/bot.py(612)rate_limit_info()
-> if user_metrics.is_limited(at_time - rule.user_rate_limit):
(Pdb) n
> /home/dgw/github/sopel/sopel/bot.py(613)rate_limit_info()
-> template = rule.user_rate_template
(Pdb) print(user_metrics.is_limited(at_time - rule.user_rate_limit))
True
(Pdb) print(rule.user_rate_limit)
0:00:00
(Pdb) n
> /home/dgw/github/sopel/sopel/bot.py(614)rate_limit_info()
-> rate_limit_type = "user"
(Pdb) n
> /home/dgw/github/sopel/sopel/bot.py(615)rate_limit_info()
-> rate_limit = rule.user_rate_limit
(Pdb) n
> /home/dgw/github/sopel/sopel/bot.py(616)rate_limit_info()
-> metrics = user_metrics
(Pdb) n
> /home/dgw/github/sopel/sopel/bot.py(630)rate_limit_info()
-> next_time = metrics.last_time + rate_limit
(Pdb) print(rate_limit_type)
user
```

The `user` rate-limit type is being triggered somehow, not the `channel` type.

Same happens for the second trigger in `test_call_rule_rate_limited_global`—the test code enters the `if user_metrics.is_limited(` block even though `rule.user_rate_limit` in this context is a `datetime.timedelta(0)`.

More tests than are throwing errors are failing with `at_time = trigger.time`; they just don't check any conditions that reveal the failure. The second message isn't sent, as expected, but _the wrong rate-limit type_ is responsible for blocking it. Only the _non-user_ `_with_message` tests reveal the problem, because there's no associated message in the (unset) `user` rate limit properties.

sopel/bot.py
```diff
@@ -592,6 +592,71 @@ def register_urls(self, urls: Iterable) -> None:
                 except plugins.exceptions.PluginError as err:
                     LOGGER.error("Cannot register URL callback: %s", err)
 
+    def rate_limit_info(self, rule: plugin_rules.Rule, trigger: Trigger) -> Tuple[bool, Optional[str]]:
+        if trigger.admin or rule.is_unblockable():
+            return False, None
+
+        is_channel = trigger.sender and not trigger.sender.is_nick()
+        channel = trigger.sender if is_channel else None
```
Exirel, line 600: Given the usage of `channel` in the `format()` down the line, I'd replace it by something that can be read, like `private-message`:

```suggestion
        channel = trigger.sender if is_channel else 'private-message'
```

The result would look like this:

```python
>>> 'You must wait before using this command in {channel} again.'.format(channel='private-message')
'You must wait before using this command in private-message again.'
```

That should be OK in most cases. Note that there is one and only one occasion to have this "private-message" is when the user is directly calling the bot in private.

sopel/plugins/rules.py
```diff
@@ -484,15 +493,11 @@ def is_limited(
             # not even started, so not limited
             return False
 
-        # detect if we just started something or if it ended
-        last_time = self.started_at
-        if self.ended_at and self.started_at < self.ended_at:
```
Exirel, line 489: This was used to check that the rule's execution ended. Because rule can be running in a thread, you have no guaranty that a second execution is not trying to do stuff at the same time, so the `last_return_value` must not apply unless it has actually ended first.

SnoopJ:
> using `at_time = trigger.time` in `Sopel.rate_limit_info()` … seems to break some of the tests for `call_rule()`

It turns out that these failures were caused by re-use of the triggers in these tests, failing to trigger the rate-limiting behavior because too much time had elapsed since the _first_ event. Re-creating the triggers associated with the event resolves the failures.

sopel/plugins/rules.py
SnoopJ: Thanks, I've restored that check in fb774ba

sopel/plugin.py
```diff
@@ -1112,10 +1112,27 @@ def rate(
...
+        - ``plugin``: the name of the plugin that hit the rate limit
+        - ``label``: the label of the plugin handler that hit the rate limit
+        - ``time_left``: the time remaining before the rate limit expires, as
+          a string
+        - ``time_left_sec``: the time remaining before the rate limit expires,
+          expressed in number of seconds
+        - ``rate_limit``: the rate limit, as a string
+        - ``rate_limit_sec``: the rate limit, expressed in number of seconds
+        - ``rate_limit_type``: the type of rate limit that was hit (one of
+          ``user, group, global``)
```
Exirel, line 1131: Restructured text doesn't use tab/space before a list item, this may even confused it. And you'll have to change the `-` by a `*`: https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#lists-and-quote-like-blocks

sopel/plugin.py
```diff
@@ -1168,10 +1185,27 @@ def rate_user(
...
```
Exirel, line 1204: Same here.

sopel/plugin.py
```diff
@@ -1213,13 +1247,29 @@ def rate_channel(
...
```
Exirel, line 1266: Ditto.

sopel/plugin.py
```diff
@@ -1248,10 +1298,30 @@ def rate_global(
...
```
Exirel, line 1319: 4th times the charm!

That makes me wonder: if they are all the same, maybe reference them only one, and use a sentence like:

```
The message can contain placeholders which will be filled in; they are the same
placeholders as :func:`that_other_rate_function`.
```

sopel/plugins/rules.py
```diff
@@ -465,16 +465,25 @@ def __init__(self) -> None:
 
     def start(self) -> None:
         """Record a starting time (before execution)."""
-        self.started_at = datetime.datetime.utcnow()
+        self.started_at = datetime.datetime.utcnow().replace(tzinfo=datetime.timezone.utc)
```
Exirel, line 468: Why not `pytz.utc.localize(datetime.datetime.utcnow())`?

sopel/plugins/rules.py
```diff
@@ -465,16 +465,25 @@ def __init__(self) -> None:
...
-        self.ended_at = datetime.datetime.utcnow()
+        self.ended_at = datetime.datetime.utcnow().replace(tzinfo=datetime.timezone.utc)
 
     def set_return_value(self, value: Any) -> None:
         """Set the last return value of a rule."""
         self.last_return_value = value
 
+    @property
+    def last_time(self):
+        # detect if we just started something or if it ended
```
Exirel, line 480: Public property should have a docstring.

```suggestion
        """Detect if we just started something or if it ended."""
```

sopel/plugins/rules.py
```diff
@@ -465,16 +465,25 @@ def __init__(self) -> None:
...
         """Record a ending time (after execution)."""
-        self.ended_at = datetime.datetime.utcnow()
+        self.ended_at = datetime.datetime.utcnow().replace(tzinfo=datetime.timezone.utc)
 
     def set_return_value(self, value: Any) -> None:
         """Set the last return value of a rule."""
         self.last_return_value = value
 
+    @property
+    def last_time(self):
```
Exirel, line 479: Should contain type annotation:

```suggestion
    def last_time(self) -> Optional[datetime.datetime]:
```

sopel/plugins/rules.py
```diff
@@ -732,58 +738,58 @@ def is_unblockable(self) -> bool:
         """
 
     @abc.abstractmethod
-    def is_user_rate_limited(self, nick: Identifier) -> bool:
+    def is_user_rate_limited(self, nick: Identifier, at_time: Optional[datetime.datetime] = None) -> bool:
         """Tell when the rule reached the ``nick``'s rate limit.
 
-        :return: ``True`` when the rule reached the limit, ``False`` otherwise
+        :param nick: the nick associated with this check.
+        :param at_time: optional time for the rate limit check. If not given, current time will be used.
```
Exirel, line 745: A datetime is not a "time". If you say "time", people may wonder if they can use `time.time()`. A `datetime` is a specific type (not to be confused with date + time either).

And, this might sound unimportant, but it actually is: it's an *aware* datetime. Since you can't compare aware datetime with naive datetime, this information matters. 

```suggestion
        :param at_time: optional aware datetime for the rate limit check.
                        If not given, ``utcnow`` will be used.
```

sopel/plugins/rules.py
```diff
@@ -732,58 +738,58 @@ def is_unblockable(self) -> bool:
...
         """
 
     @abc.abstractmethod
-    def is_channel_rate_limited(self, channel: Identifier) -> bool:
+    def is_channel_rate_limited(self, channel: Identifier, at_time: Optional[datetime.datetime] = None) -> bool:
         """Tell when the rule reached the ``channel``'s rate limit.
 
-        :return: ``True`` when the rule reached the limit, ``False`` otherwise
+        :param channel: the channel associated with this check.
+        :param at_time: optional time for the rate limit check. If not given, current time will be used.
```
Exirel, line 754: Ditto

sopel/plugins/rules.py
```diff
@@ -732,58 +738,58 @@ def is_unblockable(self) -> bool:
...
+        :return: ``True`` when the rule reached the limit, ``False`` otherwise.
         """
 
     @abc.abstractmethod
-    def is_global_rate_limited(self) -> bool:
+    def is_global_rate_limited(self, at_time: Optional[datetime.datetime] = None) -> bool:
         """Tell when the rule reached the global rate limit.
 
-        :return: ``True`` when the rule reached the limit, ``False`` otherwise
+        :param at_time: optional time for the rate limit check. If not given, current time will be used.
```
Exirel, line 762: Ditto.

sopel/plugins/rules.py
```diff
@@ -1141,53 +1147,59 @@ def is_threaded(self):
...
-            nick=nick,
-            channel=channel,
-        )
+    @property
+    def global_rate_limit(self) -> datetime.timedelta:
+        return datetime.timedelta(seconds=self._global_rate_limit)
+
+    def is_user_rate_limited(self, nick, at_time=None) -> bool:
+        if at_time is None:
+            at_time = datetime.datetime.utcnow().replace(tzinfo=datetime.timezone.utc)
```
Exirel, line 1173: Same question here for `pytz.localize`.

sopel/plugins/rules.py
```diff
@@ -1141,53 +1147,59 @@ def is_threaded(self):
...
+        metrics = self.get_global_metrics()
+        return metrics.is_limited(at_time - self.global_rate_limit)
 
-    def get_global_rate_message(self, nick):
-        template = self._global_rate_message or self._default_rate_message
+    @property
+    def user_rate_template(self):
+        template = self._user_rate_message or self._default_rate_message or None
```
Exirel, line 1191: `self._default_rate_message` should be `Optional[str]` anyway, so I don't think the `or None` is necessary:

```suggestion
        template = self._user_rate_message or self._default_rate_message
```

sopel/plugins/rules.py
```diff
@@ -1141,53 +1147,59 @@ def is_threaded(self):
...
+        return template
 
-        if not template:
-            return None
+    @property
+    def channel_rate_template(self):
+        template = self._channel_rate_message or self._default_rate_message or None
```
Exirel, line 1196: Ditto.

sopel/plugins/rules.py
```diff
@@ -1141,53 +1147,59 @@ def is_threaded(self):
...
+        return template
 
-        return template.format(
-            nick=nick,
-        )
+    @property
+    def global_rate_template(self):
+        template = self._global_rate_message or self._default_rate_message or None
```
Exirel, line 1201: Ditto.

test/test_bot.py
```diff
@@ -593,7 +593,27 @@ def url_callback_http(bot, trigger, match):
 # -----------------------------------------------------------------------------
 # call_rule
 
-def test_call_rule(mockbot):
+@pytest.fixture
+def make_hello_event(mockbot):
```
Exirel, line 597: I was confused at first when reading the tests, because "make_hello_event" uses an overloaded word: "event". Since "make" is a generic verb, and "hello" refers to the text message, it wasn't clear exactly what it does. It could either generate an event (within an event system), or generate a trigger with an `HELLO` event.

Since it appears that this doesn't generate an `HELLO` trigger, but rather a `PRIVMSG` event with an `hello` text, and then match a rule based on that... I think this is a better name that represents what this fixture actually does:

```suggestion
def match_hello_rule(mockbot):
```

It match a hello rule, and now the return value of `re.match, Trigger, SopelWrapper` makes more sense to me.

And yes this means changing the 7 or more occurrences down the chain. 😁

Exirel: You might want to use the `triggerfactory`, maybe?

Exirel: Oh, and before I forget: a docstring would be nice. This is not as trivial as creating a single object, a bunch of stuff is done in here without explicit reasons at first. You have to read the tests below to understand the reason for this fixture, and since reading code can be difficult, the reason might remain unclear.

Exirel (changes requested):
That looks better already!

A lot of nitpicking and formatting. In theory, the code works as expected, so it's not that important. The review is more about added quality than anything else.

The idea of adding a fixture to create a match for a rule and a trigger is nice. I'm not sure I would have made it that way, since I prefer when things don't have that many hard-coded text, otherwise it's cool to see the reduction in code repetition.

sopel/plugin.py
SnoopJ: I was worrying about the repetition but this straightforward approach did not occur to me. Thanks, done in `ae9a96d`

test/test_bot.py
SnoopJ: Great suggestions all, implemented in c844758 and 4eea7a1

sopel/plugins/rules.py
SnoopJ: Mostly that I didn't see any usage of it in the machinery that I poking around in, I patterned the `replace()` on the [code in `Trigger`](https://github.com/SnoopJ/sopel/blob/19605f9509f0471a1688e7cf6fc64f00790af970/sopel/trigger.py#L192-L195) which we may want to change in this PR. I'll defer to your judgment about that.

sopel/bot.py
```diff
@@ -592,6 +592,58 @@ def register_urls(self, urls: Iterable) -> None:
                 except plugins.exceptions.PluginError as err:
                     LOGGER.error("Cannot register URL callback: %s", err)
 
+    def rate_limit_info(self, rule: plugin_rules.Rule, trigger: Trigger) -> Tuple[bool, Optional[str]]:
```
Exirel, line 595:
```suggestion
    def rate_limit_info(
        self,
        rule: plugin_rules.Rule,
        trigger: Trigger,
    ) -> Tuple[bool, Optional[str]]:
```

This is the way.

sopel/bot.py
```diff
@@ -592,6 +592,58 @@ def register_urls(self, urls: Iterable) -> None:
...
+        if trigger.admin or rule.is_unblockable():
+            return False, None
+
+        is_channel = trigger.sender and not trigger.sender.is_nick()
+        channel = trigger.sender if is_channel else 'private-message'
```
Exirel, line 600: I don't like this. `channel` should be `None` here if it isn't a channel. And so is `channel_metrics`.

You could then replace the later `elif is_channel and [...]` by `elif channel_metrics is not None and [...]`.

sopel/bot.py
```diff
@@ -592,6 +592,58 @@ def register_urls(self, urls: Iterable) -> None:
...
+            metrics = global_metrics
+        else:
+            return False, None
+
+        next_time = metrics.last_time + rate_limit
+        time_left = next_time - at_time
+
+        if template:
+            message = template.format(
+                nick=trigger.nick,
```
Exirel, line 631: `trigger.nick` being used twice would mean an intermediate variable to me.

sopel/bot.py
```diff
@@ -592,6 +592,58 @@ def register_urls(self, urls: Iterable) -> None:
...
+                channel=channel,
```
Exirel, line 632: I'd ensure here that the `channel` key is `str` (goes with the previous issue of "channel should be `None`"):

```suggestion
                channel=channel or 'private message',
```

sopel/bot.py
```diff
@@ -592,6 +592,58 @@ def register_urls(self, urls: Iterable) -> None:
...
+                plugin=rule.get_plugin_name(),
+                label=rule.get_rule_label(),
+                time_left=time_left,
+                time_left_sec=time_left.total_seconds(),
+                rate_limit=rate_limit,
+                rate_limit_sec=rate_limit.total_seconds(),
+                rate_limit_type=rate_limit_type,
+            )
+        else:
+            message = None
```
Exirel, line 643: You should use this style:

```
message: Optional[str] = None

if <condition>:
    message = "Something"
```

sopel/plugins/rules.py
```diff
@@ -732,58 +741,58 @@ def is_unblockable(self) -> bool:
         """
 
     @abc.abstractmethod
-    def is_user_rate_limited(self, nick: Identifier) -> bool:
+    def is_user_rate_limited(self, nick: Identifier, at_time: Optional[datetime.datetime] = None) -> bool:
```
Exirel, line 744: This is the way:

```suggestion
    def is_user_rate_limited(
        self,
        nick: Identifier,
        at_time: Optional[datetime.datetime] = None,
    ) -> bool:
```

sopel/plugins/rules.py
```diff
@@ -732,58 +741,58 @@ def is_unblockable(self) -> bool:
...
 
-        :return: ``True`` when the rule reached the limit, ``False`` otherwise
+        :param nick: the nick associated with this check.
+        :param at_time: optional aware datetime for the rate limit check. If not given, ``utcnow`` will be used.
+        :return: ``True`` when the rule reached the limit, ``False`` otherwise.
         """
 
     @abc.abstractmethod
-    def is_channel_rate_limited(self, channel: Identifier) -> bool:
+    def is_channel_rate_limited(self, channel: Identifier, at_time: Optional[datetime.datetime] = None) -> bool:
```
Exirel, line 753: This is the way:

```suggestion
    def is_channel_rate_limited(
        self,
        channel: Identifier,
        at_time: Optional[datetime.datetime] = None,
    ) -> bool:
```

sopel/plugins/rules.py
```diff
@@ -732,58 +741,58 @@ def is_unblockable(self) -> bool:
...
         """Tell when the rule reached the ``channel``'s rate limit.
 
-        :return: ``True`` when the rule reached the limit, ``False`` otherwise
+        :param channel: the channel associated with this check.
+        :param at_time: optional aware datetime for the rate limit check. If not given, ``utcnow`` will be used.
```
Exirel, line 757:
```suggestion
        :param at_time: optional aware datetime for the rate limit check.
                        If not given, ``utcnow`` will be used.
```

sopel/plugins/rules.py
```diff
@@ -732,58 +741,58 @@ def is_unblockable(self) -> bool:
...
+        :return: ``True`` when the rule reached the limit, ``False`` otherwise.
         """
 
     @abc.abstractmethod
-    def is_global_rate_limited(self) -> bool:
+    def is_global_rate_limited(self, at_time: Optional[datetime.datetime] = None) -> bool:
```
Exirel, line 762:
```suggestion
    def is_global_rate_limited(
        self,
        at_time: Optional[datetime.datetime] = None,
    ) -> bool:
```

sopel/plugins/rules.py
```diff
@@ -732,58 +741,58 @@ def is_unblockable(self) -> bool:
...
         """Tell when the rule reached the global rate limit.
 
-        :return: ``True`` when the rule reached the limit, ``False`` otherwise
+        :param at_time: optional aware datetime for the rate limit check. If not given, ``utcnow`` will be used.
```
Exirel, line 765:
```suggestion
        :param at_time: optional aware datetime for the rate limit check.
                        If not given, ``utcnow`` will be used.
```

sopel/plugins/rules.py
```diff
@@ -1141,53 +1150,59 @@ def is_threaded(self):
     def is_unblockable(self):
         return self._unblockable
 
-    def is_user_rate_limited(self, nick):
-        metrics: RuleMetrics = self._metrics_nick.get(nick, RuleMetrics())
-        now = datetime.datetime.utcnow()
-        rate_limit = datetime.timedelta(seconds=self._user_rate_limit)
-        return metrics.is_limited(now - rate_limit)
+    def get_user_metrics(self, nick) -> RuleMetrics:
```
Exirel, line 1153: Let's go full annotations:

```suggestion
    def get_user_metrics(self, nick: Identifier) -> RuleMetrics:
```

sopel/plugins/rules.py
```diff
@@ -1141,53 +1150,59 @@ def is_threaded(self):
...
+        return self._metrics_nick.get(nick, RuleMetrics())
 
-    def is_channel_rate_limited(self, channel):
-        metrics: RuleMetrics = self._metrics_sender.get(channel, RuleMetrics())
-        now = datetime.datetime.utcnow()
-        rate_limit = datetime.timedelta(seconds=self._channel_rate_limit)
-        return metrics.is_limited(now - rate_limit)
+    def get_channel_metrics(self, channel) -> RuleMetrics:
```
Exirel, line 1156:
```suggestion
    def get_channel_metrics(self, channel: Identifier) -> RuleMetrics:
```

sopel/plugins/rules.py
```diff
@@ -1141,53 +1150,59 @@ def is_threaded(self):
...
+    def channel_rate_limit(self) -> datetime.timedelta:
+        return datetime.timedelta(seconds=self._channel_rate_limit)
 
-    def get_channel_rate_message(self, nick, channel):
-        template = self._channel_rate_message or self._default_rate_message
+    @property
+    def global_rate_limit(self) -> datetime.timedelta:
+        return datetime.timedelta(seconds=self._global_rate_limit)
+
+    def is_user_rate_limited(self, nick, at_time=None) -> bool:
```
Exirel, line 1174:
```suggestion
    def is_user_rate_limited(
        self,
        nick: Identifier,
        at_time: Optional[datetime.datetime] = None,
    ) -> bool:
```

sopel/plugins/rules.py
```diff
@@ -1141,53 +1150,59 @@ def is_threaded(self):
...
+    def is_global_rate_limited(self, at_time=None) -> bool:
+        if at_time is None:
+            at_time = pytz.utc.localize(datetime.datetime.utcnow())
+        metrics = self.get_global_metrics()
+        return metrics.is_limited(at_time - self.global_rate_limit)
 
-        if not template:
-            return None
+    @property
+    def user_rate_template(self):
```
Exirel, line 1193:
```suggestion
    def user_rate_template(self) -> Optional[str]:
```

sopel/plugins/rules.py
```diff
@@ -1141,53 +1150,59 @@ def is_threaded(self):
...
-            channel=channel,
-        )
+    @property
+    def channel_rate_template(self):
+        template = self._channel_rate_message or self._default_rate_message
+        return template
 
-    def get_global_rate_message(self, nick):
+    @property
+    def global_rate_template(self):
```
Exirel, line 1203:
```suggestion
    def global_rate_template(self) -> Optional[str]:
```

test/plugins/test_plugins_rules.py
```diff
@@ -468,7 +468,7 @@ def test_manager_has_action_command_aliases():
 # tests for :class:`Manager`
 
 def test_rulemetrics():
-    now = datetime.datetime.utcnow()
+    now = datetime.datetime.utcnow().replace(tzinfo=datetime.timezone.utc)
```
Exirel, line 471:
```suggestion
    now = pytz.utc.localize(datetime.datetime.utcnow())
```

test/test_bot.py
```diff
@@ -967,6 +922,119 @@ def testrule(bot, trigger):
...
+@pytest.mark.parametrize("msg_fmt, expected_notice", (
+    ("{nick}", "NOTICE Test :Test"),
+    ("{channel}", "NOTICE Test :#channel"),
+    ("{sender}", "NOTICE Test :#channel"),
+    ("{plugin}", "NOTICE Test :testplugin"),
+    ("{label}", "NOTICE Test :testrule"),
+    ("{rate_limit}", "NOTICE Test :0:01:40"),
+    ("{rate_limit_sec}", "NOTICE Test :100.0"),
+))
+def test_rate_limit_fixed_fields(mockbot, match_hello_rule, limit_type, msg_fmt, expected_notice):
```
Exirel, line 938:
```suggestion
def test_rate_limit_fixed_fields(
    mockbot,
    match_hello_rule,
    limit_type,
    msg_fmt,
    expected_notice,
):
```

Exirel (changes requested):
Given the amount of commits you have, good luck with the squash that will inevitably be a nightmare. :grin: 

Remember that we use `modules: short message` format for commits, not just for PR.

sopel/bot.py
dgw: (Not for me, though. Attributes are meant to be accessed!)

sopel/plugins/rules.py
dgw: Our param list style avoids capitalized sentences and full stops, though. In this case, that means using a semicolon to separate the "if not given".

Also applies to the line above (remove `.`)

sopel/plugins/rules.py
dgw: (see my reply to the previous `at_time` param's line note)

sopel/bot.py
SnoopJ: Usage twice seems like a low threshold to me, I usually define a variable in the case that either:

1) it helps the legibility (i.e. because the expr being replaced is long/complex)
2) the attribute is likely to change between retrievals

But I'm not sure I see either of those cases here. Willing to change it though, I'll let you two work that one out :sweat_smile:

Exirel: It's a bit weird to have `channel` but not `nick` tbh. 😁 

Other than that, yeah, it's fine.

sopel/plugins/rules.py
Exirel: I updated my suggestion to match.

sopel/plugins/rules.py
Exirel: (same)

sopel/bot.py
SnoopJ: > It's a bit weird to have `channel` but not `nick` tbh. grin

If `Trigger` provided a `channel` attribute, I would have happily used it here and would happily do so in many other places too :stuck_out_tongue_winking_eye: 

In this case, I defined `channel` because the expression that checks `trigger.sender` _is_ a burden to repeat and would harm the legibility of surrounding code

SnoopJ:
Squashed this down as best as I could, but if it needs to be squashed any further, I think I might just discard all of the working history and re-arrange from scratch. Let me know if that needs done.

SnoopJ:
CI failure is unrelated to the contents of this PR, seems to be an issue where CI has a more recent `urllib3` than what is in `dev-requirements.txt` causing an issue with `vcrpy`

```
AttributeError: module 'urllib3.connectionpool' has no attribute 'VerifiedHTTPSConnection'
```

This issue is being tracked upstream but

SnoopJ:
> Squashed this down as best as I could, but if it needs to be squashed any further, I think I might just discard all of the working history and re-arrange from scratch. Let me know if that needs done.

Can I get a yes/no on the current squash of this PR? 12 commits still feels like a lot, but squashing to a single commit _also_ feels excessive.

Exirel:
LGTM. Reading the list of commits, each has a purpose that is clear and separate from other commits. Sure, we can see the progress "add this one, then this one, also fix that, change this, add another one", etc. and personally, I don't mind / I like it this way.

Note that I don't mind either if you prefer to squash down to 2 or 3 commits (one for code, one for doc), I think both version of the history are acceptable. Maybe dgw has a stronger opinion, but from my perspective, LGTM!

Exirel (approved):

dgw:
I don't mind the current history either, really. I tend to commit related code, docs, & tests all at once if I'm building an ideal history as I go, but doing that in a rebase can be error-prone so it's fine if you'd rather not go there. 😸

sopel/plugins/rules.py
```diff
@@ -732,58 +741,72 @@ def is_unblockable(self) -> bool:
...
+        self,
+        nick: Identifier,
+        at_time: Optional[datetime.datetime] = None,
+    ) -> bool:
         """Tell when the rule reached the ``nick``'s rate limit.
 
-        :return: ``True`` when the rule reached the limit, ``False`` otherwise
+        :param nick: the nick associated with this check
+        :param at_time: optional aware datetime for the rate limit check
+                        if not given, ``utcnow`` will be used
```
dgw, line 753: Uh-oh, I found a missing punctuation.

```suggestion
        :param at_time: optional aware datetime for the rate limit check;
                        if not given, ``utcnow`` will be used
```

sopel/plugins/rules.py
```diff
@@ -732,58 +741,72 @@ def is_unblockable(self) -> bool:
...
+        self,
+        channel: Identifier,
+        at_time: Optional[datetime.datetime] = None,
+    ) -> bool:
         """Tell when the rule reached the ``channel``'s rate limit.
 
-        :return: ``True`` when the rule reached the limit, ``False`` otherwise
+        :param channel: the channel associated with this check
+        :param at_time: optional aware datetime for the rate limit check
+                        if not given, ``utcnow`` will be used
```
dgw, line 767: As above.

```suggestion
        :param at_time: optional aware datetime for the rate limit check;
                        if not given, ``utcnow`` will be used
```

sopel/plugins/rules.py
```diff
@@ -732,58 +741,72 @@ def is_unblockable(self) -> bool:
...
-    def is_global_rate_limited(self) -> bool:
+    def is_global_rate_limited(
+        self,
+        at_time: Optional[datetime.datetime] = None,
+    ) -> bool:
         """Tell when the rule reached the global rate limit.
 
-        :return: ``True`` when the rule reached the limit, ``False`` otherwise
+        :param at_time: optional aware datetime for the rate limit check
+                        if not given, ``utcnow`` will be used
```
dgw, line 779:
```suggestion
        :param at_time: optional aware datetime for the rate limit check;
                        if not given, ``utcnow`` will be used
```

dgw (commented):
Doc nitpicks, what would my self-identity be without them?

dgw (approved):

PR #2442, opened 2023-04-08
bot: eliminate problematic xrefs from `SopelWrapper` docs

dgw:
### Description
The formatting, in particular, is what was problematic here. Sphinx doesn't provide a way to negate the monospace formatting of a `:meth:` reference when using a custom title, so the only way to avoid having random monospace words in the middle of this paragraph is... rewriting to eliminate the xrefs.

Which is to say, if there's a better answer to [this question](https://stackoverflow.com/q/45741326/5991) in the future, we can put this back the way it was.

### Checklist
  - No code changes; didn't run `make test` locally but `make quality` was indeed fine.

### DIFF
--- a/sopel/bot.py
+++ b/sopel/bot.py
@@ -1230,9 +1230,9 @@ class SopelWrapper:
                               (e.g. plugin tag)
 
     This wrapper will be used to call Sopel's triggered commands and rules as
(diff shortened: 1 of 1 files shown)


Exirel (approved):
Aw... I thought I had a trick to prevent that, but actually I don't. So let's go with that.

PR #2446, opened 2023-04-22
tools.time: tweaks & tests

Exirel:
### Evolved description

- **Breaking:** `validate_timezone()` raises `ValueError` even if passed `None` now, instead of returning `None`
- "Duration" named tuple returned from `seconds_to_split()` function
- Added tests to cover original & fixed behaviors

### Original description

Tin. Oh, and also I added type annotations and fixed the resulting type errors.

I ensure that the tests all pass on the current version of `sopel.tools.time`, **then** I added type annotations, so there is a guarantee that it works pretty much as it did before... except for a bug I fixed where you could call a function without a `db` and it would raise an `AttributeError` because `NoneType` doesn't have the right method.

### Checklist

### DIFF
--- a/sopel/tools/time.py
+++ b/sopel/tools/time.py
@@ -9,8 +9,13 @@
 from __future__ import annotations
 
 import datetime
+from typing import Optional, overload, Tuple, TYPE_CHECKING, Union
 
 import pytz
+
+if TYPE_CHECKING:
+    from sopel.config import Config
+    from sopel.db import SopelDB
 
 
 # various time units measured in seconds; approximated for months and years
@@ -22,10 +27,20 @@ MONTHS = int(30.5 * DAYS)
 YEARS = 365 * DAYS
 
 
+@overload
+def validate_timezone(zone: None) -> None:
+    ...
+
+
+@overload
+def validate_timezone(zone: str) -> str:
+    ...
+
+
 def validate_timezone(zone):
     """Return an IETF timezone from the given IETF zone or common abbreviation.
 
-    :param str zone: in a strict or a human-friendly format
+    :param zone: in a strict or a human-friendly format
     :return: the valid IETF timezone properly formatted
     :raise ValueError: when ``zone`` is not a valid timezone
 
@@ -40,7 +55,7 @@ def validate_timezone(zone):
     becomes ``UTC``. In the majority of user-facing interactions, such
     case-insensitivity will be expected.
 
-    If the zone is not valid, ``ValueError`` will be raised.
+    If the zone is not valid, :exc:`ValueError` will be raised.
 
     .. versionadded:: 6.0
     """
@@ -55,10 +70,10 @@ def validate_timezone(zone):
     return tz.zone
 
 
-def validate_format(tformat):
+def validate_format(tformat: str) -> str:
     """Validate a time format string.
 
-    :param str tformat: the format string to validate
+    :param tformat: the format string to validate
     :return: the format string, if valid
     :raise ValueError: when ``tformat`` is not a valid time format string
 
@@ -72,13 +87,11 @@ def validate_format(tformat):
     return tformat
 
 
-def get_nick_timezone(db, nick):
+def get_nick_timezone(db: SopelDB, nick: str) -> Optional[str]:
     """Get a nick's timezone from database.
 
     :param db: Bot's database handler (usually ``bot.db``)
-    :type db: :class:`~sopel.db.SopelDB`
     :param nick: IRC nickname
-    :type nick: :class:`~sopel.tools.identifiers.Identifier`
     :return: the timezone associated with the ``nick``
 
     If a timezone cannot be found for ``nick``, or if it is invalid, ``None``
@@ -92,13 +105,11 @@ def get_nick_timezone(db, nick):
         return None
 
 
-def get_channel_timezone(db, channel):
+def get_channel_timezone(db: SopelDB, channel: str) -> Optional[str]:
     """Get a channel's timezone from database.
 
     :param db: Bot's database handler (usually ``bot.db``)
-    :type db: :class:`~sopel.db.SopelDB`
     :param channel: IRC channel name
-    :type channel: :class:`~sopel.tools.identifiers.Identifier`
     :return: the timezone associated with the ``channel``
 
     If a timezone cannot be found for ``channel``, or if it is invalid,
@@ -112,16 +123,20 @@ def get_channel_timezone(db, channel):
         return None
 
 
-def get_timezone(db=None, config=None, zone=None, nick=None, channel=None):
+def get_timezone(
+    db: Optional[SopelDB] = None,
+    config: Optional[Config] = None,
+    zone: Optional[str] = None,
+    nick: Optional[str] = None,
+    channel: Optional[str] = None,
+) -> Optional[str]:
     """Find, and return, the appropriate timezone.
 
     :param db: bot database object (optional)
-    :type db: :class:`~.db.SopelDB`
     :param config: bot config object (optional)
-    :type config: :class:`~.config.Config`
-    :param str zone: preferred timezone name (optional)
-    :param str nick: nick whose timezone to use, if set (optional)
-    :param str channel: channel whose timezone to use, if set (optional)
+    :param zone: preferred timezone name (optional)
+    :param nick: nick whose timezone to use, if set (optional)
+    :param channel: channel whose timezone to use, if set (optional)
 
     Timezone is pulled in the following priority:
 
@@ -145,40 +160,50 @@ def get_timezone(db=None, config=None, zone=None, nick=None, channel=None):
        formatting of the timezone.
 
     """
-    def _check(zone):
+    def _check(zone: Optional[str]) -> Optional[str]:
         try:
             return validate_timezone(zone)
         except ValueError:
             return None
 
-    tz = None
+    tz: Optional[str] = None
 
     if zone:
         tz = _check(zone)
-        if not tz:
+        # zone might be a nick or a channel
+        if not tz and db is not None:
             tz = _check(db.get_nick_or_channel_value(zone, 'timezone'))
-    if not tz and nick:
-        tz = _check(db.get_nick_value(nick, 'timezone'))
-    if not tz and channel:
-        tz = _check(db.get_channel_value(channel, 'timezone'))
-    if not tz and config and config.core.default_timezone:
+
+    # get nick's timezone, and if none, get channel's timezone instead
+    if not tz and db is not None:
+        if nick:
+            tz = _check(db.get_nick_value(nick, 'timezone'))
+        if not tz and channel:
+            tz = _check(db.get_channel_value(channel, 'timezone'))
+
+    # if still not found, default to core configuration
+    if not tz and config is not None and config.core.default_timezone:
         tz = _check(config.core.default_timezone)
+
     return tz
 
 
-def format_time(db=None, config=None, zone=None, nick=None, channel=None,
-                time=None):
+def format_time(
+    db: Optional[SopelDB] = None,
+    config: Optional[Config] = None,
+    zone: Optional[str] = None,
+    nick: Optional[str] = None,
+    channel: Optional[str] = None,
+    time: Optional[datetime.datetime] = None,
+) -> str:
     """Return a formatted string of the given time in the given zone.
 
     :param db: bot database object (optional)
-    :type db: :class:`~.db.SopelDB`
     :param config: bot config object (optional)
-    :type config: :class:`~.config.Config`
-    :param str zone: name of timezone to use for output (optional)
-    :param str nick: nick whose time format to use, if set (optional)
-    :param str channel: channel whose time format to use, if set (optional)
+    :param zone: name of timezone to use for output (optional)
+    :param nick: nick whose time format to use, if set (optional)
+    :param channel: channel whose time format to use, if set (optional)
     :param time: the time value to format (optional)
-    :type time: :class:`~datetime.datetime`
 
     ``time``, if given, should be a ``datetime.datetime`` object, and will be
     treated as being in the UTC timezone if it is :ref:`naïve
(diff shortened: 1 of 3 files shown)


sopel/tools/time.py
```diff
@@ -200,38 +225,44 @@ def format_time(db=None, config=None, zone=None, nick=None, channel=None,
...
-    else:
-        zone = pytz.timezone(zone)
-
-    return time.astimezone(zone).strftime(tformat)
+    # format local time with format
+    return time.astimezone(target_tz).strftime(tformat)
 
 
-def seconds_to_split(seconds):
+def seconds_to_split(seconds: int) -> Tuple[int, int, int, int, int, int]:
```
dgw, line 260: This seems like it should have been a NamedTuple, whoops.

test/tools/test_tools_time.py
```diff
@@ -2,13 +2,39 @@
...
+@pytest.fixture
+def tmpconfig(configfactory, tmpdir):
+    content = TMP_CONFIG.format(db_filename=tmpdir.join('test.sqlite'))
+    return configfactory('default.cfg', content)
+
+
+@pytest.fixture
+def db(tmpconfig):
+    db = SopelDB(tmpconfig)
+    # TODO add tests to ensure db creation works properly, too.
```
dgw, line 34: If this applies to uses of the DB in other tests, it should be a follow-up issue, not a TODO left here (in a place where it's likely to get forgotten).

sopel/tools/time.py
```diff
@@ -40,7 +55,7 @@ def validate_timezone(zone):
     becomes ``UTC``. In the majority of user-facing interactions, such
     case-insensitivity will be expected.
 
-    If the zone is not valid, ``ValueError`` will be raised.
+    If the zone is not valid, :exc:`ValueError` will be raised.
```
dgw, line 58: Arguably, an invalid zone should raise `ValueError` too instead of returning `None`.

We can make this change in 8.0, as it's a major version bump.

Exirel: What do you mean by "an invalid zone"? Unless it's `None`, only valid zone will be returned, invalid zone will trigger a `ValueError`.

I could modify that to raise a `TypeError` if `zone is None`, if that's what you are thinking about.

sopel/tools/time.py
Exirel: I can probably do that yeah.

sopel/tools/time.py
dgw: Think I managed to leave out some of my thought process, and tag the wrong lines. 😦

Seems kinda silly to return `None` if `zone is None`, when the defined semantics for "the zone is not valid" (`None` can never be valid) are `raise ValueError`. Why is `None` as input treated differently? It means having to add (as you did in this patch) `@typing.overload` definitions to special-case the return value of `None` that never happens otherwise.

The obvious change is always either return `str` or raise `ValueError`. I think removing the overloads and using `validate_timezone(zone: Optional[str])` would be OK. Just let `validate_timezone(None)` raise `ValueError` like any other invalid zone name would.

Searching internal usage of `validate_timezone()` yielded only a few call sites, one of which has a comment saying why `None` _not_ raising an exception like any other invalid value would is inconvenient:

https://github.com/sopel-irc/sopel/blob/e5bcd1fbfbc781eee990b47c0236f0772ba13124/sopel/modules/remind.py#L429-L433

Exirel: Ha yes, this way it makes sense! I agree with you, and since you are OK with this breaking change, I'll add it.

Exirel:
Did all the change as requested, ready for another review.

sopel/tools/time.py
```diff
@@ -27,22 +27,42 @@
...
+    """Seconds spent."""
+
+
+def validate_timezone(zone: Optional[str]) -> str:
     """Return an IETF timezone from the given IETF zone or common abbreviation.
 
     :param zone: in a strict or a human-friendly format
     :return: the valid IETF timezone properly formatted
     :raise ValueError: when ``zone`` is not a valid timezone
+                       (including emptry string and ``None`` value)
```
dgw, line 65:
```suggestion
                       (including empty string and ``None`` value)
```

sopel/tools/time.py
```diff
@@ -58,16 +78,23 @@ def validate_timezone(zone):
...
+
+        If ``zone`` is ``None``, raises a :exc:`ValueError` as if it was an
+        empty string or an invalid timezone instead of returning ``None``.
+
     """
     if zone is None:
-        return None
+        raise ValueError('Invalid time zone.')
 
     zone = '/'.join(reversed(zone.split(', '))).replace(' ', '_')
```
dgw, line 91: `pytz.timezone(None)` also raises `UnknownTimeZoneError`, so what do you think of simplifying this to only define the exception type & message in one location?

```py
if zone:
    zone = '/'.join(reversed(zone.split(', '))).replace(' ', '_')

try:
    tz =  # etc.
```

"Applying suggestions on deleted lines is not supported" is such a cop-out, GH. 😑

sopel/tools/time.py
```diff
@@ -58,16 +78,23 @@ def validate_timezone(zone):
...
     try:
         tz = pytz.timezone(zone)
     except pytz.exceptions.UnknownTimeZoneError:
         raise ValueError('Invalid time zone.')
-    return tz.zone
+
+    return cast(str, tz.zone)
```
dgw, line 97: Necessary because `types-pytz` defines the `zone` property as `str | None`, I take it?

I'm sure they have a reason for not defining `zone: None` on the base class and separately defining `zone: str` on its children, but… I really don't like having to use `cast()` 😩

dgw (commented):
Couple ideas… My procrastination on $work stuff is your gain!

test/tools/test_tools_time.py
```diff
@@ -47,6 +71,96 @@ def test_validate_format_none():
...
+    assert time.get_timezone(
+        db, tmpconfig, None, nick, channel
+    ) == zone_channel, 'Channel '
+    assert time.get_timezone(
+        db, tmpconfig, 'Invalid/Tz', nick, channel
+    ) == zone_channel
+
+    # nick tz has priority
+    db.set_nick_value(nick, 'timezone', zone_nick)
+    assert time.get_timezone(db, tmpconfig, None, nick, channel) == zone_nick
```
SnoopJ, line 121: This test is pretty busy, I think it makes sense to fully explode it into its cases, following the lead you've established.

This is a little more complicated-looking, but it does better reflect the complexity of this function.

```suggestion
def test_get_timezone_no_nick_or_channel_data((db: SopelDB, tmpconfig: Config):
    zone = 'Europe/Paris'
    nick = 'IronMan'
    channel = '#test'

    assert time.get_timezone(db) is None
    assert time.get_timezone(db, tmpconfig) == 'UTC', (
        'Must use default timezone from config'
    )
    assert time.get_timezone(db, tmpconfig, zone) == zone


def test_get_timezone_only_nick_data(db: SopelDB, tmpconfig: Config):
    zone_nick = 'America/New_York'
    nick = 'IronMan'
    channel = '#test'
    
    db.set_nick_value(nick, 'timezone', zone_nick)

    assert time.get_timezone(db, tmpconfig, None, nick, channel) == zone_nick
    

def test_get_timezone_only_channel_data(db: SopelDB, tmpconfig: Config):
    zone_channel = 'Asia/Tokyo'
    nick = 'IronMan'
    channel = '#test'

    db.set_channel_value(channel, 'timezone', zone_channel)
    
    assert time.get_timezone(
        db, tmpconfig, None, nick, channel
    ) == zone_channel
    assert time.get_timezone(
        db, tmpconfig, 'Invalid/Tz', nick, channel
    ) == zone_channel,  'Expected channel timezone'


def test_get_timezone_nick_and_channel_data(db: SopelDB, tmpconfig: Config):
    zone_nick = 'America/New_York'
    zone_channel = 'Asia/Tokyo'
    nick = 'IronMan'
    channel = '#test'

    db.set_nick_value(nick, 'timezone', zone_nick)
    db.set_channel_value(channel, 'timezone', zone_channel)
    
    assert time.get_timezone(db, tmpconfig, None, nick, channel) == zone_nick, (
        'Nick timezone must take priority'
    )
```

Note: I folded the removed comments into the corresponding assertion failure messages

sopel/tools/time.py
```diff
@@ -200,68 +252,92 @@ def format_time(db=None, config=None, zone=None, nick=None, channel=None,
     If ``db`` is not given or is not set up, steps 1 and 2 are skipped. If
     ``config`` is not given, step 3 will be skipped.
     """
-    utc = pytz.timezone('UTC')
-    tformat = None
+    target_tz: datetime.tzinfo = pytz.utc
```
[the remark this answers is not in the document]
dgw: Don't know how deliberate it is, but the `zone` parameter is meant to come from something like `get_timezone()`, which already does the falling back.

sopel/tools/time.py
```diff
@@ -40,25 +75,32 @@ def validate_timezone(zone):
...
+
+    .. versionchanged:: 8.0
+
+        If ``zone`` is ``None``, raises a :exc:`ValueError` as if it was an
+        empty string or an invalid timezone instead of returning ``None``.
+
     """
     if zone is None:
-        return None
+        raise ValueError('Invalid time zone.')
```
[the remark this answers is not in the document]
dgw: The idea is not to add another exception type that must be caught. Strictly speaking, yes, `NoneType != str` and this case should be a `TypeError`, but existing code that uses this function expects anything invalid to raise `ValueError`.

Then again, we're _already_ changing the semantics by raising an exception on `None` instead of just returning `None`, so maybe there's actually no valid argument here for "keeping compatibility with old code" after all.

test/tools/test_tools_time.py
```diff
@@ -47,6 +71,96 @@ def test_validate_format_none():
...
+    ) == zone_nick, 'zone argument must have priority over other arguments'
+
+    # when zone is actually a channel
+    assert time.get_timezone(db, tmpconfig, zone=channel) == zone_channel
+    assert time.get_timezone(
+        db, tmpconfig, zone=channel, nick=nick
+    ) == zone_channel, 'zone argument must have priority over nick argument'
+    assert time.get_timezone(
+        db, tmpconfig, zone=channel, nick=nick, channel=channel
+    ) == zone_channel, 'zone argument must have priority over other arguments'
```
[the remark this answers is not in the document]
SnoopJ: @dgw points out that I was reading the doc wrong and that this _is_ documented behavior. I think that probably clinches it, getting rid of this would be too disruptive to plugins without advance notice.

So, I guess my question here is more like _"how do we feel about deprecating that behavior?"_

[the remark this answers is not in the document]
Exirel: That's what convenient functions do: they are convenient for the dev-user, not the dev-writer. 😁 

I don't mind it, it is, indeed more complicated for us, and at the same time, this is now fully covered by our tests, so it doesn't bother me that much.

sopel/tools/time.py
```diff
@@ -145,40 +187,50 @@ def get_timezone(db=None, config=None, zone=None, nick=None, channel=None):
...
+        if not tz and db is not None:
             tz = _check(db.get_nick_or_channel_value(zone, 'timezone'))
-    if not tz and nick:
-        tz = _check(db.get_nick_value(nick, 'timezone'))
-    if not tz and channel:
-        tz = _check(db.get_channel_value(channel, 'timezone'))
-    if not tz and config and config.core.default_timezone:
+
+    # get nick's timezone, and if none, get channel's timezone instead
+    if not tz and db is not None:
```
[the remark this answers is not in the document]
Exirel: From a type system standpoint, I disagree. I'm not interested in the true-ishness or false-ishness of `db`, I want to know if it's an object (of the type said it should be), or `None`. Unlike `str`, which interface is not going to change, I have no reason to trust `SopelDB` to keep its boolean behavior as it is today.

sopel/tools/time.py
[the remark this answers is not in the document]
Exirel: In theory, yes. However, this is legacy code, so changes are limited to what fit best the existing behavior.

sopel/tools/time.py
Exirel: Yes. Sometimes, you have to help `mypy` a little, especially when the library has shitty type representation.

sopel/tools/time.py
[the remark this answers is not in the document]
Exirel: I'm adding tests, not changing behavior, unless I'm forced to.

sopel/tools/time.py
```diff
@@ -112,16 +150,20 @@ def get_channel_timezone(db, channel):
         return None
 
 
-def get_timezone(db=None, config=None, zone=None, nick=None, channel=None):
+def get_timezone(
```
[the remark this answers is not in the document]
Exirel: I see two issues with that approach:

* these functions repeat the same checks/transformations of `tz`
* it would create a hard coupling between `get_timezone` and these, which I don't really want

So for now, I'll keep the existing code as it is. If it has to change, that would be for another PR.

sopel/tools/time.py
Exirel: 🤦‍♂️

test/tools/test_tools_time.py
```diff
@@ -47,6 +71,119 @@
...
+    assert time.get_channel_timezone(db, channel) is None
+
+    db.set_channel_value(channel, 'timezone', 'Europe/Paris')
+    assert time.get_channel_timezone(db, channel) == 'Europe/Paris'
+
+    db.set_channel_value(channel, 'timezone', 'Invalid_TZ')
+    assert time.get_channel_timezone(db, channel) is None
+
+
+def test_get_timezone_config_only((db: SopelDB, tmpconfig: Config):
```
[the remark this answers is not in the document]
Exirel:
```suggestion
def test_get_timezone_config_only(db: SopelDB, tmpconfig: Config):
```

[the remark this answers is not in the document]
SnoopJ: oops :grimacing:

sopel/tools/time.py
Exirel: But sadly... I can't do that, because pytz defines the `timezone` function as using an `str` and does not accept `None` as a valid value.

> sopel/tools/time.py:92: error: Argument 1 to "timezone" has incompatible type "Optional[str]"; expected "str"  [arg-type]

So that won't do!

Exirel:
I squashed the fix commit on the parent's commit, so you shouldn't have issue with changes from master!

SnoopJ (approved):
I have no more remarks for this, LGTM

sopel/tools/time.py
```diff
@@ -200,68 +252,92 @@ def format_time(db=None, config=None, zone=None, nick=None, channel=None,
...
+    minutes: int = 0,
+    seconds: int = 0,
+) -> Tuple[
+    Tuple[int, str],
+    Tuple[int, str],
+    Tuple[int, str],
+    Tuple[int, str],
+    Tuple[int, str],
+    Tuple[int, str],
+]:
```
dgw, line 331: Monster return value type-hint there, but is it worth creating another custom type like `Duration`? Survey says probably not.

dgw (approved):
I'll make CliffsNotes on behavior changes to mention in the changelog by editing the OP.

PR #2458, opened 2023-05-31
test: cover obvious SASL PLAIN edge cases

dgw:
The major edge cases handled in Sopel's SASL PLAIN code are:

- Unexpected continuation message from the server (`AUTHENTICATE PLAIN` ack'd by something other than `AUTHENTICATE +`)
- Very long password that b64encodes to exactly 400 bytes
- Very long password that b64encodes to over 400 bytes

IRCv3 recently merging [a patch to add an example](https://github.com/ircv3/ircv3-specifications/pull/521) of split token negotiation inspired me to cover that in our behavior tests.

Along the way I fixed a couple typos, and realized it would also be pretty easy to validate Sopel aborting SASL PLAIN on a non-empty message from the server.

### Checklist
  - This change is literally tests, so it tested itself 😁

### Notes
Big thanks to @Exirel for writing tests of the main SASL PLAIN codepath, which I copied and adapted to cover these edge cases.

### DIFF
--- a/test/coretasks/test_coretasks_sasl.py
+++ b/test/coretasks/test_coretasks_sasl.py
@@ -31,6 +31,26 @@ auth_password = secret
 """
 
 
+TMP_CONFIG_SASL_DEFAULT_EXACTLY_400 = """
+[core]
+owner = Uowner
+nick = TestBot
+enable = coretasks
+auth_method = sasl
+auth_password = {}
+""".format('a' * 282)
+
+
+TMP_CONFIG_SASL_DEFAULT_OVER_400 = """
+[core]
+owner = Uowner
+nick = TestBot
+enable = coretasks
+auth_method = sasl
+auth_password = {}
+""".format('a' * 286)
+
+
 TMP_CONFIG_SASL_NO_PASSWORD = """
 [core]
 owner = Uowner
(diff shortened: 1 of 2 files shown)


test/coretasks/test_coretasks_sasl.py
```diff
@@ -179,6 +293,33 @@ def test_sasl_plain_not_supported(botfactory: BotFactory, tmpconfig) -> None:
...
+        'AUTHENTICATE PLAIN',
+    )
+
+    # server acknowledges PLAIN auth request in an unexpected way
+    n = len(mockbot.backend.message_sent)
+    mockbot.on_message(':irc.example.com AUTHENTICATE VGVzdEJvdABUZXN0Qm90AG')
+    assert mockbot.backend.message_sent[n:] == rawlist(
+        'AUTHENTICATE *',
+    ), ('Bot must abort SASL PLAIN auth if server reply to starting '
+        'AUTHENTICATE PLAIN is not as expected per SASL spec')
```
Exirel, line 320: What is the expected behavior from the server here? Because I'd make sure here we can send CAP END after that cancelation.

Exirel (commented):
It looks good to me however I've this nitpick about authenticate cancelation.

test/coretasks/test_coretasks_sasl.py
dgw: > The client can abort an authentication by sending an asterisk as the data. The server will send a 906 numeric.

906 (`ERR_SASLABORTED`) is handled by `sasl_fail()`. Sopel will log the failure, then quit with a "SASL Auth Failed" message.

Exirel (changes requested):
Aside from adding the ERR_SASLABORTED to check that we send the CAP END, it's all good.

test/coretasks/test_coretasks_sasl.py
dgw: Have done so!

Exirel (approved):
Perfect, let's go!

PR #2462, opened 2023-06-04
mypy: fix currently known type-checking errors

dgw:
### Description
This resolves all known errors in `mypy sopel` output, as of commit time. There are few enough other outstanding PRs, I don't think introducing new errors after this is merged will be much of an issue.

(It would help if I started checking `make mypy` before merging things, to avoid introducing new errors after fixing this set of them… I'll try to remember that step.)

### Checklist

### DIFF
--- a/sopel/plugins/rules.py
+++ b/sopel/plugins/rules.py
@@ -741,6 +741,33 @@ class AbstractRule(abc.ABC):
         """
 
     @abc.abstractmethod
+    def get_user_metrics(self, nick: Identifier) -> RuleMetrics:
+        """Get the rule's usage metrics for the given user."""
+
+    @abc.abstractmethod
+    def get_channel_metrics(self, channel: Identifier) -> RuleMetrics:
+        """Get the rule's usage metrics for the given channel."""
+
+    @abc.abstractmethod
+    def get_global_metrics(self) -> RuleMetrics:
+        """Get the rule's global usage metrics."""
+
+    @property
+    @abc.abstractmethod
+    def user_rate_limit(self) -> datetime.timedelta:
+        """The rule's user rate limit."""
+
+    @property
+    @abc.abstractmethod
+    def channel_rate_limit(self) -> datetime.timedelta:
+        """The rule's channel rate limit."""
+
+    @property
+    @abc.abstractmethod
+    def global_rate_limit(self) -> datetime.timedelta:
+        """The rule's global rate limit."""
+
+    @abc.abstractmethod
     def is_user_rate_limited(
         self,
         nick: Identifier,
@@ -799,7 +826,7 @@ class AbstractRule(abc.ABC):
         :return: A formatted string, or ``None`` if no message is set.
 
         This method is called by the bot when a trigger hits the channel rate
-        limit (i.e. for the specificed ``channel``).
+        limit (i.e. for the specified ``channel``).
         """
 
     @property
--- a/sopel/plugins/rules.py
+++ b/sopel/plugins/rules.py
@@ -481,11 +481,10 @@ class RuleMetrics:
     def last_time(self) -> Optional[datetime.datetime]:
         """Last recorded start/end time for the associated rule"""
         # detect if we just started something or if it ended
-        last_time = self.started_at
-        if self.ended_at and (self.started_at and self.started_at < self.ended_at):
-            last_time = self.ended_at
+        if (self.started_at and self.ended_at) and (self.started_at < self.ended_at):
+            return self.ended_at
 
-        return last_time
+        return self.started_at
 
     def is_limited(
         self,
--- a/sopel/plugins/rules.py
+++ b/sopel/plugins/rules.py
@@ -482,7 +482,7 @@ class RuleMetrics:
         """Last recorded start/end time for the associated rule"""
         # detect if we just started something or if it ended
         last_time = self.started_at
-        if self.ended_at and self.started_at < self.ended_at:
+        if self.ended_at and (self.started_at and self.started_at < self.ended_at):
             last_time = self.ended_at
 
         return last_time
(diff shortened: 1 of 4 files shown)


SnoopJ:
I'm seeing an additional failure on 3.7.16, looks like `mypy` surfaced a subtle bug in that version's error handling:

```
00:12 [snoopjedi@denton ~/repos/sopel-src (mypy-error-cleanup)]
$ python3 -V && mypy -V && make mypy
Python 3.7.16
mypy 0.991 (compiled: yes)
mypy sopel
sopel/plugins/capabilities.py:64: note: By default the bodies of untyped functions are not checked, consider using --check-untyped-defs  [annotation-unchecked]
sopel/plugins/capabilities.py:70: note: By default the bodies of untyped functions are not checked, consider using --check-untyped-defs  [annotation-unchecked]
sopel/plugins/capabilities.py:71: note: By default the bodies of untyped functions are not checked, consider using --check-untyped-defs  [annotation-unchecked]
sopel/plugins/capabilities.py:72: note: By default the bodies of untyped functions are not checked, consider using --check-untyped-defs  [annotation-unchecked]
sopel/irc/backends.py:316: error: Module has no attribute "exceptions"  [attr-defined]
sopel/irc/backends.py:319: error: Module has no attribute "exceptions"  [attr-defined]
sopel/plugins/rules.py:1065: note: By default the bodies of untyped functions are not checked, consider using --check-untyped-defs  [annotation-unchecked]
sopel/plugins/rules.py:1066: note: By default the bodies of untyped functions are not checked, consider using --check-untyped-defs  [annotation-unchecked]
sopel/plugins/rules.py:1263: note: By default the bodies of untyped functions are not checked, consider using --check-untyped-defs  [annotation-unchecked]
sopel/plugins/rules.py:1265: note: By default the bodies of untyped functions are not checked, consider using --check-untyped-defs  [annotation-unchecked]
sopel/bot.py:59: note: By default the bodies of untyped functions are not checked, consider using --check-untyped-defs  [annotation-unchecked]
Found 2 errors in 1 file (checked 86 source files)
make: *** [Makefile:9: mypy] Error 1
```

Looks like between 3.7 and 3.8, these exceptions were [moved into a separate submodule](https://github.com/python/cpython/commit/0baa72f4b2e7185298d09cf64c7b591efcd22af0) but are still available at the module root. If we change `asyncio.exceptions.*` to `asyncio.*` on the failing lines, this resolves the issue.

I do like the `asyncio.execeptions` spelling better, but adding ad-hoc version comparison code doesn't seem worth it here, especially as 3.7 approaches its EOL and a presumptive dropping of it from Sopel's support window.

<details><summary>clicky click for diff</summary>

```diff
diff --git a/sopel/irc/backends.py b/sopel/irc/backends.py
index 42a6781f..2a44c9cb 100644
--- a/sopel/irc/backends.py
+++ b/sopel/irc/backends.py
@@ -313,10 +313,10 @@ class AsyncioBackend(AbstractIRCBackend):
         while not self._reader.at_eof():
             try:
                 line: bytes = await self._reader.readuntil(separator=b'\r\n')
-            except asyncio.exceptions.IncompleteReadError as err:
+            except asyncio.IncompleteReadError as err:
                 LOGGER.warning('Receiving partial message from IRC.')
                 line = err.partial
-            except asyncio.exceptions.LimitOverrunError:
+            except asyncio.LimitOverrunError:
                 LOGGER.exception('Unable to read from IRC server.')
                 break
```

</details> 

Modulo this one wrinkle, this passes type-checking on all versions. Nice one!

sopel/plugins/rules.py
```diff
@@ -482,7 +482,7 @@ def last_time(self) -> Optional[datetime.datetime]:
         """Last recorded start/end time for the associated rule"""
         # detect if we just started something or if it ended
         last_time = self.started_at
-        if self.ended_at and self.started_at < self.ended_at:
+        if self.ended_at and (self.started_at and self.started_at < self.ended_at):
```
SnoopJ, line 485: Blegh, I totally understand this fix in context of the PR, but I don't like the look of it and feel like it would trip me up if I didn't have that context in my head while reading it.

After staring at it for a few minutes, the best alternative spelling I can think of is also kind of dissatisfying because it involves repetition:

```python
if not self.ended_at or not self.started_at:
    last_time = self.started_at
# else both are not-None, mypy narrows that properly
elif self.started_at < self.ended_at:
    last_time = self.ended_at
else:
    last_time = self.started_at    
```

dgw:
Even in 3.12 alpha docs, there's no hint of those exceptions being removed from the `asyncio` namespace, so I added a commit to fix it w/SnoopJ as co-author.

Looks like @SnoopJ's efforts to set up `tox` locally went well. Incoming `contrib/` PR? 😁

SnoopJ (approved):
:clipboard: :heavy_check_mark:

sopel/plugins/rules.py
dgw: It's the fix I'm least happy about in this patch, to be sure. Thinking I'll leave it for a bit and see if @Exirel weighs in with a stronger preference, or a third alternative. (Or maybe I'll have another idea? Who knows!)

Exirel: Any version will come up with will need to ensure that `self.started_at` is not `None` anyway.

Here is another version:

```
if (self.started_at and self.ended_at) and (self.started_at < self.ended_at):
        return self.ended_at

return self.started_at
```

And another one:

```
times = sorted(t for t in [self.started_at, self.ended_at] if t)
return times[-1] if times else None
```

Exirel: (but no, no strong preference here, your version is perfectly fine by me and does exactly what it should, maybe we could get rid of the `last_time` intermediate variable tho)

dgw: > Here is another version:
> 
> ```python
> if (self.started_at and self.ended_at) and (self.started_at < self.ended_at):
>         return self.ended_at
> 
> return self.started_at
> ```

I think this version is the best balance of brevity + clarity, and as a bonus it eliminates the local variable. 🚀

SnoopJ: Agreed, I like that version

SnoopJ (approved):
No further remarks, nice one!

PR #2464, opened 2023-06-06
tools.calculation: replace deprecated `ast.Num` with `ast.Constant`

dgw:
`ast.Num` and friends have been deprecated since Python 3.8, but only started to generate warnings in Python 3.12 prereleases.

Feels like stdlib made a bad trade here, simplifying Python internals in exchange for making users write uglier, more complicated type checks such as this. But it's been so long, it's not as if they'll un-deprecate the older, more-convenient-for-the-user classes.

### Compatibility
I was under the impression that `ast.Constant` would work as far back as Python 3.6, but apparently not. This change has to wait until we're ready to drop Python 3.7.

### Checklist

### DIFF
--- a/sopel/tools/calculation.py
+++ b/sopel/tools/calculation.py
@@ -57,8 +57,11 @@ class ExpressionEvaluator:
         A subclass could overwrite this to handle more nodes, calling it only
         for nodes it does not implement itself.
         """
-        if isinstance(node, ast.Num):
-            return node.n
+        if (
+            isinstance(node, ast.Constant) and
+            isinstance(node.value, (int, float))
+        ):
+            return node.value
 
         elif (isinstance(node, ast.BinOp) and
                 type(node.op) in self.binary_ops):
@@ -78,7 +81,7 @@ class ExpressionEvaluator:
             return self.unary_ops[type(node.op)](operand)
 
         raise ExpressionEvaluator.Error(
-            "Ast.Node '%s' not implemented." % (type(node).__name__,))
+            "ast.Node '%s' not implemented." % (type(node).__name__,))
 
 
 def guarded_mul(left, right):
--- a/sopel/tools/calculation.py
+++ b/sopel/tools/calculation.py
@@ -57,14 +57,24 @@ class ExpressionEvaluator:
         A subclass could overwrite this to handle more nodes, calling it only
         for nodes it does not implement itself.
         """
-        if (
-            isinstance(node, ast.Constant) and
-            isinstance(node.value, (int, float))
-        ):
+        if isinstance(node, ast.Constant):
+            if not isinstance(node.value, (int, float)):
+                raise ExpressionEvaluator.Error(
+                    "'{}' values are not supported".format(
+                        type(node.value).__name__,
+                    )
+                )
+
             return node.value
 
-        elif (isinstance(node, ast.BinOp) and
-                type(node.op) in self.binary_ops):
+        elif isinstance(node, ast.BinOp):
+            if type(node.op) not in self.binary_ops:
+                raise ExpressionEvaluator.Error(
+                    "Unsupported binary operator '{}'".format(
+                        type(node.op).__name__,
+                    )
+                )
+
             left = self._eval_node(node.left, timeout)
             right = self._eval_node(node.right, timeout)
             if time.time() > timeout:
@@ -72,8 +82,14 @@ class ExpressionEvaluator:
                     "Time for evaluating expression ran out.")
             return self.binary_ops[type(node.op)](left, right)
 
-        elif (isinstance(node, ast.UnaryOp) and
-                type(node.op) in self.unary_ops):
+        elif isinstance(node, ast.UnaryOp):
+            if type(node.op) not in self.unary_ops:
+                raise ExpressionEvaluator.Error(
+                    "Unsupported unary operator '{}'".format(
+                        type(node.op).__name__,
+                    )
+                )
+
             operand = self._eval_node(node.operand, timeout)
             if time.time() > timeout:
                 raise ExpressionEvaluator.Error(
@@ -81,7 +97,10 @@ class ExpressionEvaluator:
             return self.unary_ops[type(node.op)](operand)
 
         raise ExpressionEvaluator.Error(
-            "ast.Node '%s' not implemented." % (type(node).__name__,))
+            "Node type '{}' is not supported.".format(
+                type(node).__name__,
+            )
+        )
 
 
 def guarded_mul(left, right):


sopel/tools/calculation.py
```diff
@@ -57,8 +57,11 @@ def _eval_node(self, node, timeout):
         A subclass could overwrite this to handle more nodes, calling it only
         for nodes it does not implement itself.
         """
-        if isinstance(node, ast.Num):
-            return node.n
+        if (
+            isinstance(node, ast.Constant) and
+            isinstance(node.value, (int, float))
```
SnoopJ, line 62: I think we need to handle `complex` here as well?

```suggestion
            isinstance(node.value, (int, float, complex))
```

sopel/tools/calculation.py
```diff
@@ -78,7 +81,7 @@ def _eval_node(self, node, timeout):
             return self.unary_ops[type(node.op)](operand)
 
         raise ExpressionEvaluator.Error(
-            "Ast.Node '%s' not implemented." % (type(node).__name__,))
+            "ast.Node '%s' not implemented." % (type(node).__name__,))
```
SnoopJ, line 84: nit: there is no `ast.Node`, it would be marginally better here (imo) to drop the expression-y spelling

```suggestion
            "AST node type '%s' not implemented." % (type(node).__name__,))
```

SnoopJ (changes requested):

sopel/tools/calculation.py
dgw: Will leave this open, and apply after we've dropped Python 3.7. We already know that GH will need a kick in the pants to create a fresh merge after upstream changes, and this will be the perfect kick to give it. 😁

dgw: Everything being an `ast.Constant` now anyway, it probably needs to use something other than `type(node)`. Maybe:

```suggestion
            "'%s' values are not supported." % (type(node.v).__name__,))
```

sopel/tools/calculation.py
dgw: Ha, I had that in an earlier revision. Then I realized that `complex` being handled might be an accident, and took it out.

What do you think of the performance implications if we do leave `complex` as one of the supported types? Given the whole point of this expression evaluator class is preventing `.calc` input from pegging a whole CPU thread for more than a couple seconds.

sopel/tools/calculation.py
SnoopJ: Mmm, I agree with the generalization of the message, but I don't think we can rely on that , because not every node that might fail here is a `Constant`, and not all node types will have an attribute of this sort that we can use.

Maybe we want to split this into two cases? One that looks like what you drafted (but with `.value` rather than `.v`) for `Constant` nodes that gives a more specific report, and one for everything else that looks more like the existing report.

I.e. something like:

```python
        if isinstance(node, ast.Constant):
            msg = "'%s' values are not supported." % (type(node.value).__name__,)
        else:
            msg = "Node of type '%s' is not supported" % (type(node).__name__,)
        raise ExpressionEvaluator.Error(
            "ast.Node '%s' not implemented." % (type(node).__name__,))
```

Because I am in _"annoying nit/edge-case"_ mode, I think even this is suboptimal, because there are some narrower types of `BinOp, UnaryOp` that might not be supported, too? That one is probably easier to ignore, but maybe we want to push the narrowing `isinstance()` checks into the body of the relevant clauses, raise specific errors when those narrowing checks fail, and reword the catch-all case to be a suitably generic _"user's expression is on fire in a way we didn't anticipate"_ report. I'll make a separate suggestion for this quibble.

sopel/tools/calculation.py
SnoopJ: > What do you think of the performance implications if we do leave `complex` as one of the supported types?

"Measure it" is usually what I think about perf :stuck_out_tongue_winking_eye:

But the `isinstance` check should be pretty snappy, I don't think including one more type would be a problem.

dgw: My worry was about the performance of operations that _use_ `complex` values, but it doesn't seem to be an issue. `timeit` tests showed about 4 ns per loop for `3 * 3` and 19 ns per loop for `3 * c` (using `--setup 'c = complex(3, 5)` to exclude the object initialization from timing measurements). Doesn't seem like a big enough difference to specifically disallow the use of `complex` values.

dgw: Ooh, however, the real reason for not supporting `complex` is that `guarded_mul()` and `guarded_pow()` aren't written to support it. They explicitly accept only `int` or `float` values.

sopel/tools/calculation.py
dgw: Adapted this into an add-on commit w/co-author.

dgw:
Well, there are no direct tests of this module… Even though CI passes, I don't really trust it that much.

I've started work on a `test/tools/test_tools_calculation.py` suite. But that's out of scope of this PR, and doesn't need to be done for 8.0 specifically, so I will mark this as ready for review and just show up with the test suite later when I'm done with it.

SnoopJ (approved):
LGTM. Not having tests is I think no great sin, since that's not a change to the state of the module prior to this PR. Nice to have it covered, but doesn't need to be another thing weighing down 8.0.

Ship it!

Exirel (approved):
One more thing to add to the list of stuff to add tests for I guess.

At some point, we might also consider deprecating this tool, and extract the plugin that needs that code.

SnoopJ:
Had a go at writing a basic set of tests for `tools.calculation` if you'd like to add them to this PR. Also wrote some up for the `calc, dice` plugins, I'll open a PR separately for those though.

<details><summary>click for diff</summary>

```diff
diff --git a/test/tools/test_tools_calculation.py b/test/tools/test_tools_calculation.py
new file mode 100644
index 00000000..37180306
--- /dev/null
+++ b/test/tools/test_tools_calculation.py
@@ -0,0 +1,33 @@
+import ast
+import operator
+
+import pytest
+
+from sopel.tools.calculation import EquationEvaluator, ExpressionEvaluator
+
+
+def test_expression_eval():
+    OPS = {
+        ast.Add: operator.add,
+        ast.Sub: operator.sub,
+    }
+    evaluator = ExpressionEvaluator(bin_ops=OPS)
+
+    assert evaluator("1 + 1") == 2
+    assert evaluator("43 - 1") == 42
+    assert evaluator("1 + 1 - 2") == 0
+
+    with pytest.raises(ExpressionEvaluator.Error):
+        evaluator("2 * 2")
+
+
+def test_equation_eval():
+    evaluator = EquationEvaluator()
+
+    assert evaluator("1 + 1") == 2
+    assert evaluator("43 - 1") == 42
+    assert evaluator("(((1 + 1 + 2) * 3 / 5) ** 8 - 13) // 21 % 35") == 16.0
+    assert evaluator("-42") == -42
+    assert evaluator("-(-42)") == 42
+    assert evaluator("+42") == 42
+    assert evaluator("3 ^ 1") == 2
\ No newline at end of file
```

</details>

dgw:
> Had a go at writing a basic set of tests for `tools.calculation` if you'd like to add them to this PR.

I've [adopted them into my WIP branch](https://github.com/sopel-irc/sopel/commit/73202fb15669f39d75966c27d8c41dc7ea87da25) for adding tests to this module. This PR is already double-approved, and I'd rather not kibosh that status. 😁 But I appreciate that your quick patch gets the module to 81% coverage already! I'll have a look at the helpers and error cases at my leisure, probably for submission to the 8.1 cycle (coincidence?).

Issue #2457, opened 2023-05-30
Official support for Python 3.12


SnoopJ:
### Requested Feature

Python 3.12 has [entered beta](https://www.python.org/downloads/release/python-3120b1/), so it's a good time to start thinking about Sopel on that version.

* [ ] Check Sopel tests against 3.12
  * Installing `dev-requirements.txt` is currently blocked by [lack of support](https://github.com/python-greenlet/greenlet/issues/357) in `greenlet`, but there's a PR open for it. I'm able to install from that PR and proceed with the rest of testing
  * Looks like we may need to adjust the version requirement for `setuptools`, I see the following error with `setuptools == 67.8.0` but pinning `setuptools < 67` in `build-system.requires` resolves it.  
  ```
  AttributeError: module 'pkgutil' has no attribute 'ImpImporter'. Did you mean: 'zipimporter'?
  ```
  * There is additional upstream breakage [in `aiohttp`](https://github.com/aio-libs/aiohttp/issues/7229) and [in `frozenlist`](https://github.com/aio-libs/frozenlist/issues/433), seems that the latter has been fixed on development, but there is no PR for `aiohttp` yet. These are both induced by the dependency on `geoip`.
    * This is the point where I gave up trying to run the tests, for the time being :stuck_out_tongue: 
* [ ] Update CI to run with 3.12
  * Looks like the `python-versions` Action is [capable of providing the 3.12.0b1 pre-release](https://github.com/actions/python-versions/blob/7693cef3461cf6902845be55593a2328fef3bea3/versions-manifest.json#L3), so we could run against the beta version(s) before the official release ([scheduled](https://peps.python.org/pep-0693/) for 2 October 2023).
* [ ] Update classifiers list in `pyproject.toml`

Exirel:
Just a note here: there are feature we use in Py3.7 that are deprecated in Py3.9, Py3.10, or Py3.11, and may be removed in version as early as Py3.10 and Py3.11. So I'm pretty sure it'll be quite hard to support such a big range of Python version.

I believe it will be more reasonable to drop anything below Py3.9 to support Py3.12.

SnoopJ:
3.7 reaches its EOL in the next month, so that one is basically free. If we can continue to support 3.8, that would be nice to our downstream, although I don't know what similar issues there might be for that version.

SnoopJ:
[...]

PR #2468, opened 2023-06-08
Replace uses of deprecated `datetime.utcnow()`

dgw:
### Description
As @SnoopJ mentioned on IRC right after opening #2457, `datetime.utcnow()` is deprecated in Python 3.12 (and also `datetime.utcfromtimestamp()`). The recommended replacements for those items have been usable for most of Python 3's history—though we don't get to use`datetime.UTC` instead of `datetime.timezone.utc` until Python 3.11 is our minimum version.

### Checklist

### Notes
As often happens when I poke at stuff, there are some minor documentation corrections lumped in here that I didn't think warranted (a) separate PR(s).

I wanted to go further with this and get rid of `pytz`, but [`zoneinfo`](https://docs.python.org/3.9/library/zoneinfo.html) isn't available until Python 3.9. (There is a [`backports.zoneinfo`](https://pypi.org/project/backports.zoneinfo/) package, but it hasn't been updated in a while and its issue tracker is starting to fill with reports of errors & warnings.)

### DIFF
--- a/docs/source/plugin/time.rst
+++ b/docs/source/plugin/time.rst
@@ -14,8 +14,6 @@ Here is a full example of that, adapted from the built-in ``.t`` command::
 
     import datetime
 
-    import pytz
-
     from sopel import plugin
     from sopel.tools.time import format_time, get_timezone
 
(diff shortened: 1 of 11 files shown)


Exirel (approved):
Nice.

Issue #2461, opened 2023-06-04
Rollout plan for `mypy` checks as part of general workflow


dgw:
Some `mypy` errors that made it into `master`, which I've been working on this week (patches that became #2462), served as a good demonstration to me that it's probably time to start running type checks as part of CI and `make qa`.

Assuming I haven't forgotten anything, we'll want a PR that rolls out these changes:

- Add `mypy` target to `make qa` sequence
- Update pull request template checklist to mention `mypy` in addition to `quality` and `test` targets
- Add Actions step to run `mypy` at the correct point in CI (after `flake8`?)

Care will have to be taken with previously-approved PRs, but any that receive changes after merging the updated CI workflow will have any new errors flagged.

----

_Future idea: Add a Makefile target to run the full GHA suite, using something like [nektos/act](https://github.com/nektos/act)_
_Not opening its own issue just yet because it's a **whole** kettle of fish to set up `act`, which itself requires Docker, which has its own system requirements—then select the correct images to actually behave like a GHA runner, and those images are **not** small._

dgw:
I've started working on the errors left behind now that #2462 is merged, by checking the output of `mypy --check-untyped-defs sopel`. That's in a `mypy-untyped-defs` branch, and I think if we're going to start enforcement we might could wait until the rest of the glaring issues are solved. (But if someone else wants to press ahead with CI enforcement sans flag, sure, go ahead.)

dgw:
Oh, another reason to delay enforcement a little longer is that there's at least one error in `sopel/__init__.py` that requires a fix to mypy that hasn't been included in a release yet. It was merged a week before mypy 1.3, but wasn't included in that tag; it's been pulled into the upstream `release-1.4` branch, though.

dgw:
`mypy` is up to v1.5.0 on PyPI now, so the fix we were waiting for should be present going forward. And we already have verification that CI passes with `mypy` type-checking as part of the process thanks to #2502.

PR #2470, opened 2023-06-12
remind: fix import streamlining gone haywire

dgw:
"Unexpected AttributeError: 'str' object has no attribute 'utc'" begone.

Fixes a bug I introduced in #2468 by not being careful enough with my imports. Code review or not, I should've caught it. (And I was reminded to finish/open this PR by `mypy --check-untyped-defs` flagging these errors, so that's all the more reason to work toward #2461 in the longer term.)

### Checklist

### Notes
`mypy --check-untyped-defs` still reports errors in `sopel/modules/remind.py` after this patch, but they'll be fixed separately. @SnoopJ and I have been chipping away at that list over the weekend.

### DIFF
--- a/sopel/modules/remind.py
+++ b/sopel/modules/remind.py
@@ -9,7 +9,7 @@ https://sopel.chat
 from __future__ import annotations
 
 import collections
-from datetime import datetime, timezone
+import datetime
 import io  # don't use `codecs` for loading the DB; it will split lines on some IRC formatting
 import logging
 import os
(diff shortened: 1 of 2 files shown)


SnoopJ (approved):
Changes look fine to me, although this would also be a good opportunity to add a test that would have caught the bug

dgw:
> this would also be a good opportunity to add a test that would have caught the bug

Went for the minimum-necessary-change approach because:

1. As discovered shortly after, `mypy --check-untyped-defs` (the new default behavior of `make mypy`, soon) flags this problem
2. More importantly, the not-so-long-term plan is to replace this built-in plugin with an external [sopel-remind](https://github.com/sopel-irc/sopel-remind) package that already exists and has much better test coverage

PR #2473, opened 2023-06-15
bot: don't add duplicate timestamp into error log line

dgw:
If a Trigger causes an error in a plugin, we already know when the error happened because the logger call is timestamped. The error as reported to IRC also is timestamped, and `datetime.now()` is not useful data.

(It is, however, confusing data. Considered replacing it with `trigger.time` but decided against that, since `server-time` means those could be in the past/future, making the timestamp embedded in the log message even *more* confusing instead of less.)

### Checklist

### DIFF
--- a/sopel/bot.py
+++ b/sopel/bot.py
@@ -8,7 +8,6 @@
 from __future__ import annotations
 
 from ast import literal_eval
-from datetime import datetime, timezone
 import inspect
 import itertools
 import logging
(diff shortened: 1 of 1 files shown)


Exirel (approved):

SnoopJ (approved):
Agreed, emitting this data on IRC is mostly noise, and we're not losing anything since it's in the logs anyway.

Issue #2401, opened 2023-01-18
ValueError: invalid file open mode 'U' on Python 3.11


mock:
### Description

I'm logging this here for reference per the advice given by those in the IRC support channel.

After upgrading my OS, my default Python is now 3.11.1. When I tried to start up my bots, I received multiple errors for the my custom modules. The error manifested as:

```
[2023-01-18 14:24:24,238] sopel.bot            ERROR    - Error loading weather: invalid file open mode 'U'
Traceback (most recent call last):
  File "/home/mock/.local/lib/python3.11/site-packages/sopel/bot.py", line 323, in setup_plugins
    plugin.load()
  File "/home/mock/.local/lib/python3.11/site-packages/sopel/plugins/handlers.py", line 471, in load
    self._module = self._load()
                   ^^^^^^^^^^^^
  File "/home/mock/.local/lib/python3.11/site-packages/sopel/plugins/handlers.py", line 436, in _load
    mod = imp.load_module(self.name, mod, self.path, description)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.11/imp.py", line 230, in load_module
    raise ValueError('invalid file open mode {!r}'.format(mode))
ValueError: invalid file open mode 'U'
```

The full error file is included below for a complete trace.

It was suggested for the time being to pull a copy of the **master** branch and then make any adjustments to the code to avoid the `'U'` mode. (The line causing the problems was located here: https://github.com/sopel-irc/sopel/blob/v7.1.9/sopel/plugins/handlers.py#L435.)

### Reproduction steps

1. Upgrade/Install Fedora 37
2. pip install --upgrade sopel
3. sopel start -c bot.cfg -d

### Expected behavior

Loading of my custom modules as with Python 3.10.x. Custom modules then will listen for triggers.

### Relevant logs

```shell
[2023-01-18 09:16:02,039] sopel.irc            ERROR    - Fatal error in core, please review exceptions log.
[2023-01-18 09:16:01,492] sopel.irc.backends   ERROR    - Server timeout detected after 550.133971s; closing.
[2023-01-18 09:16:23,122] sopel.bot            ERROR    - Cannot register URL callback: Bugzilla URL callback requires "bugzilla.domains" to be configured; check your config file.
[...]

PR #2474, opened 2023-06-15
Cherry-pick: Drop `imp` usage on 7.1.x branch

SnoopJ:
### Description

This PR cherry-picks 596adc4 to the `7.1.x` branch, fixing crash on startup with Python 3.11 as reported by #2401. I think this is the majority of what we'd need in order to do an interim release, if we decide to do one before `8.0` is out.

**As-is, this changeset breaks compatibility with Python 2.7**

### Checklist
  - There are errors in some of the tests for `sopel/modules`, but they are also present on the current `HEAD` of the branch.
    ```1137 passed, 8 xfailed, 10 errors in 31.04s```
  - I'm able to start the bot, I didn't take it for much of a test-drive beyond that, though.

### DIFF
--- a/sopel/plugins/handlers.py
+++ b/sopel/plugins/handlers.py
@@ -44,21 +44,17 @@ away from the rest of the application.
 # Licensed under the Eiffel Forum License 2.
 from __future__ import absolute_import, division, print_function, unicode_literals
 
-import imp
+import abc
 import importlib
+import importlib.util
 import inspect
 import itertools
 import os
+import sys
+from typing import Optional
 
 from sopel import loader
 from . import exceptions
-
-try:
-    reload = importlib.reload
-except AttributeError:
-    # py2: no reload function
-    # TODO: imp is deprecated, to be removed when py2 support is dropped
-    reload = imp.reload
 
 
 class AbstractPluginHandler(object):
@@ -301,7 +297,7 @@ class PyModulePlugin(AbstractPluginHandler):
 
         This method assumes the plugin is already loaded.
         """
-        self._module = reload(self._module)
+        self._module = importlib.reload(self._module)
 
     def is_loaded(self):
         return self._module is not None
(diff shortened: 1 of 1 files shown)


dgw:
Well I don't really have the time (or spoons) to look at this moment, but we _should_ still have CI running on Travis for this branch… Where are the checks, GitHub?

Logging into Travis, it doesn't even seem to recognize that I'm part of the sopel-irc organization, and I get different errors (out of credits, too many users, etc.) depending on which page I'm on. I'm glad we left them behind. 😑

dgw:
Also per IRC discussion, it sounds like this isn't a viable change for 7.1.x, due to breaking older Pythons. (Unless I misunderstood the comments there.)

SnoopJ:
I agree, I don't think it's worth doing more than a cherry-pick for an interim release, and this looks like it would need more labor to support 2.7.

I think our time is better spent either bumping things from the 8.0 milestone or merging them, but I figured opening a cherry-pick PR was a good way to get started on poking the hornet's nest :grin:

dgw:
> and this looks like it would need more labor to support 2.7.

That's the biggest reason for not bothering to finish this off, plus the inability to run our old CI pipeline against the deprecated Python versions. We should just tidy up the last few loose ends for 8.0.0 and :shipit:, dropping all Python<3.8 along the way.

ekohl:
For what it's worth: `imp` has been removed in Python 3.12 so it's currently not possible to use a released version of sopel with Python 3.12. [8.0.0](https://github.com/sopel-irc/sopel/milestone/15) has a target date in 2025, so what's the plan in the mean time?

dgw:
8.0.0's 2025 target date is the result of an in-joke amongst the project's IRC regulars; it's _actually_ coming out as soon as we finish compiling the changelog, migration guide, and other release miscellanea.

PR #2477, opened 2023-06-17
meetbot: remove in favor of external package

dgw:
### Description
Tin. Why?

```irc
<+xrl> I'm reviewing the type thing PR. I think we should remove the meetbot plugin.
<+xrl> Reason: I've never seen anyone using it, it's not well tested, and I'm not even sure it actually works.
```

Draft for now, because:

- Merging #2471 (the aforementioned "type thing PR") when it's finished will create a conflict
- I'll want to update the extracted plugin file at https://github.com/sopel-irc/sopel-meetbot with the PR's changes to it, anyway
- We probably shouldn't _finalize_ this removal until a package is published, and I haven't converted the new plugin repo to a package yet (because I'm waiting until the first two points are taken care of)

This change leaves the references to `meetbot` in documentation for how to use the plugin-related settings in the bot's `core` config section, as they're just examples. No real reason to change them.

### Checklist

### Notes
As usual I will add a review request to this PR when it's actually ready. I just realized tonight that I'm likely to forget the branch exists if I don't tie it to an open pull request. 😅


dgw:
If anyone wants to test-run the package, it's in process at sopel-irc/sopel-meetbot#1

SnoopJ (approved):
The perfect PR doesn't exi-

Easy to approve this one, all red lines!

Might be worth making this the first thing that gets put into NEWS for 8.0 in this changeset, but not having it isn't a big deal for the actual changes.

dgw:
> Might be worth making this the first thing that gets put into NEWS for 8.0

Reminded me that this should get the Breaking Change label, which is exactly what tells me should be mentioned first in the news, upgrade guide, etc. Thanks.

PR #2478, opened 2023-06-18
remind: remove plugin reimplemented as external package w/better tests

dgw:
### Description
Tin. Development moved to https://github.com/sopel-irc/sopel-remind, with more comprehensive tests and support for running the plugin on Python 3.11+.(built-in has a regex pattern error).

Starting as a draft because of impending merge conflict from #2471 and waiting for a migration path for existing reminders to the new plugin (see sopel-irc/sopel-remind#1).

### Checklist
- [x] I have read [CONTRIBUTING.md](https://github.com/sopel-irc/sopel/blob/master/CONTRIBUTING.md)
- [x] I can and do license this contribution under the EFLv2
  - (It's very easy to license deletions 😼)
- [x] No issues are reported by `make qa` (runs `make quality` and `make test`)
- [x] I have tested the functionality of the things this change touches
  - By which I mean I've tested that `remind` is no longer included with the base Sopel installation, and the test suite was correctly removed (no leftovers causing failures post-deletion)

PR #2480, opened 2023-06-27
mypy: use generic types instead of `typing` aliases where possible

dgw:
### Description
It turns out that "where possible" is equivalent to "almost everywhere". There are a _very_ few places where we still have to leave `typing` aliases in for compatibility [with Python versions before 3.9](https://peps.python.org/pep-0585/).

### Checklist
  - This also includes running `mypy` 1.4 under Python 3.8, to make sure none of the changes breaks that version. (I did not bother to test Python 3.7 since it's EOL any day now. Tentatively it's EOL _today_, but I haven't seen any formal announcement or update to [the Python Versions page](https://devguide.python.org/versions/) yet.)

### Notes
Follow-up to #2471; prerequisite for finishing #2477 (so I can extract a very nice, clean file for packaging externally).

Haven't touched `remind` as it's due to be removed shortly (#2478, replaced by external package).

### DIFF
--- a/sopel/bot.py
+++ b/sopel/bot.py
@@ -18,11 +18,8 @@ import time
 from types import MappingProxyType
 from typing import (
     Any,
-    Dict,
-    Iterable,
     Mapping,
     Optional,
-    Tuple,
     TYPE_CHECKING,
     TypeVar,
     Union,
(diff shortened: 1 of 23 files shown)


Exirel (approved):
If mypy doesn't complain, I'm OK with that.

Issue #2488, opened 2023-07-11
Importing callables from other plugins silently duplicates bot commands


SnoopJ:
### Description

The [loader machinery](https://github.com/sopel-irc/sopel/blob/6af4f23d2ea011aa1e5fc9ce5449b2f2bce1bd1c/sopel/loader.py#L239-L240) treats Sopel callables in a plugin as belonging to that plugin even if the callable is imported from another plugin, which can result in duplication of a command when one plugin is importing from another.

### Reproduction steps

I discovered this while writing a plugin that aliases the command `w` to `wp`

```
# wikipedia_alias.py
from sopel import plugin
from sopel.modules.wikipedia import wikipedia as original_wikipedia_cmd


@plugin.command("w")
@plugin.output_prefix("[wikipedia ]")
def wiki_alias(bot, trigger)
    return original_wikipedia_cmd(bot, trigger)
```

This "unboxing" of the callable from the original plugin can lead to duplication of the `wp` command.

### Expected behavior

The loader should be able to distinguish between callables that are indigenous to a plugin and callables that came from somewhere else, by comparing the `__module__` attribute of the callable against the plugin module under consideration. 

It is probably too disruptive to skip callables that came from somewhere else ([users may be relying on this behavior](https://www.hyrumslaw.com/)), but the core could issue a warning about this to make the failure marginally less mysterious.

### Sopel version

6af4f23

### Installation method

`pip install`

### Python version

3.9.16

### Operating system

Ubuntu 20.04

SnoopJ:
@Exirel points out that we can't look at `__module__` because plugins may have substructure that the core should respect. It may be possible to introspect if a callable was defined "under" the relevant plugin that is being processed, but the check described above would be too aggressive.

dgw:
It's like what I said earlier on IRC about "or a submodule"—but with more nuance, I guess? Plugins that import different commands from submodules of themselves to build the "real" plugin file that Sopel loads were what I had in mind. 😁

Exirel:
To report and to extend on what I said on IRC: I'm not sure to see an issue to resolve here, and I don't think solving said issue would be a good idea, especially given the complexity of the problem.

[...]

PR #2489, opened 2023-07-12
docs: add example of plugin re-use

SnoopJ:
### Description

This changeset adds an example to the "tips and tricks" for plugins of importing another plugin and re-using a command from therein, as well as a warning about the potential for the duplication problem described in #2488.

### Checklist
- [N/A] No issues are reported by `make qa` (runs `make quality` and `make test`)

### DIFF
--- a/docs/source/plugin/advanced.rst
+++ b/docs/source/plugin/advanced.rst
@@ -121,6 +121,37 @@ Note that you don't specifically need to use ``@plugin.thread(False)``, but
 it is still recommended to prevent any race condition.
 
 
+Re-using commands from other plugins
+====================================
+
+.. TODO: this example should be updated when the Wikipedia plugin is removed
+.. as part of https://github.com/sopel-irc/sopel/issues/1291
+
+Because plugins are just Python modules it is possible to import functionality
+from other plugins, including commands. For example, this can be used to add
+an alias for an existing command::
+
+    from sopel import plugin
+    from sopel.modules import wikipedia as wp
+
+    @plugin.command("wiki")
+    @plugin.output_prefix(wp.wikipedia.output_prefix)
+    def wiki_alias(bot, trigger):
+        wp.wikipedia(bot, trigger)
+
+.. warning::
+
+    Any callables imported from other plugins will be treated as if they were
+    exposed in the current plugin. This can lead to duplication of plugin
+    rules. For the most predictable results, import the other plugin as a
+    module rather than unpacking its callables using a ``from`` import.
+
+.. warning::
+
+    While this example shows off loading a built-in plugin, some plugins may
+    not be as easy to import. For example, a :term:`Single file plugin` may
+    not be available on ``sys.path`` without extra handling not shown here.
+
 Managing Capability negotiation
 ===============================
 


docs/source/plugin/advanced.rst
```diff
@@ -121,6 +121,28 @@ Note that you don't specifically need to use ``@plugin.thread(False)``, but
...
+
+    from sopel import plugin
+    from sopel.modules import wikipedia as wp
+
+    @plugin.command("wiki")
+    @plugin.output_prefix(wp.wikipedia.output_prefix)
+    def wiki_alias(bot, trigger):
+        wp.wikipedia(bot, trigger)
+
+.. warning::
```
dgw, line 139:
```suggestion
.. warning::

```

docs/source/plugin/advanced.rst
```diff
@@ -121,6 +121,28 @@ Note that you don't specifically need to use ``@plugin.thread(False)``, but
...
+    Any callables imported from other plugins will be treated as if they were
+    defined in the current plugin. This can lead to duplication of plugin
+    rules. For the most predictable results, import the other plugin as a
+    module rather than unpacking callables from it using a 'from' import.
```
dgw, line 143:
```suggestion
    module rather than unpacking its callables using a ``from`` import.
```

dgw (commented):
These are just style nits. Bigger thing is that you can import a plugin's module even if it's not loaded, so the example could also be used to "rename" a preexisting command.

docs/source/plugin/advanced.rst
```diff
@@ -121,6 +121,29 @@ Note that you don't specifically need to use ``@plugin.thread(False)``, but
...
+
+    @plugin.command("wiki")
+    @plugin.output_prefix(wp.wikipedia.output_prefix)
+    def wiki_alias(bot, trigger):
+        wp.wikipedia(bot, trigger)
+
+.. warning::
+
+    Any callables imported from other plugins will be treated as if they were
+    defined in the current plugin. This can lead to duplication of plugin
```
Exirel, line 142:
```suggestion
    exposed in the current plugin. This can lead to duplication of plugin
```

Exirel (changes requested):
I'd approve this, however I think there is at least one note to be added: we should say that the single-file plugin type may not work this way, because it may not be importable. You can link to single-file plugin definition with:

```rst
Link to :term:`Single file plugin` definition.
```

SnoopJ:
> we should say that the single-file plugin type may not work this way, because it may not be importable

I'd love to see information of this sort added to Sopel, but I think it's a deeper subject than the warning I wanted to add in this case. I figure that users who are importing not-a-builtin-plugin are probably going to find out about the distinction anyway, and I'm not really sure how to expand this section to tackler the larger subject of how one might go about locating and importing plugins.

Do you think it makes sense to file a separate issue for fleshing out the subject?

Exirel:
It's an advanced tips & tricks section, we can direct the reader to more resources without having to explain everything. Beside, there is an explanation about [single-file plugin already](https://sopel.chat/docs/plugin/what.html#id3). I don't think it's that deeper that it requires another issue: a single line in this section is enough in my opinion.

Exirel (approved):
🚢

dgw (approved):
:shipit: (but squash first, please)

PR #2494, opened 2023-07-15
docs: add docstring to library root module

### LINKED ISSUE #2493: Root module description is a mini-rant about LC_ALL rather than a description of the library
### Description

Looking at the `sopel` module with `pydoc` in an interactive prompt) exposes the user to [a short rant](https://github.com/sopel-irc/sopel/blob/c26914b68913bc25bdd1f5fed9c5942a87fdfee6/sopel/__init__.py#L1-L4) about the behavior of `LC_ALL` and instructions to use only ASCII in this module.

I'm sympathetic to the frustration over #984 that led to this, but it will be an improvement to add a docstring to the module with a short description.

### Reproduction steps

Run `python3 -m pydoc sopel` or `import sopel; help(sopel)` in an interactive prompt.

### Expected behavior

The user should see a short description of Sopel

### Relevant logs

_No response_

### Notes

_No response_

### Sopel version

c26914b

### Installation method

`pip install`

### Python version

_No response_

### Operating system

_No response_

### IRCd

_No response_

### Relevant plugins

_No response_

SnoopJ:
### Description
This PR fixes #2493 by adding a module-level docstring to `sopel/__init__.py`. I just copied the first two sentences from https://sopel.chat

### Checklist
- [N/A] No issues are reported by `make qa` (runs `make quality` and `make test`)
- [N/A] I have tested the functionality of the things this change touches

### DIFF
--- a/sopel/__init__.py
+++ b/sopel/__init__.py
@@ -3,11 +3,7 @@ Sopel is a simple, easy-to-use, open-source IRC utility bot, written in Python.
 
 It’s designed to be easy to use, easy to run, and easy to extend.
 """
-# ASCII ONLY IN THIS FILE THOUGH!!!!!!!
-# Python does some stupid bullshit of respecting LC_ALL over the encoding on the
-# file, so in order to undo Python's ridiculous fucking idiocy, we have to have
-# our own check.
-
+#
 # Copyright 2008, Sean B. Palmer, inamidst.com
 # Copyright 2012, Elsie Powell, http://embolalia.com
 # Copyright 2012, Elad Alfassa <elad@fedoraproject.org>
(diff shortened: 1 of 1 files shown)


SnoopJ:
And actually, digging further into this, the issue is no longer actually present after Python 3.7, when [PEP 538](https://peps.python.org/pep-0538/) was implemented to coerce the C locale to UTF-8. I believe we could remove the restriction to use only ASCII text in that module, and the [associated warning](https://github.com/sopel-irc/sopel/blob/c26914b68913bc25bdd1f5fed9c5942a87fdfee6/sopel/__init__.py#L38-L43) might be superfluous now as well.

I'll leave that out of this PR unless y'all agree that they can go, but the default behavior since 3.7 is to coerce these problematic locales to UTF-8, so the problem just doesn't exist in versions we support now anyway.

Exirel (changes requested):
Let's remove the rant now. 😁

SnoopJ:
Done. I preserved the warning about non-UTF-8 locale but softened the language of it a bit. I'm not really sure it's a concern anymore with the changes that happened since Python 3.7, but I'm not confident enough to remove it entirely for the handful of users who will have a locale that is not using UTF-8 and is not coerced to same.

Exirel (approved):

dgw (approved):
Sure, why not. (I dismissed the CodeQL alert about running `print()` at the module level. This is "something's seriously screwed up in the environment" code, so I think we can leave it be.)

Issue #2498, opened 2023-08-02
version: fallback behavior if no module- or file-level `__version__`


dgw:
One can `.version pluginname` for any loaded plugin, but e.g. a single-file module stored in `~/.sopel/plugins/pluginname.py` will simply return `[version] pluginname (unknown)` if it doesn't contain a `__version__` member.

I propose that, at least for single-file plugins, the plugin machinery should fall back on the file modification timestamp in lieu of a version number.\* It'd probably be reasonable to format the file's modification time as `yyyy.mm.dd`, and ignore hours or smaller units.\*\*

----

<table>
<tr>
 <td>*
 <td>It gets trickier for Folder or Namespace Package plugins, but we're discouraging the use of those now.
<tr>
 <td>**
 <td>Our Sphinx theme for 8.x, Furo, is even versioned this way <a href="https://pypi.org/project/furo/2023.7.26/">on PyPI</a>, though packaging strips <a href="https://github.com/pradyunsg/furo/releases/tag/2023.07.26">the tags' leading zeroes</a>.

I'm sure someone will come along and say we should just use the Unix timestamp as the version number, but come on.
</table>

SnoopJ:
Here's a small patch to `version` that would implement the desired behavior, although it relies on accessing `Sopel._plugins` directly, so it will need some support work to do it the Right Way™, either adding the modification time to the plugin metadata or adding a mechanism for retrieving the plugin from the bot.

<details><summary>click for patch</summary>

```diff
diff --git a/sopel/modules/version.py b/sopel/modules/version.py
index 536eaa83..ad2bca5e 100644
--- a/sopel/modules/version.py
+++ b/sopel/modules/version.py
@@ -9,6 +9,7 @@ https://sopel.chat
 from __future__ import annotations
 
 import datetime
+import importlib.util
 import os
 import platform
 
@@ -51,6 +52,13 @@ def version(bot, trigger):
         meta = bot.get_plugin_meta(plugin)
         if meta["version"] is None:
             version = "(unknown)"
+            # TODO: this uses implementation details of Sopel ;(
+            spec = importlib.util.find_spec(bot._plugins[plugin].module_name)
+            if spec.origin:
+                mtime = os.stat(spec.origin).st_mtime
+                timestamp = datetime.datetime.fromtimestamp(mtime).strftime("%d %b %Y")
+                version += " — last modified {ts}".format(ts=timestamp)
         else:
             version = "v" + str(meta["version"])
```

</details>

[...]

PR #2500, opened 2023-08-02
build, meta: drop Python 3.7 support, checks, and CI job

dgw:
### Description
This fulfills the last "Finally letting the dead snakes lie." part of [the 8.0.0 milestone](https://github.com/sopel-irc/sopel/milestone/15)'s description. Python 3.7 was officially EOL'd as of a month ago, on June 27.

### Checklist
  - (As of when I prepped the patch.)
  - When I wrote it, I did, with `make test` etc. (Now it needs a rebase for retesting.)

### Conflict notes
@SnoopJ and @Exirel arrived at similar conclusions for dealing with aliases in `plugins/rules.py`; this has a conflict with 098d8c309058dc794dd04d0d464598482ca94505, but the approaches are quite similar. That and b0693b7b0da1047aa0f509a30f9d15ef6b71065d are both from #2471, which touched a bunch of stuff (so I knew the risk of conflicts with this relatively long-lived housekeeping branch was inevitable).

The only other little conflict is more directly my fault. While I set @SnoopJ loose on my todo list for #2471 to get it done faster, #2480—specifically, 85f8b82d3917135e309b6ce4efc601d4c79560a7—is all me. 🙈

Guess I have to rebase this one, huh? Damn, I love merging branches that diverged long before `master` because the history graph looks cool, but sometimes it's just not meant to be. 😅

### Special thanks
…to the other maintainers for humoring me when I said I wanted to reach #2500 and make this PR a numerical milestone as well as a development one. You're all awesome for tossing in extra issues & PRs to get us here! ❤️

### DIFF
--- a/sopel/bot.py
+++ b/sopel/bot.py
@@ -1255,8 +1255,8 @@ class Sopel(irc.AbstractBot):
             The Python documentation for the `re.search`__ function and
             the `match object`__.
 
-        .. __: https://docs.python.org/3.7/library/re.html#re.search
-        .. __: https://docs.python.org/3.7/library/re.html#match-objects
+        .. __: https://docs.python.org/3.11/library/re.html#re.search
+        .. __: https://docs.python.org/3.11/library/re.html#match-objects
 
         """
         for regex, function in self._url_callbacks.items():
--- a/sopel/modules/announce.py
+++ b/sopel/modules/announce.py
@@ -25,12 +25,8 @@ def _chunks(items, size):
     # This approach is safer than slicing with non-subscriptable types,
     # for example `dict_keys` objects
     iterator = iter(items)
-    # TODO: Simplify to assignment expression (`while cond := expr`)
-    # when dropping Python 3.7
-    chunk = tuple(itertools.islice(iterator, size))
-    while chunk:
+    while (chunk := tuple(itertools.islice(iterator, size))):
         yield chunk
-        chunk = tuple(itertools.islice(iterator, size))
 
 
 @plugin.command('announce')
--- a/sopel/plugins/handlers.py
+++ b/sopel/plugins/handlers.py
@@ -52,9 +52,6 @@ import os
 import sys
 from typing import Optional, TYPE_CHECKING, TypedDict
 
-# TODO: refactor along with usage in sopel.__init__ in py3.8+ world
-import importlib_metadata
-
 from sopel import __version__ as release, loader, plugin as plugin_decorators
 from . import exceptions
 
@@ -614,9 +611,13 @@ class EntryPointPlugin(PyModulePlugin):
         """
         version: Optional[str] = super().get_version()
 
-        if version is None and hasattr(self.module, "__package__"):
+        if (
+            version is None
+            and hasattr(self.module, "__package__")
+            and self.module.__package__ is not None
+        ):
             try:
-                version = importlib_metadata.version(self.module.__package__)
+                version = importlib.metadata.version(self.module.__package__)
             except ValueError:
                 # package name is probably empty-string; just give up
                 pass
--- a/sopel/plugins/handlers.py
+++ b/sopel/plugins/handlers.py
@@ -50,7 +50,7 @@ import inspect
 import itertools
 import os
 import sys
-from typing import Optional, TYPE_CHECKING
+from typing import Optional, TYPE_CHECKING, TypedDict
 
 # TODO: refactor along with usage in sopel.__init__ in py3.8+ world
 import importlib_metadata
@@ -62,6 +62,26 @@ from . import exceptions
 if TYPE_CHECKING:
     from sopel.bot import Sopel
     from types import ModuleType
+
+
+class PluginMetaDescription(TypedDict):
+    """Meta description of a plugin, as a dictionary.
+
+    This dictionary is expected to contain specific keys:
+
+    * name: a short name for the plugin
+    * label: a descriptive label for the plugin; see
+      :meth:`~sopel.plugins.handlers.AbstractPluginHandler.get_label`
+    * type: the plugin's type
+    * source: the plugin's source
+      (filesystem path, python module/import path, etc.)
+    * version: the plugin's version string if available, otherwise ``None``
+    """
+    name: str
+    label: str
+    type: str
+    source: str
+    version: Optional[str]
 
 
 class AbstractPluginHandler(abc.ABC):
@@ -107,22 +127,14 @@ class AbstractPluginHandler(abc.ABC):
         """
 
     @abc.abstractmethod
-    def get_meta_description(self) -> dict:
+    def get_meta_description(self) -> PluginMetaDescription:
         """Retrieve a meta description for the plugin.
 
-        :return: meta description information
+        :return: Metadata about the plugin
         :rtype: :class:`dict`
 
-        The expected keys are:
-
-        * name: a short name for the plugin
-        * label: a descriptive label for the plugin
-        * type: the plugin's type
-        * source: the plugin's source
-          (filesystem path, python import path, etc.)
-        * version: the plugin's version string if available, otherwise ``None``
+        The expected keys are detailed in :class:`PluginMetaDescription`.
         """
-        # TODO: change return type to a TypedDict when dropping py3.7
 
     @abc.abstractmethod
     def get_version(self):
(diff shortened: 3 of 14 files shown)


dgw:
> - [x] No issues are reported by make qa (runs make quality and make test)
>   - (As of when I prepped the patch.)

Well, now the latest version of `pyflakes` (updated from 3.0.1 to 3.1.0) raises "F811 Redefinition of unused name" for:

https://github.com/sopel-irc/sopel/blob/e63afba333ea47ca02248e83b495544bc8b256f0/sopel/plugins/rules.py#L1782

The name `execute_handler` is defined earlier, but never used:

https://github.com/sopel-irc/sopel/blob/e63afba333ea47ca02248e83b495544bc8b256f0/sopel/plugins/rules.py#L1766

Upstream seems to consider this as "Working as intended" (PyCQA/pyflakes#780), so I guess we'll have to `if: ... else:` this one now. 🙁

sopel/plugins/rules.py
```diff
@@ -1780,6 +1780,10 @@ def from_callable(cls, settings, handler):
             @functools.wraps(handler)
             def execute_handler(bot, trigger):
                 return handler(bot, trigger, match=trigger)
+        else:
+            # if/else to avoid F811 "Redefinition of unused name" error in
+            # pyflakes 3.1.0+
+            execute_handler = handler
```
SnoopJ, line 1786: This seems like a bug in `pyflakes` to me. The report is incorrect for our code, because `execute_handler` is definitely being used in this scope. Seems like [the upstream claims this is working as intended](https://github.com/PyCQA/pyflakes/issues/780). Although the example in that report does not use the redefined name, adding a statement that _does_ use it causes the same error.

I would rather that we leave it the way it was (which I find more legible) and add a `# noqa:F811` to the class definition line, maybe along with a remark about why we need to work around pyflakes's bug.

sopel/plugins/handlers.py
```diff
@@ -64,6 +61,26 @@
     from types import ModuleType
 
 
+class PluginMetaDescription(TypedDict):
```
SnoopJ, line 64: I was wondering if it would make sense to do this while looking at #2498, I'm in favor of it! :grin:

SnoopJ (approved):
LGTM, most of this is pretty straightforward housekeeping.

sopel/plugins/handlers.py
```diff
@@ -609,26 +613,28 @@ def get_version(self) -> Optional[str]:
 
         if version is None and hasattr(self.module, "__package__"):
             try:
-                version = importlib_metadata.version(self.module.__package__)
+                version = importlib.metadata.version(self.module.__package__)
```
dgw, line 616: Just noticed that this creates a mypy error now:

> sopel/plugins/handlers.py:616: error: Argument 1 to "version" has incompatible type "str | None"; expected "str"  [arg-type

It would make mypy happy if we ensure that `self.module.__package__ is not None` first, but if there's a less verbose way to spell type-safety here I'd be happy to use that instead:

```py
        if (
            version is None
            and hasattr(self.module, "__package__")
            and self.module.__package__ is not None
        ):
```

(I'd expect the simpler check `getattr(self.module, "__package__", None) is not None` to achieve the same effect, but mypy doesn't interpret that as ensuring that `self.module.__package__` isn't None.)

SnoopJ: Two other options that occur to me:

1) drop the `hasattr()` check, catch the `AttributeError` in the handler
2) spell the guard condition as `if version is None and version := getattr(self.module, "__package__"):`

I don't really like option (2) (it reeks of cleverness) but (1) seems like it should be fine. Not that there's anything wrong with your suggestion of an additional clause in the conditional.

dgw: Option (2) assigns the package name to the variable named `version`, but why? Maybe it "reeks of cleverness" because it's trying to be _too_ clever? 😁

Option (1) doesn't silence mypy; it emits the same error with:

```diff
+        if version is None:
             try:
                 version = importlib.metadata.version(self.module.__package__)
+            except AttributeError:
+                # module probably doesn't have a __package__; skip
+                pass
```

SnoopJ: oh, hrm, right. I think the Right Way™ to do it is the additional condition, then.

sopel/plugins/rules.py
dgw: I don't think it's a bug. If you keep following the issue chain, you'll eventually arrive at:

> a function overriding a variable is almost always an error -- if you want to do exceptional things you can utilize `# noqa` (you're already using flake8)

_Originally posted by @asottile in https://github.com/PyCQA/pyflakes/issues/760#issuecomment-1380461009_

… which matches what triggered the F811 error here: a `def` overriding a regular assignment to the same name that is _not used between assignment and redefinition_. That's not to say I'm exactly _happy_ about upstream updating behavior to detect this case as an error, but I think it _is_ doing what they intended it to do.

If we could arrive at a sensible spelling of something like the below (I guess most of what it needs is a good descriptive name for the intermediate `def` that isn't `execute_handler`), would you take that instead of a `# noqa`?

```py
# ...

execute_handler = handler
argspec = inspect.getfullargspec(handler)

if len(argspec.args) >= match_count:
    @functools.wraps(handler)
    def some_other_name(bot, trigger):
        return handler(bot, trigger, match=trigger)

    execute_handler = some_other_name
```

(I'd say moving the `execute_handler = handler` assignment closer to its redefinition is clearer regardless of the verdict.)

sopel/plugins/rules.py
```diff
@@ -87,9 +87,10 @@ def _clean_rules(rules, nick, aliases):
 
 def _compile_pattern(pattern, nick, aliases=None):
     if aliases:
-        nicks = list(aliases)  # alias_nicks.copy() doesn't work in py2
-        nicks.append(nick)
-        nick = '(?:%s)' % '|'.join(re.escape(n) for n in nicks)
+        nick = '(?:%s)' % '|'.join(
+            re.escape(n)
+            for n in itertools.chain(aliases, (nick,))
```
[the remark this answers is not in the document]
dgw: I think `*aliases` wins over `(nick,)` tbh. I'll adopt this, thanks.

sopel/bot.py
```diff
@@ -1255,8 +1255,8 @@ def search_url_callbacks(self, url):
             The Python documentation for the `re.search`__ function and
             the `match object`__.
 
-        .. __: https://docs.python.org/3.7/library/re.html#re.search
-        .. __: https://docs.python.org/3.7/library/re.html#match-objects
+        .. __: https://docs.python.org/3.11/library/re.html#re.search
+        .. __: https://docs.python.org/3.11/library/re.html#match-objects
```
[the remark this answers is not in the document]
dgw: I'm worried in this case about anchor names changing between Python releases. Could be a baseless fear, but the names of stdlib modules are pretty damn stable over time. Anchor names changing, however, has tripped us up before.

It's a weak argument against just linking to `/3/`, but it's what I got.

dgw:
Having whittled down most of the open threads from @SnoopJ and myself, I figured it's worth seeing if Exi has free time later this week(end) to chime in.

sopel/plugins/rules.py
SnoopJ: Summarizing discussion from IRC: @dgw is right, this is not a bug in the tool, even though we both disagree with what the tool is checking for. The behavior is indeed specifically about redefinition using `def`, so using an intermediate name works. I'm still team `# noqa` but don't object to the name-swap workaround.

asottile: please when quoting don't mention me into a thread. I already get enough notifications and I don't need the back and forth of some random PR in my inbox

SnoopJ:
> * I think `importlib_metadata` can be removed from a few more places, unless 3.8 doesn't have `importlib.metadata.EntryPoint()`

It _does_ have `EntryPoint` but I believe the behavior of `entry_points()` changes subtly between 3.9 and 3.10 which keeps the dependency around? I don't know the details myself, but see [this comment](https://github.com/sopel-irc/sopel/blob/271b306ac328ccea32da333787c00b739021bb39/sopel/plugins/__init__.py#L36-L40)

dgw:
> * `modules/tld.py` has a "py<3.5" comment

My pattern missed that because of the `<`, whoops. Thanks.

half-duplex (approved):
> It _does_ have `EntryPoint` but I believe the behavior of `entry_points()` changes subtly between 3.9 and 3.10 which keeps the dependency around? I don't know the details myself, but see [this comment](https://github.com/sopel-irc/sopel/blob/271b306ac328ccea32da333787c00b739021bb39/sopel/plugins/__init__.py#L36-L40)

I read that as being about `entry_points(filter=)` only, not the `EntryPoint` class itself, but I don't have py3.8 handy to test with.

Either way, we still need the dependency for now, so eh. :rocket:

dgw:
I decided to throw replacing `importlib_metadata` usage under `test/` at the CI for fun. Nothing broke on 3.8/3.9, so 🚀 I guess.

sopel/modules/announce.py
```diff
@@ -25,12 +25,8 @@ def _chunks(items, size):
     # This approach is safer than slicing with non-subscriptable types,
     # for example `dict_keys` objects
     iterator = iter(items)
-    # TODO: Simplify to assignment expression (`while cond := expr`)
-    # when dropping Python 3.7
-    chunk = tuple(itertools.islice(iterator, size))
-    while chunk:
+    while (chunk := tuple(itertools.islice(iterator, size))):
```
Exirel, line 28: I'm just here to say that I love it.

Exirel (approved):
I love `importlib.metadata`. I love the walrus operator (`:=`). I love removing Py3.7.

PR #2504, opened 2023-08-21
modules: rename built-in plugins to "builtins"

dgw:
### Description

Finally does away with deadnaming core plugins as "modules". The alternative we were thinking of a while ago was `sopel.internals` (to mirror how custom plugins not part of a Python package emit logs under `sopel.externals.*`), but I liked `builtins` better than `internals`.

Updated relevant documentation examples and test code.

Have not renamed any functions/methods/attributes/variables in code that manipulates plugins from the `sopel.modules` (now `sopel.builtins`) module namespace.

Opening as draft first, since I made most of these modifications in a terminal emulator on my phone and have _definitely_ not had a chance to double check everything yet. (Also the reason that half the checklist below is blank for now.)

Milestoned for 8.0.0 to start, but I'll be the first one to bump it into 8.1.0 if everything else is done but this isn't ready yet for any reason.

### Checklist

### Open question

Do we need to keep shims in `sopel.modules` in case custom plugins import from those, and set a timeline for removing them for real later (in 9.0, probably)? Or can we just call this a non-breaking change because the core plugins aren't part of our public API?

I know I'm in favor of just moving the files, without any shimming or deprecation cycle, but I'm curious if anyone disagrees with that position.


Exirel:
> Or can we just call this a non-breaking change because the core plugins aren't part of our public API?

I think it's a breaking change if we really want to prevent breaking [someone's heating system](https://xkcd.com/1172/). Otherwise, I say we go for it.

sopel/plugins/__init__.py
```diff
@@ -62,12 +62,12 @@ def find_internal_plugins():
...
     :return: yield instances of :class:`~.handlers.PyModulePlugin`
-             configured for ``sopel.modules.*``
+             configured for ``sopel.builtins.*``
 
-    Internal plugins can be found under ``sopel.modules``. This list does not
+    Internal plugins can be found under ``sopel.builtins``. This list does not
     include the ``coretasks`` plugin.
     """
-    modules = importlib.util.find_spec('sopel.modules')
+    modules = importlib.util.find_spec('sopel.builtins')
```
SnoopJ, line 70: Nit: Seems to me that we could drop the name `modules` here as well. I would propose `builtins_spec` as a replacement name.

SnoopJ (approved):
LGTM, passes QA for me locally.

Exirel:
Just a quick note to say that I fully approve of this PR btw.

dgw:
Rebased on `master` to resolve merge conflicts and tweaked the `find_internal_plugins()` code after all.

Will most likely merge this next because it touches so damned many files; other open PRs can be manually merged (if approved) or rebased as necessary. This is just one of those situations where "the perfect merge opportunity" will never come. Big PRs cause conflicts, but it doesn't make sense to do this kind of thing piecemeal.

Exirel (approved):

PR #2505, opened 2023-09-04
test: add capabilities test cases for trailing space handling

dgw:
Some IRCds include trailing space in capability lists, and this is not a spec violation. Default `.split()` behavior in Python should handle it fine, but it's also good to be explicit that we expect the default behavior, through tests.

### Checklist

  - `Success: no issues found in 82 source files`
  - `1322 passed, 8 xfailed, 1 warning in 70.41s (0:01:10)`

### Notes
Inspired by ircv3/ircv3-specifications#530, which is not yet merged. However, since this is de facto behavior by UnrealIRCd and InspIRCd that isn't likely to change very soon even if the spec amendment is rejected, I think we should just add the tests.

Review requested from @Exirel specifically because the new capability system is his baby, and he's the most likely one to point out any incorrect assumptions I might have had about how to adapt existing test cases for this purpose.

### DIFF
--- a/test/irc/test_irc_capabilities.py
+++ b/test/irc/test_irc_capabilities.py
@@ -83,6 +83,20 @@ def test_capabilities_ls_multiline(mockbot, triggerfactory):
     assert manager.available == {'away-notify': None, 'account-tag': None}
 
 
+def test_capabilities_ls_trailing_space(mockbot, triggerfactory: TriggerFactory):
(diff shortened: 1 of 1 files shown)


Exirel (approved):
Yeah, all good as far as I can tell.

dgw:
Spec clarification was merged 🎉

PR #2512, opened 2023-10-01
currency: replace `exchangerate.host` -> `open.er-api.com`

dgw:
exchangerate.host was bought and changed to require an API key. The new owner also allegedly broke compatibility with the original endpoints, based on GitHub comments from other developers, so I didn't even bother signing up for a key and testing it. I just found a new keyless API.

ExchangeRate-API.com also provides key-based services, but there's an open endpoint that doesn't require any key we can use for now.

VCR file for `currency` example test has been updated, as well.

Note that this will be a "breaking" change for users who specify the `currency.fiat_provider` option in their settings as `exchangerate.host` since that's no longer a valid value. (Ask me how I know... I forgot to update the default at first and the test suite failed because of it. :D)

### Checklist

  - Technically, my local lint reports a weird mypy error in `sopel/tools/memories.py`, but it's unrelated to this patch.
  - Both interactively and via updated VCR file for the new network request.


dgw:
OK, I'm at a bit of a loss. The bundled VCR cassettes work in our CI, but fail for me locally with a weird Brotli error:

```
_____________________________________________ test_example_exchange_cmd_0 ______________________________________________
/home/dgw/github/sopel/sopel/tests/pytest_plugin.py:148: in test
    ???
E   AssertionError: Output does not match the regex:
E   Pattern: 100 USD is [\d\.]+ BTC, [\d\.]+ CAD, [\d\.]+ EUR, \(unsupported: CAN, AUX\)
E   Output: Something went wrong while I was getting the exchange rate.
-------------------------------------------------- Captured log call ---------------------------------------------------
ERROR    sopel.modules.currency:currency.py:127 Error in GET request: ('Received response with content-encoding: br, but
failed to decode it.', error('BrotliDecoderDecompressStream failed while processing the stream'))
```

Guessing there's something wrong with my local environment, given that there's also a weird error from mypy when I run `make lint` (as mentioned in the PR description here) that doesn't happen in CI.

Whoever reviews this, if you have a moment to checkout and test locally, it's not part of the PR but I'd appreciate knowing whether this failure also happens for you locally.

SnoopJ (approved):
Plugin change looks fine to me, I'm pretty impressed that the change is _just_ the URL, but I guess it makes sense that providers of this information would naturally settle on similar response formats. And I trust that you did the right thing with the VCR changes as long as the associated tests pass.

> Whoever reviews this, if you have a moment to checkout and test locally, it's not part of the PR but I'd appreciate knowing whether this failure also happens for you locally.

QA passes for me on 3.8-3.11 :+1:

dgw:
> > Whoever reviews this, if you have a moment to checkout and test locally, it's not part of the PR but I'd appreciate knowing whether this failure also happens for you locally.
> 
> QA passes for me on 3.8-3.11 👍

I should probably just budget time for a nuke-and-rebuild of my local environment before we finalize 8.0. This plus the weird `mypy` check that nobody else sees makes me just not trust anything my terminal tells me.

Thanks for checking!

PR #2516, opened 2023-10-05
Add Python 3.12 support

### LINKED ISSUE #2457: Official support for Python 3.12

SnoopJ:
### Description

This changeset adds Python 3.12 support to Sopel, which mostly means some chore updating of version lists. The most notable change is the adjustment of the pinned version of `setuptools` to avoid a [known incompatibility with 3.12](https://github.com/pypa/setuptools/issues/3631).

~At one point it looked like `aiohttp` would be a blocker for this support (as they still have no issued a new release with a compat bugfix :grimacing:), but the problem disappeared between the last check of a Python 3.12 RC against Sopel and the official release. :shrug:~  Edit: nope, `aiohttp` is indeed [a problem](https://github.com/sopel-irc/sopel/actions/runs/6415020971/job/17416271455?pr=2516#step:4:613) (failure in CI associated with the [known upstream bug](https://github.com/aio-libs/aiohttp/issues/7229)), so we will be waiting for them to do a new release after all. We _can_ workaround the build failure by `AIOHTTP_NO_EXTENSIONS=1` in the environment, but there isn't a compelling reason to do this in CI, we can just wait on upstream to do a release. See [related issue](https://github.com/aio-libs/aiohttp/issues/7639), hopefully that will happen in the next week or so.

Edit: since the closure of #2523, the above no longer blocks this PR. Users who want both Python 3.12 and `sopel-iplookup` will be stuck waiting on `aiohttp`, but they can add `--pre` to their install command or set the above environment variable to get around it.

Closes #2457

### Checklist

  - `1323 passed, 8 xfailed, 591 warnings in 57.03s`

### Supplemental checklist (from #2457)

* [x] Check Sopel tests against 3.12
* [x] Update CI to run with 3.12
* [x] Update classifiers list in `pyproject.toml`
* ~~[ ] Check on 3.12 support concerns for official non-core plugins~~ NAK, not this PR's problem — @dgw

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -39,6 +39,7 @@ classifiers = [
     "Programming Language :: Python :: 3.9",
     "Programming Language :: Python :: 3.10",
     "Programming Language :: Python :: 3.11",
+    "Programming Language :: Python :: 3.12",
     "Topic :: Communications :: Chat :: Internet Relay Chat",
 ]
 requires-python = ">=3.8"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,5 +1,5 @@
 [build-system]
-requires = ["setuptools~=63.0", "wheel"]
+requires = ["setuptools~=66.1", "wheel"]
 build-backend = "setuptools.build_meta"
 
 [tool.setuptools]
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,6 +24,7 @@ jobs:
           - "3.9"
           - "3.10"
           - "3.11"
+          - "3.12"
     steps:
       - uses: actions/checkout@v3
       - name: Set up Python ${{ matrix.python-version }}
--- a/Makefile
+++ b/Makefile
@@ -12,20 +12,20 @@ lint-type:
 
 .PHONY: test test_norecord test_novcr vcr_rerecord
 test:
-	coverage run -m py.test -v .
+	coverage run -m pytest -v .
 
 test_norecord:
 	# error if VCR recording for a web request is missing (useful on CI)
-	coverage run -m py.test -v . --vcr-record=none
+	coverage run -m pytest -v . --vcr-record=none
 
 test_novcr:
 	# disable VCR completely; useful to check if recordings are outdated
-	coverage run -m py.test -v . --disable-vcr
+	coverage run -m pytest -v . --disable-vcr
 
 vcr_rerecord:
 	# clear VCR cassettes and run tests to record fresh ones
 	rm -rf ./test/vcr/*
-	coverage run -m py.test -v . --vcr-record=all
+	coverage run -m pytest -v . --vcr-record=all
 
 .PHONY: coverage_report coverage_html coverages
 coverage_report:
--- a/contrib/tox.ini
+++ b/contrib/tox.ini
@@ -2,12 +2,12 @@
 package_root = .
 min_version = 4.3.3
 envlist =
-    py{38,39,310,311}-qa
+    py{38,39,310,311,312}-qa
 skip_missing_interpreters = true
 ignore_base_python_conflict = true
 labels =
-    lint = py{38,39,310,311}-lint
-    test = py{38,39,310,311}-test
+    lint = py{38,39,310,311,312}-lint
+    test = py{38,39,310,311,312}-test
 
 
 [testenv]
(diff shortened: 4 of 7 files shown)


pyproject.toml
```diff
@@ -1,5 +1,5 @@
 [build-system]
-requires = ["setuptools~=63.0", "wheel"]
+requires = ["setuptools~=66.1", "wheel"]
```
SnoopJ, line 2: @dgw I made the minimal adjustment to this pin from the value set by 557b44c, but it seems prudent to ask if we have any interest in relaxing the pin now. I don't really know what the original issue was, but I can perform an `--editable` install with both 3.9 and 3.12 with the pin as I have it here.

dgw:
Rebasing this to fix the merge conflict from #2520 will be the perfect excuse to make GHA retest after aiohttp gets its Python 3.12 ducks lined up. Which should be any day now… right? (Lots of recent activity on aio-libs/aiohttp#7675 about some shutdown-timeout bug that seems to be holding back a stable release.)

SnoopJ:
It does look like `aiohttp` is going to be in pre-release for a while longer. @Exirel mentioned on IRC that we _could_ move the `ip` plugin (which gives us that dependency via `geoip2`) to an external package, which would soften how blocked we are viz. this PR. See #2523 for that changeset.

SnoopJ:
The story upstream in `aiohttp` seems to be getting more complicated, so I think our best way forward is to get #2523 merged. Users who want 3.12 _and_ the `sopel-iplookup` plugin will need to use `--pre` for the time being when installing the plugin, but everything else here is ready once that PR is merged.

dgw:
In the interest of helping this along, I chose #2523 as one of today's merges. Based on local testing yesterday, installing and running on Python 3.12 appeared to be fine with that branch merged into this one. We're left with a bunch of deprecation warnings emitted in `pytest` itself—about 46,700 of them—that were resolved for me by upgrading `pytest` from 7.1.3 to 7.4.3.

While I didn't test the `pytest` upgrade against everything Python \<3.12, a cursory check showed that `pytest` 7.4.3 is installable all the way back to Python 3.8, and the test suite passed on py3.8.\* Passing on 3.8 and 3.12 presumably means all the intermediate Python versions will work, too; we can probably just raise [the requirement](https://github.com/sopel-irc/sopel/blob/166de9f3d5f89778f804ecf4689c199748a256d4/dev-requirements.txt#L10) from `pytest~=7.1.0` 🡒 `pytest~=7.4.0` and be fine.

\* — _Passed with my manually-upgraded `urllib3` 2.0.7. As mentioned in IRC chatter yesterday, I have a solution in mind for working around that problem as part of #2519. That'll get pushed once I steel myself for a potential unexpected ❌ from CI. 😅_

Exirel:
Nothing more to say than "yes, good!" to your last comment @dgw:

* Pytest to 7.4 => yes, good!
* urllib3 thing => yes, also good!

SnoopJ:
With closure of #2523, this PR is now as final as I forsee it being. The only outstanding issue is checking on plugins external to core, but I think even if there are issues there, those will be issues _with those plugins_ and won't affect the contents of this PR.

dgw (approved):
Looks like you even caught all of the `py.test` references everywhere. Well done!

dgw:
It's official, y'all:
image

PR #2520, opened 2023-10-12
contrib: drop Python 3.7 from tox config

SnoopJ:
### Description

This changeset drops Python 3.7 from the `tox` config, since this version is now EOL and no longer targeted by Sopel.

### Checklist


dgw (approved):
This is gonna piss off #2516, but 1) you're only making trouble for yourself 😝 and 2) that'll need some more attention later anyway, after aiohttp sorts out its 3.12-compatible releases.

PR #2523, opened 2023-10-14
modules: replace ip plugin with external package

SnoopJ:
### Description

This PR replaces the built-in `ip` plugin with the [newly packageized](https://github.com/sopel-irc/sopel-iplookup) `sopel-iplookup` plugin, as part of #1291. Note that since #2504 hasn't landed yet, I'm still using the older `modules` terminology here, but if that PR lands first, this one should be reworked to use the new terminology.

This **should not be merged** until `sopel-iplookup` has been uploaded to PyPI.

I have marked this as a breaking change because, per discussion on IRC, we decided that it makes sense for the new package to name the plugin `iplookup` rather than the more generic `ip`. If users mention `ip` in their bot configuration (i.e. to exclude it), they will need to change the name to account for that.

### Checklist


pyproject.toml
```diff
@@ -53,6 +53,7 @@ dependencies = [
     "importlib_metadata>=3.6",
     "packaging",
     "sopel-help>=0.4.0",
+    "sopel-iplookup",
```
dgw, line 56: Generally speaking, if we're pulling a plugin out of core, it doesn't become a requirement. The `help` plugin has been the only exception to that rule, because it's actually pretty important for users to be able to get command lists & usage information out of the box.

For cases like this where we're extracting the plugin due to a dependency problem, we _really_ shouldn't make the new external package a requirement. The new plugin package will still depend on geoip2, which depends on aiohttp, which doesn't yet install cleanly on Python 3.12 and would still leave #2516 blocked after this change.

dgw (changes requested):

dgw (approved):
Derps all sorted out from my end, including the correct title format based on the last half-dozen times we've done this. Just need a squash & commit-message edit.

dgw:
We've published [sopel-iplookup 1.0.0](https://pypi.org/project/sopel-iplookup/1.0.0/). I had a soft-block on this being merged before publication, despite @SnoopJ's "YOLO" above, but the plugin's up now so :shipit:

SnoopJ:
Glad we didn't merge yet, we'd missed that this PR didn't drop the `geoip2` requirement! With that change added in bd1e6f1, the path to #2516 should be unblocked, as the breakage caused by the upstream troubles in `aiohttp` will be quarantined to the third-party package.

Exirel (approved):
Yay let's go!

dgw:
I'd ideally like the removal to be a single operation, deleting the plugin and its requirement in one go. I'd also like the Coveralls check statuses to report properly… They're getting stuck on Pending a lot lately and it always seems to be on commits that don't touch any code. 🤔

SnoopJ:
> I'd ideally like the removal to be a single operation, deleting the plugin and its requirement in one go.

Rebased, should be the case now.

> I'd also like the Coveralls check statuses to report properly

Yea, dunno what's going on with them :shrug:

PR #2525, opened 2023-10-20
memories: test/fix interactions between `SopelIdentifierMemory` ↔ `dict`

### LINKED ISSUE #2524: SopelIdentifierMemory pop() does not cast
### Description

`SopelIdentifierMemory` automatically casts strings to `Identifier`s when accessing contents with `mem["foo"]`, but `pop()` doesn't seem to.

### Reproduction steps

```python
>>> from sopel.tools import Identifier, SopelIdentifierMemory
>>> mem = SopelIdentifierMemory()
>>> mem["Foo"] = "bar"
>>> mem["Foo"]
'bar'
>>> mem.pop("Foo")
KeyError: 'Foo'
>>> mem.pop(Identifier("Foo"))
'bar'
```

Note the capital F in "Foo" - hashing Identifier("foo") and "foo" seems to match, but not Identifier("Foo") and "Foo".

### Expected behavior

`pop()` performs the same cast as `mem["Foo"]` and works.

### Sopel version

271b306ac328ccea32da333787c00b739021bb39

### Installation method

`pip install`

### Python version

3.11.5

dgw:
### Description
Tin. Will resolve #2524.

### Checklist

  - Copiously.

### Notes
Started as a partial, untested fix from my tablet to reserve the PR number. New day, new task list:

- [x] `del d['KeY']`
- [x] `d.clear()` (_appears_ to work, but the object becomes a regular `dict` and loses the `Identifier` magic)
- [x] `d.copy()`
     Made a copy `m` and added another item to it, verifying that the new item did not appear in the original `d`.
     However, the new object `m` is a `dict`, not a `SopelIdentifierMemory`.
- [x] `d.get('KeY')`
     Returns the default passed to `.get()`, not `key`'s value.
- [x] `d.pop('KeY')` (currently raises `KeyError`)
- [x] `d.pop('KeY', 'default')` (currently always returns `'default'`)
- [x] `d.pop('NotInDict')`
      Returns `None` but **should** raise `KeyError`.
- [x] `d.setdefault('KeY', 'default')`
     With `'key' in d`, returns `'default'` & adds `'KeY'` to `d` as a regular string, _not_ `Identifier`.
- [x] `d.update(a_plain_dict)`
     If `d` has `Identifier('thing1')` as a key, and `a_plain_dict` has `'ThinG1'` (a regular string), they will not be merged correctly and `d` will now also have a regular string key called `'ThinG1'`.
- [x] `d | a_plain_dict`
     Same problem as `d.update(a_plain_dict)`.
- [x] `d |= a_plain_dict`
     Same problem as `d.update(a_plain_dict)` and `d | a_plain_dict`.
- ~~[ ] `d == a_plain_dict`~~
     If `a_plain_dict` has the same key/value pairs but the key capitalization doesn't match `Identifier`'s canonical (lowercase) internal representation, the equality check returns `False`.
     **Decided to NAK this one and let them be unequal.** It's an easily reversible decision, since this just preserves the existing behavior (so there's no worry about plugin authors screaming that we keep changing the rules 😁).
- [x] `result = SopelIdentifierMemory({'KeY': 'foo'})`
     `result` is of type `SopelIdentifierMemory`, but without the magic. Existing keys are not converted to `Identifier`, and `'key' in result` doesn't work as expected for that type.
     In fact, even `'KeY' in result` returns `False` for the example above, and `result['KeY'] raises `KeyError`. Even basic functionality of the resulting fake identifier memory is completely broken.

### DIFF
--- a/sopel/tools/memories.py
+++ b/sopel/tools/memories.py
@@ -246,3 +246,27 @@ class SopelIdentifierMemory(SopelMemory):
 
     def update(self, maybe_mapping=tuple()):
         super().update(self._convert_keys(maybe_mapping))
+
+    def __or__(self, other):
+        if not isinstance(other, dict):
+            return NotImplemented
+
+        # self on the left, so other's keys overwrite
+        new = self.copy()
+        new.update(other)
+        return new
+
+    def __ror__(self, other):
+        if not isinstance(other, dict):
+            return NotImplemented
+
+        # self on the right, so keep only new keys from other
+        new = self.copy()
+        new.update((k, v) for k, v in other.items() if k not in self)
+        return new
+
+    def __ior__(self, other):
+        if not isinstance(other, dict):
+            return NotImplemented
+        self.update(other)
+        return self
--- a/sopel/tools/memories.py
+++ b/sopel/tools/memories.py
@@ -183,6 +183,11 @@ class SopelIdentifierMemory(SopelMemory):
     def copy(self):
         return type(self)(self, identifier_factory=self.make_identifier)
 
+    def get(self, key: str, default=_NO_DEFAULT):
+        if default is _NO_DEFAULT:
+            return super().get(self._make_key(key))
+        return super().get(self._make_key(key), default)
+
     def pop(self, key: str, default=_NO_DEFAULT):
         if default is _NO_DEFAULT:
             return super().pop(self._make_key(key))
--- a/sopel/tools/memories.py
+++ b/sopel/tools/memories.py
@@ -228,6 +228,9 @@ class SopelIdentifierMemory(SopelMemory):
     def __setitem__(self, key: Optional[str], value):
         super().__setitem__(self._make_key(key), value)
 
+    def setdefault(self, key: str, default=None):
+        return super().setdefault(self._make_key(key), default)
+
     def __delitem__(self, key: str):
         super().__delitem__(self._make_key(key))
 
--- a/sopel/tools/memories.py
+++ b/sopel/tools/memories.py
@@ -13,7 +13,10 @@ from typing import Any, Optional, TYPE_CHECKING, Union
 from .identifiers import Identifier, IdentifierFactory
 
 if TYPE_CHECKING:
-    from collections.abc import Generator, Mapping
+    from collections.abc import Iterable, Mapping
+    from typing import Tuple
+
+    MemoryConstructorInput = Union[Mapping[str, Any], Iterable[Tuple[str, Any]]]
 
 
 class _NO_DEFAULT:
@@ -179,26 +182,42 @@ class SopelIdentifierMemory(SopelMemory):
                 'SopelIdentifierMemory expected at most 1 argument, got {}'
                 .format(len(args))
             )
-        if len(args) == 1:
-            # help out mypy
-            mapping: Union[Mapping, Generator[tuple[Any, Any], None, None]]
-            # prepare keys
-            if hasattr(args[0], 'items'):
-                mapping = {identifier_factory(k): v for k, v in args[0].items()}
-            else:
-                mapping = ((identifier_factory(k), v) for k, v in args[0])
-
-            super().__init__(mapping)
-        else:
-            super().__init__()
 
         self.make_identifier = identifier_factory
         """A factory to transform keys into identifiers."""
+
+        if len(args) == 1:
+            super().__init__(self._convert_keys(args[0]))
+        else:
+            super().__init__()
 
     def _make_key(self, key: Optional[str]) -> Optional[Identifier]:
         if key is None:
             return None
         return self.make_identifier(key)
+
+    def _convert_keys(
+        self,
+        data: MemoryConstructorInput,
+    ) -> Iterable[tuple[Identifier, Any]]:
+        """Ensure input keys are converted to ``Identifer``.
+
+        :param data: the data passed to the memory at init or update
+        :return: a generator of key-value pairs with the keys converted
+                 to :class:`~.identifiers.Identifier`
+
+        This private method takes input of a mapping or an iterable of key-value
+        pairs and outputs a generator of key-value pairs ready for use in a new
+        or updated :class:`self` instance. It is designed to work with any of the
+        possible ways initial data can be passed to a :class:`dict`, except that
+        ``kwargs`` must be passed to this method as a dictionary.
+        """
+        # figure out what to generate from
+        if hasattr(data, 'items'):
+            data = data.items()
+
+        # return converted input data
+        return ((self.make_identifier(k), v) for k, v in data)
 
     def __getitem__(self, key: Optional[str]):
         return super().__getitem__(self._make_key(key))
@@ -224,3 +243,6 @@ class SopelIdentifierMemory(SopelMemory):
         if default is _NO_DEFAULT:
             return super().pop(self._make_key(key))
         return super().pop(self._make_key(key), default)
+
+    def update(self, maybe_mapping=tuple()):
+        super().update(self._convert_keys(maybe_mapping))
--- a/sopel/tools/memories.py
+++ b/sopel/tools/memories.py
@@ -38,6 +38,12 @@ class SopelMemory(dict):
     them at the same time from different threads, we use a blocking lock in
     ``__setitem__`` and ``__contains__``.
 
+    .. note::
+
+        Unlike the :class:`dict` on which they are based, ``SopelMemory`` and
+        its derivative types do not accept key-value pairs as keyword arguments
+        at construction time.
+
     .. versionadded:: 3.1
         As ``Willie.WillieMemory``
     .. versionchanged:: 4.0
@@ -74,6 +80,12 @@ class SopelMemory(dict):
 
 class SopelMemoryWithDefault(defaultdict):
     """Same as SopelMemory, but subclasses from collections.defaultdict.
+
+    .. note::
+
+        Unlike the :class:`~collections.defaultdict` on which it is based,
+        ``SopelMemoryWithDefault`` does not accept key-value pairs as keyword
+        arguments at construction time.
 
     .. versionadded:: 4.3
         As ``WillieMemoryWithDefault``
--- a/sopel/tools/memories.py
+++ b/sopel/tools/memories.py
@@ -149,9 +149,9 @@ class SopelIdentifierMemory(SopelMemory):
         """A factory to transform keys into identifiers."""
 
     def _make_key(self, key: Optional[str]) -> Optional[Identifier]:
-        if key is not None:
-            return self.make_identifier(key)
-        return None
+        if key is None:
+            return None
+        return self.make_identifier(key)
 
     def __getitem__(self, key: Optional[str]):
         return super().__getitem__(self._make_key(key))
--- a/sopel/tools/memories.py
+++ b/sopel/tools/memories.py
@@ -161,3 +161,6 @@ class SopelIdentifierMemory(SopelMemory):
 
     def __setitem__(self, key: Optional[str], value):
         super().__setitem__(self._make_key(key), value)
+
+    def __delitem__(self, key: str):
+        super().__delitem__(self._make_key(key))
--- a/sopel/tools/memories.py
+++ b/sopel/tools/memories.py
@@ -273,3 +273,12 @@ class SopelIdentifierMemory(SopelMemory):
             return NotImplemented
         self.update(other)
         return self
+
+    def __eq__(self, other):
+        if not isinstance(other, dict):
+            return NotImplemented
+        return super().__eq__(other)
+
+    def __ne__(self, other):
+        ret = self.__eq__(other)
+        return ret if ret is NotImplemented else not ret
--- a/sopel/tools/memories.py
+++ b/sopel/tools/memories.py
@@ -8,9 +8,12 @@ from __future__ import annotations
 
 from collections import defaultdict
 import threading
-from typing import Optional
+from typing import Any, Optional, TYPE_CHECKING, Union
 
 from .identifiers import Identifier, IdentifierFactory
+
+if TYPE_CHECKING:
+    from collections.abc import Generator, Mapping
 
 
 class _NO_DEFAULT:
@@ -159,8 +162,25 @@ class SopelIdentifierMemory(SopelMemory):
         *args,
         identifier_factory: IdentifierFactory = Identifier,
     ) -> None:
-        super().__init__(*args)
-        self.make_identifier: IdentifierFactory = identifier_factory
+        if len(args) > 1:
+            raise TypeError(
+                'SopelIdentifierMemory expected at most 1 argument, got {}'
+                .format(len(args))
+            )
+        if len(args) == 1:
+            # help out mypy
+            mapping: Union[Mapping, Generator[tuple[Any, Any], None, None]]
+            # prepare keys
+            if hasattr(args[0], 'items'):
+                mapping = {identifier_factory(k): v for k, v in args[0].items()}
+            else:
+                mapping = ((identifier_factory(k), v) for k, v in args[0])
+
+            super().__init__(mapping)
+        else:
+            super().__init__()
+
+        self.make_identifier = identifier_factory
         """A factory to transform keys into identifiers."""
 
     def _make_key(self, key: Optional[str]) -> Optional[Identifier]:
--- a/sopel/tools/memories.py
+++ b/sopel/tools/memories.py
@@ -164,3 +164,6 @@ class SopelIdentifierMemory(SopelMemory):
 
     def __delitem__(self, key: str):
         super().__delitem__(self._make_key(key))
+
+    def copy(self):
+        return type(self)(self, identifier_factory=self.make_identifier)
--- a/sopel/tools/memories.py
+++ b/sopel/tools/memories.py
@@ -13,6 +13,21 @@ from typing import Optional
 from .identifiers import Identifier, IdentifierFactory
 
 
+class _NO_DEFAULT:
+    """Private class to help with overriding C methods like ``dict.pop()``.
+
+    Some Python standard library features are implemented in pure C, and can
+    have a ``null`` default value for certain parameters that is impossible to
+    emulate at the Python layer. This class is our workaround for that.
+
+    .. warning::
+
+        Plugin authors **SHOULD NOT** use this class. It is not part of Sopel's
+        public API.
+
+    """
+
+
 class SopelMemory(dict):
     """A simple thread-safe ``dict`` implementation.
 
@@ -167,3 +182,8 @@ class SopelIdentifierMemory(SopelMemory):
 
     def copy(self):
         return type(self)(self, identifier_factory=self.make_identifier)
+
+    def pop(self, key: str, default=_NO_DEFAULT):
+        if default is _NO_DEFAULT:
+            return super().pop(self._make_key(key))
+        return super().pop(self._make_key(key), default)
--- a/test/tools/test_tools_memories.py
+++ b/test/tools/test_tools_memories.py
@@ -1,6 +1,8 @@
 """Tests for Sopel Memory data-structures"""
 from __future__ import annotations
 
+import pytest
+
 from sopel.tools import identifiers, memories
 
 
@@ -93,3 +95,46 @@ def test_sopel_identifier_memory_copy():
     del copied['LowerOnly']
     assert 'LowerOnly' not in copied
     assert 'LowerOnly' in memory
+
+
+def test_sopel_identifier_memory_pop():
+    memory = memories.SopelIdentifierMemory()
+    memory['SomeCamelCase'] = True
+    assert len(memory) == 1
+
+    # verify SopelIdentifierMemory.pop('key') behavior
+    pop_me = memory.copy()
+    assert len(pop_me) == 1
+    assert pop_me.pop('someCAMELcase') is True
+    assert len(memory) == 1  # sanity check
+    assert len(pop_me) == 0
+
+    # verify SopelIdentifierMemory.pop('key', None) behavior
+    pop_me = memory.copy()
+    assert len(pop_me) == 1
+    assert pop_me.pop('someCAMELcase', None) is True
+    assert len(pop_me) == 0
+
+    # verify SopelIdentifierMemory.pop('key', 'default_value') behavior
+    pop_me = memory.copy()
+    assert len(pop_me) == 1
+    assert pop_me.pop('someCAMELcase', 'DEFAULT') is True
+    assert len(pop_me) == 0
+
+    # verify SopelIdentifierMemory.pop('missing_key') behavior
+    pop_me = memory.copy()
+    assert len(pop_me) == 1
+    assert 'missing_key' not in pop_me
+    with pytest.raises(KeyError):
+        pop_me.pop('missing_key')
+    assert len(pop_me) == 1
+
+    # verify SopelIdentifierMemory.pop('missing_key', None) behavior
+    assert 'missing_key' not in pop_me
+    assert pop_me.pop('missing_key', None) is None
+    assert len(pop_me) == 1
+
+    # verify SopelIdentifierMemory.pop('missing_key', 'default_value') behavior
+    assert 'missing_key' not in pop_me
+    assert pop_me.pop('missing_key', 'DEFAULT') == 'DEFAULT'
+    assert len(pop_me) == 1
--- a/test/tools/test_tools_memories.py
+++ b/test/tools/test_tools_memories.py
@@ -6,6 +6,11 @@ import pytest
 from sopel.tools import identifiers, memories
 
 
+def test_sopel_default_memory_contains():
+    memory = memories.SopelMemoryWithDefault(list)
+    assert 'key' not in memory
+
+
 def test_sopel_identifier_memory_none():
     memory = memories.SopelIdentifierMemory()
     assert None not in memory
--- a/test/tools/test_tools_memories.py
+++ b/test/tools/test_tools_memories.py
@@ -449,3 +449,62 @@ def test_sopel_identifier_memory_ior_op_other():
         memory |= 5
     with pytest.raises(TypeError):
         memory |= 'str'
+
+
+def test_sopel_identifier_memory_eq():
+    """Test equality checks between two `SopelIdentifierMemory` instances."""
+    memory = memories.SopelIdentifierMemory({'Foo': 'bar', 'Baz': 'Luhrmann'})
+    other_memory = memories.SopelIdentifierMemory((('Foo', 'bar'), ('Baz', 'Luhrmann')))
+
+    assert memory == other_memory
+    assert other_memory == memory  # cover our bases vis-a-vis operand precedence
+
+    memory = memories.SopelIdentifierMemory({'fOO': 'bar', 'bAZ': 'Luhrmann'})
+
+    assert memory == other_memory
+    assert other_memory == memory  # cover our bases vis-a-vis operand precedence
+
+
+def test_sopel_identifier_memory_eq_dict_id():
+    """Test ``SopelIdentifierMemory`` comparison to ``dict[Identifier, Any]``.
+
+    These can be equivalent, just like two ``dict`` objects with identical
+    ``Identifier`` keys can be.
+    """
+    memory = memories.SopelIdentifierMemory({'Foo': 'bar', 'Baz': 'Luhrmann'})
+    dictionary = dict(memory)
+
+    assert dictionary == memory
+    assert memory == dictionary  # cover our bases vis-a-vis operand precedence
+
+
+def test_sopel_identifier_memory_eq_dict_str():
+    """Test ``SopelIdentifierMemory`` comparison to ``dict[str, Any]``.
+
+    These are intentionally not considered equivalent.
+    """
+    dictionary = {'Foo': 'bar', 'Baz': 'Luhrmann'}
+    memory = memories.SopelIdentifierMemory(dictionary)
+
+    assert dictionary != memory
+    assert memory != dictionary  # cover our bases vis-a-vis operand precedence
+
+    memory = memories.SopelIdentifierMemory((('Spam', 'eggs'), ('GreenEggs', 'ham')))
+
+    assert dictionary != memory
+    assert memory != dictionary  # cover our bases vis-a-vis operand precedence
+
+
+def test_sopel_identifier_memory_eq_other():
+    """Test ``SopelIdentifierMemory`` comparison to non-dict types.
+
+    Never equivalent (``NotImplemented``).
+    """
+    memory = memories.SopelIdentifierMemory({'Foo': 'bar', 'Baz': 'Luhrmann'})
+
+    assert memory != 'string'
+    assert memory != 5
+    assert memory != ('Foo', 'Bar')
+
+    # operator precedence ass-covering
+    assert 'string' != memory
--- a/test/tools/test_tools_memories.py
+++ b/test/tools/test_tools_memories.py
@@ -267,3 +267,52 @@ def test_sopel_identifier_memory_clear():
     assert memory['SpamEggs'] == 'spameggs'
     assert memory['spaMeggS'] == 'spameggs'
     assert memory['spameggs'] == memory['sPAmeGGs']
+
+
+def test_sopel_identifier_memory_update():
+    memory = memories.SopelIdentifierMemory({
+        'FromIdentifierMemory': True,
+    })
+    tuple_ = (('FromTuplePairs', True), ('FromIdentifierMemory', False))
+    list_ = [('FromTuplePairs', False), ('FromListPairs', True)]
+    dict_ = {'FromDict': True, 'FromListPairs': False}
+    set_ = set((('FromDict', False), ('FromSet', True)))
+
+    assert len(memory) == 1
+    assert memory['froMidentifieRmemorY'] is True
+    assert 'froMtuplEpairS' not in memory
+    assert 'froMlisTpairS' not in memory
+    assert 'froMdicT' not in memory
+    assert 'froMseT' not in memory
+
+    memory.update(tuple_)
+    assert len(memory) == 2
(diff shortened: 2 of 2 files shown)


SnoopJ (commented):
Fix for `pop()` looks reasonable to me. Agree that we want some tests to enforce the expected behavior, subclassing `dict` is tricky!

dgw:
> Fix for `pop()` looks reasonable to me.

Fun fact: It wasn't reasonable, and in fact didn't raise `KeyError` if the key is missing and no `default` was provided.

Fixed that, and checked off some more of my task list w/fixes for `.copy()`, `del memory['KeY']`, and `.get()`.

Still draft, though. I have several more behaviors to retest and then fix if necessary.

dgw:
Think I'm about burned out on this for the day. But if my notes are correct, only `__eq__()`/`__ne__()` and `setdefault()` are left to check/unit-test/maybe-fix. 🎉

dgw:
Turns out I lied about being done for the day. `setdefault()` was a pretty simple one, and then after toying with `__eq__()` for a bit I just decided not to bother.

**Objects of different types probably _should_ be considered unequal**, unless there's a really good use-case I'm missing. (Anyone who really wants to force a comparison between a regular `dict` and a `SopelIdentifierMemory` can test `identifier_memory == SopelIdentifierMemory(my_normal_dict)` instead, since this patch fixes that behavior.)

sopel/tools/memories.py
```diff
@@ -166,3 +205,46 @@ def __contains__(self, key):
...
+            return super().pop(self._make_key(key))
+        return super().pop(self._make_key(key), default)
+
+    def update(self, maybe_mapping=tuple()):
+        super().update(self._process_arg(maybe_mapping))
+
+    def __or__(self, other):
+        if not isinstance(other, dict):
+            return NotImplemented
+        new = type(self)(self, identifier_factory=self.make_identifier)
```
SnoopJ, line 237: Is it possible to use `copy()` here? If not, we should probably use the same spelling of referencing the type (i.e. use `type(self)` in both places or `self.__class__` in both places, but not mix the two)

```suggestion
        new = self.copy()
```

sopel/tools/memories.py
```diff
@@ -166,3 +205,46 @@ def __contains__(self, key):
...
+        new.update(other)
+        return new
+
+    def __ror__(self, other):
+        if not isinstance(other, dict):
+            return NotImplemented
+        new = type(self)(other, identifier_factory=self.make_identifier)
```
SnoopJ, line 244: See previous remark

```suggestion
        new = self.copy()
```

test/tools/test_tools_memories.py
```diff
@@ -58,3 +62,412 @@ def test_sopel_identifier_memory_channel_str():
...
+    assert memory['fOo'] is None
+
+    memory.setdefault('bAr', 'DEFAULT')
+    assert len(memory) == 2
+    assert 'FoO' in memory
+    assert 'BaR' in memory
+    assert memory['Bar'] == 'DEFAULT'
+
+    assert memory.setdefault('baR', 'different') == 'DEFAULT'
+    assert len(memory) == 2  # still, that is
```
SnoopJ, line 84: The wording of the comment is slightly weird, but I think the code speaks for itself anyway.

```suggestion
    assert len(memory) == 2
```

test/tools/test_tools_memories.py
```diff
@@ -58,3 +62,412 @@ def test_sopel_identifier_memory_channel_str():
...
+    memory['SomeCamelCase'] = True
+    memory['loweronly'] = False
+    assert len(memory) == 2
+    assert type(memory).__name__ == 'SopelIdentifierMemory'
+
+    copied = memory.copy()
+    assert len(copied) == 2
+    assert 'SomeCamelCase' in copied
+    assert 'loweronly' in copied
+    assert type(copied).__name__ == 'SopelIdentifierMemory'
```
SnoopJ, line 98: Hmm, any reason to test the name of the type over the value? I.e. why not `assert type(copied) is memories.SopelIdentifierMemory` or even `assert isinstance(copied, memories.SopelIdentifierMemory)`?

I'm not sure we need to test this at all on the original, either, but I assume you did it for symmetry?

test/tools/test_tools_memories.py
```diff
@@ -58,3 +62,412 @@ def test_sopel_identifier_memory_channel_str():
...
+    # verify key-missing behavior w/explicit default of None
+    assert 'missing_key' not in memory
+    got = memory.get('missing_key', None)
+    assert got is None
+    del got
+
+    # verify key-missing behavior w/explicit "real" default value
+    assert 'missing_key' not in memory
+    got = memory.get('missing_key', 'DEFAULT')
+    assert got == 'DEFAULT'
```
SnoopJ, line 146: I hate the `del got` sprinkled over this test, so I had a go at rewriting. It could be simplified further if not for the mixed case in the keys (i.e. as assign `'SomeCamelCase'` but we test `somecamelcase`), maybe it would be better to write a test dedicated to testing `get()`'s behavior over case, or maybe this test itself should be several (parametrized?) tests.

Anyway, just scratching my own itches here.

Note that I dropped the repetition of `assert 'missing_key' not in memory` in the last clause, which I think is only necessary if we're trying to test that `get()` does not mutate the memory object (which is maybe another point in favor of breaking this into multiple tests)

```suggestion
def test_sopel_identifier_memory_get():
    memory = memories.SopelIdentifierMemory()
    memory['SomeCamelCase'] = True
    memory['loweronly'] = False
    assert len(memory) == 2

    # This could be used for the above, too, except for the
    # case change. That might be something that could be
    # tested elsewhere, though…
    pairs = [
        ("somecamelcase", True),
        ("LowerOnly", False),
    ]

    for key, expected in pairs:
        # verify key-exists behavior w/implicit default
        assert memory.get(key) == expected

        # verify key-exists behavior w/explicit default of None
        assert memory.get(key, None) == expected

        # verify key-exists behavior w/explicit "real" default value
        assert memory.get(key, 'DEFAULT') == expected

    # verify key-missing behavior w/implicit default
    assert 'missing_key' not in memory
    assert memory.get('missing_key') is None
    # verify key-missing behavior w/explicit default of None
    assert memory.get('missing_key', None) is None
    # verify key-missing behavior w/explicit "real" default value
    assert memory.get('missing_key', 'DEFAULT') == 'DEFAULT'
```

test/tools/test_tools_memories.py
```diff
@@ -58,3 +62,412 @@ def test_sopel_identifier_memory_channel_str():
...
+    assert len(pop_me) == 1
+
+    popped = pop_me.pop('someCAMELcase')
+    assert len(memory) == 1  # sanity check
+    assert len(pop_me) == 0
+    assert popped is True
+
+    # reset
+    del pop_me
+    del popped
```
SnoopJ, line 165: I similarly hate the `del` here but hopefully the rewrite for the `get()` test gets that across well enough that you can see what I would have written here instead.

test/tools/test_tools_memories.py
```diff
@@ -58,3 +62,412 @@ def test_sopel_identifier_memory_channel_str():
...
+    assert 'missing_key' not in pop_me
+    popped = pop_me.pop('missing_key', 'DEFAULT')
+    assert popped == 'DEFAULT'
+
+
+def test_sopel_identifier_memory_del_key():
+    memory = memories.SopelIdentifierMemory()
+    memory['KeY'] = True
+    assert 'KeY' in memory
+    del memory['KeY']
```
SnoopJ, line 216: This `del` is fine since that's what we're testing here! 😁

test/tools/test_tools_memories.py
```diff
@@ -58,3 +62,412 @@ def test_sopel_identifier_memory_channel_str():
...
+
+    assert memory == other_memory
+    assert other_memory == memory  # cover our bases vis-a-vis operand precedence
+
+
+def test_sopel_identifier_memory_eq_dict():
+    """Test equality checks between `dict` and `SopelIdentifierMemory`.
+
+    Sopel's memory types are actually NOT guaranteed to compare as identical
+    with a `dict` containing the same key-value pairs.
```
SnoopJ, line 467: Nit: Not crazy about the wording of this docstring, because we do not in general have _"the same key-value pairs"_ because the whole point of the type is that we make `Identifier` objects. But I understand the code that it goes with, so it's not a huge deal for me.

test/tools/test_tools_memories.py
```diff
@@ -58,3 +62,412 @@ def test_sopel_identifier_memory_channel_str():
...
+    result = memory | dictionary
+    assert len(result) == 2
+    assert result['fROmMemorY'] is False
+    assert result['fROmDicT'] is True
+
+    # __ror__
+    result = dictionary | memory
+    assert len(result) == 2
+    assert result['fROmMemorY'] is True
+    assert result['fROmDicT'] is False
```
SnoopJ, line 402: It feels like these tests have a decent amount of repetition in them to cover case inflections, and would be a lot more compact and to-the-point if that functionality was tested elsewhere. As I've gone through this PR I've become more convinced that a test (or series of tests) dedicated to case would help the legibility of the rest of the added tests.

test/tools/test_tools_memories.py
```diff
@@ -58,3 +62,412 @@ def test_sopel_identifier_memory_channel_str():
...
+
+# TODO: remove condition when dropping Python 3.8
+@pytest.mark.skipif(sys.version_info < (3, 9), reason="'|=' operator added in py3.9")
+def test_sopel_identifier_memory_ior_op():
+    memory = memories.SopelIdentifierMemory({'FromMemory': True})
+    assert len(memory) == 1
+    assert memory['FromMemory'] is True
+    assert memory['frommemory'] is True
+    assert memory['froMmemorY'] is True
+    assert 'FromDict' not in memory
```
SnoopJ, line 413: Copy-paste leftovers?

```suggestion
```

test/tools/test_tools_memories.py
```diff
@@ -58,3 +62,412 @@ def test_sopel_identifier_memory_channel_str():
...
+    assert len(other_memory) == 2
+    assert 'FromOther' in other_memory
+    assert other_memory['FromOther'] is True
+    assert 'fromother' in other_memory
+
+    memory |= other_memory
+    assert len(memory) == 2
+    assert memory['fROmMemorY'] is False
+    assert memory['fROmoTHEr'] is True
+    assert 'FromTuple' not in memory
```
SnoopJ, line 428: Copy-paste leftovers?

```suggestion
```

test/tools/test_tools_memories.py
```diff
@@ -58,3 +62,412 @@ def test_sopel_identifier_memory_channel_str():
...
+    assert len(dictionary) == 1
+    assert 'FromDict' in dictionary
+    assert dictionary['FromDict'] is True
+    assert 'fromdict' not in dictionary
+
+    memory |= dictionary
+    assert len(memory) == 2
+    assert memory['fROmMemorY'] is True
+    assert memory['fROmDicT'] is True
+    assert 'FromTuple' not in memory
```
SnoopJ, line 451: Copy-paste leftovers?

```suggestion
```

test/tools/test_tools_memories.py
```diff
@@ -58,3 +62,412 @@ def test_sopel_identifier_memory_channel_str():
...
+
+# TODO: remove condition when dropping Python 3.8
+@pytest.mark.skipif(sys.version_info < (3, 9), reason="'|' operator added in py3.9")
+def test_sopel_identifier_memory_or_op():
+    memory = memories.SopelIdentifierMemory({'FromMemory': True})
+    assert len(memory) == 1
+    assert memory['FromMemory'] is True
+    assert memory['frommemory'] is True
+    assert memory['froMmemorY'] is True
+    assert 'FromDict' not in memory
```
SnoopJ, line 359: Copy-paste leftovers?

```suggestion
```

test/tools/test_tools_memories.py
```diff
@@ -58,3 +62,412 @@ def test_sopel_identifier_memory_channel_str():
...
+    assert len(other_memory) == 2
+    assert 'FromOther' in other_memory
+    assert other_memory['FromOther'] is True
+    assert 'fromother' in other_memory
+
+    result = memory | other_memory
+    assert len(result) == 2
+    assert result['fROmMemorY'] is False
+    assert result['fROmoTHEr'] is True
+    assert 'FromTuple' not in result
```
SnoopJ, line 374: Copy-paste leftovers?

```suggestion
```

test/tools/test_tools_memories.py
dgw: Admittedly, we already have tests of case inflections for both `str` and `Identifier` keys; take a look near the top of the file. I even got pretty dang tired of coming up with alternative versions of each key while writing the tests, but it felt necessary.

Many of the behaviors this patch touches needed fixing because they broke handling of case inflections in one way or another (letting non-`Identifier` keys into the object; returning a `dict` instead of `SopelIdentifierMemory`; etc.). The test cases check multiple inflections as a defense-in-depth against breaking the key-casting behavior again.

I'm also not convinced that what's here is enough in all cases. When you asked about using `self.copy()` in `__ror__()` above, I changed it before thinking, ran the test suite, and it still passed. (Next push will include a modification to catch this in at least one of the `or_op` test cases.)

test/tools/test_tools_memories.py
dgw: I'm going to adapt some, but perhaps not all, of the ideas from this rewrite for both the `get()` and `pop()` tests, and give you co-author credit for the inspiration. Stay tuned!

> maybe it would be better to write a test dedicated to testing `get()`'s behavior over case

This basically _is_ said test dedicated to testing `get()`'s behavior over case. The problem this patch fixes is that accessing a key with arbitrary casing through `__getitem__()` worked properly, but `get()` didn't.

We don't _really_ need to check that `get()` actually gets something, per se, assuming we trust the Python runtime's test suite to validate the behavior of `super().get()`. What we need to check is whether our overridden `get()` modifies the input correctly, to retrieve values in those cases where `dict.get()` itself wouldn't. That's why I so often create or assign values using one casing of a key, but retrieve them with a different variation: It checks the behavior _we_ control, which is key transformation.

(But the tests do check the whole range of `get()` behaviors as documented in Python's docs, for completeness. Just in case a future code change breaks something else. Regressions suck.)

dgw (commented):
Probably not _quite_ time for you to look at the code again, since I haven't pushed any tweaks yet, but I thought it'd be useful to publish my comment responses before I get really into addressing the bigger line notes that will lead to any refactoring.

sopel/tools/memories.py
```diff
@@ -13,11 +13,27 @@
...
 
 if TYPE_CHECKING:
-    from typing import Callable, Optional
+    from collections.abc import Generator, Mapping
+    from typing import Any, Callable, Optional, Union
 
     IdentifierFactory = Callable[[str], Identifier]
 
 
+class _NO_DEFAULT:
```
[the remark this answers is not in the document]
SnoopJ: If it makes you feel better, the stdlib does it in a whole bunch of places, but it's not really a big deal in this case since no instances are ever constructed, and it does give us a convenient place to hang a docstring.

sopel/tools/memories.py
```diff
@@ -149,14 +165,37 @@ def __init__(
         *args,
         identifier_factory: IdentifierFactory = Identifier,
     ) -> None:
-        super().__init__(*args)
+        if len(args) > 1:
+            raise TypeError(
+                'SopelIdentifierMemory expected at most 1 argument, got {}'
+                .format(len(args))
+            )
```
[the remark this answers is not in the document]
SnoopJ: Sounds like good fodder for a follow-up issue. You want to file that, or should I?

sopel/tools/memories.py
```diff
@@ -149,14 +165,37 @@ def __init__(
...
+
     def _make_key(self, key: Optional[str]) -> Optional[Identifier]:
-        if key is not None:
-            return self.make_identifier(key)
-        return None
+        if key is None:
+            return None
+        return self.make_identifier(key)
+
+    def _process_arg(self, maybe_mapping=None):
```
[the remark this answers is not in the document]
SnoopJ: > I now can't think of any reason not to just always return `Generator[tuple[Identifier, Any], None, None]`. Thoughts?

I'm pretty luke-warm on `Generator[T, None, None]` and usually opt for `Iterable[T]` instead since it conveys about the same information, but I agree that it makes sense for this internal helper to return a consistent type, and doing a lazy stream of converted-key/val pairs makes sense.

sopel/tools/memories.py
```diff
@@ -149,14 +165,37 @@ def __init__(
...
+        if maybe_mapping is None:
+            return tuple()
+
+        # help out mypy
+        processed: Union[Mapping, Generator[tuple[Any, Any], None, None]]
+        # prepare keys
+        if hasattr(maybe_mapping, 'items'):
+            processed = {self.make_identifier(k): v for k, v in maybe_mapping.items()}
+        else:
+            processed = ((self.make_identifier(k), v) for k, v in maybe_mapping)
```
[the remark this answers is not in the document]
SnoopJ: `mapping_or_iterable` is fine to me, if not a little verbose. I would probably personally call it `data` and sort out the details in the docstring, but I have no objection to the wordier form that doesn't require its own documentation.

dgw:
Aha, need to decide what to do about those "assert statement has a side effect" errors from CodeQL. Now that I eliminated the intermediate variables from those tests, I don't really want to put them back. 😅

sopel/tools/memories.py
SnoopJ: > Do you place higher value on using `self.copy()` here because it's shorter, or on using `type(self)()` for parity with the sister method?

I would prefer `self.copy()` because it's more clearly expressing what's going on _and_ avoiding repetition.

sopel/tools/memories.py
SnoopJ: Oops, yea, that should have been `new = other.copy()` in my suggestion. Copy/paste error of my own…

test/tools/test_tools_memories.py
SnoopJ: While reading on a first pass, I had some thoughts about how we might try to push the case inflection testing into parameters or fixtures, I will mull that over a bit more for whenever I come back to this.

test/tools/test_tools_memories.py
SnoopJ: Repeating what I said on IRC here now that I'm at a proper computer again:

The wording still makes me uneasy because the equivalence could be misunderstood as being about the underlying `str` behind the `Identifier`. I would probably document here that the case in question IS guaranteed to NOT be an equivalence, and that's intended behavior.

sopel/tools/memories.py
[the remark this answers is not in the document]
dgw: Why don't you tackle that one? I've just realized another hole in the yet-to-be-renamed `_process_arg()` logic. You can also construct a regular `dict` like `dict(foo='bar', spam='eggs')` and get back `{'foo': 'bar', 'spam': 'eggs'}`, but the current patch doesn't handle that case.

sopel/tools/memories.py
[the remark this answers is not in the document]
dgw: I waffled on this for quite a while, eventually choosing to call it `data` + add a docstring in my upcoming push.

dgw (commented):
Batched responses related to my upcoming push. Will look forward to the next round of feedback tomorrow or later in the week 🙂

Exirel:
> **Objects of different types probably _should_ be considered unequal**

In this case, absolutely, yes. A SopelIdentifierMemory is a very specific structure that has a very specific behavior. If people are going to do weird stuff with it, that weird stuff shouldn't work.

sopel/tools/memories.py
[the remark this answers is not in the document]
Exirel: About the `dict(foo='bar', spam='eggs')`: this class doesn't handle it and I don't think it should, or that it has to. It's OK to have a different behavior and init than everything else, it is *like* a dict, but not *exactly* a dict.

sopel/tools/memories.py
Exirel: Why not just: 

```python
new = self.copy()
new.update({k: v for k, v in other.items() if k not in self)
```

dgw: > Why not just:

I think it's less clear what that spelling does. Taking a copy of `other` and then updating it with `self` exactly mirrors how the `__or__` method works, and (to me) seems clearer even with the `type(self)(...)` shenanigans. Adding a comment as @SnoopJ suggested isn't a bad idea, especially since I think _either_ spelling would need one.

More importantly, the dict comprehension means iterating over `other` _twice_: once to build the argument to `update()` in an ephemeral `dict`, and then again when `update()` calls `_convert_keys()` on its input.

(One couldn't even bypass the second iteration by pre-converting the keys in the dict comprehension; `update()` will call `_convert_keys()` regardless of whether any keys in its input actually need conversion, since checking first would introduce _another_ iteration over all keys in the input.)

Exirel: Oh oh that's because I forgot that you can pass an iterable of tuple (k, v) to update and it'll work:

```python
new = self.copy()
new.update((k, v) for k, v in other.items() if k not in self)
```

Now it's a generator expression, so it's not evaluated twice: it's evaluated only once!

And I think it is clearer that way: get a copy of the current state, and add any key that isn't already there from the other. It mimics almost exactly `__or__` by adding a slight difference.

dgw: Ah, after remembering generator expressions in one of my other commits this week, I forgot about them again when looking at this feedback. You, sir, earned a co-author credit. :)

dgw:
Welp, looks like no more CI builds on this branch until I rebase. Y'all mind if I squash out the fixups at the same time?

Exirel:
Yes go ahead, this will fix the conflict too.

dgw:
> this will fix the conflict too

Yep, that's why I wanted to rebase. Hard to continue bikeshedding new test cases if CI won't run any more. 😝

Ran 3.10 & 3.8 tests locally; pushing in a moment, once I review the final history for rebase mistakes… I wound up doing three passes on it so there'd be less chance of putting fixups in the wrong order by accident, and I'm still not quite positive that I put everything in the right order. 😅

dgw:
Got carried away and used `git rebase --exec` to run lint + tests on Python 3.11 for every commit in the cleaned-up history.

But because of that, I know every intermediate step works as expected and doesn't have any dangling imports or commits that were reordered incorrectly. :)

sopel/tools/memories.py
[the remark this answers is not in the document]
Exirel: I think it's OK. If someone find a better solution in the future, it's only a PR away. I'm OK with that!

sopel/tools/memories.py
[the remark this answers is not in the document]
Exirel: Thinking about it, if you want to change the signature, this should work:

```
def __init__(
    self,
    # single optional arg
    data: <...type to figure out...> = None,
    *,  # ensure factory is a kwarg only
    identifier_factory: IdentifierFactory = Identifier):
```

As it makes the first arg optional, but you can't use a kwarg for it, and it keeps `identifier_factory` as a kwarg only.

Alternatively, you can keep the current signature. Either way, both work and do what we want at the moment. And we can still change that later!

test/tools/test_tools_memories.py
```diff
@@ -58,3 +62,415 @@ def test_sopel_identifier_memory_channel_str():
...
+    memory['SomeCamelCase'] = True
+    memory['loweronly'] = False
+    assert len(memory) == 2
+    assert isinstance(memory, memories.SopelIdentifierMemory)
+
+    copied = memory.copy()
+    assert len(copied) == 2
+    assert 'SomeCamelCase' in copied
+    assert 'loweronly' in copied
+    assert isinstance(memory, memories.SopelIdentifierMemory)
```
Exirel, line 98: This might need another step:

```python
    copied['newKey'] = 'value'
    assert 'newKey' in copied
    assert 'newKey' not in memory
```

Just to make sure that the "copy" actually create a copy, and not a reference to the original object.

test/tools/test_tools_memories.py
```diff
@@ -58,3 +62,415 @@ def test_sopel_identifier_memory_channel_str():
...
+    del memory['KeY']
+    assert 'KeY' not in memory
+
+
+def test_sopel_identifier_memory_from_dict():
+    dictionary = {'CamelCaseKey': True, 'lowercasekey': False}
+    assert 'CamelCaseKey' in dictionary
+    assert 'camelcasekey' not in dictionary
+    assert 'lowercasekey' in dictionary
+    assert 'LowerCaseKey' not in dictionary
```
Exirel, line 194: Don't need those here, that's how `dict`  are supposed to work already. Trust Python!

```suggestion
    dictionary = {'CamelCaseKey': True, 'lowercasekey': False}
```

Exirel (changes requested):
Alright, that sound pretty good so far.

Aside from various recommendations or requests, my repeating points are:

* you can use `memory[key] == dict[key]` to check that the value is the same/expected when you create from dict or by using copy
* check that a copy doesn't alter the original, and vice versa

And I agree with SnoopJ on the left over "key not in memory", they don't look relevant.

sopel/tools/memories.py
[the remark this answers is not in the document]
dgw: > `data: <...type to figure out...> = None,`

That should be easy as `Optional[MemoryConstructorInput]`, though it'll need modification to allow for non-`str` keys on the non-`Identifier`-coercing memory types.

test/tools/test_tools_memories.py
```diff
@@ -58,3 +62,415 @@ def test_sopel_identifier_memory_channel_str():
...
+
+    # verify key-missing behavior w/explicit default of None
+    assert 'missing_key' not in memory
+    assert memory.get('missing_key', None) is None
+    assert len(memory) == 2
+
+    # verify key-missing behavior w/explicit "real" default value
+    assert 'missing_key' not in memory
+    assert memory.get('missing_key', 'DEFAULT') == 'DEFAULT'
+    assert len(memory) == 2
```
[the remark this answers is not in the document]
dgw: True, subscripting is checked in earlier test cases.

test/tools/test_tools_memories.py
```diff
@@ -58,3 +62,415 @@ def test_sopel_identifier_memory_channel_str():
...
+def test_sopel_identifier_memory_eq_dict_str():
+    """Test ``SopelIdentifierMemory`` comparison to ``dict[str, Any]``.
+
+    These are intentionally not considered equivalent.
+    """
+    dictionary = {'Foo': 'bar', 'Baz': 'Luhrmann'}
+    memory = memories.SopelIdentifierMemory(dictionary)
+
+    assert dictionary != memory
+    assert memory != dictionary  # cover our bases vis-a-vis operand precedence
```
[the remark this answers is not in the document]
dgw: Look for that in the next push :)

dgw:
I have another rebase in my future, apparently. Hopefully @SnoopJ can revisit the one or two open threads he started by the time Exi comes back from being "AFK" (as he said) for a few days, and I can rebase one last time with all the feedback/fixups then.

sopel/tools/memories.py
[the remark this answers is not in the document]
SnoopJ: Filed as #2541

sopel/tools/memories.py
```diff
@@ -161,3 +230,57 @@ def __contains__(self, key):
...
+        new.update(other)
+        return new
+
+    def __ror__(self, other):
+        if not isinstance(other, dict):
+            return NotImplemented
+
+        # self on the right, so keep only new keys from other
+        new = self.copy()
+        new.update((k, v) for k, v in other.items() if k not in self)
```
SnoopJ, line 271: Am I mistaken, or can this be done more symmetrically by copying `other` first? Keys from `self` should override any shared keys, and everything else is in the disjoint and belongs in the union.

```suggestion
        # self on the right, so keep only new keys from other
        new = other.copy()
        new.update(self)
```

test/tools/test_tools_memories.py
```diff
@@ -58,3 +62,424 @@ def test_sopel_identifier_memory_channel_str():
...
+    assert dictionary['FromMemory'] is False
+    assert dictionary['AlsoFromDict'] is True
+    assert 'alsofromdict' not in dictionary
+
+    # __or__
+    result = memory | dictionary
+    assert len(result) == 3
+    assert result['fROmMemorY'] is False
+    assert result['fROmDicT'] is True
+    assert result['alsOfroMdicT'] is True
```
SnoopJ, line 394: Nit: this part of this test should still be run in Python 3.8, only `__ror__(dict)` is version-gated.

(probably the following clause can be wrapped in a version guard?)

test/tools/test_tools_memories.py
```diff
@@ -58,3 +62,424 @@ def test_sopel_identifier_memory_channel_str():
...
+    # __ror__
+    result = dictionary | memory
+    assert len(result) == 3
+    assert result['fROmMemorY'] is True
+    assert result['fROmDicT'] is False
+    assert result['alsOfroMdicT'] is True
+
+
+# TODO: remove condition when dropping Python 3.8
+@pytest.mark.skipif(sys.version_info < (3, 9), reason="'|=' operator added in py3.9")
```
SnoopJ, line 405: Nit: This guard could be dropped now, this test should run fine in 3.8

```suggestion
```

test/tools/test_tools_memories.py
```diff
@@ -58,3 +62,424 @@ def test_sopel_identifier_memory_channel_str():
...
+    assert other_memory['froMotheR'] is True
+
+    memory |= other_memory
+    assert len(memory) == 2
+    assert memory['fROmMemorY'] is False
+    assert memory['fROmoTHEr'] is True
+
+
+# TODO: remove condition when dropping Python 3.8
+@pytest.mark.skipif(sys.version_info < (3, 9), reason="'|=' operator added in py3.9")
```
SnoopJ, line 426: Nit: This guard could be dropped now, this test should run fine in 3.8

```suggestion
```

sopel/tools/memories.py
```diff
@@ -161,3 +230,57 @@ def __contains__(self, key):
...
+        self.update(other)
+        return self
+
+    def __eq__(self, other):
+        if not isinstance(other, dict):
+            return NotImplemented
+        return super().__eq__(other)
+
+    def __ne__(self, other):
+        return not self.__eq__(other)
```
SnoopJ, line 286: Nit: This _could_ be omitted, what's written here [is the default behavior](https://docs.python.org/3/reference/datamodel.html#object.__ne__):

> For `__ne__()`, by default it delegates to `__eq__()` and inverts the result unless it is NotImplemented.

```suggestion
```

SnoopJ (approved):
I had a few more nits, but this is good enough to get a :heavy_check_mark: from me

sopel/tools/memories.py
dgw: [Deja vu](https://github.com/sopel-irc/sopel/pull/2525#discussion_r1369588596)? Previous thread tl;dr: `other` might be a normal `dict`, and a copy of it wouldn't be a `SopelIdentifierMemory` if that's the case. That's how we got to the `new = type(self)(other)` construct that @Exirel proposed replacing with a `new = self.copy()` + the generator-expression-based `new.update()` call.

SnoopJ: d'oh, right

sopel/tools/memories.py
dgw: Since we're nitpicking, that's the default behavior for `object.__ne__()` and not necessarily `dict.__ne__()`, which is what we'd get if this explicit (is better than implicit) override didn't exist.

A better question is, does this behave correctly `if not isinstance(other, dict)`? Even if the override here is removed? (The answer is no. Need to fix that and add test cases for it.)

test/tools/test_tools_memories.py
dgw: This appears to be a case where the nit is technically wrong (`__ror__` isn't version-gated) but it still pointed out something useful. It's true that you can't do `dict | dict` in Python 3.8, but you damn well can do:

- `SopelIdentifierMemory | SopelIdentifierMemory`
- `SopelIdentifierMemory | dict`
- `dict | SopelIdentifierMemory`

Plus the `|=` versions of those. Just having the relevant dunder methods makes it work. Next push will remove the version constraints from these tests entirely.

test/tools/test_tools_memories.py
dgw: Any results of said mulling to address in this PR, or should we resolve the thread? I ask because you did approve the changes after all, so maybe the result was "it's not worth the refactoring effort".

dgw:
The fixups I just pushed (plus one technically unrelated new test I'm shoehorning in) catch even more edge cases that were missed, which @SnoopJ's nitpicks nudged me to find either directly or indirectly. Another very helpful review, thanks!

sopel/tools/memories.py
SnoopJ: Aha, I stand corrected!

Exirel (approved):
I'm going to be honest: the tests look good. The code looks good. Is there anything missing? Maybe, but I can't tell. All I know is that it looks very thorough, and if a bug arises... then we'll fix it. I don't see anything that would prevent us from fixing any issue, as the design sounds sane to me, and that's the most important part.

Note: if you really *really* want a nitpick at this point, I still don't think it's necessary to do any assertion about how a builtin dict works. But it's already here and it works, so there is no argument to have here.

dgw:
Latest force-push to squash out all the fixups moved one of the tests within the file, but meh. I made sure that the commit order is sane using `git rebase master --exec '<pip3 deps> && <run lint/test>'` and all the intermediate stages pass. I verified each commit's diff manually to be even more sure that the intermediate steps make logical sense. Can't imagine what else I could double check at this point, so let's :shipit:

> Note: if you really _really_ want a nitpick at this point, I still don't think it's necessary to do any assertion about how a builtin dict works.

I reduced the number of those, if not completely eliminated them. You're right that it isn't worth delaying the patch any longer.

PR #2519, opened 2023-10-12
build, test: upgrade to vcrpy>=5 & refresh broken cassettes

dgw:
This is *probably* the ideal solution to my issues with certain VCR cassettes failing locally, since we were on vcrpy 2.x for quite a long time. The error I saw on my local environment went away after upgrading vcrpy (to 5.x) & urllib3 (to 2.x).

Skipping two whole major versions of vcrpy probably isn't ideal, but I like this solution because:

* No more pinning urllib3 at runtime to make a dev requirement happy
* No more snarky incompatibility warning from pip that types-requests wants urllib3 2.x but we have 1.x (seen in CI logs)
* No more weird BrotliDecoderDecompressStream exceptions in my local environment (which didn't happen in CI)

Note that this commit only updates half a dozen YAML cassettes for two plugins whose tests *failed* after upgrading vcrpy. Working cassettes have been left untouched.

### Checklist

- [x] I have read [CONTRIBUTING.md](https://github.com/sopel-irc/sopel/blob/master/CONTRIBUTING.md)
- [x] I can and do license this contribution under the EFLv2
- [x] No issues are reported by `make qa` (runs `make lint` and `make test`)
- [x] I have tested the functionality of the things this change touches

### Notes

I'm sort of expecting some **new and exciting error** to happen in CI that I don't see locally, because that's just what happens when I start messing with anything related to VCR.

All the same, I figured I should probably try to see if updating some dependencies would fix that weird Brotli error I ran into locally while creating new cassettes for #2512 (which worked in CI). Needing to update some cassettes for this patch isn't a huge surprise, since I expect some degree of change in cassette behavior across major versions of the VCR library.

PR #2531, opened 2023-10-27
config: space before zone in `default_time_format`'s default

dgw:
### Description

Dates like `2023-10-26 - 23:45CDT` look a bit weird, and should be presented as `2023-10-26 - 23:45 CDT`. (The default timezone is still UTC, don't worry.)

### Checklist

  - Technically I do get 2 local failures in `make test`, from the `vcrpy` problem that #2519 is meant to fix. CI seems to bypass that, so I should be fine. _\*knocks on wood\*_

### Notes

While making this, I noticed that Sphinx does not output any of the attribute values. Because of that, we have the same default value defined in three different places. The header line for each attribute _ideally_ would look like a module-level constant, e.g. `name_of_config_attribute = ListAttribute('name_of_config_attribute', default=['list', 'of', 'defaults'])` (and `ListAttribute` there would link to the type, of course). It'd save us having to duplicate default values in docstrings, at least.

This patch can't do anything about it, and I couldn't figure out in half an hour of research how to change the docs so Sphinx _would_ do it—but now I'm annoyed, so I wanted to pass that along to everyone else. 🤪

### DIFF
--- a/sopel/config/core_section.py
+++ b/sopel/config/core_section.py
@@ -489,10 +489,10 @@ class CoreSection(StaticSection):
     """
 
     default_time_format = ValidatedAttribute('default_time_format',
-                                             default='%Y-%m-%d - %T%Z')
+                                             default='%Y-%m-%d - %T %Z')
     """The default format to use for time in messages.
 
-    :default: ``%Y-%m-%d - %T%Z``
+    :default: ``%Y-%m-%d - %T %Z``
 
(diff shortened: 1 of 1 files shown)


Exirel (approved):
Hm... this will probably break some tests for my plugins that uses the default time format... but heh, it's not that important tbh.

SnoopJ (approved):
Seems like an easy win for legibility of the default. +1

PR #2532, opened 2023-10-27
dice: bugfixes, type hints, and additional tests

dgw:
What it says on the tin, really; but the highlights (other than added test cases) are:

- Removed dead code in `DicePouch` left over from a refactor in ~2013 (more details in c35c24697e424393b53160ad00ea33307dbad5e5's commit message).
- Fixed fallthrough from "can't drop the lowest _n_ dice" to outputting a result without dropping anything.
- Handle the range of inputs between when `eval_equation` raises `ValueError` (_really_ big) and when the result is small enough to calculate but still too big for Python's `int` -> `str` conversion.
- Refactor control flow and remove responsibility for IRC output from `_roll_dice()` helper.
- Type annotations!

Follow-up to https://github.com/sopel-irc/sopel/pull/2503#issuecomment-1676205118.

### Checklist

### Notes

Sadly, I believe 100% test coverage is impossible using `@plugin.example` alone, without further modifying the plugin's logic. I got `sopel/modules/dice.py` up to 97%, which is still pretty good. A more in-depth explanation of where I ran into trouble is in the commit message for 7ab4be177f5708a4b5a740a46735b03ff33ae458.

This is easily bumped to 8.0.1 if it takes too long to get settled, but I would _rather_ change the constructor interface of `DicePouch` in 8.0.0 even though the `sopel.modules.*` subpackages aren't really part of the public API. The bugfixes are nice, too.


Exirel (approved):
Nice.

sopel/modules/dice.py
```diff
@@ -140,39 +133,50 @@ def _roll_dice(bot, dice_expression):
 
     # Dice can't have zero or a negative number of sides.
     if dice_type <= 0:
-        bot.reply("I don't have any dice with %d sides. =(" % dice_type)
-        return None  # Signal there was a problem
+        raise ValueError("I don't have any dice with %d sides. =(" % dice_type)
```
SnoopJ, line 136: Hmm, something feels slightly gross about "I" in an exception message, here. I understand that the intent is for this message to propagate to the user, but it feels a bit off that the exception carries this payload, in a way that I have a hard time putting my finger on.

sopel/modules/dice.py
```diff
@@ -246,4 +251,10 @@ def _get_pretty_str(dice):
...
         return
 
-    bot.say("%s: %s = %d" % (arg_str_raw, pretty_str, result))
+    try:
+        bot.say("%s: %s = %d" % (arg_str_raw, pretty_str, result))
+    except ValueError:
+        # Converting the result to a string can also raise ValueError if it has
+        # more than int_max_str_digits digits (4300 by default on CPython)
+        # See https://docs.python.org/3.12/library/stdtypes.html#int-max-str-digits
+        bot.reply("I can't display a number that big. =(")
```
SnoopJ, line 260: Niiice.

SnoopJ (approved):
I had one nit about using an exception as a string payload carrier, but it's only a nit. LGTM

sopel/modules/dice.py
dgw: You're gonna laugh… This nit sent me on another refactor of the error handling, touching even more edge cases, and now at 99% file coverage with just one missed line.

But the exceptions no longer carry the message, only data _about_ the error. 😄

Exirel: It's OK, these are very edge cases, and people using IRC are used to weird errors. I'd post-pone any serious refactoring once the plugin is extracted out of core.

sopel/modules/dice.py
```diff
@@ -99,89 +99,153 @@ def get_compressed_string(self):
...
-        d
-        (?P<dice_type>-?\d+)
-        (v(?P<drop_lowest>-?\d+))?
-        $""",
-        dice_expression,
-        re.IGNORECASE | re.VERBOSE)
+class DiceError(Exception):
+    """Custom base exception type."""
+    def __init__(self, *args):
+        super().__init__(*args)
```
Exirel, line 123: Is this necessary? It feels like you don't need more than just the class name and the docstring tbh.

sopel/modules/dice.py
```diff
@@ -99,89 +99,153 @@ def get_compressed_string(self):
...
 
 @plugin.command('roll', 'dice', 'd')
 @plugin.priority("medium")
+@plugin.example(".roll", "No dice to roll.")
+@plugin.example(".roll 2d6+4^2&",
+                "I don't know how to process that. "
+                "Are the dice as well as the algorithms correct?")
+@plugin.example(".roll 65(2)", "I couldn't find any valid dice expressions.")
+@plugin.example(".roll 1d0", "I don't have any dice with 0 sides.")
+@plugin.example(".roll -1d6", "I can't roll -1 dice.")
```
Exirel, line 226: Both are good but can you do `.roll 1d-2`?

Exirel (changes requested):
A couple of nitpicks, while we are at it. Of course I made my comment about not needed a full refactoring *before* realizing you already done it. Well play.

sopel/modules/dice.py
dgw: Hm, I could drop this, yeah. I started out with a defined `message: str` argument (and corresponding property), and `ValueError` as the base class, but clearly things changed and what's left here is useless boilerplate.

sopel/modules/dice.py
dgw: The check is `<= 0` and the `1d0` example hits that. But it's definitely not hard to add an explicit test case for negative faces if you think it's important.

dgw (commented):
At least one fixup incoming next time I'm in my terminal, I see. If force-pushing the squashed history is OK, just ping me on IRC later :)

sopel/modules/dice.py
Exirel: Well it's digital dices, who know what they can do or represent?

Also, I'm pretty sure it'll raise a different error than the 1d0 case!

sopel/modules/dice.py
Exirel: Ah! I see, I had the same idea: start with a `ValueError` as a base class! :grin:

SnoopJ: I'm usually okay with `Exception` as the base class when I'm defining a base for a variety of context-specific exceptions, especially if there's any possibility that they might be analogous to different parts of the built-in hierarchy (i.e. a `DiceError` that is _not_ associated with a bad value).

Don't have a strong preference either way here, but I agree that the initializer can go :stuck_out_tongue:

SnoopJ (approved):
New changes look good to me. Particularly fond of the named capture groups, I love those!

sopel/modules/dice.py
dgw: > Also, I'm pretty sure it'll raise a different error than the 1d0 case!

It doesn't change the error (except for the number). This would be the new example, which I just tested locally:

```py
@plugin.example(".roll 1d-2", "I don't have any dice with -2 sides.")  # verified would pass locally
```

Did you mean `.roll -1d-2`, maybe? For that, `_roll_dice()` checks for `InvalidDiceFacesError` before `NegativeDiceCountError`, so the output is the same as it would be for `.roll 1d-2`.

```py
@plugin.example(".roll -1d-2", "I don't have any dice with -2 sides.")  # verified would pass locally
```

sopel/modules/dice.py
dgw: I've chosen to keep `DiceError` based on `Exception` instead of `ValueError`, lest the custom subtypes be expected to conform to the semantics of how `ValueError` works (i.e., you construct it with a message). Yeah, I know stuff in plugins is basically internal, but I still want it to be sane-ish.

Regardless, the useless initializer is now gone. 🥳

sopel/modules/dice.py
Exirel: Oh! I thought the regex didn't catch the negative sign in front of the dice. Well, apparently it does! It is a different case than 0 anyway, because if we do `<= 0` at the moment, it doesn't mean we'll always do that in the future. And negative number, 0, and positive number are 3 different cases.

dgw: Added a `2d-2` case just for you. 🙂

Exirel (approved):
LGTM

dgw:
Rebasing after #2504, to fix merge conflict, and also #2516, to get the new Python 3.12 status check.

Exirel (approved):

PR #2535, opened 2023-10-29
Pin `packaging` version

SnoopJ:
@Exirel discovered that we have some problems with type analysis on `master` if one has a `packaging` version that is 'too old'.

```
$ mypy --check-untyped-defs sopel
sopel/lifecycle.py:147: error: Unsupported operand types for >= ("tuple[int, ...]" and "None")  [operator]
sopel/lifecycle.py:147: note: Right operand is of type "tuple[int, ...] | None"
```

The above was observed with `packaging==21.3`.

I tried to look in their changelog to see if there was a specific version that fixed the issue we saw, but came up empty-handed. Pinning the lower bound to the latest version is the next best thing to do, and this package has excellent compatibility, so it shouldn't be a problem.

### Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -51,7 +51,7 @@ dependencies = [
     "dnspython<3.0",
     "sqlalchemy>=1.4,<1.5",
     "importlib_metadata>=3.6",
-    "packaging",
+    "packaging>=23.2",
     "sopel-help>=0.4.0",
 ]
 


Exirel (approved):

dgw:
After some question about dropping back to earliest-good after I found [the source of `mypy` considering](https://github.com/pypa/packaging/blob/b10b90a7d1c3afac6852462fa2548db231810adf/packaging/version.py#L135-L137) the version's `release` property optional (`tuple[...] | None`), we elected to just keep this latest-good release constraint. It's fine.

PR #2496, opened 2023-07-21
docs, dev-requirements: upgrade sphinx and try to fix some type issues with documentation

Exirel:
### Description

First, I upgraded Sphinx to the latest version, which requires to upgrade Furo as well. That went well so that's good! Then I added links to the general index and the Python module index, because they are nice.

Then I tried to fix Sphinx autodoc's warnings regarding type annotations. I more or less figure out a way to fix one or two... but then I got stuck very hard: Sphinx, currently, does not handle type annotation properly. For some reason it works for `TypedRule` now but not with some `datetime.datetime` annotation? There are clearly bugs (that are not new to version 7.0 of Sphinx by the way), and I tried few options (such as using `autodoc_type_aliases`, or for a moment I switch to another autodoc extension, with the same issues). Sopel is not an isolated case, [as this issue shows it](https://github.com/sphinx-doc/sphinx/issues/9813).

So yeah, since it's mostly Sphinx's fault, I decided to give up on the remaining warnings. I don't think we can do much: we could, hypothetically, tinker for hours with the code... but I don't see the value here. I already spent quite a few hours on this and I'm not happy so let's not waste any more time.

**Note**: this won't work with Python 3.7 as I use a feature from Python 3.8 (`typing.Protocol`). So until @dgw publish his PR to remove Python 3.7 support, this will be a draft only.

### Checklist

- [x] I have read [CONTRIBUTING.md](https://github.com/sopel-irc/sopel/blob/master/CONTRIBUTING.md)
- [x] I can and do license this contribution under the EFLv2
- [x] No issues are reported by `make qa` (runs `make quality` and `make test`)
- [x] I have tested the functionality of the things this change touches

PR #2539, opened 2023-11-01
plugin: clean up redundant in-docstring parameter type annotations

dgw:
### Description

I realized that leaving parameter types in the docstrings of these decorators is a maintenance nightmare waiting to happen. Our docs show the type-hints in the function signatures very nicely now. Putting types in the parameter list too is just redundant.

### Checklist

  - I mean there'd better not be any, since I didn't touch code.
  - By this I mean I test-built the docs locally. Sphinx still warns about some type annotations/x-refs that it has trouble resolving, but those aren't new (discussed [in #2496](https://github.com/sopel-irc/sopel/pull/2496#issuecomment-1741076230) and variously on IRC).

### DIFF
--- a/sopel/plugin.py
+++ b/sopel/plugin.py
@@ -1454,7 +1454,7 @@ def require_chanmsg(
 def require_account(
     message: Union[Callable, Optional[str]] = None,
     reply: bool = False,
-) -> Callable:  # lgtm [py/similar-function]
+) -> Callable:
     """Decorate a function to require services/NickServ authentication.
 
     :param message: optional message to say if a user without authentication
@@ -1573,7 +1573,7 @@ def require_privilege(
 def require_admin(
     message: Union[Callable, Optional[str]] = None,
     reply: bool = False,
-) -> Callable:  # lgtm [py/similar-function]
+) -> Callable:
     """Decorate a function to require the triggering user to be a bot admin.
 
     :param message: optional message said to non-admin user
@@ -1633,7 +1633,7 @@ def require_admin(
 def require_owner(
     message: Union[Callable, Optional[str]] = None,
     reply: bool = False,
-) -> Callable:  # lgtm [py/similar-function]
+) -> Callable:
     """Decorate a function to require the triggering user to be the bot owner.
 
     :param message: optional message said to non-owner user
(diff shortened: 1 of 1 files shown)


sopel/plugin.py
```diff
@@ -1365,8 +1361,8 @@ def require_chanmsg(
 ) -> Callable:
     """Decorate a function to only be triggerable from a channel message.
 
-    :param str message: optional message said if triggered in private message
-    :param bool reply: use :meth:`~.bot.Sopel.reply` instead of
+    :param message: optional message said if triggered in private message
```
SnoopJ, line 1364: This one is kinda interesting, since the annotation is _more_ misleading than the doc was, but I don't think it poses any actual problem for user understanding. Still an improvement to have one source of truth.

SnoopJ (approved):

sopel/plugin.py
dgw: Yeah, that `Union` is an unfortunate side effect of the established decorator dimorphism (usage supported both with and without `()`; can we call it that?).

But you didn't point out the real mistake in my work two lines below, forgetting to rewrap something 😛

Exirel: Oh, I see what it was about you said on IRC.

My recommendation here is to add a usage example in the docstring that shows that you can use it without parenthesis, as it is a valid syntax of Python and it is supported by this decorator.

Exirel (approved):
Aside from a nitpick mentioned by SnoopJ, I really like that!

sopel/plugin.py
dgw: I've done examples for rather more than just this one: https://github.com/sopel-irc/sopel/pull/2539/commits/fd84b3e73ce2ab910e2a02572c573542e377804e

Enough to warrant re-reviews, I think. Sorry! But no force-push or rebase, so it should be easy to look at only the new changes. 😁

Exirel (approved):
Yeah, still LGTM!

dgw:
I'm going to rebase with a few more tweaks ahead of @SnoopJ's rereview. Found some more things I could clean up, in particular some of the more confusing decorator type-hints (`plugin` no longer even imports `typing.Any` in my local copy 🥳).

Exirel (approved):
Let's go.

dgw:
```irc
13:20:57 <+dgw> SnoopJ, did you want to take another look at #2539 or should I just ship it?
13:21:38 <+SnoopJ> nah, ship it
```

PR #2469, opened 2023-06-09
Add optional tox configuration

SnoopJ:
### Description
This PR adds a `tox` configuration for use with Sopel. It's [dgw's fault](https://github.com/sopel-irc/sopel/pull/2462#issuecomment-1581903253)!

### Checklist
- [x] I have read [CONTRIBUTING.md](https://github.com/sopel-irc/sopel/blob/master/CONTRIBUTING.md)
- [x] I can and do license this contribution under the EFLv2
  - `toxfile.py` does include some plugin code licensed under MIT, not sure if that is a problem in `contrib/`
- [x] No issues are reported by `make qa` (runs `make quality` and `make test`)
- [x] I have tested the functionality of the things this change touches

PR #2540, opened 2023-11-01
privileges, plugin, tools.target, docs: channel privileges as IntFlag

### LINKED ISSUE #2538: docs: privileges now on another page and not linked
### Description

[On this page](https://web.archive.org/web/20231031205435/https://unstable.docs.sopel.chat/plugin/decorators.html), `require_privilege` and `require__bot_privilege` mention needing an `int` for permissions, but those are no longer on that page ([they were in stable](https://web.archive.org/web/20231031205300/https://sopel.chat/docs/plugin/decorators.html)); they are now on [this page](http://web.archive.org/web/20231031205517/https://unstable.docs.sopel.chat/plugin/privileges.html#module-sopel.privileges).

It would be nice if there was a link to the relevant docs or something.

### Reproduction steps

N/A

### Expected behavior

It would be nice if there was a link to the relevant docs or something.

### Relevant logs

```shell
N/A
```

### Notes

N/A

### Sopel version

N/A

### Installation method

`pip install`

### Python version

N/A

### Operating system

N/A

### IRCd

N/A

### Relevant plugins

N/A

dgw:
### Description

Rather than [have someone open an issue](https://github.com/sopel-irc/sopel/issues/2538#issuecomment-1788027761), I decided to Just Do It. I realized that the time to change how these constants are defined/accessed is _now_, as they've been moved to a new location but the new location hasn't appeared in a release yet. After we release 8.0, it would be that much more painful to change this a second time.

**I'm open to bikeshedding on the class name here.** If nothing else, maybe it should be plural to match `sopel.formatting.colors`. I'm keeping in mind the fact that we keep thinking about better access controls on bot functions, and other stuff that might also logically have constants living in `sopel.privileges`. `AccessLevel` is perhaps not the most future-proof choice for avoiding ambiguity. Other ideas included just plain `Level`, `Channel`, `ChannelPrivilege`, `ChanPriv`, and a few more I forgot.

If we do this, the type annotation changes in `plugin` will resolve #2538. Decorators that take this type will link to it in their function signatures' type annotations:

image

### Checklist

  - Thank you @SnoopJ for #2469 and also teaching me how to use it this week.

### Notes

Honestly, `privileges.Channel` made the most sense to me as far as a future collection of different kinds of privileges went. Having two different types named `Channel`—here, and in `tools.target`—sounds like a terrible idea in the long term, though. `tools.target.Channel` even has logic to manipulate the hypothetical `privileges.Channel` type. Getting into "wait, which kind of `Channel` is this?" sounded very not-fun.

### DIFF
--- a/sopel/privileges.py
+++ b/sopel/privileges.py
@@ -15,26 +15,17 @@ __all__ = [
 class AccessLevel(enum.IntFlag):
     """Enumeration of available user privilege levels.
 
-    Comparing privileges
-    ====================
-
-    This class represents privileges as powers of two, with higher values
-    assigned to higher-level privileges::
+    This class represents privileges as comparable, combinable flags. Lower
+    privilege levels compare as *less than* (``<``) higher ones::
 
         >>> from sopel.privileges import AccessLevel
         >>> AccessLevel.VOICE < AccessLevel.HALFOP < AccessLevel.OP \\
         ... < AccessLevel.ADMIN < AccessLevel.OWNER
         True
 
-    Then a user's privileges are represented as a sum of privilege levels::
+    A user's privileges are represented as a combination of privilege levels::
 
-        >>> AccessLevel.VOICE
-        1
-        >>> AccessLevel.OP
-        4
         >>> priv = AccessLevel.VOICE | AccessLevel.OP
-        >>> priv
-        5
 
     This allows using comparators and bitwise operators to compare privileges.
     Here, ``priv`` contains both VOICE and OP privileges, but not HALFOP::
@@ -46,9 +37,9 @@ class AccessLevel(enum.IntFlag):
 
     .. important::
 
-        Do not refer directly to the integer value of a privilege level in your
-        code; the values may change. Use the appropriate member of this class as
-        a reference point instead.
+        Do not hard-code the value of a privilege level in your code; the values
+        may change. Always reference or compare to the appropriate member of
+        this class directly.
 
     """
     # values should behave as ints, but their string representations should
@@ -77,9 +68,7 @@ class AccessLevel(enum.IntFlag):
 
     .. important::
 
-        Not all IRC networks support this privilege mode. If you are writing a
-        plugin for public distribution, ensure your code behaves sensibly if
-        only ``+v`` (voice) and ``+o`` (op) modes exist.
+        Beware: This is one of the `nonstandard privilege levels`_.
 
     """
 
@@ -105,9 +94,7 @@ class AccessLevel(enum.IntFlag):
 
     .. important::
 
-        Not all IRC networks support this privilege mode. If you are writing a
-        plugin for public distribution, ensure your code behaves sensibly if
-        only ``+v`` (voice) and ``+o`` (op) modes exist.
+        Beware: This is one of the `nonstandard privilege levels`_.
 
     """
 
@@ -122,9 +109,7 @@ class AccessLevel(enum.IntFlag):
 
     .. important::
 
-        Not all IRC networks support this privilege mode. If you are writing a
-        plugin for public distribution, ensure your code behaves sensibly if
-        only ``+v`` (voice) and ``+o`` (op) modes exist.
+        Beware: This is one of the `nonstandard privilege levels`_.
 
     """
 
@@ -142,8 +127,6 @@ class AccessLevel(enum.IntFlag):
 
     .. important::
 
-        Not all IRC networks support this privilege mode. If you are writing a
-        plugin for public distribution, ensure your code behaves sensibly if
-        only ``+v`` (voice) and ``+o`` (op) modes exist.
+        Beware: This is one of the `nonstandard privilege levels`_.
 
     """
--- a/sopel/privileges.py
+++ b/sopel/privileges.py
@@ -1,146 +1,174 @@
-"""Constants for user privileges in channels.
-
-Privilege levels
-================
-
-Historically, there were two user privileges in channels:
-
-* :data:`OP`: channel operator, or chanop, set and unset by ``+o`` and ``-o``
-* :data:`VOICE`: the privilege to send messages to a channel with the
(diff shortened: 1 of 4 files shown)


sopel/privileges.py
```diff
@@ -1,146 +1,172 @@
...
-* :data:`ADMIN`: channel admin, above OP and below OWNER, set and unset by
-  ``+a`` and ``-a``
-* :data:`OWNER`: channel owner, above ADMIN and OP, set and unset by ``+q`` and
-  ``-q``
+class AccessLevel(enum.IntFlag):
+    """Enumeration of available user privilege levels.
 
-.. important::
+    Privilege levels
+    ================
```
Exirel, line 17: I think this should stay at the module level, as these explanations are not tied to the implementation details of "AccessLevel is an ENUM now". It was thought out as a standalone explanation for "what are privilege levels for Sopel", giving more context than the implementation details could.

sopel/privileges.py
```diff
@@ -1,146 +1,172 @@
...
 
-    Not all IRC networks support these added privilege modes. If you are
-    writing a plugin for public distribution, ensure your code behaves sensibly
-    if only +v (voice) and +o (op) modes exist.
+    Historically, there were two user privilege levels in a channel:
 
-Compare privileges
-==================
+    * :data:`~AccessLevel.OP`: channel operator, set and unset by modes ``+o``
```
Exirel, line 21: I haven't tested this, but maybe this would work as well?

```suggestion
    * :attr:`~AccessLevel.OP`: channel operator, set and unset by modes ``+o``
```

sopel/privileges.py
dgw: You make a lot of sense, as usual. I pulled this section out to the `docs/source/plugin/privileges.rst` intro section, before we get into how to work with privileges _within Sopel_; see b8c2d99e2890dc7f06ec8a65c0c8722779163472 and/or updated [deploy preview](https://deploy-preview-2540--sopel-unstable-docs.netlify.app/plugin/privileges#user-privileges). 🙂

sopel/privileges.py
```diff
@@ -1,146 +1,147 @@
...
 
-.. important::
+    Comparing privileges
+    ====================
 
-    Not all IRC networks support these added privilege modes. If you are
-    writing a plugin for public distribution, ensure your code behaves sensibly
-    if only +v (voice) and +o (op) modes exist.
+    This class represents privileges as powers of two, with higher values
+    assigned to higher-level privileges::
```
SnoopJ, line 20: Nit: I'm not sure how I feel about recapitulating the behavior of `auto()` here, it seems like a detail that could just as easily be omitted.

sopel/privileges.py
```diff
@@ -1,146 +1,147 @@
...
-This allows to use comparators and bitwise operators to compare privileges::
+    .. important::
 
-    >>> priv >= OP
-    True
-    >>> bool(priv & HALFOP)
-    False
+        Do not refer directly to the integer value of a privilege level in your
+        code; the values may change. Use the appropriate member of this class as
+        a reference point instead.
```
SnoopJ, line 49: I think this is part of my ambivalence about describing the _powers of two_ implementation: it reads to me as nudging users a bit more towards thinking about flags as integers.

sopel/privileges.py
```diff
@@ -1,146 +1,147 @@
...
 
-.. versionadded:: 4.1
-.. versionchanged:: 8.0
-   Moved into :mod:`sopel.privileges`.
+        Not all IRC networks support this privilege mode. If you are writing a
+        plugin for public distribution, ensure your code behaves sensibly if
+        only ``+v`` (voice) and ``+o`` (op) modes exist.
 
-.. important::
+    """
```
SnoopJ, line 110: Nit: is it possible to express this as a cross-reference to the top-level description of why these flags may not be supported by all servers? Bit beyond my Sphinx-fu, but it seems like it might be slightly better to do `Not all IRC networks support this privilege mode, see <ref>`

SnoopJ (approved):

sopel/privileges.py
```diff
@@ -1,146 +1,147 @@
-"""Constants for user privileges in channels.
+"""Constants for user privileges in channels."""
+# Copyright 2023 dgw, technobabbl.es
```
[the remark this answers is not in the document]
dgw: The commit message for 12b00b0b424e8fb9896ad48b8368cc1fcb51379b explains it. The file might have already existed, but git said "rewrite sopel/privileges.py (97%)" so it's basically all me. Some of the old documentation is left over, I guess, so that number's perhaps a bit high, but I still think it's justified. I did forget the license part, though, so I'll add that.

Poking around, I saw that there are plenty of other files without copyright notices. We should probably fix that, though analyzing each file's history to decide what year(s) and name(s) go in the header could be annoying. Sounds like a good lazy-patch project for 8.1 to me. 😉

sopel/privileges.py
dgw: A few lines below you can see the original documentation for the module-level constants, which had this exact language. There's an argument to be made that this whole introduction to the idea of privilege flags can be improved, but I hope you don't pin the wording on me. I just moved around what was already there. 😝

If you really want to take care of it now, I'll see what I can come up with later in the week.

sopel/privileges.py
dgw: It's a hedge against Sphinx showing the actual value of each enum member in generated docs. I'm not sure how to turn it off, but would it help to omit the "integer" reference here and only talk about not hard-coding flag values (of an unspecified type)?

sopel/privileges.py
dgw: Psych, I had a brainstorm tonight after all.

SnoopJ (approved):
Looks great :shipit:

sopel/privileges.py
SnoopJ: That would resolve my ambivalence about the wording, yea

PR #2530, opened 2023-10-27
calc: improve help output, test coverage (now 100%)

dgw:
Error cases weren't covered by the `@plugin.example` tests. Fixed that.

Also declared a few more examples that should appear in the help output, to showcase more of the plugin's capabilities.

Follow-up to https://github.com/sopel-irc/sopel/pull/2503#issuecomment-1676205118.

### Checklist

- [x] I have read [CONTRIBUTING.md](https://github.com/sopel-irc/sopel/blob/master/CONTRIBUTING.md)
- [x] I can and do license this contribution under the EFLv2
- [x] No issues are reported by `make qa` (runs `make lint` and `make test`)
- [x] I have tested the functionality of the things this change touches

Issue #2508, opened 2023-09-21
tools.calculation: `ExpressionEvaluator.Error` should not be defined inside another class


dgw:
Because the `Error` raised by `ExpressionEvaluator` is defined here, inside the `ExpressionEvaluator` class itself:

https://github.com/sopel-irc/sopel/blob/274470ec29429a26bd0faed12984ce6ac09cf949/sopel/tools/calculation.py#L22-L24

Downstream consumers of the submodule's main (relevant) utility function can't just cleanly import the error type. They have to access it either as an attribute of one of the classes (`ExpressionEvaluator` or its child `EquationEvaluator`), or as an attribute of the "function" `eval_equation()` (which is actually an instance of `EquationEvaluator`, but 🤫).

This is more than a little weird as Python exception types go. Plus, if someone just lets Sopel's built-in error handler catch the exception, you get "Unexpected Error" in the output, because the class name is just "Error". Not very descriptive, that.

### Proposal
Let's move the `ExpressionEvaluator.Error` type out into its own standalone class. In fact, it should probably become a full error hierarchy to cover the range of different error cases available. (WIP tests for the `tools.calculation` module currently have to substring-match the exception message to make sure the "right" `Error` is being raised.)

I've toyed with ideas for keeping the `ExpressionEvaluator.Error` name around for a while in case anyone else is using it (like I've just proposed doing in #2507), but am not dead set on doing so. The new standalone error type (hierarchy root, tentatively `CalculationError`) could start its life as a subclass of the old one, and then switch to subclassing `Exception` directly in the next major version. There won't be an effective hook point to show any deprecation warnings, though, so anyone using code that relies on the old (current) name will have to read release notes. 🤷‍♂️

PR #2543, opened 2023-11-03
tools, tools.calculation: docs/type-hint improvements, API fixes, better test coverage

dgw:
### Description

This starts with some tweaks to the docstring data in `tools` (`.. versionadded::` kinda stuff) and `tools.calculation` that I found sitting in a forgotten branch; apparently those smaller tweaks never reached the point of a full PR.

That stalled work's finally done: `sopel/tools/calculation.py` is now fully type-annotated and has 100% test coverage. Along the way I discovered a few API glitches (e.g. incorrectly documented parameter types, missing parameters) and fixed 'em.

I admit some of the test assertions are pretty ugly, but with only one custom exception type it's necessary to make sure that _the right kind_ of `ExpressionEvaluator.Error` is being raised. A future refactoring opportunity, that is (which #2508 already tracks).

### Checklist

### Notes

The framework for fully exercising `tools.calculation` in tests came from both #2530 and https://github.com/sopel-irc/sopel/pull/2464#issuecomment-1676170000 — thanks, @SnoopJ! Getting around to this would have taken a lot longer if you didn't kick-start it.

Just one thing I haven't cracked: Two new Sphinx warnings from type annotations that it can't resolve to cross-reference links:

```
/home/dgw/github/sopel-irc/sopel/sopel/tools/calculation.py:docstring of
sopel.tools.calculation.ExpressionEvaluator:1: WARNING: py:class reference
target not found: ast.operator
/home/dgw/github/sopel-irc/sopel/sopel/tools/calculation.py:docstring of
sopel.tools.calculation.ExpressionEvaluator:1: WARNING: py:class reference
target not found: ast.unaryop
```

Those are, however, the types required to satisfy `mypy`. Trying to use the CamelCase classes in `ast`'s documentation ([`BinOp`](https://docs.python.org/3.12/library/ast.html#ast.BinOp) and [`UnaryOp`](https://docs.python.org/3.12/library/ast.html#ast.UnaryOp)) never passed type-checking. The class hierarchy there doesn't work like one might think it does (or should).

At the moment, I'm ready to give up on these like what eventually happened with the unfixed warnings left over in #2496. Really seems like Sphinx just can't handle certain things. Even the Python docs themselves [reference these](https://github.com/python/cpython/blob/0d3df272fbd131bff7f02d4d4279ad1e35081121/Doc/library/ast.rst?plain=1#L2151C71-L2151C71), and [the rendered HTML](https://docs.python.org/3.12/library/ast.html#ast.AsyncFor) shows them as `xref py py-class` but they're not linked.

### DIFF
--- a/sopel/tools/calculation.py
+++ b/sopel/tools/calculation.py
@@ -13,6 +13,10 @@ import ast
 import numbers
 import operator
 import time
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from typing import Callable, Optional
 
 __all__ = ['eval_equation']
 
@@ -30,16 +34,20 @@ class ExpressionEvaluator:
 
     class Error(Exception):
         """Internal exception type for :class:`ExpressionEvaluator`\\s."""
-        pass
 
-    def __init__(self, bin_ops=None, unary_ops=None):
+    def __init__(
+        self,
+        bin_ops: Optional[dict[type[ast.operator], Callable]] = None,
+        unary_ops: Optional[dict[type[ast.unaryop], Callable]] = None
+    ):
         self.binary_ops = bin_ops or {}
         self.unary_ops = unary_ops or {}
 
-    def __call__(self, expression_str, timeout=5.0):
+    def __call__(self, expression_str: str, timeout: float = 5.0):
         """Evaluate a Python expression and return the result.
 
-        :param str expression_str: the expression to evaluate
+        :param expression_str: the expression to evaluate
+        :param timeout: timeout for processing the expression, in seconds
         :raise SyntaxError: if the given ``expression_str`` is not a valid
                             Python statement
         :raise ExpressionEvaluator.Error: if the instance of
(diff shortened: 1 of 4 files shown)


sopel/tools/calculation.py
```diff
@@ -176,13 +186,11 @@ def pow_complexity(num, exp):
         return exp ** 1.590 * num.bit_length() ** 1.73 / 36864057619.3
 
 
-def guarded_pow(num, exp):
+def guarded_pow(num: Union[int, float], exp: Union[int, float]):
```
SnoopJ, line 189: Nit/:fire:fun:fire: fact: you can just annotate `float` here, because `int` coerces to `float` just fine as far as typing is concerned, which is [explicitly-defined behavior in PEP 484](https://peps.python.org/pep-0484/#the-numeric-tower):

> when an argument is annotated as having type float, an argument of type int is acceptable

(Unless I'm wrong and you checked that and we actually _do_ need the `Union`, in which case this needs a comment explaining why :stuck_out_tongue_winking_eye:)

```suggestion
def guarded_pow(num: float, exp: float):
```

sopel/tools/calculation.py
```diff
@@ -223,8 +238,8 @@ def __init__(self):
             unary_ops=self.__unary_ops
         )
 
-    def __call__(self, expression_str):
-        result = ExpressionEvaluator.__call__(self, expression_str)
+    def __call__(self, expression_str: str, timeout: Union[float, int] = 5.0):
```
SnoopJ, line 241: Same nit as above, `float` allows `int`

```suggestion
    def __call__(self, expression_str: str, timeout: float = 5.0):
```

SnoopJ (approved):
LGTM, some fine Sopelunking here! 100% coverage train, choo choo! :steam_locomotive::dash:

sopel/tools/calculation.py
dgw: > (Unless I'm wrong and you checked that and we actually _do_ need the `Union`, in which case this needs a comment explaining why 😜)

Nah, I just didn't think of the inheritance there. Found several places with an unnecessary `Union` of this kind—all of them, in fact. The file no longer needs `Union` at all.

PR #2552, opened 2023-11-09
irc, bot, builtins: add & use `AbstractBot.make_identifier_memory()` helper

dgw:
### Description

Here is a shortcut way of getting a new `SopelIdentifierMemory` with the `identifier_factory` kwarg filled in for you so it automatically uses the bot's awareness of casemapping from the IRC server's ISUPPORT tokens.

Manual casemapping selection is still possible by using the `SopelIdentifierMemory` constructor directly, of course, as before.

Using this new method in the bot to initialize its `users` and `channels` automatically adds this to test coverage:

image

_(NB: That is an older coverage run, from before I expanded the docstring.)_

### Checklist

### DIFF
--- a/sopel/bot.py
+++ b/sopel/bot.py
@@ -75,9 +75,7 @@ class Sopel(irc.AbstractBot):
         self.modeparser = modes.ModeParser()
         """A mode parser used to parse ``MODE`` messages and modestrings."""
 
-        self.channels = tools.SopelIdentifierMemory(
-            identifier_factory=self.make_identifier,
-        )
+        self.channels = self.make_identifier_memory()
         """A map of the channels that Sopel is in.
 
         The keys are :class:`~sopel.tools.identifiers.Identifier`\\s of the
@@ -85,9 +83,7 @@ class Sopel(irc.AbstractBot):
         which contain the users in the channel and their permissions.
         """
 
-        self.users = tools.SopelIdentifierMemory(
-            identifier_factory=self.make_identifier,
-        )
+        self.users = self.make_identifier_memory()
         """A map of the users that Sopel is aware of.
 
         The keys are :class:`~sopel.tools.identifiers.Identifier`\\s of the
--- a/sopel/builtins/find.py
+++ b/sopel/builtins/find.py
@@ -18,14 +18,11 @@ import re
 
 from sopel import plugin
 from sopel.formatting import bold
-from sopel.tools import SopelIdentifierMemory
 
 
 def setup(bot):
     if 'find_lines' not in bot.memory:
-        bot.memory['find_lines'] = SopelIdentifierMemory(
-            identifier_factory=bot.make_identifier,
-        )
+        bot.memory['find_lines'] = bot.make_identifier_memory()
 
 
 def shutdown(bot):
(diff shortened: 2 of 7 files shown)


dgw:
That'll teach me to skip at least `make lint` because "it's only a docs patch". 😅

SnoopJ (approved):
LGTM, nice little ergonomics improvement for the core, builtins, _and_ plugin developers.

Exirel (changes requested):
Code-wise, I'm OK. I think however that this would be better with at least one automated test, probably in `test/test_coretasks.py`, by following the same ideas as the `test_handle_issuport_casemapping` test.

dgw:
> this would be better with at least one automated test, probably in `test/test_coretasks.py`, by following the same ideas as the `test_handle_issuport_casemapping` test.

I'm unclear on what you want to test for that isn't already covered by use of the new method in other already-tested locations.

If it's about the memory's existing keys updating with the new casemapping behavior when `CASEMAPPING` comes in from the server, _only_ the bot's own `_nick` is rebuilt in that case. Previously-constructed `Identifier`s anywhere else in the program retain their old `casemapping` (and their old `chantypes`), and this isn't unique to the `SopelIdentifierMemory` type. There's no reasonable way to `bot.rebuild_nick()` for all existing `Identifier` objects out there.

About the best I can do—without rewriting a bunch of stuff and taking this PR far away from the original scope—is to check adding "the same" nick that now is "not the same" under the new casemapping rules. Even that's problematic, though: if `Test[a]` is added to the memory before `CASEMAPPING` comes in (using the default `rfc1459` behavior), and then `CASEMAPPING=ascii` comes in, you still won't be able to add `test{a}` to the memory because the _old_ `rfc1459` casemapping already matches that:

```py
# CASEMAPPING token change doesn't affect previously existing Identifiers
assert memory['Test{a}'] == False

memory['tEST{a}'] = True  # this overwrites the previous value; it doesn't add a new key
assert len(memory) == 2  # this fails; len(memory) is (still) 1
```

Plugins probably just shouldn't construct `Identifer` before the bot has connected (e.g. in `setup()`). Do we need a guard on these methods that raises an error if the bot isn't ready yet? I've just thought that we could (in the future) give `make_identifier()` a manager that keeps track of all `Identifier`s it's created and then all of them could be updated if the `CASEMAPPING` changes. But that kinda sounds like overkill…

dgw:
I pushed new stuff based on IRC discussion, not just out of the blue. I promise!

Exirel:
> I'm unclear on what you want to test for that isn't already covered by use of the new method in other already-tested locations.

As discussed on IRC, the TL;DR is: this method need to be covered by direct invocation in the test suite, so if we stop using it internally, it is still tested and properly covered.

For everything else, the new commit with the added tests is the way to go.

Exirel (approved):

dgw:
> As discussed on IRC, the TL;DR is: this method need to be covered by direct invocation in the test suite, so if we stop using it internally, it is still tested and properly covered.

I added the specific tests of `bot.make_identifier()` and `bot.make_identifier_memory()` in `test/test_irc.py` while I was here, to give _both_ related methods a direct invocation from the test suite that isn't incidental to other test cases like `test/test_coretasks.py`'s previously-existing `test_handle_isupport_casemapping` unit, in addition to the newly added `test_handle_isupport_casemapping_identifiermemory` test.

If you have another idea before merge (or after, idc), just holler @Exirel :)

Exirel:
I already approved that *years* ago. :sunglasses:

dgw:
We had ourselves a spirited discussion about the idea of having `make_identifier_memory()` take an argument like the `SopelIdentifierMemory()` constructor now can (after #2525), with a lot of shifting opinions as more considerations came up.

I _think_ the three of us (@Exirel, @SnoopJ, and myself) landed on the idea that it's fine to ship this as-is and come back to add an optional argument in 8.1.0. At the same time, we can clean up the messy `*args` argspecs throughout `tools.memories`.

Assuming we don't change our consensus in the next few days, this will get merged and **this comment should be the basis of a new issue for 8.1** regarding signatures both here and in the `Memory` types.

PR #2533, opened 2023-10-28
docs: install and running instance documentation

Exirel:
### Description

This is my attempt at working on #2497 with a version that I think could be good enough for Sopel 8.0.0.

I keep  it as a draft to get feedfback more than review i.e. more about what content should be or shouldn't be in here.

I took a lot from [Installing](https://sopel.chat/usage/installing/) to know what to put in here. I think the next step (here or in a future PR?) would be to add more from the [per-channel config](https://sopel.chat/usage/per-channel-configuration/) page.

What I mean is: what do you think?

### Checklist

- [x] I have read [CONTRIBUTING.md](https://github.com/sopel-irc/sopel/blob/master/CONTRIBUTING.md)
- [x] I can and do license this contribution under the EFLv2
- [x] No issues are reported by `make qa` (runs `make lint` and `make test`)
- [x] I have tested the functionality of the things this change touches

Note: couldn't run `make qa` because mypy 1.6.1 fails on a line that is already in current master.

PR #2558, opened 2023-11-12
docs: set `maxdepth` or `titlesonly` on overly-long TOCs

dgw:
### Description

I set out to clean up some of the documentation appearance & make things more usable, taking into consideration the fact that we use the Furo theme now and detailed in-page TOCs are kinda redundant in a lot of places thanks to the dual sidebars.

In addition to the original impetus for this patch—`maxdepth` or `titlesonly` options on TOCs where needed—this patch also cleans up some warnings generated by `SopelIdentifierMemory` inheriting docstrings from `dict`, tells autodoc not to evaluate function argument defaults before display (eliminates `<class 'sopel.foo.nonsense.ClassName'>` from being shown, which is very ugly), and also improves the docstring of `tools.get_sendable_message()` because I don't want to open a tiny PR for that.

I'm also aware of what #2533 plans to change, and therefore avoided touching any of that PR's renamed/modified rST files at all. The comment in `index.rst` about adding install docs here is also gone; they're coming soon!

### Checklist

  - `make test` not so relevant, as no code changed; but I still ran `make lint` since I [learned my lesson](https://github.com/sopel-irc/sopel/pull/2552#issuecomment-1804700159) after skipping it earlier this week. 😅
  - Oh so many rebuilds… I'm as mad as I'll ever be now about those spurious `:py:class: reference not found` warnings we get from using type aliases for things like `Casemapping` or `IdentifierFactory`. Sphinx can't fix that soon enough.

### DIFF
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -20,6 +20,7 @@ Documentation
 
 .. toctree::
     :maxdepth: 2
+    :titlesonly:
 
     run
     plugin
--- a/docs/source/package/config.rst
+++ b/docs/source/package/config.rst
@@ -2,7 +2,14 @@
 Configuration functionality
 ===========================
 
+The description of :mod:`.config` below is a quick guide to the basic structure
+of configuration settings in Sopel. Consult the :mod:`.config.types` module for
+more specific information about the available kinds of options, and
+:mod:`.config.core_section` for examples of how the different setting types are
+practically used.
+
 .. toctree::
+   :titlesonly:
 
    config/types
    config/core_section
--- a/docs/source/package/irc.rst
+++ b/docs/source/package/irc.rst
@@ -2,13 +2,20 @@
 IRC Core Implementation
 =======================
 
+Operation of the :mod:`sopel.irc` core is supported by these utility submodules:
+
 .. toctree::
+    :titlesonly:
 
     irc/backends
     irc/modes
     irc/isupport
     irc/capabilities
     irc/utils
+
+
+sopel.irc
+=========
 
 .. automodule:: sopel.irc
     :members:
--- a/docs/source/package/plugins.rst
+++ b/docs/source/package/plugins.rst
@@ -9,6 +9,7 @@ Plugin Internal Machinery
    section exists primarily for the aid of Sopel’s core development.
 
 .. toctree::
+    :maxdepth: 2
 
     loader
     plugins/plugins
--- a/docs/source/package/tools.rst
+++ b/docs/source/package/tools.rst
@@ -2,7 +2,11 @@
 Additional Tools
 ================
 
+Sopel provides utilities for some commonly used plugin functionality. Most of
+these utility features have been collected into submodules for convenience:
+
 .. toctree::
+   :titlesonly:
 
    tools/calculation
    tools/events
@@ -12,6 +16,8 @@ Additional Tools
    tools/target
    tools/time
    tools/web
+
+Miscellaneous tools that don't fit any particular category appear below.
 
 
 sopel.tools
--- a/docs/source/tests.rst
+++ b/docs/source/tests.rst
@@ -2,11 +2,20 @@
 Testing tools
 =============
 
+The best code comes with unit tests. Sopel provides a :doc:`pytest plugin
+<tests/pytest>` with standard fixtures; :doc:`factories <tests/factories>` and
+:doc:`mocks <tests/mocks>` to support writing tests; and the
+:func:`~.tests.rawlist` function to help analyze the mocked-up bot's output.
+
 .. toctree::
+   :titlesonly:
 
    tests/pytest
    tests/factories
    tests/mocks
+
+sopel.tests
+===========
 
 .. automodule:: sopel.tests
    :members:
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -125,6 +125,7 @@ autodoc_type_aliases = {
     'ModeDetails': 'sopel.irc.modes.ModeDetails',
     'PrivilegeDetails': 'sopel.irc.modes.PrivilegeDetails',
 }
+autodoc_preserve_defaults = True
 
 
 # -- Options for HTML output ---------------------------------------------------
--- a/docs/source/configuration.rst
+++ b/docs/source/configuration.rst
@@ -232,13 +232,13 @@ In that case:
 
 .. seealso::
 
-   Sopel uses the built-in :func:`ssl.wrap_socket` function to wrap the socket
-   used for the IRC connection.
+   Sopel uses the built-in :meth:`ssl.SSLContext.wrap_socket` function to wrap
+   the socket used for the IRC connection.
 
 .. note::
 
    Sopel will try to look at one of these files for the CA certs pem file
-   required by :func:`ssl.wrap_socket`:
+   required by :meth:`ssl.SSLContext.wrap_socket`:
 
    * ``/etc/pki/tls/cert.pem``
    * ``/etc/ssl/certs/ca-certificates.crt`` (Debian)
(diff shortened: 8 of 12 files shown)


docs/source/package/plugins.rst
```diff
@@ -9,6 +9,7 @@ Plugin Internal Machinery
    section exists primarily for the aid of Sopel’s core development.
 
 .. toctree::
+    :titlesonly:
```
SnoopJ, line 12: This also feels a little terse, I prefer the output of `:maxdepth: 2` here as well

```suggestion
    :maxdepth: 2
```

The output is definitely still busy, but it's 1.5 pages of busy for me instead of the 7 pages that current unstable docs produces.

image

docs/source/package/tools.rst
```diff
@@ -2,7 +2,11 @@
 Additional Tools
 ================
 
+Sopel provides utilities for some commonly used plugin functionality. Most of
+these utility features have been collected into submodules for convenience:
+
 .. toctree::
+   :titlesonly:
```
SnoopJ, line 9: This one I think I agree with, but the rendered docs would be _much_ more helpful if the `On this page` bar listed every anchor reachable on the page.

docs/source/tests.rst
```diff
@@ -2,11 +2,20 @@
 Testing tools
 =============
 
+The best code comes with unit tests. Sopel provides a :doc:`pytest plugin
+<tests/pytest>` with standard fixtures; :doc:`factories <tests/factories>` and
+:doc:`mocks <tests/mocks>` to support writing tests; and the
+:func:`~.tests.rawlist` function to help analyze the mocked-up bot's output.
+
```
SnoopJ, line 9: I approve of the brief narrative added, do we also want to mention here that we do some test generation from plugin examples?

SnoopJ (approved):
I had some nits, but overall LGTM

docs/source/index.rst
```diff
@@ -12,14 +12,11 @@ Quick links
...
 * `Source code <https://github.com/sopel-irc/sopel>`_
 
-..
-    Eventually, add installation instructions to Sphinx docs?
-
 Documentation
 -------------
 
 .. toctree::
-    :maxdepth: 2
```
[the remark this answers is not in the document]
dgw: Played with this a bit more, and will probably switch it to `:titlesonly:` with the original `:maxdepth: 2`.

What really bugged me was the inclusion of non-page stuff, and just adding `:titlesonly:` solves that while still giving decent detail in the overview.

docs/source/package/irc.rst
```diff
@@ -3,6 +3,7 @@ IRC Core Implementation
 =======================
 
 .. toctree::
+    :hidden:
```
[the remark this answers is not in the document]
dgw: "On this page" includes only stuff that's, well, "On this page". Furo specifically splits what's _on_ the page (right sidebar) from _what pages_ exist (left sidebar).

I still think `:maxdepth: 2` is too verbose for the top of a module page. (In fact, the more of these I look at, the stronger my feeling that the toctrees should come _after_ the main module content. But to your points about navigability, that would make things less glanceable.)

What I like here is a short introductory sentence explaining why there's a list of stuff up top, followed by just the page titles:

image

docs/source/package/plugins.rst
dgw: You're probably right, this can afford to be a bit more verbose as it isn't sitting on top of any "real content" and is just a index page for the rest of the plugin machinery docs.

docs/source/package/tools.rst
dgw: What anchor(s) is it missing? It doesn't include the top-level heading (page title), but looks like everything else is there. (The subpages are not "anchor[s] reachable on the page"; they are subpages.)

docs/source/tests.rst
dgw: Hmmmmmmmmmm, perhaps not _yet_. We still have that whole thing with `vcrpy` compatibility hanging over us at the moment, so I don't _really_ want to encourage wider use of the `@example` decorator just now. 😅

(I also go back and forth on whether `@example` even deserves to be part of the plugin API, beyond the basic "Here's how you call the command, and here's some sample output." for help reasons. Overloading it to generate tests is cool, but not something I'm positive that we'll want to keep long-term.)

docs/source/package/tools.rst
SnoopJ: Oh, you're right, I guess that's something I've been slightly misunderstanding. I know that the entire tree is in practice visible in the left side-bar, but I think it's kinda hard to drill down into those and that is probably influencing my remarks here.

dgw:
@SnoopJ "Back to you, Chet" to resolve or respond to the outstanding review conversations.

SnoopJ (approved):
Thoroughly de-nitted as far as I'm concerned :shipit:

docs/source/tests.rst
Exirel: Sorry to hack the conversation to say: I even proposed, a year ago or so, to have 2 separate decorators, one for auto-test, one for actual usage examples, so the test framework doesn't clash with the documentation and the plugin definition. 😁

docs/source/package/config.rst
```diff
@@ -2,7 +2,14 @@
...
 ===========================
 
+The description of :mod:`.config` below is a quick guide to the basic structure
+of configuration settings in Sopel. Consult the :mod:`.config.types` module for
+more specific information about the available kinds of options, and
+:mod:`.config.core_section` for examples of how the different setting types are
+practically used.
+
 .. toctree::
+   :hidden:
```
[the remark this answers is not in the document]
dgw: The main subject of each subpage is linked in the introductory paragraph; is that insufficient? It seems silly to put multiple links to (effectively) the same location clustered together like:

image

I can make the links go to the pages instead easily enough (pending wording tweaks):

image

docs/source/tests.rst
```diff
@@ -2,11 +2,20 @@
...
 .. toctree::
+   :hidden:
```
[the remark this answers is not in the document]
dgw: As suggested above in `package/config.rst`, here the page titles _are_ linked directly in the intro paragraph.

Same reaction from me about putting duplicate links to those same locations in bulleted-list form right below.

docs/source/package/config.rst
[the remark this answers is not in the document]
Exirel: It is, because a paragraph is a good way to get an introduction, to have a smooth landing in unknown territory. But it is very ineffective when someone wants to navigate quickly from section to section: scanning a list is far easier and faster for the brain than it is finding a link in a paragraph.

And I'm talking from experience: I absolutely didn't realize there were link to subpages in the first paragraph. Especially with the light mode, they are far less noticeable than with the dark mode. The first screenshot you have is exactly what I'd prefer.

It is not silly: different presentations, in documentation, have different purpose. They are used by different readers, and the brain scan differently each format. My brain, for instance, will automatically skip any paragraph of text when I'm traversing the documentation for a specific module/function reference, and I'm focusing only on finding the appropriate menu or list. When I'm discovering a module (for example in the Python documentation), I read the paragraph, and more or less ignore the menu.

docs/source/tests.rst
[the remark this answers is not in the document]
Exirel: And same response from me. 😁

dgw:
I lack the energy to fight for using `:hidden:` in a few selected locations, so the win goes to @Exirel.

Latest branch update added a fix for Sphinx warnings about being unable to find <tt>:py:class:\`datetime\`</tt> in function signatures, as well. Now we have "only" 18 warnings (14 from that pesky issue with type aliases; 2 for internals of `ast`; and 2 for bad links to `ssl.wrap_socket`, which I might try to fix next).

Exirel:
> Sphinx warnings about being unable to find :py:class:`datetime` in function signatures

I hate you Sphinx. I really hate you.

Exirel (approved):

PR #2559, opened 2023-11-12
docs: add warning when rendering preview docs

SnoopJ:
### Description

I had not realized that the [unstable documentation](https://unstable.docs.sopel.chat/) was being regularly built against `master`, and I thought it would be helpful if these preview docs included a warning about their preview-ness as well as a reference to the specific commit they were built against.

The warning added by this changeset is included on all pages using Sphinx's [`rst_prolog` directive](https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-rst_prolog), but only rendered when the Sopel version is a pre-release (i.e. `.devN`) version.

Below is an example of how preview documentation renders (the link points to the relevant commit on GitHub)

image

And here's what the docs look like for a release version:

image

### Checklist

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -10,7 +10,10 @@ from __future__ import generator_stop
 # All configuration values have a default; values that are commented out
 # serve to show the default.
 
+import subprocess
 from datetime import date
+
+from packaging.version import Version
 
 from sopel import __version__
 
@@ -308,3 +311,26 @@ texinfo_documents = [
 
 # How to display URL addresses: 'footnote', 'no', or 'inline'.
 #texinfo_show_urls = 'footnote'
+
+
+try:
+    if Version(__version__).is_prerelease:
+        tags.add("preview")
+
+    is_dirty = bool(subprocess.check_output(["git", "status", "--untracked-files=no", "--porcelain"], text=True).strip())
+    commit_hash = subprocess.check_output(["git", "rev-parse", "HEAD"], text=True).strip()
+    git_describe = subprocess.check_output(["git", "describe", "--tags"], text=True).strip() + ("-dirty" if is_dirty else "")
+    github_ref = "https://github.com/sopel-irc/sopel/commit/{commit_hash}".format(commit_hash=commit_hash)
+    build_info = "(built against `{git_describe} <{github_ref}>`_)".format(git_describe=git_describe, github_ref=github_ref)
+except Exception as exc:
+    build_info = "(built against an unknown commit)"
+
+
+rst_prolog = """
+.. only:: preview
+
+    .. warning:: This is preview documentation for Sopel |version| {build_info}.
+
+                 Click `here <https://sopel.chat/docs/>`_ for the latest stable documentation.
+
+""".format(build_info=build_info)


docs/source/conf.py
```diff
@@ -308,3 +311,26 @@
...
 #texinfo_show_urls = 'footnote'
+
+
+try:
+    if Version(__version__).is_prerelease:
+        tags.add("preview")
+
+    is_dirty = bool(subprocess.check_output(["git", "status", "--untracked-files=no", "--porcelain"], text=True).strip())
+    commit_hash = subprocess.check_output(["git", "rev-parse", "HEAD"], text=True).strip()
+    git_describe = subprocess.check_output(["git", "describe", "--tags"], text=True).strip() + ("-dirty" if is_dirty else "")
```
SnoopJ, line 322: I have been very much enjoying the [`versioningit`](https://pypi.org/project/versioningit/) project over the last year or so, which can automate away this sort of stuff (at the expense of reading the version from a `_version.py` file or somesuch instead of using `importlib` metadata). But I tried to write this patch using only what's already present, and it should work fine as long as `git` is present and the docs are being built from the repository, and degrade gracefully if there are problems (i.e. you'll still get a warning box if `git` is not present or the CWD violates this code's assumptions)

Exirel (commented):
At the moment I'm only commenting because I don't know if I like the result more than I hate the process: using subprocess in Python is something I'm very averse to for many reasons.

SnoopJ:
> using subprocess in Python is something I'm very averse to for many reasons.

Is that a generalized aesthetic concern, or are you thinking of specific failure modes here?

An alternative that occurred to me while putting this PR together would be to add a `Makefile` target for the preview documentation that retrieves the relevant commit information and passes it to Sphinx. I don't know what the specific mechanism would be, but I'm sure we could pass it in.

docs/source/conf.py
```diff
@@ -308,3 +311,29 @@
...
+    build_info = "(built against `{commit_hash}{dirty} <{github_ref}>`_)".format(
+        commit_hash=commit_hash[:7],
+        dirty="-dirty" if is_dirty else "",
+        github_ref=github_ref
+    )
+else:
+    build_info = "(built against an unknown commit)"
+
+if Version(__version__).is_prerelease or commit_hash:
+    tags.add("preview")
```
SnoopJ, line 329: Note: with the change to an environment variable, the `preview` tag is added if the user set the environment variable, even if the `__version__` is not a pre-release. It seems prudent that the output of `make docs_preview` would always set the tag

docs/source/conf.py
```diff
@@ -308,3 +311,29 @@
 
 # How to display URL addresses: 'footnote', 'no', or 'inline'.
 #texinfo_show_urls = 'footnote'
+
+
+commit_hash = os.environ.get("SOPEL_GIT_COMMIT", "")[:7]
+is_dirty = bool(os.environ.get("SOPEL_GIT_DIRTY"))
+if commit_hash:
+    github_ref = "https://github.com/sopel-irc/sopel/commit/{commit_hash}".format(commit_hash=commit_hash)
```
SnoopJ, line 319: It occurred to me while preparing my revisions that this GitHub link isn't guaranteed to be valid when building the preview docs against commits that aren't pushed to the repo, but I'm not sure that's worth worrying about, it should be valid in general for work we do in this repo (and the docs that end up on the website), and anybody hacking on a local clone of the source probably will understand if the link is invalid.

docs/source/conf.py
Exirel: Hm... this might sound stupid but: why do we need to check the commit hash when we can just check for the version number? It feels like an awful lot of works for something that shouldn't happen outside of the commit that bumps Sopel's version to a stable one.

So actually, I'm :-1: on the whole commit hash check.

PR #2561, opened 2023-11-12
coretasks: fix SASL EXTERNAL without password set

### LINKED ISSUE #2560: SASL EXTERNAL incorrectly requires a password to be set
### Description

SASL EXTERNAL doesn't work correctly unless e.g. `auth_password` is set.

Big capability refactor (#2341, specifically 0401ae9f9e6745ee7b899f8620fcea0d0a61c056) likely caused this, but I didn't go back and test an older revision. Just building on @half-duplex's sleuthing from a session working on #2362 earlier tonight.

### Reproduction steps

1. Provide a `client_cert_file` path (`.pem` file), `auth_method = sasl` and `auth_target = EXTERNAL` in Sopel's config file.
  a. I registered the certificate fingerprint with a NickServ account for completeness, but this shouldn't be required to repro.
2. Start Sopel and wait for it to connect.
3. Sopel will quit with an error:

```
[2023-11-12 01:11:12,806] sopel.bot            INFO     - Client capability negotiation list: away-notify, chghost, invite-notify, multi-prefix, sasl, userhost-in-names
[2023-11-12 01:11:12,839] sopel.coretasks      ERROR    - Configuration error on ACK capability "sasl": ConfigurationError: SASL authentication required but no password available; please check your configuration file.
[2023-11-12 01:11:12,871] sopel.irc            ERROR    - ERROR received from server: Closing Link: your.host.name ()
```

### Expected behavior

Successful SASL EXTERNAL authentication:

```
[2023-11-12 01:11:42,497] sopel.irc            INFO     - Connected, initiating setup sequence
[2023-11-12 01:11:43,763] sopel.bot            INFO     - Client capability negotiation list: away-notify, chghost, invite-notify, multi-prefix, sasl, userhost-in-names
[2023-11-12 01:11:43,920] sopel.coretasks      INFO     - Successful SASL Auth.
[2023-11-12 01:11:43,920] sopel.bot            INFO     - End of client capability negotiation requests.
[2023-11-12 01:11:44,001] sopel.coretasks      INFO     - Enabled client capabilities: chghost, multi-prefix, userhost-in-names, away-notify, sasl
```

### Relevant logs

_No response_

### Notes

@half-duplex added a very simple patch for this already in #2362 (see below). I'm just opening this issue so we don't forget about this bug, regardless of whether SCRAM-SHA-256 support makes it into 8.0.0 or not.

```patch
diff --git a/sopel/coretasks.py b/sopel/coretasks.py
index 22f7fe1f..97c44849 100644
--- a/sopel/coretasks.py
+++ b/sopel/coretasks.py
@@ -112,7 +112,7 @@ def _handle_sasl_capability(
         return plugin.CapabilityNegotiation.ERROR
     # Check SASL configuration (password is required)
     password, mech = _get_sasl_pass_and_mech(bot)
-    if not password:
+    if mech != "EXTERNAL" and not password:
         raise config.ConfigurationError(
             'SASL authentication required but no password available; '
             'please check your configuration file.',
```

### Sopel version

51300a1ab854d6ec82d90df1bc876188c03335ff

### Installation method

`pip install`

### Python version

_No response_

### Operating system

_No response_

### IRCd

_No response_

### Relevant plugins

_No response_

half-duplex:
### Description

Fix #2560, now separate and with test (verified to fail without patch)

### Checklist

### DIFF
--- a/test/coretasks/test_coretasks_sasl.py
+++ b/test/coretasks/test_coretasks_sasl.py
@@ -59,6 +59,15 @@ enable = coretasks
 auth_method = sasl
 """
 
+TMP_CONFIG_SASL_EXTERNAL = """
+[core]
+owner = Uowner
+nick = TestBot
+enable = coretasks
+auth_method = sasl
+auth_target = EXTERNAL
+"""
+
 
 @pytest.fixture
 def tmpconfig(configfactory: ConfigFactory) -> Config:
@@ -251,6 +260,67 @@ def test_sasl_plain_no_password(
         'CAP END',
         'QUIT :Configuration error.',
     ), 'No password is a configuration error and the bot must quit.'
+
+
+def test_sasl_external_no_password(
+    botfactory: BotFactory,
+    configfactory: ConfigFactory,
+) -> None:
+    tmpconfig = configfactory("conf.ini", TMP_CONFIG_SASL_EXTERNAL)
+    mockbot = botfactory.preloaded(tmpconfig, preloads=["coretasks"])
+    mockbot.backend.connected = True
+
+    # connect and capability negotiation
+    mockbot.on_connect()
+    mockbot.on_message(":irc.example.com CAP * LS :sasl=PLAIN,EXTERNAL")
+    n = len(mockbot.backend.message_sent)
+    mockbot.on_message(":irc.example.com CAP * ACK :sasl")
+    assert mockbot.backend.message_sent[n:] == rawlist(
+        "AUTHENTICATE EXTERNAL",
+    ), "The bot should initiate SASL EXTERNAL authentication without a password set."
+    n += 1
+    mockbot.on_message("AUTHENTICATE +")
+    assert mockbot.backend.message_sent[n:] == rawlist(
+        "AUTHENTICATE +",
+    ), "SASL EXTERNAL authentication should continue without a password set."
+    n += 1
+
+    mockbot.on_message(":irc.example.com 903 SopelTest :SASL authentication successful")
+    assert mockbot.backend.message_sent[n:] == rawlist(
+        "CAP END",
+    ), "SASL success must resume capability negotiation."
+
+
+def test_sasl_external_fail(
+    botfactory: BotFactory,
+    configfactory: ConfigFactory,
+) -> None:
+    tmpconfig = configfactory("conf.ini", TMP_CONFIG_SASL_EXTERNAL)
+    mockbot = botfactory.preloaded(tmpconfig, preloads=["coretasks"])
+    mockbot.backend.connected = True
+
+    # connect and capability negotiation
+    mockbot.on_connect()
+    mockbot.on_message(":irc.example.com CAP * LS :sasl=PLAIN,EXTERNAL")
+    n = len(mockbot.backend.message_sent)
+
+    mockbot.on_message(":irc.example.com CAP * ACK :sasl")
+    assert mockbot.backend.message_sent[n:] == rawlist(
+        "AUTHENTICATE EXTERNAL",
+    ), "The bot should initiate SASL EXTERNAL authentication without a password set."
+    n += 1
+
+    mockbot.on_message("AUTHENTICATE +")
+    assert mockbot.backend.message_sent[n:] == rawlist(
+        "AUTHENTICATE +",
+    ), "SASL EXTERNAL authentication should continue without a password set."
+    n += 1
+
+    mockbot.on_message(":irc.example.com 904 SopelTest :SASL authentication failed")
+    assert mockbot.backend.message_sent[n:] == rawlist(
+        "CAP END",
+        "QUIT :SASL Auth Failed",
+    ), "SASL EXTERNAL failure should trigger CAP END and QUIT."
 
 
 def test_sasl_plain_bad_password(botfactory: BotFactory, tmpconfig) -> None:
--- a/sopel/coretasks.py
+++ b/sopel/coretasks.py
@@ -107,9 +107,9 @@ def _handle_sasl_capability(
         )
         return plugin.CapabilityNegotiation.ERROR
 
-    # Check SASL configuration (password is required)
+    # Check SASL configuration (password is required for PLAIN/SCRAM)
     password, mech = _get_sasl_pass_and_mech(bot)
-    if not password:
+    if mech != "EXTERNAL" and not password:
         raise config.ConfigurationError(
             'SASL authentication required but no password available; '
             'please check your configuration file.',


test/coretasks/test_coretasks_sasl.py
```diff
@@ -253,6 +253,26 @@ def test_sasl_plain_no_password(
...
+    mockbot.backend.connected = True
+
+    # connect and capability negotiation
+    mockbot.on_connect()
+    mockbot.on_message(":irc.example.com CAP * LS :sasl=PLAIN,EXTERNAL")
+    n = len(mockbot.backend.message_sent)
+    mockbot.on_message(":irc.example.com CAP * ACK :sasl")
+    assert mockbot.backend.message_sent[n:] == rawlist(
+        "AUTHENTICATE EXTERNAL"
+    ), "SASL EXTERNAL did not continue without a password configured."
```
Exirel, line 273: Since this is not a fail state, I'd like to see the test to do the same thing as the SASL plain test, i.e. see the server sends a "all done" so the bot can reply with a CAP END, like so (I'm pretty sure this should pass as-is):

```python
    # Server accept authentication
    mockbot.on_message(
        ':irc.example.com 900 TestBot TestBot!sopel@example.com sopel '
        ':You are now logged in as TestBot')
    mockbot.on_message(
        ':irc.example.com 903 TestBot :SASL authentication successful')
    assert mockbot.backend.message_sent[n:] == rawlist(
        'CAP END',
    )
```

This we know that if the external mechanism works as intended between Sopel and the server without explicit interaction at the IRC protocol level, then we have proof that this doesn't stop Sopel from sending the requested CAP END.

test/coretasks/test_coretasks_sasl.py
```diff
@@ -253,6 +253,26 @@ def test_sasl_plain_no_password(
     ), 'No password is a configuration error and the bot must quit.'
 
 
+# https://github.com/sopel-irc/sopel/issues/2560
+def test_sasl_external_no_password(
+    botfactory: BotFactory,
+    configfactory: ConfigFactory,
+) -> None:
+    tmpconfig = configfactory("conf.ini", TMP_CONFIG_SASL_NO_PASSWORD)
```
Exirel, line 261: I'd rather see a new `TMP_CONFIG_SASL_EXTERNAL` constant here, because yes, it doesn't have a password, but it's using a different SASL mechanism, which is more important than "not having a password".

test/coretasks/test_coretasks_sasl.py
Exirel, line 273: Nit: don't tell what you think happened, tell what should happen but didn't.

```suggestion
    ), "EXTERNAL authentication should continue even without password"
```

test/coretasks/test_coretasks_sasl.py
```diff
@@ -253,6 +253,26 @@ def test_sasl_plain_no_password(
     ), 'No password is a configuration error and the bot must quit.'
 
 
+# https://github.com/sopel-irc/sopel/issues/2560
```
Exirel, line 256: Nit: I don't think this is necessary, otherwise we would have hundreds of comments like this in the tests.

```suggestion
```

If you want to keep it, I'd rather see it in the test's docstring.

Exirel (changes requested):
Alright, looking good. Smaller modifications requested for the test, and a couple of nitpicks.

test/coretasks/test_coretasks_sasl.py
dgw: I don't believe SASL EXTERNAL is compliant if it proceeds without an exchange of `AUTHENTICATE +`. See raw logs from my testing of this patch as it existed at the tip of #2362 last night:

```
2023-11-12 01:10:31,825 <<      ':irc.rizon.io CAP SopelTest ACK :sasl\r\n'
2023-11-12 01:10:31,825 >>      'AUTHENTICATE EXTERNAL\r\n'
2023-11-12 01:10:31,885 <<      'AUTHENTICATE +\r\n'
2023-11-12 01:10:31,886 >>      'AUTHENTICATE +\r\n'
2023-11-12 01:10:31,948 <<      ':irc.rizon.io 900 SopelTest SopelTest!sopel@my-host.name SopelTest :You are now logged in as SopelTest.\r\n'
2023-11-12 01:10:31,949 <<      ':irc.rizon.io 903 SopelTest :SASL authentication successful\r\n'
2023-11-12 01:10:31,950 >>      'CAP END\r\n'
```

Then again, I could be completely off-base. Actual examples of the EXTERNAL mechanism are really hard to find; the IRCv3 spec pages say that EXTERNAL is one of the defined mechanisms, but only gives example client-server exchanges for PLAIN or SCRAM-SHA-1. Without diving into IRCd source code, the easiest way to investigate this is to observe how other implementations behave.

I didn't bother registering a certificate fingerprint at Libera, but it too performs a round of `AUTHENTICATE +` as part of the EXTERNAL mechanism:

```
2023-11-12 17:44:01,017 <<      ':molybdenum.libera.chat CAP SopelTest ACK :sasl\r\n'
2023-11-12 17:44:01,018 >>      'AUTHENTICATE EXTERNAL\r\n'
2023-11-12 17:44:02,148 <<      'AUTHENTICATE +\r\n'
2023-11-12 17:44:02,149 >>      'AUTHENTICATE +\r\n'
2023-11-12 17:44:03,232 <<      ':molybdenum.libera.chat 904 SopelTest :SASL authentication failed\r\n'
2023-11-12 17:44:03,234 >>      'CAP END\r\n'
2023-11-12 17:44:03,234 >>      'QUIT :SASL Auth Failed\r\n'
```

(Testing 904 w/EXTERNAL is probably a good test case, too, btw.)

test/coretasks/test_coretasks_sasl.py
```diff
@@ -253,6 +262,51 @@ def test_sasl_plain_no_password(
...
+    mockbot.on_message(":irc.example.com CAP * LS :sasl=PLAIN,EXTERNAL")
+    n = len(mockbot.backend.message_sent)
+    mockbot.on_message(":irc.example.com CAP * ACK :sasl")
+    mockbot.on_message("AUTHENTICATE +")
+    mockbot.on_message(":irc.example.com 903 SopelTest :SASL authentication successful")
+    assert mockbot.backend.message_sent[n:] == rawlist(
+        "AUTHENTICATE EXTERNAL",
+        "AUTHENTICATE +",
+        "CAP END",
+    ), "SASL EXTERNAL authentication should continue without a password set."
```
Exirel, line 284: This need to be a bit more thorough in its testing of step orders: it is important to check that we first have the 2 `AUTHENTICATE` messages, then the `CAP END`, because Sopel **must not** send the `CAP END` before the `903` (SASL sucess) event is received from the server.

```suggestion
    mockbot.on_message(":irc.example.com CAP * ACK :sasl")
    mockbot.on_message("AUTHENTICATE +")
    assert mockbot.backend.message_sent[n:] == rawlist(
        "AUTHENTICATE EXTERNAL",
        "AUTHENTICATE +",
    ), "SASL EXTERNAL authentication should continue without a password set."
    n = n + 2
    # assume it all went well
    mockbot.on_message(":irc.example.com 903 SopelTest :SASL authentication successful")
    assert mockbot.backend.message_sent[n:] == rawlist(
        "CAP END",
    ), "SASL success must resume capability negotiation"
```

test/coretasks/test_coretasks_sasl.py
```diff
@@ -253,6 +262,51 @@ def test_sasl_plain_no_password(
...
+    n = len(mockbot.backend.message_sent)
+    mockbot.on_message(":irc.example.com CAP * ACK :sasl")
+    mockbot.on_message("AUTHENTICATE +")
+    mockbot.on_message(":irc.example.com 904 SopelTest :SASL authentication failed")
+    assert mockbot.backend.message_sent[n:] == rawlist(
+        "AUTHENTICATE EXTERNAL",
+        "AUTHENTICATE +",
+        "CAP END",
+        "QUIT :SASL Auth Failed",
+    ), "SASL EXTERNAL failure should trigger CAP END and QUIT."
```
Exirel, line 307: Ditto:

```suggestion
    mockbot.on_message(":irc.example.com CAP * ACK :sasl")
    mockbot.on_message("AUTHENTICATE +")
    assert mockbot.backend.message_sent[n:] == rawlist(
        "AUTHENTICATE EXTERNAL",
        "AUTHENTICATE +",
    ), "SASL EXTERNAL authentication should continue without a password set."
    n = n + 2
    # assume it didn't go well
    mockbot.on_message(":irc.example.com 904 SopelTest :SASL authentication failed")
    assert mockbot.backend.message_sent[n:] == rawlist(
        "CAP END",
        "QUIT :SASL Auth Failed",
    ), "SASL EXTERNAL failure should trigger CAP END and QUIT."
```

Exirel (approved):
Yay! Thank you!

Issue #2562, opened 2023-11-13
MODE between WHO and RPL_WHOREPLY is lost


half-duplex:
### Description

I recently encountered an issue where a channel op (+o) was not being treated as an op by the bot. Checking the raw log, I found this when they last joined the channel:
```
<<      ':User!user@user.host JOIN :#channel\r\n'
>>      'WHO User\r\n'
<<      ':ChanServ!ChanServ@services MODE #channel +o User\r\n'
<<      ':server 352 Sopel #channel user user.host foo.server User Hr :0 John Smith\r\n'
<<      ':server 315 Sopel User :End of /WHO list.\r\n'
```
I'm not really sure how to handle this correctly, or if it's even Sopel's responsibility.

### Reproduction steps

Run this test:
```py
def test_bot_who_mode_order(mockbot, ircfactory):
    irc = ircfactory(mockbot)
    irc.bot._isupport = isupport.ISupport(chanmodes=("o", "", "", "", tuple()))
    irc.bot.modeparser.chanmodes = irc.bot.isupport.CHANMODES
    irc.channel_joined("#test")
    mockbot.on_message(":Alex!alex@alex.local JOIN :#test")
    assert mockbot.backend.message_sent == rawlist("WHO Alex")
    mockbot.on_message(":ChanServ!ChanServ@services.local MODE #test +o Alex")
    mockbot.on_message(":server.local 352 Sopel #test alex alex.local specific.server.local Alex Hr :0 Alex")
    mockbot.on_message(":server.local 315 Sopel Alex:End of /WHO list.")
    assert mockbot.channels["#test"].privileges[Identifier("Alex")] == OP
```
```
test/test_coretasks.py:129: in test_bot_who_mode_order
    assert mockbot.channels["#test"].privileges[Identifier("Alex")] == OP
E   assert 0 == <AccessLevel.OP: 4>
```

### Expected behavior

`bot.channels["#channel"].privileges["User"] = OP`

### Sopel version

Roughly 51300a1ab854d6ec82d90df1bc876188c03335ff

### Installation method

`pip install`

### IRCd

Unreal3.2.10.6

half-duplex:
I wonder if modern Unreal has the same interleaving behavior, considering 3.2 was released in 2004 and EOL in 2016...
https://www.unrealircd.org/docs/UnrealIRCd_releases

dgw:
[...]

PR #2563, opened 2023-11-15
tools.target: don't clobber existing permissions in `Channel.add_user()`

dgw:
### Description

Here's a potential fix for #2562, including a slightly modified version of the test case @half-duplex wrote for that issue.

The race-condition feeling I get from the bug report is a strong one. I can only assume that the IRCd on which this was reported (Unreal 3.2, which is admittedly EOL) suffers from a race condition between its `who` and `mode` modules, something like:

- New user joins `#channel`
- IRCd notifies all `#channel` members, including Sopel and ChanServ
- Sopel immediately sends MODE and WHO for the newly joined channel
  - Near-simultaneously, ChanServ notices that the new user should be auto-opped and sends MODE +o
- The `who` module finds the user, builds the status prefix, and sends the reply to Sopel
  - Meanwhile, the `mode` module (possibly on a different server in the network, adding more delay) processes (and possibly has to relay) the MODE change sent by ChanServ
- By the time the MODE change is processed/received, Sopel's server has already prepared the WHO response with an outdated status field, so Sopel gets a reply that doesn't indicate the new user's chanop privileges

Lots of educated guesses there based on skimming the C source for Unreal 3.2, but even if the exact sequence is wrong the result walks/quacks like a race condition, so we have to treat it like one.

I've also bundled a tweak to `coretasks`, moving the duplicate `MODE` and `WHO` invocations for a new channel to a single location (new helper function) and adding some more commentary on the join-queue logic, simply because those tweaks don't warrant a pull request of their own.

### Checklist

### Notes

No milestone up front; I'm not positive that we should shoehorn this particular issue & patch into 8.0.0. Depends on the reaction from other maintainers whether it goes in 8.0.0, 8.0.1, or 8.1.0. 😉

### DIFF
--- a/sopel/coretasks.py
+++ b/sopel/coretasks.py
@@ -194,12 +194,16 @@ def shutdown(bot):
 
 
 def _join_event_processing(bot):
-    """Process a batch of JOIN event from the ``join_events_queue`` queue.
+    """Process a batch of JOIN events from the ``join_events_queue`` queue.
 
-    Every time this function is executed, it processes at most
-    ``throttle_join`` JOIN events. For each JOIN, it sends a WHO request to
-    know more about the channel. This will prevent an excess of flood when
-    there are too many channels to join at once.
+    Every time this function is executed, it processes up to ``throttle_join``
+    JOIN events. For each queued JOIN event, it invokes ``_new_channel_info()``
+    to send requests for more information about the channel, such as channel
+    modes, online users, and their privileges.
+
+    This queue is intended to protect the bot from being kicked for "excess
+    flood" when there are too many channels to join at once. Scheduling this
+    function to run is handled by `coretasks.setup()`.
     """
     batch_size = max(bot.settings.core.throttle_join, 1)
     for _ in range(batch_size):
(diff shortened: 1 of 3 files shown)


dgw:
Ah, well, maybe this _isn't_ a good solution. It depends on whether the behavior of `Channel.add_user()` replacing any data that already exists is important.

half-duplex:
add_user is called in three places: JOIN, RPL_NAMREPLY, and _record_who(). Could add a clear=False to add_user and call it with clear=True from JOIN? Idk, all of these feel gross when NAMREPLY and WHOREPLY should be absolute and trustworthy.

I guess whether we do something like this or consider it an Unreal bug depends on whether it's likely to happen elsewhere. The possibility of different threads — or even servers with link latency — answering things makes it feel not entirely unlikely. The network where I originally observed this has several servers.

dgw:
> Idk, all of these feel gross when NAMREPLY and WHOREPLY should be absolute and trustworthy.

And that's why I initially didn't even plan to work on this. The patch here was an experiment, and the only reasonable way to put it "out there" for discussion was to open a PR. 🤷‍♂️

> add_user is called in three places: JOIN, RPL_NAMREPLY, and _record_who(). Could add a clear=False to add_user and call it with clear=True from JOIN?

That could be an interesting line of discussion, though I think it's less about whether to `clear` (clear _what_ is not obvious from the kwarg name) and more about "was this user directly observed JOINing or not?"

If we can reasonably trust NAMREPLY/WHOREPLY for users previously in the channel when Sopel joins, and a race condition is only a concern for users who join _after_ Sopel, we can possibly develop sensible logic. But it all still comes down to whether we can trust the server's NAMREPLY/WHOREPLY data, and none of this overthinking should be needed because ultimately, _what the server says is supposed to be trustworthy_.

(If I come off as frustrated, it's because I am. Inability to trust state expressed in the server's responses could break a _lot_ more than privilege tracking.)

half-duplex:
Set a flag for "we have a WHO out" on the User object, and if anything weird (MODE) happens before ENDOFWHO for it, re-WHO?

dgw:
Hmm, I'm going to take this discussion back to the original issue. This doesn't seem to be the way forward.

PR #2572, opened 2023-11-28
make: run flake8 on specific directories

SnoopJ:
### Description

Tin. Run the linter on only the files we are concerned with. This is slightly friendlier to developers who have throwaway scripts or a virtual environment (other than one in `env/`) in their repository directory.

### Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,7 +5,7 @@ qa: lint test coverages
 lint: lint-style lint-type
 
 lint-style:
-	flake8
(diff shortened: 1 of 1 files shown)


dgw (approved):
Looked a couple of times, didn't find any more declarative (in a config file somewhere) way to do this. The current state also bit me earlier this week, though, so let's accept this Good solution in lieu of a hypothetical Perfect one yet to be found.

PR #2574, opened 2023-11-29
docs: FAQ entry on logging + some minor clarification

### LINKED ISSUE #2395: docs: include how-to information for getting help/reporting bugs
### Requested Feature

It's common for errors reported in a reply by a Sopel instance to be inadequate for debugging what went wrong, so that the first step in getting a user the assistance they need is helping them locate and share the logs for their instance. It would be handy to have more of a formal "how to get help" narrative to make the experience more consistent for both users and helpers, and there may be other common advice/procedures that could be laid out in a document like that.

### Problems Solved

_No response_

### Alternatives

_No response_

### Notes

There's [some notes in `CONTRIBUTING.md`](https://github.com/sopel-irc/sopel/blob/master/CONTRIBUTING.md?plain=1#L1-L12) and [some information about logger configuration in the docs](https://sopel.chat/docs/configuration.html#logging) that are good starting points for this request.

Some 'prior art' in other projects:
* The Littlest JupyterHub's [Looking at Logs](https://tljh.jupyter.org/en/latest/troubleshooting/logs.html)
* PyInstaller's [When Things Go Wrong](https://pyinstaller.org/en/stable/when-things-go-wrong.html)

dgw:
### Description

After I closed #2395 there was some question whether the FAQ added in #2533 actually covered everything that was requested.

Deciding that we could always do it better, @SnoopJ convinced me (it wasn't hard) to add more info about logs to the FAQ.

I also clarified a piece of the "Logging" section docs under "The [core] configuration section" because it confused me, too.

### Checklist


docs/source/faq.rst
```diff
@@ -31,10 +34,53 @@ configured some templates to help and guide you through the process and
...
+  enabled by default; turn it if needed on using the
+  :attr:`~.config.core_section.CoreSection.log_raw` setting in your config file.
+
+.. warning::
+
+    The ``raw`` log may contain sensitive information, e.g. your bot's NickServ
+    account credentials, its IP address, or channel keys.
+
+    Always check the portion of your ``raw`` log that you are sharing for
+    secrets, and censor any you find!
```
SnoopJ, line 81: I appreciate the treatment of the distinct flavors of logs here! However, I think we should have a cleaner separation between the two, so that the canonical answer to _"how do I get my logs?"_ is _"find `~/sopel/logs/whatever.sopel.log` or equivalent"_ and we have some supplementary information about how to get the raw stuff should it be needed.

SnoopJ (commented):
You're a good sport, thank you for tackling this one!

Modulo my one remark about more cleanly separating _"do this when filing a report"_ from more thorough information about raw logs, this LGTM.

docs/source/faq.rst
```diff
@@ -31,10 +34,53 @@ configured some templates to help and guide you through the process and
...
+:option:`--config <sopel start --config>` argument passed when starting Sopel.
+
+There are two main log types you might be asked to provide:
+
+* ``config_name.sopel.log``: Everything Sopel does from startup to shutdown.
+  In normal operation, the ``INFO`` (default) or ``WARNING``
+  :attr:`~.config.core_section.CoreSection.logging_level` is sufficient, but you
+  might be asked to enable ``DEBUG`` logging and reproduce your problem to help
+  Sopel's developers determine what's happening.
+* ``config_name.raw.log``: Raw traffic between Sopel and the IRC server. Not
```
Exirel, line 71: I believe `<value to change>` is the common format to use:

```suggestion
* ``<config_name>.sopel.log``: Everything Sopel does from startup to shutdown.
  In normal operation, the ``INFO`` (default) or ``WARNING``
  :attr:`~.config.core_section.CoreSection.logging_level` is sufficient, but you
  might be asked to enable ``DEBUG`` logging and reproduce your problem to help
  Sopel's developers determine what's happening.
* ``<config_name>.raw.log``: Raw traffic between Sopel and the IRC server. Not
```

docs/source/faq.rst
```diff
@@ -31,10 +34,53 @@ configured some templates to help and guide you through the process and
...
+===================
+
+When requesting help, you'll probably be asked to provide logs illustrating the
+problem. If you :ref:`report a bug <faq-bug-report>`, the form includes a field
+for "Relevant logs" that you should fill in if possible.
+
+By default you will find Sopel's logs in the ``~/.sopel/logs`` folder. If Sopel's
+config file specifies the :attr:`~.config.core_section.CoreSection.homedir`
+and/or :attr:`~.config.core_section.CoreSection.logdir` setting(s), you will find
+the log files in ``/path/to/homedir/logdir``.
```
Exirel, line 59: I wonder if you could add a reference to the "install Sopel as a service", because in that install doc we suggest a specific place for Sopel's logs, that are more appropriate for a systemd service.

And by a reference I mean a link and a couple of words, not necessarily a full on `.. note`.

docs/source/faq.rst
```diff
@@ -31,10 +34,53 @@ configured some templates to help and guide you through the process and
...
+
+There are two main log types you might be asked to provide:
+
+* ``config_name.sopel.log``: Everything Sopel does from startup to shutdown.
+  In normal operation, the ``INFO`` (default) or ``WARNING``
+  :attr:`~.config.core_section.CoreSection.logging_level` is sufficient, but you
+  might be asked to enable ``DEBUG`` logging and reproduce your problem to help
+  Sopel's developers determine what's happening.
+* ``config_name.raw.log``: Raw traffic between Sopel and the IRC server. Not
+  enabled by default; turn it if needed on using the
```
Exirel, line 72: I'm not sure here, is "if needed on using" the right phrasing here? I'm not used to "needed on"? Maybe it's supposed to "turn it on if needed using"?

docs/source/faq.rst
```diff
@@ -31,10 +34,53 @@ configured some templates to help and guide you through the process and
...
+for "Relevant logs" that you should fill in if possible.
+
+By default you will find Sopel's logs in the ``~/.sopel/logs`` folder. If Sopel's
+config file specifies the :attr:`~.config.core_section.CoreSection.homedir`
+and/or :attr:`~.config.core_section.CoreSection.logdir` setting(s), you will find
+the log files in ``/path/to/homedir/logdir``.
+
+Log files' names start with the name of their associated config file, i.e. the
+:option:`--config <sopel start --config>` argument passed when starting Sopel.
+
```
Exirel, line 63: There is a specific ref link named `logging-basename` for that!

Exirel (changes requested):
Minor nitpicking, aside from that, great idea!

docs/source/faq.rst
SnoopJ: Well spotted! I agree with your rewording.

docs/source/faq.rst
dgw: The best part is I had it more or less this way in an earlier revision and then dropped the brackets because I thought they might be confusing… 😂

docs/source/faq.rst
dgw: Yikes, rewording that line apparently got left halfway. Gonna have to turn in my pedantic-editor card after this!

```suggestion
  enabled by default; turn it on if needed using the
```

docs/source/faq.rst
dgw: True, though that note refers to the config's `basename` attribute rather than the value passed to `--config`. Hmm…

docs/source/faq.rst
dgw: What exactly are you suggesting? A second FAQ titled (e.g.) "How to obtain raw logs?"

Or maybe reformatting this to be less bullet-y, maybe using subheadings instead?

SnoopJ: Basically, push the stuff that's talking about raw logs _further_ away from the path most users are going to take. Subheadings would be good, with the regular logs first.

Exirel (approved):
I still think this is too verbose and would be better with less compact paragraphs. However, I don't have the proper wording/phrasing in mind, and it's better to have all that information in the FAQ now than wait another month to get the "best" version.

That will do, and I'm happy about it!

SnoopJ (approved):

dgw:
I should note that merging this should come with a follow-up PR (draft) updating the relevant issue template(s) to link to this entry from the "Relevant logs" field description. Assuming links are allowed there, that is. I haven't checked into that yet.

PR #2575, opened 2023-11-29
wikipedia: fix urlparse usage & handle Special: namespace

### LINKED ISSUE #2573: wikipedia: results for non-existent namespaced articles are surprising
### Description

The `wikipedia` plugin supports redirection and "near miss" queries by talking to the `action=query` part of the MediaWiki API. This has surprising results if the user's query is namespaced (i.e. `SomeNamespace:*`) and the target page does not exist. Oftentimes, the result will be something outside the intended namespace, which is fairly surprising from the user's side.

```
10:22 <+SnoopJ> a less profane test case: https://en.wikipedia.org/wiki/Category:Spatulas
10:22 <+Sopel> [wikipedia] Spatulas | "A spatula is a broad, flat, flexible blade used to mix, spread and lift material including foods, drugs, plaster and paints. In medical applications, "spatula" may also be used synonymously with tongue depressor.The word spatula derives from the Latin word for a flat piece of wood or splint, a diminutive form of the Latin spatha, meaning 'broadsword', and hence can also refer to a tongue depressor. The words spade (digging […]"
10:22 <+SnoopJ> what's interesting about that one is…
10:22 <+SnoopJ> .wp Category:Spatulas
10:22 <+Sopel> [wikipedia] Category:Spatula (genus) | "" | https://en.wikipedia.org/wiki/Category%3ASpatula_%28genus%29
10:23 <+SnoopJ> …that there *is* a fairly close category, but I guess the remote API is either not sending it, or it's after the other one
```

(I checked the `Special:*` namespace as well, it seems all namespaces are affected)

This is a regression caused by #2414 which introduced the use of `urlparse()` on a non-URL string, which causes the namespace to be confused for a URL scheme. In simple terms, my code for addressing #2412 is holding `urlparse()` wrong.

### Reproduction steps

1) Query the bot with a Wikipedia URL or `.wp` command for a non-existent namespaced page
2) There is no (2)

### Expected behavior

The namespace should not be ignored, if I asked for `Category:Spatulas`, I don't want to know about `Spatulas`

### Relevant logs

_No response_

### Notes

We can `urlparse()` the entire URL and drop the prefixing `/wiki`, but there might be a better way to do it that involves slicing, since dropping the prefix is slightly annoying in Python 3.8. Shouldn't be hard to fix, though.

### Sopel version

7693af3

### Installation method

`pip install`

### Python version

3.8.18

### Operating system

Ubuntu 20.04

### IRCd

_No response_

### Relevant plugins

wikipedia

SnoopJ:
### Description

Closes #2573.

This changeset corrects errant usage of `urllib.parse.urlparse()` in the `wikipedia` plugin, fixing a regression introduced by #2414. This fix prevents namespaces from being erroneously discarded.

In the process of this fix, I discovered that the `Special:*` namespace is a special case as far as the MediaWiki API is concerned, queries to the API for pages in this namespace don't do what you might expect, so these pages cannot be located and turned into snippets for IRC. So, instead, the URL rule ignores these pages, and issues a specific explanation if a user tries to query them using a command.

### Checklist

### DIFF
--- a/sopel/builtins/wikipedia.py
+++ b/sopel/builtins/wikipedia.py
@@ -327,12 +327,16 @@ def mw_image_description(server, image):
 def mw_info(bot, trigger, match=None):
     """Retrieves and outputs a snippet of the linked page."""
     server = match.group(1)
-    page_info = urlparse(match.group(2))
-    article = unquote(page_info.path)
+    page_info = urlparse(match.group(0))
+    # in Python 3.9+ this can be str.removeprefix() instead, but we're confident that
+    # "/wiki/" is at the start of the path anyway since it's part of the pattern
+    trim_offset = len("/wiki/")
+    article = unquote(page_info.path)[trim_offset:]
     section = unquote(page_info.fragment)
 
     if section:
-        if section.startswith('cite_note-'):  # Don't bother trying to retrieve a snippet when cite-note is linked
+        if section.startswith('cite_note-'):
+            # Don't bother trying to retrieve a section snippet if cite-note is linked
             say_snippet(bot, trigger, server, article, show_url=False)
         elif section.startswith('/media'):
             # gh2316: media fragments are usually images; try to get an image description
--- a/sopel/builtins/wikipedia.py
+++ b/sopel/builtins/wikipedia.py
@@ -334,6 +334,12 @@ def mw_info(bot, trigger, match=None):
     article = unquote(page_info.path)[trim_offset:]
     section = unquote(page_info.fragment)
 
+    if article.startswith("Special:"):
+        # The MediaWiki query API does not include pages in the Special:
+        # namespace, so there's no point bothering when we know this will error
+        LOGGER.debug("Ignoring page in Special: namespace")
+        return False
+
     if section:
         if section.startswith('cite_note-'):
             # Don't bother trying to retrieve a section snippet if cite-note is linked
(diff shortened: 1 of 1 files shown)


sopel/builtins/wikipedia.py
```diff
@@ -327,12 +327,22 @@ def mw_image_description(server, image):
 def mw_info(bot, trigger, match=None):
     """Retrieves and outputs a snippet of the linked page."""
     server = match.group(1)
-    page_info = urlparse(match.group(2))
-    article = unquote(page_info.path)
+    page_info = urlparse(match.match.string)
```
dgw, line 330: We should try to get away from using the `match` argument. The behavior of `trigger` is now identical, because the `match` is also the `trigger`:

https://github.com/sopel-irc/sopel/blob/7693af3356f59be7c825cafd91251dc88f5caf2a/sopel/plugins/rules.py#L1778-L1792

Don't think we've gotten around to formally deprecating `match`, but we should. Maybe in 8.1.

dgw (changes requested):
As I've already played with the patched plugin in your bot's test channel, I only have nitpicks left.

dgw (approved):
Thought this PR had been made ready to go, but apparently not. Good to squash, and if you're feeling extra on-top-of-things you can make sure the draft changelog you've been working on reflects both bugfixes from here. 😉

Issue #2399, opened 2023-01-14
RFC: Make `require_privilege` imply `require_chanmsg`


dgw:
This is an idea that I keep thinking of, briefly, every so often. I always forget about it again before getting around to opening an issue though. Tonight I finally decided, enough.

I think it's a bit of a potential foot-gun to have `@plugin.require_privilege` be ignored in PMs, even though it's documented. Seems like it would make more sense to make `@plugin.require_privilege` imply `@plugin.require_chanmsg`.

We could also raise an error during plugin loading if a callable is decorated with both `@plugin.require_privmsg` and either of `@plugin.require_chanmsg` or `@plugin.require_privilege`.

Any thoughts from other maintainers about this? Am I failing to consider some obscure but very useful edge case where it would be useful to do this on purpose?

_Note: 8.0.0 milestone for attention, since it's a breaking change that would have to wait another who-knows-how-many-years until 9.0 if not done in v8._

Exirel:
I think it's safe to enforce require chanmsg when requiring privileges.

As for the other problem: developers are humans, they will make mistakes. The question is: is that really a problem?
They'll probably discover that they can't use their command with both decorator, so we shouldn't bother that much.

PR #2580, opened 2023-12-06
plugin: `require_bot_privilege()` implies `require_chanmsg()`

dgw:
### Description

If the bot needs a channel privilege for a given callable, it probably wants to do something related to channel operations, and triggering such a callable in response to a private message makes no sense.

Sorry about trying to add something into 8.0.0 this late… I've been writing the 8.0 migration guide tonight, and when I ran across #2405 I realized that #2399 should have contemplated doing this too. If we're gonna "breaking change" privilege-decorator stuff, we should do it _all_ in one major version.

If others agree that this is a good idea we can just merge the patch that's already ready to go and get on with finishing the release. (That's why I didn't open an issue first, to shorten the discuss -> implement -> review cycle.)

### Checklist

### DIFF
--- a/sopel/plugin.py
+++ b/sopel/plugin.py
@@ -1716,16 +1716,18 @@ def require_bot_privilege(
     <.bot.Sopel.say>`, but when ``reply`` is ``True`` it uses :meth:`bot.reply()
     <.bot.Sopel.reply>` instead.
 
-    Privilege requirements are ignored in private messages.
+    Use of ``require_bot_privilege()`` implies :func:`require_chanmsg`.
 
     .. versionadded:: 7.1
+    .. versionchanged:: 8.0
+        Decorated callables no longer run in response to private messages.
     """
     def actual_decorator(function):
         @functools.wraps(function)
         def guarded(bot, trigger, *args, **kwargs):
-            # If this is a privmsg, ignore privilege requirements
+            # If this is a privmsg, do not trigger
             if trigger.is_privmsg:
-                return function(bot, trigger, *args, **kwargs)
+                return
 
             if not bot.has_channel_privilege(trigger.sender, level):
                 if message and not callable(message):
--- a/test/test_plugin.py
+++ b/test/test_plugin.py
@@ -511,19 +511,21 @@ def test_require_bot_privilege_private_message(configfactory,
     def mock(bot, trigger):
         return True
 
-    assert mock(bot, bot._trigger) is True
+    assert mock(bot, bot._trigger) is not True, (
+        'Callable requiring bot channel privilege must be ignored in private.')
 
     @plugin.command('ban')
     @plugin.require_bot_privilege(plugin.OP)
     def mock(bot, trigger):
         return True
 
-    assert mock(bot, bot._trigger) is True
+    assert mock(bot, bot._trigger) is not True, (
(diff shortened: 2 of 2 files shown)


sopel/plugin.py
```diff
@@ -1716,16 +1716,18 @@ def require_bot_privilege(
...
+        Decorated callables no longer run in response to private messages.
     """
     def actual_decorator(function):
         @functools.wraps(function)
         def guarded(bot, trigger, *args, **kwargs):
-            # If this is a privmsg, ignore privilege requirements
+            # If this is a privmsg, do not trigger
             if trigger.is_privmsg:
-                return function(bot, trigger, *args, **kwargs)
+                return
```
SnoopJ, line 1730: This seems like a good place for a logger message (I settled on `warning` since this is a breaking change, but I can see an argument for `debug` instead)

```suggestion
                LOGGER.warning("Skipping evaluation of channel-only callable in private message: %r", function)
                return
```

test/test_plugin.py
```diff
@@ -511,19 +511,19 @@ def test_require_bot_privilege_private_message(configfactory,
     def mock(bot, trigger):
         return True
 
-    assert mock(bot, bot._trigger) is True
+    assert mock(bot, bot._trigger) is not True
```
SnoopJ, line 514: This assertion seems like it should have an associated message with this change.

test/test_plugin.py
```diff
@@ -511,19 +511,19 @@ def test_require_bot_privilege_private_message(configfactory,
...
 
     @plugin.command('ban')
     @plugin.require_bot_privilege(plugin.OP)
     def mock(bot, trigger):
         return True
 
-    assert mock(bot, bot._trigger) is True
+    assert mock(bot, bot._trigger) is not True
```
SnoopJ, line 521: This assertion seems like it should have an associated message with this change.

SnoopJ (approved):
Had some nits, but I don't fundamentally object to this change since the decorator is _fundamentally_ about channels, so I say let's do it.

sopel/plugin.py
dgw: Should be `debug`, yeah. But it makes sense to do only if the non-bot version (`require_privilege`) does the same.

Exirel: We don't have a logger for the `require_privilege`, and I don't think it is necessary: the `callable` may be already protected by the appropriate decorator, but this one is reached first because of the order of declaration.

Beside, if someone put that on the generic rule for `.*`, this will span the debug logs for no reason.

So, -0.5 here from me.

Exirel (commented):
I agree with @SnoopJ's review regarding the assert message. I have a slight negative opinion of the debug line. Other than that, this looks good to me.

dgw:
Each assertion now has the message, and I skipped the logger call given @Exirel's valid point about "span" (🇫🇷? 😁) in the logs.

Ready to be merged over the weekend unless one of y'all swings by with a last-minute objection. 🙂

Exirel (approved):
> valid point about "span" (🇫🇷? 😁)

Nah, just a stupid typo from a 🧠 💨 

> Ready to be merged over the weekend unless one of y'all swings by with a last-minute objection. 🙂

GO

PR #2582, opened 2023-12-11
tell: fix edge case with `@` as the tellee

bronzeguy:
### Description

There's an annoying bug where an oversight in code that checks if a message is prepended with an @ allows for a user to make the bot spit out error messages forever until the bot is restarted. I have fixed this oversight here

### Checklist

- [x] I have read [CONTRIBUTING.md](https://github.com/sopel-irc/sopel/blob/master/CONTRIBUTING.md)
- [x] I can and do license this contribution under the EFLv2
- [x] No issues are reported by `make qa` (runs `make lint` and `make test`)
- [x] I have tested the functionality of the things this change touches

Issue #2583, opened 2023-12-11
tell: check if a tellee is a valid identifier


SnoopJ:
### Requested Feature

Filing as follow-up from #2582.

Currently, the `tell` plugin does no verification that a tellee is a valid identifier, but it's something we may want to consider for longer term purposes. Example cases where we will store reminders for a nick we will never see:

```
22:49 <+SnowJ> .tell 猫 nyaaa
22:49 <+Sopel> SnowJ: I'll pass that on when 猫 is around.
22:50 <+SnowJ> .tell shinji@NERV get in the robot
22:50 <+Sopel> SnowJ: I'll pass that on when shinji@NERV is around.
```

PR #2584, opened 2023-12-11
tell: avoid errors produced by empty tellee

SnoopJ:
### Description

This changeset is an alternative to #2582 that transforms the `tellee` as early as possible.

The use of `lstrip("@")` in this patch means that an arbitrary number of prefixing `@`s will be removed, so that `@@Someone` is transformed into `Someone` where it would previously have been transformed to `@Someone`, but since this is an edge case to begin with, I don't think it's any problem.

Longer term, we'd like to have better awareness of if `tellee` is a valid nick or not. I've filed #2583 for this purpose.

### Checklist

### DIFF
--- a/sopel/builtins/tell.py
+++ b/sopel/builtins/tell.py
@@ -178,10 +178,13 @@ def f_remind(bot, trigger):
     verb = trigger.group(1)
 
     if not trigger.group(3):
+        tellee = ""
+    else:
+        tellee = trigger.group(3).rstrip('.,:;').lstrip('@')
+
+    if not tellee:
         bot.reply("%s whom?" % verb)
         return
-
-    tellee = trigger.group(3).rstrip('.,:;')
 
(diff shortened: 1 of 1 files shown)


dgw:
Note to self: Manually close #2582 when merging, since GH artificially doesn't allow merging a PR to auto-close another PR. (I say "artificially" because you can auto-close _issues_, and "all PRs are issues but not all issues are PRs" according to the GH API.)

Exirel (changes requested):
I'd add an `@plugin.example` for the new cases. Like:

```
@plugin.example('.ask @', 'ask whom?')
@plugin.example('.tell @', 'tell whom?')
@plugin.example('.tell @Exirel:', 'tell Exirel what?')
```

That way we can test that the command removes some characters from the `tellee`.

Exirel (approved):

PR #2585, opened 2023-12-11
bot, coretasks, url: `search_url_callbacks()` as a public wrapper

dgw:
### Description

Follow-up to #2581, which will let us keep using the deprecated mechanism internally without logging warnings about it, while warning third party plugin authors as expected.

### Checklist

I don't remember if I ran the linter/tests when making this branch a couple days ago, so I won't check those boxes just yet. 😅 Plenty of time to sort it out, as this will be an 8.1 change.

### DIFF
--- a/sopel/bot.py
+++ b/sopel/bot.py
@@ -1255,6 +1255,10 @@ class Sopel(irc.AbstractBot):
         .. __: https://docs.python.org/3.11/library/re.html#match-objects
 
         """
+        return self._search_url_callbacks_impl(url)
+
+    def _search_url_callbacks_impl(self, url):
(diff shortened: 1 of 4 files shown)


dgw:
No, you're right. I double-checked your points:

- The deprecation warning works.
- There are no remaining calls to `bot.search_url_callbacks()` outside of test code.
- Tests pass silently, without exposing the deprecation warning unless a case breaks (in which case the "captured log call" will include applicable deprecation warning(s)—a little bit of clutter, but no problem).

Let's just skip this. The reason for doing it—packed-in code still using the deprecated method—is gone.

PR #2586, opened 2023-12-16
version: retrieve commit hash for git worktrees

SnoopJ:
### Description

Tin and personal wish fulfillment. With this changeset, the `version` command will report commit information if Sopel is running from a [`git-worktree`](https://git-scm.com/docs/git-worktree).

The current revision of the `version` plugin only supports reading commit information when the host Sopel is running inside of a 'real' git repository. I find it convenient to run Sopel from a worktree for a "near to development" copy, and it's nice to still have the commit information in this use-case.

### Checklist

  - I did see some errors when running `mypy` because `types-urllib3` was not installed. I'm not sure if that's missing from `dev-requirements.txt` or if my local environment was in a weird state.
  - Also saw some test failures in `test_example_suggest_*` caused by content decoding errors. I only vaguely recognize the issue, but I'm confident it's not this changeset's fault.

### DIFF
--- a/sopel/builtins/version.py
+++ b/sopel/builtins/version.py
@@ -18,7 +18,6 @@ from sopel import __version__ as release, plugin
 
 logger = logging.getLogger(__name__)
 
-
 PROJECT_DIR = os.path.dirname(os.path.dirname(os.path.dirname(__file__)))
 GIT_DIR = os.path.join(PROJECT_DIR, '.git')
 
@@ -69,7 +68,7 @@ def _resolve_git_dirs(path: str) -> tuple[str, str]:
     return gitdir, head
 
 
-def git_info() -> str | None:
+def sopel_git_commit() -> str | None:
     """Determine the git commit hash of this Sopel, if applicable
     """
     gitdir, head = _resolve_git_dirs(GIT_DIR)
@@ -110,7 +109,7 @@ def version(bot, trigger):
         'Python: %s' % platform.python_version()
     ]
     try:
-        sha = git_info()
+        sha = sopel_git_commit()
     except OSError:
         logger.warning("Failed to retrieve git commit hash", exc_info=True)
         sha = None
--- a/sopel/builtins/version.py
+++ b/sopel/builtins/version.py
@@ -9,27 +9,71 @@ https://sopel.chat
 from __future__ import annotations
 
 import datetime
+import logging
 import os
 import platform
 
 from sopel import __version__ as release, plugin
 
 
+logger = logging.getLogger(__name__)
+
+
 PROJECT_DIR = os.path.dirname(os.path.dirname(os.path.dirname(__file__)))
 GIT_DIR = os.path.join(PROJECT_DIR, '.git')
 
 
-def git_info():
-    head = os.path.join(GIT_DIR, 'HEAD')
+def _read_commit(gitdir: str, head: str) -> str | None:
+    """Given paths to ``.git/`` and ``HEAD``, determine the associated commit hash
+
+    :param gitdir: path to a ``.git/`` directory
+    :param head: path to ``HEAD`` file
+    """
+    result = None
+
     if os.path.isfile(head):
         with open(head) as h:
             head_loc = h.readline()[5:-1]  # strip ref: and \n
-        head_file = os.path.join(GIT_DIR, head_loc)
+        head_file = os.path.join(gitdir, head_loc)
         if os.path.isfile(head_file):
             with open(head_file) as h:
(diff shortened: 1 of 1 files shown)


dgw:
> * I did see some errors when running `mypy` because `types-urllib3` was not installed. I'm not sure if that's missing from `dev-requirements.txt` or if my local environment was in a weird state.

Well it's _not_ in `dev-requirements.txt`, and CI doesn't appear to install it from anywhere, but in CI its absence doesn't cause any `make lint` errors. Curious. Locally, `pipdeptree` doesn't even identify any packages as depending on `types-urllib3` either, yet it's present in my environment. Let's chase this one down if it happens to anyone else.

> * Also saw some test failures in `test_example_suggest_*` caused by content decoding errors. I only vaguely recognize the issue, but I'm confident it's not this changeset's fault.

It's the same issue I described in a couple of my PRs, and is _probably_ because you haven't reinstalled/`--upgrade`d the `dev-requirements.txt` since #2519 (if I had to guess).

To me, this change is low-stakes and could be shipped in 8.0.0 while I still lack the time (😩) to work on release notes. _(It's just nuts how much stuff has kept coming up for me this month! I thought I had everything planned out for $work etc. but noooo.)_

SnoopJ:
> It's the same issue I described in a couple of my PRs, and is _probably_ because you haven't reinstalled/`--upgrade`d the `dev-requirements.txt` since #2519 (if I had to guess).

Huh, I had an old `urllib3`. Might be why I saw `types-urllib3` weirdness too. I could have sworn I installed the requirements file, but I guess I only manually installed the stubs. Maybe I left my environment in a bad state while checking something against 7.1.9.

sopel/builtins/version.py
```diff
@@ -19,17 +19,55 @@
...
+
+    return gitdir, head
+
+
+def git_info() -> str | None:
+    """
+    Determine the git commit hash of this Sopel, if applicable
+    """
+    gitdir, head = _resolve_git_dirs(GIT_DIR)
+    return _read_commit(gitdir, head)
```
Exirel, line 70: I feel like it's a good time to add a `try/except` block either here or in the command callable, since this relies on I/O operations that can fail.

sopel/builtins/version.py
```diff
@@ -19,17 +19,55 @@
...
+                head = os.path.join(rest, "HEAD")
+            # else: we can't make sense of this file, stick to the default
+
+    return gitdir, head
+
+
+def git_info() -> str | None:
+    """
+    Determine the git commit hash of this Sopel, if applicable
+    """
```
Exirel, line 68:
```suggestion
    """Determine the git commit hash of this Sopel, if applicable"""
```

sopel/builtins/version.py
```diff
@@ -19,17 +19,55 @@
 GIT_DIR = os.path.join(PROJECT_DIR, '.git')
 
 
-def git_info():
-    head = os.path.join(GIT_DIR, 'HEAD')
+def _read_commit(gitdir: str, head: str) -> str | None:
+    """
+    Given paths to ``.git/`` and ``HEAD``, determine the associated commit hash
+    """
```
Exirel, line 25:
```suggestion
    """Determine the associated commit hash from the git dir and its HEAD"""
```

If you want, you can detail each parameter with `:param gitdir: <desc>` stanza in the docstring's body.

Exirel (changes requested):
That's looking good, feature wise. Code wise, it's mostly style related.

I didn't put a comment on the `f` variable, because it isn't important, however I tend to use `fd` (for file descriptor) instead of just `f`. In general, I prefer more expressive code and names. I don't think reducing the character count on variable names that are less than 10 characters matters.

Exirel:
I don't feel comfortable having this in 8.0.1. I'm sure this can wait Sopel 8.1 - beside, it should be useful mostly when working with an unstable version from git, so something that would fit nicely in master and unreleased for now.

SnoopJ:
> I'm sure this can wait Sopel 8.1

Agreed

sopel/builtins/version.py
```diff
@@ -19,17 +19,55 @@
...
             with open(head_file) as h:
                 sha = h.readline()
                 if sha:
-                    return sha
+                    result = sha
+
+    return result
+
+
+def _resolve_git_dirs(pth: str) -> tuple[str, str]:
```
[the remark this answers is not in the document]
SnoopJ: I habitually use this as a name to avoid conflict with `Path` in code where both are present, but I agree that it's more obvious in a function signature. No problem to change it here

PR #2589, opened 2024-01-27
find: support escaping backslash, fix double-bolding, and add basic tests

dgw:
### Description

Allowing `\|` or `\/` is all well and good, but if one can't escape the escape character, things get very confusing! So let's make it possible to escape `\` itself as `\\`.

With the test scaffolding now written (OK, much of it stolen from the preexisting `test/builtins/test_builtins_isup.py`), it'll be easier to add more cases (edge or otherwise) as we think of them.

### Checklist

### Notes

Early 8.0.x bugfix candidate, unless someone else feels strongly about including it in 8.0.0. Feels late in the game to add One More Thing into that milestone…

### DIFF
--- a/test/builtins/test_builtins_find.py
+++ b/test/builtins/test_builtins_find.py
@@ -43,6 +43,7 @@ REPLACES_THAT_WORK = (
     ("An escaped / line.", r"s/\//slash/", f"An escaped {bold('slash')} line."),
     ("A piped line.", r"s|line|replacement|", f"A piped {bold('replacement')}."),
     ("An escaped | line.", r"s|\||pipe|", f"An escaped {bold('pipe')} line."),
+    ("An escaped \\ line.", r"s/\\/backslash/", f"An escaped {bold('backslash')} line."),
 )
 
 
--- a/test/builtins/test_builtins_find.py
+++ b/test/builtins/test_builtins_find.py
@@ -102,8 +102,6 @@ def test_replace_the_replacement(bot, irc, user, channel):
             channel, user.nick, bold('eggs'),
         ),
         "PRIVMSG %s :%s meant to say: %s" % (
-            channel, user.nick, bold(bold('bacon')),
-            # the test is accurate, even though the behavior here (doubled bold
-            # control characters) is less than ideal
+            channel, user.nick, bold('bacon'),
         ),
     )
--- a/test/builtins/test_builtins_find.py
+++ b/test/builtins/test_builtins_find.py
@@ -34,6 +34,11 @@ def user(userfactory):
 
 
 @pytest.fixture
+def other_user(userfactory):
+    return userfactory('other_user')
+
+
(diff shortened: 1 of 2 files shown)


test/builtins/test_builtins_find.py
```diff
@@ -0,0 +1,61 @@
...
+    ("An escaped / line.", r"s/\//slash/", f"An escaped {bold('slash')} line."),
+    ("A piped line.", r"s|line|replacement|", f"A piped {bold('replacement')}."),
+    ("An escaped | line.", r"s|\||pipe|", f"An escaped {bold('pipe')} line."),
+    ("An escaped \\ line.", r"s/\\/backslash/", f"An escaped {bold('backslash')} line."),
+)
+
+
+@pytest.mark.parametrize('original, command, result', REPLACES_THAT_WORK)
+def test_valid_replacements(bot, irc, user, channel, original, command, result):
+    irc.channel_joined(channel, ['User'])
```
Exirel, line 52:
```suggestion
    irc.channel_joined(channel, [user.nick])
```

test/builtins/test_builtins_find.py
```diff
@@ -0,0 +1,61 @@
...
+
+    irc.say(user, channel, original)
+    irc.say(user, channel, command)
+
+    assert len(bot.backend.message_sent) == 1, (
+        "The bot should respond with exactly one line.")
+    assert bot.backend.message_sent == rawlist(
+        "PRIVMSG %s :User meant to say: %s" % (channel, result),
```
Exirel, line 60:
```suggestion
        "PRIVMSG %s :%s meant to say: %s" % (channel, user.nick, result),
```

Exirel (changes requested):
First, I love that you are using the testing tools for that. I'm happy to see it works quite well. I've suggested a couple of quick changes, that makes the test less reliable on hard-coded value (the user's nick).

I think it would be valuable to add two other tests:

1. one where there are 2 users, and one say something, the other use the replace feature
2. one where a user replace the replaced line

And with that, I believe we would have a pretty good set of tests.

test/builtins/test_builtins_find.py
dgw: GH doesn't know I accepted this, but I did. I just did it (including a `Co-authored-by` line to squash into the history later) manually.

test/builtins/test_builtins_find.py
dgw: Ditto the other suggestion.

dgw (commented):
Points well taken about additional tests. I added a few more test cases, and also fixed another bug: the double-`\x02` wrapping of replacements inside previous replacements.

Exirel (approved):
Love to see it!

dgw:
Took care of a `squashme:` commit. No changes to the patch itself; the commit history is just cleaner.

PR #2592, opened 2024-02-14
GHA: Update/fix actions

dgw:
Both `actions/checkout` and `actions/setup-python` were using deprecated Node runtimes and generating warnings in our CI logs. Oh, and they'd also eventually stop working entirely when the deprecated runtime is removed. That's probably the best reason to update them.

I also discovered that there is no `v1` tag for `pypa/gh-action-pypi-publish`, so our existing `pypi.yml` workflow from #2328 wouldn't ever work. Because tags can be easily changed if a malicious someone gains access to the action's repository, the release action is pinned to a specific, verified, commit's SHA.

### Checklist

  - This will be handled for the `ci.yml` workflow by PR checks, and if the checkout & setup-python actions work there we can assume the same new versions will also work in the `pypi.yml` workflow.
  - I can't really test the `pypi.yml` workflow but that's no change (it was already untested), and we'll have a chance to test it for real soon enough.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,9 +26,9 @@ jobs:
           - "3.11"
           - "3.12"
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
       - name: Set up Python ${{ matrix.python-version }}
-        uses: actions/setup-python@v4
+        uses: actions/setup-python@v5
         with:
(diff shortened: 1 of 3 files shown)


dgw:
Noticed that https://github.blog/changelog/2024-01-12-code-scanning-deprecation-of-codeql-action-v2/ had slipped by. Better to catch it now than having the analysis suddenly break in December.

PR #2594, opened 2024-02-24
Fix version check for entrypoint plugins with unequal project/package names

### LINKED ISSUE #2593: Inequal 'project' and 'package' name in `EntryPointPlugin` causes various crashes
### Description

I have observed that an entrypoint-style Sopel plugin cannot be reloaded if its project name does not exactly match the top-level import name.

I believe this is caused by catching `ValueError` specifically in the [`get_version` logic](https://github.com/sopel-irc/sopel/blob/973a489355540d68b95db01a49e983ac7a740bcc/sopel/plugins/handlers.py#L619-L624), where other exceptions may bubble up out of the `importlib` machinery.

## Log error report

```
[2024-02-23 19:55:26,265] sopel.bot            ERROR    - Unexpected PackageNotFoundError (sopel_dummy) from SnoopJ. Message was: !reload dummy
  File "/var/lib/gitea/data/gitea-repositories/snoopj/sopel.git/sopel/bot.py", line 695, in call_rule
    rule.execute(sopel, trigger)
  File "/var/lib/gitea/data/gitea-repositories/snoopj/sopel.git/sopel/plugins/rules.py", line 1274, in execute
  File "/var/lib/gitea/data/gitea-repositories/snoopj/sopel.git/sopel/plugin.py", line 1399, in guarded
    return function(bot, trigger, *args, **kwargs)
  File "/var/lib/gitea/data/gitea-repositories/snoopj/sopel.git/sopel/builtins/reload.py", line 109, in pm_f_reload
    f_reload(bot, trigger)
  File "/var/lib/gitea/data/gitea-repositories/snoopj/sopel.git/sopel/plugin.py", line 1640, in guarded
    return function(bot, trigger, *args, **kwargs)
  File "/var/lib/gitea/data/gitea-repositories/snoopj/sopel.git/sopel/builtins/reload.py", line 53, in f_reload
    bot.reload_plugin(name)
  File "/var/lib/gitea/data/gitea-repositories/snoopj/sopel.git/sopel/bot.py", line 365, in reload_plugin
    meta = plugin_handler.get_meta_description()
  File "/var/lib/gitea/data/gitea-repositories/snoopj/sopel.git/sopel/plugins/handlers.py", line 647, in get_meta_description
    data = super().get_meta_description()
  File "/var/lib/gitea/data/gitea-repositories/snoopj/sopel.git/sopel/plugins/handlers.py", line 321, in get_meta_description
    'version': self.get_version(),
  File "/var/lib/gitea/data/gitea-repositories/snoopj/sopel.git/sopel/plugins/handlers.py", line 620, in get_version
    version = importlib.metadata.version(self.module.__package__)
  File "/home/snoopjedi/.pyenv/versions/3.8.18/lib/python3.8/importlib/metadata.py", line 530, in version
    return distribution(distribution_name).version
  File "/home/snoopjedi/.pyenv/versions/3.8.18/lib/python3.8/importlib/metadata.py", line 503, in distribution
    return Distribution.from_name(distribution_name)
  File "/home/snoopjedi/.pyenv/versions/3.8.18/lib/python3.8/importlib/metadata.py", line 177, in from_name
    raise PackageNotFoundError(name)
importlib.metadata.PackageNotFoundError: sopel_dummy
```

### Reproduction steps

## Broken `setup.cfg`

For a plugin whose top-level import name is `sopel_dummy`

```
[metadata]                                                                                                                                                              
name = sopel-dummy-plugin                                                                                                                                               
version = 1.0.0                                                                                                                                                         
description = A dummy plugin for Sopel                                                                                                                                  
author = SnoopJ                                                                                                                                                         
author_email = snoopjedi@gmail.com                                                                                                                                                                                                                                                                
license = Eiffel Forum License, version 2                                                                                                                               
classifiers =                                                                                                                                                           
    Intended Audience :: Developers                                                                                                                                     
    Intended Audience :: System Administrators                                                                                                                          
    Topic :: Communications :: Chat :: Internet Relay Chat                                                                                                              
                                                                                                                                                                        
[options]                                                                                                                                                               
packages = find:                                                                                                                                                        
zip_safe = false                                                                                                                                                        
include_package_data = true                                                                                                                                             
install_requires =                                                                                                                                                      
    sopel>=7.1                                                                                                                                                          
                                                                                                                                                                        
[options.entry_points]                                                                                                                                                  
sopel.plugins =                                                                                                                                                         
    dummy = sopel_dummy
```

### Expected behavior

I expect the plugin to be reloaded

### Relevant logs

_No response_

### Notes

_No response_

### Sopel version

973a489

### Installation method

`pip install`

### Python version

3.8.18

### Operating system

Ubuntu 20.04

### IRCd

_No response_

### Relevant plugins

_No response_

SnoopJ:
### Description

Fixes #2593.

This changeset swallows all exceptions when trying to determine the version of an `EntryPointPlugin`, as there are more failure cases than what catching `ValueError` specifically covers, and allowing those exceptions to propagate breaks plugin reloading.

### Checklist

  - `1388 passed, 8 xfailed, 1 warning in 45.44s`

### DIFF
--- a/sopel/plugins/handlers.py
+++ b/sopel/plugins/handlers.py
@@ -48,6 +48,7 @@ import importlib
 import importlib.util
 import inspect
 import itertools
+import logging
 import os
 import sys
 from typing import Optional, TYPE_CHECKING, TypedDict
@@ -59,6 +60,9 @@ from . import exceptions
 if TYPE_CHECKING:
     from sopel.bot import Sopel
     from types import ModuleType
+
+
+LOGGER = logging.getLogger(__name__)
 
 
 class PluginMetaDescription(TypedDict):
@@ -620,14 +624,20 @@ class EntryPointPlugin(PyModulePlugin):
 
         if (
             version is None
-            and hasattr(self.module, "__package__")
-            and self.module.__package__ is not None
+            and hasattr(self.entry_point, "dist")
+            and hasattr(self.entry_point.dist, "name")
         ):
+            dist_name = self.entry_point.dist.name
             try:
-                version = importlib.metadata.version(self.module.__package__)
-            except ValueError:
-                # package name is probably empty-string; just give up
-                pass
+                version = importlib.metadata.version(dist_name)
+            except (ValueError, importlib.metadata.PackageNotFoundError):
+                LOGGER.warning("Cannot determine version of %r", dist_name)
+            except Exception:
+                LOGGER.warning(
+                    "Unexpected error occurred while checking the version of %r",
+                    dist_name,
+                    exc_info=True,
+                )
 
         return version
 
--- a/test/plugins/test_plugins_handlers.py
+++ b/test/plugins/test_plugins_handlers.py
@@ -142,3 +142,21 @@ def test_folder_plugin_imports(plugin_folder):
     handler = handlers.PyFilePlugin(plugin_folder)
     handler.load()
     assert handler.module.foo == 'bar baz'
+
+
+def test_get_version_entrypoint_package_does_not_match(plugin_tmpfile):
+    # See gh-2593, wherein an entrypoint plugin whose project/package names
+    # are not equal raised an exception that propagated too far
+    distrib_dir = os.path.dirname(plugin_tmpfile.strpath)
+    sys.path.append(distrib_dir)
+
+    try:
+        entry_point = importlib.metadata.EntryPoint(
+            'test_plugin', 'file_mod', 'sopel.plugins')
+        plugin = handlers.EntryPointPlugin(entry_point)
+        plugin.load()
+        plugin.module.__package__ = "FAKEFAKEFAKE"
+        # Under gh-2593, this call raises a PackageNotFound error
+        assert plugin.get_version() is None
+    finally:
+        sys.path.remove(distrib_dir)


sopel/plugins/handlers.py
```diff
@@ -618,8 +618,9 @@ def get_version(self) -> Optional[str]:
         ):
             try:
                 version = importlib.metadata.version(self.module.__package__)
-            except ValueError:
-                # package name is probably empty-string; just give up
+            except:
```
dgw, line 621: Linter won't be happy unless you use:

```suggestion
            except Exception:
```

dgw:
I dug into the original issue _just a bit_ (literally, this time) and found that we probably could differentiate the `EntryPointPlugin` type from its parent class `PyModulePlugin` a bit more.

This PR works as a quick fix for the exact issue you ran into, but I do think the real solution should be actually handling entry points correctly (i.e. not just as "spicy modules" like the current impl).

dgw:
Bonked on the underlying issue here with my newest plugin project, which bundles multiple related plugins into a single package as `the_collection.thing1.plugin` and so on. Sopel of course fails to get the version of such plugins (`PackageNotFoundError` for `the_collection.thing1`) because of trying to use the plugin `module`'s `__package__` attr.

My patch below is not tested thoroughly, but if you want to come back to this branch it already has a test ready and you probably have Tox set up locally to run said test suite across all supported Python versions. 😸

GitHub still has that annoying limitation on how far away from the actual change you can start a line note, so unfortunately I can't make this into a one-click suggestion:

```py
        # inside `EntryPointPlugin.get_version()`; you already know where,
        # since this PR already touches the same area 😉
        if (
            version is None
            and hasattr(self.entry_point, "dist")
            and hasattr(self.entry_point.dist, "name")
        ):
            try:
                version = importlib.metadata.version(self.entry_point.dist.name)
            except Exception:
                # fine, just give up
                pass
```

Exirel:
> This PR works as a quick fix for the exact issue you ran into, but I do think the real solution should be actually handling entry points correctly (i.e. not just as "spicy modules" like the current impl).
> 
> For checking the version in particular, an `EntryPoint` object has a `dist` attribute, which itself has a `name` that is most likely what would be expected by the `importlib.metadata.version()` call whose failure in this mismatched-name case spawned #2593.

I was looking for that solution, and yes, this is exactly the right solution here, as demonstrated by this snippet of code (running with Python 3.10):

```
>>> from importlib.metadata import entry_points, version
>>> for entry_point in entry_points(group='console_scripts'):
...     print(entry_point.name, entry_point.dist.name, version(entry_point.dist.name))
...
pip pip 23.0.1
pip3 pip 23.0.1
pip3.9 pip 23.0.1
```

Once you have an `EntryPoint` object, you can ask its `dist`, which has a `name`, and that's the name you want when calling `version(name)`.

Alternatively, I don't mind if we separate the plugin version from the package version, i.e. allow a Plugin author to use `__version__` in the plugin file (as exposed by the entry point).

dgw:
> Alternatively, I don't mind if we separate the plugin version from the package version, i.e. allow a Plugin author to use `__version__` in the plugin file (as exposed by the entry point).

I think the intent is already there in the code, but checking whether the version is `None` after calling `super().get_version()` before `EntryPointPlugin` tries its own logic. But we can test it when fixing this dist-name bug!

SnoopJ:
After discussing on IRC, what's left here is for me to integrate @dgw's suggested tweak, and also to see about adding a logger instance here to strike a balance between more broadly catching exceptions and falling back on the `None` base case for `get_version()` and recording what went wrong.

sopel/plugins/handlers.py
```diff
@@ -618,8 +618,9 @@ def get_version(self) -> Optional[str]:
         ):
             try:
                 version = importlib.metadata.version(self.module.__package__)
-            except ValueError:
-                # package name is probably empty-string; just give up
+            except Exception:
+                # Just give up
+                # Can be caused by empty package name, mismatch between package/project names
```
dgw, line 623: We've been talking about this one on IRC this evening, and I believe @SnoopJ and I have found a good middle ground:

If logging works here (we haven't tested how logging interacts with the CLI interfaces like `sopel-plugins`), it would be good to catch ``ValueError` and `PackageNotFoundError` which the `importlib.metadata.version()` call might raise, and just give up. Maybe log a warning, but that's it. As a fallback, catch `Exception` and log the full exception traceback for investigation. We'd like to know about other failure cases if they exist, but they shouldn't crash the CLI or the `.version` command called from IRC.

dgw:
I had a partial implementation of what's discussed in https://github.com/sopel-irc/sopel/pull/2594#discussion_r1767683659 and committed it on top of this branch to clean out uncommitted changes in my Gitpod workspace (they are apparently discontinuing their turnkey cloud IDE product, which sucks). CI failures appear unrelated; the error is in `url`:

```
mypy --check-untyped-defs sopel
sopel/builtins/url.py:440: error: Argument 1 to "ip_address" has incompatible type "Rdata"; expected "Union[int, str, bytes, IPv4Address, IPv6Address]"  [arg-type]
```

Must have been an update somewhere since the last type-checking cleanup in #2614.

test/plugins/test_plugins_handlers.py
```diff
@@ -100,3 +100,21 @@ def test_get_label_entrypoint(plugin_tmpfile):
     assert meta['label'] == 'plugin label'
     assert meta['type'] == handlers.EntryPointPlugin.PLUGIN_TYPE
     assert meta['source'] == 'test_plugin = file_mod'
+
+
+def test_entrypoint_plugin_get_version(plugin_tmpfile):
```
Exirel, line 105: This test case doesn't have any ` assert`. What is it checking?

Exirel (changes requested):

test/plugins/test_plugins_handlers.py
dgw: I'm not sure that it checks anything meaningful. When I couldn't break this new test case, even by modifying the code being tested to call `importlib.metadata.version()` with a nonexistent distribution name or to remove the try/except structure (in case it was catching an error and just `pass`ing), I got suspicious.

In the context of this test, the EntryPoint's `dist` attribute is `None`, so because of the `and hasattr(self.entry_point.dist, "name")` clause of the `if` statement edited in `EntryPointPlugin.get_version()`, that whole block is skipped and the return value is just the value from `super().get_version()`.

My conclusion is that this test doesn't set up the EntryPoint sufficiently. We can't meaingfully test "get the version of a plugin whose entry point name doesn't match its parent distribution's name" if the entry point isn't part of a distribution. It needs to also create a mock distribution, which I started looking into but ran out of time for.

SnoopJ: The check is that exceptions other than `ValueError` do not propagate beyond the version checking code.

Running the same test on `master` (2e8de023) produces the `PackageNotFoundError` that this test covers, because `master` has a too-narrow exception handler.

```
test/plugins/test_plugins_handlers.py:160: in test_entrypoint_plugin_get_version
    plugin.get_version()
sopel/plugins/handlers.py:627: in get_version
    version = importlib.metadata.version(self.module.__package__)
/home/jgerity/.pyenv/versions/3.8.14/lib/python3.8/importlib/metadata.py:530: in version
    return distribution(distribution_name).version
/home/jgerity/.pyenv/versions/3.8.14/lib/python3.8/importlib/metadata.py:503: in distribution
    return Distribution.from_name(distribution_name)
/home/jgerity/.pyenv/versions/3.8.14/lib/python3.8/importlib/metadata.py:177: in from_name
    raise PackageNotFoundError(name)
E   importlib.metadata.PackageNotFoundError: FAKEFAKEFAKE
```

However, I didn't remember any of this immediately from looking at the code again, so the test can be rewritten to make this more clear.

dgw: I think @Exirel's point is that the test only verifies the absence of an exception, rather than also verifying that correct version metadata is fetched for this case.

Ideally we'd have several test cases related to this:

1. The "normal" case: Distribution name and package/module name match (e.g. `sopel_testplugin` installable with `pip install sopel_testplugin`)
   Assert: plugin version matches the distribution version.
2. Distribution name and package name don't match (e.g. `sopel_myplugins.test` installable with `pip install my_sopel_plugins`), where `sopel_myplugins.test` has its own version that would be returned by `super().get_version()`
   Assert: plugin version matches the version of `sopel_myplugins.test`
3. Same as 2, but `sopel_myplugins.test` does _not_ have its own version string so we fall through to getting the distribution's version instead.
   Assert: plugin version matches the distribution version
4. Same as 3, but `entry_point` does not have a `dist` (or the `dist` does not have a `name`).
   Assert: plugin version is `None`; all attempts to get one from the module or distribution failed

As the test is currently written, the code that would have raised `PackageNotFoundError` on `master` is simply skipped over. The `if` statement now checks `hasattr(self.entry_point, "dist") and hasattr(self.entry_point.dist, "name")`, which evaluates as `False` because no distribution is mocked up; it's a bare entry point with no parent.

That's a fine case to check, but we also need at least some behavior tests where the distribution-related conditions actually pass, and a (mock) distribution's metadata can be inspected.

SnoopJ: > rather than also verifying that correct version metadata is fetched for this case.

See 7d6c7b1 which adds an assertion, if not a particularly strong one.

I agree that there are probably several tests for write for **the feature**, but this PR is concerned primarily with (2), the bug it is meant to fix. In other words, this is a regression test to cover the bug I found and *only* this case.

sopel/plugins/handlers.py
SnoopJ: Done in d5f4c1a. The logger does "work" on the command-line (i.e. `sopel-plugins list`) in the sense that it produces output. I think it happens because the logging hierarchy is unconfigured and Python falls back on assigning a default handler to the root of the hierarchy. At any rate, it doesn't fail silently which is the main thing we were after.

test/plugins/test_plugins_handlers.py
Exirel: I'm fine with the PR not testing everything. However, I still want a test that represent an existing use-case, and describe what output is expected in said use case (i.e. what would be the result of `get_version()`).

Looking at the latest diff, it looks like you added the appropriate assert, and so I am happy!

test/plugins/test_plugins_handlers.py
```diff
@@ -100,3 +100,21 @@ def test_get_label_entrypoint(plugin_tmpfile):
     assert meta['label'] == 'plugin label'
     assert meta['type'] == handlers.EntryPointPlugin.PLUGIN_TYPE
     assert meta['source'] == 'test_plugin = file_mod'
+
+
+def test_entrypoint_plugin_get_version_does_not_raise(plugin_tmpfile):
```
Exirel, line 105: To be on brand with the rest of the file, I suggest this nitpicky modification:

```suggestion
def test_get_version_entrypoint_package_does_not_match(plugin_tmpfile):
```

Exirel (approved):
I approve on the premise that you'll rebase on master and solve the merge conflict. I added a nitpick but it's not blocking anything.

test/plugins/test_plugins_handlers.py
```diff
@@ -102,19 +102,19 @@ def test_get_label_entrypoint(plugin_tmpfile):
...
     distrib_dir = os.path.dirname(plugin_tmpfile.strpath)
     sys.path.append(distrib_dir)
 
-    # load the entry point
     try:
         entry_point = importlib.metadata.EntryPoint(
             'test_plugin', 'file_mod', 'sopel.plugins')
         plugin = handlers.EntryPointPlugin(entry_point)
-        plugin.load()
         plugin.module.__package__ = "FAKEFAKEFAKE"
```
dgw, line 115: Trying to access `plugin.module` before calling `plugin.load()` should raise `RuntimeError`:

https://github.com/sopel-irc/sopel/blob/2e8de02325c4842086eec16f4ee532437643bd5d/sopel/plugins/handlers.py#L279-L283

Before loading and setting the module:

https://github.com/sopel-irc/sopel/blob/2e8de02325c4842086eec16f4ee532437643bd5d/sopel/plugins/handlers.py#L610-L611

… the module is initialized to `None` by the superclass:

https://github.com/sopel-irc/sopel/blob/2e8de02325c4842086eec16f4ee532437643bd5d/sopel/plugins/handlers.py#L277

SnoopJ: D'oh, I got that backwards. At least it's in an easily-reverted commit.

SnoopJ:
Sorry about the force-push noise, but this should clear the unrelated typing problems in CI now and is ready for final review. Happy to squash it further if desired.

dgw (approved):
I agree with @Exirel about squashing to a single commit. While I could suggest merging some changes together and rearranging the chronology to make a multi-commit history that tells a coherent story about the feature, this patch isn't big enough to make it worth the time it'd take to carefully rewrite history that way.

Oh, I also played around a little and found that `self.entry_point.dist` (if it exists) has a `version` attr. I have _not_ looked at all the edge cases that `importlib.metadata.version()` should theoretically handle for us, though, so don't mess with the patch any more. This is a curiosity I wanted to mention, not patch feedback!

SnoopJ:
Squishy squashed again, ready to rock

dgw:
Realized it's been over a week since this got squashed and readied for re-review… Had to tear myself away from working on updates to a Factorio mod (balance changes for 2.0, etc.) 😅

PR #2601, opened 2024-03-18
adminchannel: rework topic-mask commands

### LINKED ISSUE #2596: adminchannel: empty `.tmask` handled incorrectly
### Description

Running `.tmask` without any argument technically sets an empty topic mask, but in a broken way.

### Reproduction steps

0. Enable `adminchannel` plugin if not enabled
1. As chanop, do just `.tmask` (no argument)
2. Do `.showmask`
3. Sopel responds with an exception:
   `Unexpected TypeError (can only concatenate str (not "NoneType") to str) from dgw. Message was: .showmask`

### Expected behavior

`adminchannel` should properly _clear_ the topic mask in this case, such that doing `.showmask` yields the same output as if `.tmask` had never been set at all.

Additionally, `.showmask` should handle `None` as if the topic mask is unset. Defense in depth!

### Relevant logs

_No response_

### Notes

I've a half-assed patch for this in `adminchannel-tmask-none` branch, but it's untested and probably needs some more work before becoming a pull request.

### Sopel version

973a489355540d68b95db01a49e983ac7a740bcc

### Installation method

`pip install`

### Python version

3.10.7

### Operating system

_No response_

### IRCd

_No response_

### Relevant plugins

adminchannel

dgw:
### Description

- `.tmask get` shows the current topic mask
  - `.tmask` without arguments assumes the `get` subcommand
- `.tmask set` now sets the topic mask, and provides a usage hint if called without further arguments
- `.tmask clear` is added to explicitly clear the topic mask

In combination, the above should be sufficient to resolve #2596.

### Checklist

### Notes

If no one minds adding this to 8.0.0 while the release notes are still under peer review, I'm happy to get it shipped sooner.

_(Because it's modifying command behavior, if we don't ship this patch in 8.0.0, I believe the semver-appropriate thing to do would be to ship the first commit in 8.0.1 and hold the second for 8.1.0.)_

### DIFF
--- a/sopel/builtins/adminchannel.py
+++ b/sopel/builtins/adminchannel.py
@@ -365,31 +365,35 @@ def topic(bot, trigger):
 @plugin.require_privilege(plugin.OP, ERROR_MESSAGE_NO_PRIV)
 @plugin.command('tmask')
 def set_mask(bot, trigger):
-    """Set the topic mask to use for the current channel
+    """Set or get the topic mask to use for the current channel.
 
-    Within the topic mask, {} is used to allow substituting in chunks of text.
+    This mask is used when running the 'topic' command. `{}` is used to allow
+    interpolating chunks of text within the topic mask.
 
-    This mask is used when running the 'topic' command.
+    If called without arguments, prints the channel's current topic mask.
     """
     new_mask = trigger.group(2)
 
     if new_mask is None:
-        bot.db.delete_channel_value(trigger.sender, 'topic_mask')
-        bot.reply('Cleared topic mask.')
+        mask = bot.db.get_channel_value(
+            trigger.sender, 'topic_mask', default_mask(trigger))
+        bot.reply('Current topic mask: {}'.format(mask))
         return
 
     bot.db.set_channel_value(trigger.sender, 'topic_mask', new_mask)
     message = (
         'Topic mask set. '
-        'Use `{prefix}topic <args>` to set topic '
-        'and `{prefix}showmask` to see current mask.'
+        'Use `{prefix}topic <args>` to set topic, '
+        '`{prefix}tmask` without arguments to see the current mask, '
+        'and `{prefix}cleartmask` to clear the topic mask.'
     ).format(prefix=bot.settings.core.help_prefix)
     bot.reply(message)
 
 
 @plugin.require_chanmsg
 @plugin.require_privilege(plugin.OP, ERROR_MESSAGE_NO_PRIV)
-@plugin.command('showmask')
-def show_mask(bot, trigger):
-    """Show the topic mask for the current channel."""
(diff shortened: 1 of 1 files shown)


sopel/builtins/adminchannel.py
```diff
@@ -366,24 +365,35 @@ def topic(bot, trigger):
...
     bot.reply(message)
 
 
 @plugin.require_chanmsg
 @plugin.require_privilege(plugin.OP, ERROR_MESSAGE_NO_PRIV)
-@plugin.command('showmask')
-def show_mask(bot, trigger):
-    """Show the topic mask for the current channel."""
-    bot.say(bot.db.get_channel_value(trigger.sender, 'topic_mask', default_mask(trigger)))
+@plugin.command('cleartmask')
```
Exirel, line 395: What do you say about that alias?

```suggestion
@plugin.command('cleartmask', 'ctmask')
```

Exirel (approved):
I'm all for it. I made an optional suggestion for a "ctmask" alias.

sopel/builtins/adminchannel.py
half-duplex: My general preference is still one command - No arguments to show the current mask, empty quotes or `-clear` to clear, and content to set; or get(or empty)/set/clear subcommands.
Otherwise, agreed - I don't think clearing the mask is something that will need to be done often, so I don't see a significant benefit to a shorter alias. I would end up looking it up every time anyway.

dgw:
Refactored to use subcommands. I kind of like this style, actually. The `.help` situation is a bit messy, though the details of that would be a discussion for the `sopel-help` issue tracker.

sopel/builtins/adminchannel.py
```diff
@@ -362,28 +361,48 @@ def topic(bot, trigger):
...
-        'Topic mask set. '
-        'Use `{prefix}topic <args>` to set topic '
-        'and `{prefix}showmask` to see current mask.'
-    ).format(prefix=bot.settings.core.help_prefix)
-    bot.reply(message)
+    if subcommand == 'set':
+        if not trigger.group(2):
+            message = (
+                'I need a non-empty topic mask to set. '
+                'To delete the saved topic mask, use {prefix}tmask clear.'
```
Exirel, line 390:
```suggestion
                'To delete the saved topic mask, use `{prefix}tmask clear`.'
```

Exirel (approved):
Nitpicking aside, LGTM.

sopel/builtins/adminchannel.py
dgw: Good catch; I amended the patch. Thanks!

dgw:
Last night, @half-duplex said in our IRC: "reading only the strings and decorators in 2601: lgtm" — and that puts to rest the earlier discussion about `.cleartmask` vs. `.tmask clear` or other single-command structure. Shipping… :shipit:

sins921 (approved):

PR #2600, opened 2024-03-16
8.0.0 changelog

dgw:
Since I only wanted to work on this if I could do a whole page of 25 PRs in one sitting, it sometimes took over a week between sessions. So here we are in March despite me starting on this in earnest two months ago.

I don't know how far along @SnoopJ has gotten with putting feedback on sopel-irc/sopel.chat#46 but this and that should probably be reviewed somewhat together. The changelog (this) summarizes the highlights, and the upgrade guide (that) gives all the details with suggested replacements for deprecated/removed features.

This **will** be a lesson for future changelogs, and we're going to have an `8.1-NEWS`, `9.0-NEWS`, etc. branch for the next releases. I've learned that the way we did it for 7.0.0 was Good, Actually!

And yes, I am very very tempted to add this PR itself to the changelog. We like being meta around here, right?

PR #2606, opened 2024-04-24
Final 8.0 release prep

dgw:
### Description

My own branch protection settings won't let me commit a quick fix directly to `master`, so I guess I'll open a catch-all PR for this and any other stuff that was forgotten as part of #2600.

### Checklist

  - Didn't touch code, so if something breaks it's not my fault 🙃
  - Ditto above, didn't touch code


SnoopJ:
Not sure if this is a draft because you intend to add other things, or in case _we_ spot something that needs adding. Either way, changes in it as of now LGTM.

dgw:
It is a draft for both reasons, until the website stuff is done. Then I'll have to version-bump as part of the PR (because of those branch protection rules already mentioned). 😁

dgw:
I uploaded a build of 3261d9c714fdc05ea373cc71842eb70e2f3cbb6b to https://test.pypi.org/project/sopel/8.0.0/ and everything seems to work.

Speaking of 3261d9c714fdc05ea373cc71842eb70e2f3cbb6b, when I examine (Test)PyPI's JSON API output, I can see that `sopel-pypi` will show the whole author list ("dgw, Florian Strzelecki, Sean B. Palmer, [etc.]"). Previous releases have only listed me as the author, and we are correcting that in the 8.x series. (Let's think of this as an opportunity to find `sopel-pypi` output improvements.)

The whole list is _arguably_ still better than "(unknown name)", which is what it does for projects with both a name and email, like many of the plugins I/we have switched over to `pyproject.toml` metadata. There seems to be an unsolved issue in Python's ecosystem relating to `pyproject.toml` / PEP 621 metadata and non-empty `email`; see pypi/warehouse#9400 (and also pypi/warehouse#14813, a related but distinct issue with specifying `email` for some-but-not-all authors or maintainers).

dgw:
> I used the test index to install Sopel 8 and it installed successfully, without any problem for the help plugin.

Is this a reference to sopel-irc/sopel-help#11? I'm sure it doesn't affect _all_ (or even _most_) Python environments; how common is it not to have `setuptools` installed at all? 😂 But that's also an issue for `sopel-help` to solve if necessary, and definitely shouldn't be a blocker for the main bot itself. That was the whole point of starting to extract nonessential plugins from core 😁

Exirel:
Not even a ref to that, I just made sure that the install worked as intended because sopel-help must not break Sopel's install.

SnoopJ (approved):
Does it count as super-approving if I accidentally approve it from my work account first? :facepalm:

PR #2613, opened 2024-08-13
search: remove built-in plugin with external replacement

dgw:
### Description

The built-in `search` has been unreliable, often yielding no results because it probably gets blocked from scraping Bing or DuckDuckGo SERPs. (This mechanism isn't entirely clear, as it still does work from some machines—including places like Gitpod cloud development servers that presumably live in a higher-profile public cloud than any of the bots with problems.)

However, a near-scratch rewrite at [sopel-irc/sopel-search](https://github.com/sopel-irc/sopel-search) using a library to access DDG seems to be significantly more reliable, and lets us drop `xmltodict` as a core dependency; `search.py` was the last place using it.

### Checklist

  - I have realized that `xkcd` still wants to use `duck_search()` as a fallback, so am opening this as a draft with linting expected to fail.
  - Ditto above re: `xkcd`

### Notes

I see no problem with doing this in 8.1, but am open (as always) to other opinions 😸

Happy for feedback on anything other than the `xkcd` bits I have to take care of when I'm next at a computer.


dgw:
Staying in draft until type-checking errors are fixed, which should be accomplished by #2614.

dgw:
Rebased onto `master` after merging #2614. Finally ready to review 😁

sopel/builtins/xkcd.py
```diff
@@ -26,19 +26,21 @@
...
-    'wiki.xkcd.com',
-    'what-if.xkcd.com',
-]
-sites_query = ' site:xkcd.com -site:' + ' -site:'.join(ignored_sites)
+
+class SearchXkcdError(Exception):
+    """Generic exception to raise if there was a problem contacting the API."""
+
+
+class SearchFailedError(SearchXkcdError):
```
SnoopJ, line 34: Nitpick: That seems a little specialized to me, I would probably have used the base exception to cover this.

sopel/builtins/xkcd.py
```diff
@@ -70,19 +62,20 @@ def searchxkcd_search(query):
...
         hits = response.json()['results']['hits']
         if not hits:
-            return None
+            raise NoResultsError
         first = hits[0]['objectID']
     except (JSONDecodeError, LookupError):
-        LOGGER.debug("Data format from searchxkcd API could not be understood.")
-        return None
+        msg = "Data format from searchxkcd API could not be understood."
+        LOGGER.debug(msg)
```
SnoopJ, line 77: Huh, I would have expected a warning here

test/vcr/builtins/search/test_example_duck_0.yaml
SnoopJ, line 1: Bumper YAML crop this year!

sopel/builtins/xkcd.py
```diff
@@ -124,13 +118,16 @@ def xkcd(bot, trigger):
...
-                    # generic web-search engine as fallback
-                    number = web_search(query)
-                if not number:
-                    bot.reply('Could not find any comics for that query.')
+                try:
+                    number = searchxkcd_search(query)
+                except NoResultsError:
+                    bot.reply("Sorry, I couldn't find any comics for that query.")
                     return
+                except (ResponseFormatError, SearchFailedError, SearchXkcdError):
```
SnoopJ, line 126: Nitpick: Why catch the subclasses here?

sopel/builtins/xkcd.py
```diff
@@ -98,22 +91,23 @@ def xkcd(bot, trigger):
...
+
     """
-    # get latest comic for rand function and numeric input
+    # get latest comic, for random selection and validating numeric input
     latest = get_info()
     max_int = latest['num']
 
     # if no input is given (pre - lior's edits code)
-    if not trigger.group(2):  # get rand comic
+    if not trigger.group(2):  # get random comic
```
SnoopJ, line 104: Thank you for making that easier to read! 😁

SnoopJ (approved):
I'm on mobile so my review is going to be lower quality, but this LGTM

sopel/builtins/xkcd.py
dgw: I'm gonna say it's because I did this at some ungodly wee hour and forgot that they were subclasses despite just creating the new exception types mere moments earlier. Possibly also related to the fact that I started out creating an `Enum` type for the search function to return on errors, before deciding exceptions were better.

tl;dr: I dunno, weird brain stuff. 😹

sopel/builtins/xkcd.py
dgw: Maybe it needs a better name? The idea was to shadow `requests.exceptions.ConnectionError` with a plugin-specific error type—note where `SearchFailedError` is used below. It's an attempt to differentiate between errors of the "I talked to the API but couldn't understand what it said back" kind and more serious "the API might be gone???" issues.

I might look at this again when I go through to fix up things like your logging comment and reevaluate the exception hierarchy, even though you did say "LGTM".

sopel/builtins/xkcd.py
dgw: To be fair I often err on the side of "don't spew stuff into logs if the user isn't actively debugging", but this one doesn't carry a big dump of stuff with it. What do you think of a warning with the user-friendly error message here and then a debug logging call with the API response?

dgw (commented):
I love how SnoopJ makes an excuse (being on mobile) for the quality of his review, then makes more comments than most reviewers I work with. 😂

sopel/builtins/xkcd.py
```diff
@@ -124,13 +118,16 @@ def xkcd(bot, trigger):
...
-                    bot.reply('Could not find any comics for that query.')
+                try:
+                    number = searchxkcd_search(query)
+                except NoResultsError:
+                    bot.reply("Sorry, I couldn't find any comics for that query.")
                     return
+                except (ResponseFormatError, SearchFailedError, SearchXkcdError):
+                    bot.reply(
+                        "A technical problem prevented me from searching. "
+                        "Please ask my owner to check my logs.")
```
dgw, line 129:
```suggestion
                        "Please ask my owner to check my logs.")
                        return
```

Don't want to proceed without a search result, do we?

sopel/builtins/xkcd.py
Exirel: I think a warning is a good thing here.

Exirel (approved):

PR #2614, opened 2024-08-13
Miscellaneous type-checking updates

dgw:
### Description

`types-pkg-resources` was yanked, which I discovered when submitting #2613. `pkg_resources` itself is typed in setuptools>=71.1, negating the need to install type stubs.

When testing that fix, I updated `mypy` and friends (`pip install -U -r dev-requirements.txt`) and found a few more new type-check errors caught by the newer tools:

- `SopelIdentifierMemory.update()`'s signature isn't compatible with its supertype (fixed by adding `typing_extensions.override` decorator; can't go stdlib-only until py3.12…)
- `tools.OutputRedirect` calls methods on `sys.__stdout__` and `sys.__stderr__`, which are allowed to be `None` and therefore have to be guarded with `if`s (I'm glad this class is deprecated)

### Checklist

### Notes

All of this needs backporting to the `8.0.x` maintenance branch, which is why this has the 8.0.1 milestone.

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -22,7 +22,9 @@ sphinx-rfcsection~=0.1.1
 vcrpy @ git+https://github.com/sopel-irc/vcrpy@uncap-urllib3
 # type check
 mypy>=1.3,<2
+# for `pkg_resources`; first version in which it's typed
+# we don't use `pkg_resources` directly, but mypy still cares
+setuptools>=71.1
 sqlalchemy[mypy]>=1.4,<1.5
-types-pkg-resources~=0.1.3
 types-pytz
 types-requests>=2,<3
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -52,6 +52,7 @@ dependencies = [
     "importlib_metadata>=3.6",
     "packaging>=23.2",
     "sopel-help>=0.4.0",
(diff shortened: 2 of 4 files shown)


dgw:
Manually re-running CI jobs got them to pass, but I do _not_ like the random failure in 3.9's `sopel/builtins/rand.py::test_example_rand_4` test that canceled the remaining unfinished jobs. I believe this error should be impossible, but it happened somehow and probably needs its own investigation (😩):

```
=================================== FAILURES ===================================
_____________________________ test_example_rand_4 ______________________________
sopel/tests/pytest_plugin.py:137: in test
    assert len(outputs) == expected_output_count
E   AssertionError: assert 9 == 10
E    +  where 9 = len(['random(0, 2): 1', 'random(0, 2): 1', 'random(0, 2): 1', 'random(0, 2): 2', 'random(0, 2): 1', 'random(0, 2): 1', ...])
```

`test_example_rand_4` is one of these two in `sopel/builtins/rand.py`, probably the second:

https://github.com/sopel-irc/sopel/blob/1237568f05a4155f80e12ff0fd3711e8ed97d976/sopel/builtins/rand.py#L20-L21

Exirel (approved):
Yeah all good, I'm glad we have the mypy check in the CI.

And also this:

> `removed_in='8.1'`

Can't wait for that!

Issue #2603, opened 2024-03-21
Gradual enforcement plan for stricter `mypy` checks


dgw:
As I've started to work with modern Sopel in the context of plugins that want to do their own type-checking ([e.g. sopel-iplookup](https://github.com/sopel-irc/sopel-iplookup/blob/5db42c9968e3496bbf8ae91c4afb862fc77a2a74/Makefile#L6)), I've also built up an opinion that we should really give ourselves a concrete target for officially declaring `sopel` as supporting type-checks, i.e. adding the `py.typed` file that it needs to silence mypy's `missing-imports` warning.

I propose that the target be "every piece of Sopel's public API has complete type hints". We can go above and beyond that with type hints for internal functions or plugin callables, but anything that we expect a plugin developer to import or touch (including the types passed through the `bot` and `trigger` args and all of the things they contain) should be fully hinted.

To help us find shortcomings, the following [`mypy` options](https://mypy.readthedocs.io/en/latest/command_line.html#untyped-definitions-and-calls) are available:

- `--disallow-incomplete-defs`: raises an error for function definitions where some, but not all, parameters are typed
- `--disallow-untyped-decorators`: errors when a type-hinted function is decorated by an untyped decorator
  (not sure how useful this one is for us, but it's around)
- `--disallow-untyped-defs`: superset of `--disallow-incomplete-defs`; _any_ `def` without type hints is an error
  This would be our final endgame, once EVERYTHING is typed, even the internals and built-in plugins.

For 8.0.1 we should definitely do a pass with `--disallow-incomplete-defs` and fix those as a first step. The goal—or at least, my goal—is to do the rest of the typing work along the road to 9.0 and ship `py.typed` with `--disallow-untyped-defs` enforcement by then.

_(In case you're wondering who to blame for this "make more work" issue, in my own defense, it was @SnoopJ who said "probably a good thing to start tracking for 9.0.0 milestone tho" in our IRC channel when I mentioned `--disallow-incomplete-defs`. 😜)_

Exirel:
I'm on it for the 8.0.1 milestone, and we are starting with few errors:

> Found 61 errors in 15 files (checked 81 source files)

That looks doable, I'll see if it's that easy.

Exirel:
The 8.0.1 portion is done. I'm moving the target to 8.1.0!

Exirel:
Just for information, as of today:

> Found 674 errors in 66 files (checked 81 source files)

Yeah OK. There is some work to be done here.

PR #2616, opened 2024-09-01
mypy: add --disallow-incomplete-defs option

Exirel:
### Description

Tin, with the necessary type definition fix. Related to #2603 for Sopel 8.0.1.

Note: I used `Any` in several places, in particular in `sopel.tools.identifiers`. I'm not really happy about it. However, the best fix is to properly type annotate instances, i.e. doing something like `users: dict[str, User] = SopelIdentifierMemory()` as it should properly type check the `users` sopel memory object.

### Checklist

### DIFF
--- a/sopel/bot.py
+++ b/sopel/bot.py
@@ -17,7 +17,9 @@ import time
 from types import MappingProxyType
 from typing import (
     Any,
+    Callable,
     Optional,
+    Sequence,
     TYPE_CHECKING,
     TypeVar,
     Union,
@@ -36,6 +38,8 @@ from sopel.trigger import Trigger
 
 if TYPE_CHECKING:
     from collections.abc import Iterable, Mapping
+
+    from sopel.plugins.handlers import AbstractPluginHandler
     from sopel.trigger import PreTrigger
 
 
@@ -182,7 +186,6 @@ class Sopel(irc.AbstractBot):
 
         :return: the bot's current hostmask if the bot is connected and in
                  a least one channel; ``None`` otherwise
-        :rtype: Optional[str]
         """
         if not self.users or self.nick not in self.users:
             # bot must be connected and in at least one channel
@@ -198,11 +201,11 @@ class Sopel(irc.AbstractBot):
         """
         return MappingProxyType(self._plugins)
 
-    def has_channel_privilege(self, channel, privilege) -> bool:
+    def has_channel_privilege(self, channel: str, privilege: int) -> bool:
         """Tell if the bot has a ``privilege`` level or above in a ``channel``.
 
-        :param str channel: a channel the bot is in
-        :param int privilege: privilege level to check
+        :param channel: a channel the bot is in
+        :param privilege: privilege level to check
         :raise ValueError: when the channel is unknown
 
         This method checks the bot's privilege level in a channel, i.e. if it
@@ -339,10 +342,10 @@ class Sopel(irc.AbstractBot):
 
     # plugins management
 
-    def reload_plugin(self, name) -> None:
+    def reload_plugin(self, name: str) -> None:
         """Reload a plugin.
 
-        :param str name: name of the plugin to reload
+        :param name: name of the plugin to reload
         :raise plugins.exceptions.PluginNotRegistered: when there is no
             ``name`` plugin registered
 
@@ -391,22 +394,24 @@ class Sopel(irc.AbstractBot):
 
     # TODO: deprecate both add_plugin and remove_plugin; see #2425
 
-    def add_plugin(self, plugin, callables, jobs, shutdowns, urls) -> None:
+    def add_plugin(
+        self,
+        plugin: AbstractPluginHandler,
+        callables: Sequence[Callable],
+        jobs: Sequence[Callable],
+        shutdowns: Sequence[Callable],
+        urls: Sequence[Callable],
+    ) -> None:
         """Add a loaded plugin to the bot's registry.
 
         :param plugin: loaded plugin to add
-        :type plugin: :class:`sopel.plugins.handlers.AbstractPluginHandler`
         :param callables: an iterable of callables from the ``plugin``
-        :type callables: :term:`iterable`
         :param jobs: an iterable of functions from the ``plugin`` that are
                      periodically invoked
-        :type jobs: :term:`iterable`
         :param shutdowns: an iterable of functions from the ``plugin`` that
                           should be called on shutdown
-        :type shutdowns: :term:`iterable`
         :param urls: an iterable of functions from the ``plugin`` to call when
                      matched against a URL
-        :type urls: :term:`iterable`
         """
         self._plugins[plugin.name] = plugin
         self.register_callables(callables)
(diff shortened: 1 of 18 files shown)


Exirel:
I rebased that over master and it passes the 3.13 mypy check as well!

dgw (changes requested):
It's a constant battle to keep the docs clean, with _only_ the type annotations _or_ hints in the `:param:` declarations, but not both.

While you go through and look for those (I didn't annotate every one, especially since many are outside where GH will allow line notes), also double check that merging #2285 didn't introduce any new wrinkles for this mypy option? 🙏

Exirel:
> but not both.

I fixed the functions I touched, and then fixed some for good measure!

> double check that merging #2285 didn't introduce any new wrinkles for this mypy option? 🙏

That would require a rebase first, so you tell me if it's OK and I'll do.

dgw (commented):
Lots of good stuff in here, but I do have a few questions (the same one in multiple places, actually) about the choice of type annotation syntax. Overall very impressed with how thorough you were with cleaning up the docstrings!

> > double check that merging #2285 didn't introduce any new wrinkles for this mypy option? 🙏
> 
> That would require a rebase first, so you tell me if it's OK and I'll do.

Now that I think about it, your fixups were tested against the latest `master`, including the merged changes from #2285. It passed, so you can just squash out the fixups after we settle the syntax question.

sopel/db.py
```diff
@@ -980,8 +982,8 @@ def get_nick_or_channel_value(
         self,
         name: str,
         key: str,
-        default=None
-    ) -> typing.Optional[typing.Any]:
+        default: typing.Any | None = None
+    ) -> typing.Any | None:
```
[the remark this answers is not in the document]
Exirel: Nope because it's an annotation. We can't use this syntax for type aliases, but we can use it in arg & return type annotations.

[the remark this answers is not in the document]
dgw: Hm, Sphinx even seems to handle it in the preview docs despite being stuck on py3.8, even.

OK, I'm convinced. Squash your fixups 😸

sopel/bot.py
```diff
@@ -1030,12 +1034,11 @@ def error(
         self,
         trigger: Optional[Trigger] = None,
         exception: Optional[BaseException] = None,
-    ):
+    ) -> None:
         """Called internally when a plugin causes an error.
 
         :param trigger: the ``Trigger``\\ing line (if available)
```
SnoopJ, line 1040: Nit: could probably drop the reference to the type here? I would probably also omit the word "line"

sopel/builtins/units.py
```diff
@@ -71,14 +77,16 @@ def temperature(bot, trigger):
...
-        return
+        return None
 
     bot.say("{:.2f}°C = {:.2f}°F = {:.2f}K".format(
         celsius,
         fahrenheit,
         kelvin,
     ))
 
+    return None
```
SnoopJ, line 88: Hmm, the explicit return here feels weird. No objection to it, but I would probably have left it off.

sopel/builtins/units.py
```diff
@@ -199,3 +209,5 @@ def mass(bot, trigger):
         stupid_part = '{:.2f} oz'.format(ounce)
 
     bot.say('{} = {}'.format(metric_part, stupid_part))
+
+    return None
```
SnoopJ, line 213: Same here: I probably would have left off the explicit return

sopel/lifecycle.py
```diff
@@ -13,21 +13,21 @@
...
 
 from sopel import __version__
 
 
 def deprecated(
-    reason: Optional[str] = None,
-    version: Optional[str] = None,
-    removed_in: Optional[str] = None,
-    warning_in: Optional[str] = None,
+    reason: str | Callable | None = None,
```
SnoopJ, line 24: Should also update the doc for `reason` to mention that it can be a callable.

sopel/lifecycle.py
```diff
@@ -13,21 +13,21 @@
 import inspect
 import logging
 import traceback
-from typing import Callable, Optional
+from typing import Callable
```
SnoopJ, line 16: Same here: Is there a reason to prefer the deprecated `typing.Callable` over `collections.abc.Callable`?

sopel/bot.py
```diff
@@ -17,7 +17,9 @@
 from types import MappingProxyType
 from typing import (
     Any,
+    Callable,
     Optional,
+    Sequence,
```
SnoopJ, line 22: Any reason not to import these from `collections.abc` since the `typing` counterparts are deprecated since Python 3.9?

sopel/plugins/rules.py
```diff
@@ -26,6 +26,7 @@
 import threading
 from typing import (
     Any,
+    Callable,
```
SnoopJ, line 29: Same comment here about `collections.abc` vs. `typing`

sopel/tools/calculation.py
```diff
@@ -42,7 +42,11 @@ def __init__(
         self.binary_ops = bin_ops or {}
         self.unary_ops = unary_ops or {}
 
-    def __call__(self, expression_str: str, timeout: float = 5.0):
+    def __call__(
+        self,
+        expression_str: str,
+        timeout: float = 5.0,
+    ) -> int | float:
```
SnoopJ, line 49: IIRC, `int | float` is redundant because `float` can annotate both per [PEP 484](https://peps.python.org/pep-0484/#the-numeric-tower)

I'm okay with explicitly annotating it as a way to document the specific type that will be present but would personally prefer an alias for the union in that case, as a place to leave a comment about why we don't do it the "usual" way.

sopel/tools/memories.py
```diff
@@ -221,19 +224,19 @@ def _convert_keys(
...
-    def __getitem__(self, key: Optional[str]):
+    def __getitem__(self, key: str | None) -> Any:
         return super().__getitem__(self._make_key(key))
 
-    def __contains__(self, key):
+    def __contains__(self, key: Any) -> Any:
         return super().__contains__(self._make_key(key))
 
-    def __setitem__(self, key: Optional[str], value):
+    def __setitem__(self, key: str | None, value: Any) -> Any:
```
SnoopJ, line 233: Nit: that can be `None`

```suggestion
    def __setitem__(self, key: str | None, value: Any) -> None:
```

SnoopJ (approved):
Had some remarks, but if the type checker is happy then so am I!

sopel/bot.py
dgw: Maybe this would work, but I wouldn't want to take too much context out of the docstring.

```suggestion
        :param trigger: the ``Trigger`` that led to the error, if available
```

sopel/builtins/units.py
dgw: Leaving it out is a mypy error:

```
sopel/builtins/units.py:58: error: Missing return statement  [return]
```

sopel/lifecycle.py
dgw: This is a "magic" behavior that allows both `@deprecated` and `@deprecated()` to work.

I'm not convinced that it would be useful to explicitly document passing a `Callable` for this argument. It isn't meant to be done explicitly.

sopel/bot.py
dgw: The only reason would be "not touching more than necessary", I think.

Switching from the deprecated imports to their replacements is unrelated to the PR's goal of making `--disallow-incomplete-defs` work.

sopel/lifecycle.py
SnoopJ: Ah, right.

sopel/builtins/units.py
dgw: I thought [my comment on the other thread](https://github.com/sopel-irc/sopel/pull/2616#discussion_r1801069390) would cover this, but it's the same here. The function returns `int | None`, so it must _always_ explicitly return a value.

Personally I don't mind the implicit `return None` at the end of a function, but the type-checker clearly does. 😸

sopel/tools/calculation.py
Exirel: Good call.

sopel/tools/memories.py
Exirel: Oh, nice catch.

sopel/bot.py
Exirel: Or maybe... ?

```suggestion
        :param trigger: the line leading to that error (if available)
```

dgw: Another idea:

```suggestion
        :param trigger: the IRC line that caused the error (if available)
```

This feels a lot like something I always have to spend a lot of time on at work 😂

sopel/bot.py
```diff
@@ -1030,12 +1034,11 @@ def error(
...
-    ):
+    ) -> None:
         """Called internally when a plugin causes an error.
 
-        :param trigger: the ``Trigger``\\ing line (if available)
-        :type trigger: :class:`sopel.trigger.Trigger`
-        :param Exception exception: the exception raised by the error (if
+        :param trigger: the IRC line that caused the error (if available)
+        :param exception: the exception raised by the error (if
                                     available)
```
dgw, line 1042: Oops, I missed this one before (it needs rewrapping after removing the type in `:param:`)

```suggestion
        :param exception: the exception raised by the error (if available)
```

Not much of this PR touched multi-line `:param:` descriptions, so I must have glossed over it. 😅

Issue #1222, opened 2017-09-15
Scheduled callable events


deathbybandaid:
Hi, not really an issue, but I wanted to know if the functionality existed and how to use it.

I have an event that is at 7pm daily, I want sopel to announce it to the channel.

I also want sopel to check an rss feed for updates at regular intervals, and if there is new content, to announce it to the channel.

I see there is something called "jobs", but I'm not sure how to add "jobs"

Thanks!

kwaaak:
https://github.com/RebelCodeBase/sopel-rss

Try .help at and .help in for your other question

deathbybandaid:
I looked at .help in and .help at

I'm not looking for something that I call from within IRC, but directly in the module.


I found a workaround, but I feel like there should be a better way:

````
import sopel.module
import datetime

# scheduled event is at 23:00 UTC daily
@sopel.module.interval(60)
def scheduledevent(bot):
    for channel in bot.channels:
        now = datetime.datetime.utcnow()
        if now.hour == 23 and now.minute == 10:
            bot.msg(channel, "This is the scheduled event")
````

#### I feel like what Sopel needs is something along the lines of:

````
import sopel.module
import datetime

@sopel.module.schedule(23,10)
def scheduledevent(bot):
    for channel in bot.channels:
        bot.msg(channel, "This is the scheduled event")
````

SpiceBot:
Additional Ideas:


## Idea 1, optionally pass  timezone along with it.

````
@sopel.module.schedule(23,10, timezone=UTC)
````

Wouldn't be often needed, but I could see it coming in handy.


## Idea 2

I have some stuff that I like to run when the bot is starting. Right now, I use a poor method:
(The nines are there in an attempt to not actually run every 5 seconds)

````

startupinterval = 5

# Startup check
@sopel.module.interval(startupinterval )
def startupcheck(bot):

    if "startup_op_check" in bot.memory:
        startupinterval += 999999999999999999
        return
    bot.memory["startup_op_check"] = True
    
    startupinterval += 999999999999999999

    """ Run The Code """

````

What I'd suggest is something more like:


````
@sopel.module.startup()
def startupevent(bot):
    """ do the thing """
````


## Idea 3, what I sometimes do for the bot joining a new channel

````
# watch Joins
@event('JOIN')
@rule('.*')
@sopel.module.thread(True)
def joindetect(bot, trigger):
    if trigger.nick == bot.nick:
        """ do the thing """
````

should have something like:

````
[...]

PR #2617, opened 2024-09-02
docs: add advanced tip about scheduling function calls

dgw:
How to do this was discussed fairly well in #1222 a while ago, and I just cribbed an example from there.

I also found an apparently obsolete note about using `sched` with `@plugin.rate` and friends while poking around the topic of scheduled callables, because I tested what it said and found that the caveat no longer applies (see #825).

### Checklist

### Notes

Along the way I also briefly considered giving the `bot` a `schedule` instance that it provides to plugins, but I'm not ready to commit to that kind of API addition yet. Putting this into the tips & tricks section is better than pointing to a random issue comment on the occasions when "how do I schedule a function at a specific time?" comes up, and it doesn't preclude API additions at a later time if we come up with a sane decorator interface for it.

### DIFF
--- a/docs/source/plugin/advanced.rst
+++ b/docs/source/plugin/advanced.rst
@@ -15,6 +15,45 @@ If something is not in here, feel free to ask about it on our IRC channel, or
 maybe open an issue with the solution if you devise one yourself.
 
 
+Running a function on a schedule
+================================
+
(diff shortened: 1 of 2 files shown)


Exirel (approved):
I love how simple that is. Nice!

dgw:
Ramping up to 8.0.1 soon, don't see why this should be held back for 8.1.

PR #2620, opened 2024-09-14
irc.utils, test: also remove null (\x00) in safe()

### LINKED ISSUE #2566: NUL (`\x00`) is not filtered from IRC args/text
### Requested Feature

`irc.utils.safe()` strips CR and LF from its input, but not NUL (`\x00`) even though all three octets [are disallowed in IRC lines](https://datatracker.ietf.org/doc/html/rfc2812#section-2.3.1).

https://github.com/sopel-irc/sopel/blob/3ff58c160ecc7a9d51f1e06426d456c7db4d985a/sopel/irc/utils.py#L20-L48

### Problems Solved

Plugins accidentally sending NUL will most likely cause the server to disconnect the bot with an error ("malformed data" or similar).

### Alternatives

Plugin authors must manually ensure that anything their code sends to IRC does not contain the NUL byte.

### Notes

_No response_

dgw:
### Description

Fixes #2566.

In `test_irc_utils` suite, rewrote `test_safe` and `test_safe_bytes` to generate a bunch more permutations of invalid characters automatically rather than writing them all out. The horse is definitely beaten *far* past death at this point... but the tests sure are thorough(bred).

Also renamed `test_safe_null` -> `test_safe_none` to avoid confusion; `test_safe` and `test_safe_bytes` are the cases that check behavior with null characters/bytes. Python's docs technically do refer to `None` as the "null object" once or twice, but most of the docs (and most of us devs) simply call it `None`.

### Checklist

  - Ran `make lint` myself, and tried out the test runner integration in VS Code / Gitpod. Very slick; made iterating my updated tests much easier without always tabbing back to the terminal.

### Notes

This could go into 8.0.1, but I don't really want to mess around with hotfixes unless there's a major bug. I'm open to suggestions for improving the maintenance-branch workflow; since `master` is _de facto_ for 8.1.0 dev now, the options are to keep the `8.0.x` branch unprotected so I can push cherry-picks to it directly, or make @sopel-irc/rockstars review the same PR twice (once to `master`, once to `8.0.x`) for full CI coverage. I don't love either of those.

### DIFF
--- a/test/irc/test_irc_utils.py
+++ b/test/irc/test_irc_utils.py
@@ -1,6 +1,8 @@
 """Tests for core ``sopel.irc.utils``"""
 from __future__ import annotations
 
+from itertools import permutations
+
 import pytest
 
 from sopel.irc import utils
@@ -8,15 +10,19 @@ from sopel.irc import utils
 
 def test_safe():
     text = 'some text'
-    assert utils.safe(text + '\r\n') == text
-    assert utils.safe(text + '\n') == text
-    assert utils.safe(text + '\r') == text
-    assert utils.safe('\r\n' + text) == text
-    assert utils.safe('\n' + text) == text
-    assert utils.safe('\r' + text) == text
-    assert utils.safe('some \r\ntext') == text
-    assert utils.safe('some \ntext') == text
-    assert utils.safe('some \rtext') == text
+    variants = permutations(('\n', '\r', '\x00'))
+    for variant in variants:
+        seq = ''.join(variant)
+        assert utils.safe(text + seq) == text
+        assert utils.safe(seq + text) == text
+        assert utils.safe('some ' + seq + 'text') == text
+        assert utils.safe(
+            variant[0]
+            + 'some '
+            + variant[1]
+            + 'text'
+            + variant[2]
+        ) == text
 
 
 def test_safe_empty():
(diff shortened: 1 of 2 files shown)


test/irc/test_irc_utils.py
```diff
@@ -1,43 +1,52 @@
...
-    assert utils.safe(text + '\n') == text
-    assert utils.safe(text + '\r') == text
-    assert utils.safe('\r\n' + text) == text
-    assert utils.safe('\n' + text) == text
-    assert utils.safe('\r' + text) == text
-    assert utils.safe('some \r\ntext') == text
-    assert utils.safe('some \ntext') == text
-    assert utils.safe('some \rtext') == text
+    variants = permutations(('\n', '\r', '\x00'))
+    for variant in variants:
```
Exirel, line 14: This calls for parametrized tests!

test/irc/test_irc_utils.py
```diff
@@ -1,43 +1,52 @@
...
-    assert utils.safe(text + b'\n') == text.decode('utf-8')
-    assert utils.safe(text + b'\r') == text.decode('utf-8')
-    assert utils.safe(b'\r\n' + text) == text.decode('utf-8')
-    assert utils.safe(b'\n' + text) == text.decode('utf-8')
-    assert utils.safe(b'\r' + text) == text.decode('utf-8')
-    assert utils.safe(b'some \r\ntext') == text.decode('utf-8')
-    assert utils.safe(b'some \ntext') == text.decode('utf-8')
-    assert utils.safe(b'some \rtext') == text.decode('utf-8')
+    variants = permutations((b'\n', b'\r', b'\x00'))
+    for variant in variants:
```
Exirel, line 41: And same here, of course.

Exirel (changes requested):
Even tho I don't like when tests have too much logic in them, I draw the line at assert inside for-in bloc. I think `@pytest.mark.parametrize` is a much better option for these tests.

Exirel (approved):

SnoopJ (approved):
Looks good :horse:

Is it maybe also worth mentioning in the `safe()` docstring that we strip `NUL` because it's explicitly disallowed by the spec?

Exirel:
Hey, this one could easily go into Sopel 8.0.1 to me. Like a hotfix of something that should be done but isn't.

dgw:
> Like a hotfix of something that should be done but isn't.

Yeah, it is, isn't it? I will tweak the `safe()` docstring though, as it really should mention NUL in the topline and also explain why the method exists at all.

dgw:
Netlify still does branch previews of our docs (for now… they are still stuck on Python 3.8 with no apparent plans to upgrade) but including a preview screenshot of the updated docs for this function anyway didn't seem like a bad idea.

image

Note I also fixed the `\x00` references in the docstring, which needs to use `\\x00` for the text to display as intended.

PR #2624, opened 2024-09-25
config.types.ChoiceAttribute: log the invalid value if validation fails

dgw:
### Description

This should make debugging a bad configuration easier, especially if the value is read from an environment variable.

It should also help with figuring out weird errors related to invisible characters in the configuration file, but that's not the primary motivation here.

### Checklist


Exirel (approved):
Could go in Sopel 8.0.1.

dgw:
> Could go in Sopel 8.0.1.

It could. Easy cherry-pick, so why not.

PR #2628, opened 2024-10-10
ci, contrib, dice, meta: enable testing on Python 3.13 + fix warnings

### LINKED ISSUE #2608: Python 3.13 support
### Requested Feature

Sopel should support Python 3.13 when possible. 3.13 entered beta on 7 May and is now feature complete, so it's a good time to start tracking what's needed for support. 3.13.0 final is scheduled for October 2024, and release candidates are scheduled for late July and early September.

### Problems Solved

_No response_

### Alternatives

_No response_

### Notes

Known problems:

* ~required `setuptools` version needs to be updated to 66.0 or higher (usage of deprecated `pkgutil.ImpImporter` by `setuptools.version`)~ done in #2516
* ~`greenlet` (required by `SQLAlchemy`) does not yet support Python 3.13 ([issue](https://github.com/python-greenlet/greenlet/issues/392))~
  * ~There _is_ a PR to add support, but I can't build it against 3.13.0b1: https://github.com/python-greenlet/greenlet/pull/396~
  * Fixed in `greenlet` 3.1.0: https://github.com/python-greenlet/greenlet/pull/396

dgw:
### Description

- ci: Added Python 3.13 to CI workflow
- contrib: Added Python 3.13 envs to tox config
- dice: removed unnecessary `count` parameter to `re.sub()` and pass `flags` as a kwarg
  (py3.13 deprecates these as positional args)
- url: fix type-check error with `dnspython` 2.7
  Use `Rdata.to_text()` to get each result as a string, which is one of the types that `ipaddress.ip_address()` accepts.
- meta: add Python 3.13 classifier for PyPI releases of 8.1+

With the above changes, the only warning emitted during `pytest` runs under Python 3.13 is from SQLAlchemy, which isn't new; it's because we haven't fully migrated to their 2.0 style yet, and not related to which Python version is used.

The type-check error from `dnspython` 2.7 isn't strictly Python 3.13-related, but this patch wouldn't pass linting without it. That change doesn't seem like it needs its own PR (though it might need to be backported for 8.0.1; we'll see).

Closes #2608.

### Checklist

  - Shockingly, I even ran the whole `tox` job on every version listed, after updating them all to the latest patch release available in `pyenv`. Maybe not necessary, but I was making sure the `tox` config still worked after I edited it.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,6 +25,7 @@ jobs:
           - "3.10"
           - "3.11"
           - "3.12"
+          - "3.13"
     steps:
       - uses: actions/checkout@v4
       - name: Set up Python ${{ matrix.python-version }}
--- a/contrib/tox.ini
+++ b/contrib/tox.ini
@@ -2,12 +2,12 @@
 package_root = .
 min_version = 4.3.3
 envlist =
-    py{38,39,310,311,312}-qa
+    py{38,39,310,311,312,313}-qa
 skip_missing_interpreters = true
 ignore_base_python_conflict = true
 labels =
-    lint = py{38,39,310,311,312}-lint
-    test = py{38,39,310,311,312}-test
+    lint = py{38,39,310,311,312,313}-lint
+    test = py{38,39,310,311,312,313}-test
 
 
 [testenv]
@@ -21,12 +21,14 @@ envname =
     py310: py310
     py311: py311
     py312: py312
+    py313: py313
 envdir =
     py38: {toxinidir}/.tox/py38
     py39: {toxinidir}/.tox/py39
     py310: {toxinidir}/.tox/py310
     py311: {toxinidir}/.tox/py311
     py312: {toxinidir}/.tox/py312
+    py313: {toxinidir}/.tox/py313
 depends =
     base
 deps =
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -40,6 +40,7 @@ classifiers = [
     "Programming Language :: Python :: 3.10",
     "Programming Language :: Python :: 3.11",
     "Programming Language :: Python :: 3.12",
+    "Programming Language :: Python :: 3.13",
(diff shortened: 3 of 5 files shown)


SnoopJ (approved):
Let's do it!

SnoopJ:
If we wanted to add 3.13 free-threading support, it looks like there's a feature request for this in GHA: https://github.com/actions/setup-python/issues/771

(I didn't yet test Sopel on a free-threaded build FWIW, but I can try it)

dgw:
> (I didn't yet test Sopel on a free-threaded build FWIW, but I can try it)

I just finished a passing run on Python 3.13.0t (setting `PYTHON_GIL=0` before running `make test`; an sqlalchemy submodule will trigger the GIL if you don't force-disable it, at the moment).

SnoopJ:
I also see passing tests in both modes, although weirdly I did not see a faster suite with nogil, but @dgw did. Either way, good enough for this PR

dgw:
I don't think it's that weird. Without installing a plugin like `pytest-parallel` or `pytest-xdist`, tests are run serially.

The better thing to think about a bit here is: Should we wait for a nogil build to be available in GHA? Is it worth running a second CI job for Python 3.13 when that feature is still experimental?

Because if we aren't going to wait, this can just ship to `master` so it's ready to keep going.

SnoopJ:
> Should we wait for a nogil build to be available in GHA? Is it worth running a second CI job for Python 3.13t when that feature is still experimental?

-1 on both of those IMO, I think our manual checks are sufficient that there's nothing _fundamentally_ wrong

dgw:
And it's easy enough to run through another PR like this later if we find a compelling reason to add nogil to the CI matrix, too. Just wanted to ask in case you had a different idea :)

Exirel (approved):

dgw:
Note: Adding the `url` changes to my `backports/8.0.1` branch. When the time comes to ship those updates, CI will likely fail on the `8.0.x` branch unless type-check fixes are included.

PR #2630, opened 2024-10-13
pytest_plugin: disable flood protection for example tests

dgw:
### Description

_This patch feels almost too simple to work, but it does._

For quite a while, we've had three tests from the `rand` plugin that take several seconds each. I had a Sopelunk with the Git version of [`pytest-profiling`](https://github.com/man-group/pytest-plugins/tree/master/pytest-profiling) (because they haven't released in about 5 years and the last package is broken on modern Pythons) and discovered that about 98% of the time spent on the five tests selected by `pytest -k rand.py` was in `time.sleep()`.

Drilling down, I saw that was being called from `bot.say()`, which led pretty quickly to the real culprit: Flood protection.

There's no server to flood during tests, so we can just turn that protection off. With this change, those five tests from `rand` went from ~12 seconds on my test system down to about 0.25 sec.

### Checklist

### DIFF
--- a/sopel/tests/pytest_plugin.py
+++ b/sopel/tests/pytest_plugin.py
@@ -18,6 +18,8 @@ TEMPLATE_TEST_CONFIG = """
 nick = {name}
 owner = {owner}
 admin = {admin}
+# avoid wasting cycles in time.sleep() during `repeat`ed tests
+flood_max_wait = 0
 """
 
 


dgw:
Super ironic that the CI run for this PR had some of the longest job times this repo's ever seen… 3/6 jobs went over 2m, where normally they are all done in about 90 seconds.

Still, I promise this made local tests on my machine a lot faster. GHA hosted runner performance might be too inconsistent to draw any conclusions from, especially based on only a single run.

Exirel (approved):

PR #2629, opened 2024-10-12
bot, rules: fix rate limiting rules without a rate limit

Exirel:
### Description

Fix #2627 by checking if the rate limit is actually set (equal or less than 0s).

I decided to make the `at_time` parameter mandatory, as it's all internal stuffs and it makes it clearer what time we are comparing it to. That's how I figured out that it was about the repeating call to the same rule for a single trigger: it would check against the same trigger object, but compare it to the latest call to `now()`. That's why the bug was visible on `url` rules, and it affected `find` rules as well.

### Checklist

- [x] I have read [CONTRIBUTING.md](https://github.com/sopel-irc/sopel/blob/master/CONTRIBUTING.md)
- [x] I can and do license this contribution under the EFLv2
- [x] No issues are reported by `make qa` (runs `make lint` and `make test`)
- [x] I have tested the functionality of the things this change touches

PR #2631, opened 2024-10-14
Pin `mypy` to 1.11.2

dgw:
### Description

Tin. CI fails at the moment, and it would be nice if it kept working.

I have started a `mypy-1.12` branch to work on the broken things. but it'll require some coordination with #2616 and #2629.

### Checklist

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -21,7 +21,8 @@ sphinx-rfcsection~=0.1.1
 # (or until py3.9 EOL... in 10/2025, I HOPE NOT)
 vcrpy @ git+https://github.com/sopel-irc/vcrpy@uncap-urllib3
 # type check
-mypy>=1.3,<2
+# often breaks CI on master, so pin and update deliberately, on our own terms
+mypy==1.11.2
 # for `pkg_resources`; first version in which it's typed
 # we don't use `pkg_resources` directly, but mypy still cares
 setuptools>=71.1


dev-requirements.txt
```diff
@@ -21,7 +21,8 @@ sphinx-rfcsection~=0.1.1
 # (or until py3.9 EOL... in 10/2025, I HOPE NOT)
 vcrpy @ git+https://github.com/sopel-irc/vcrpy@uncap-urllib3
 # type check
-mypy>=1.3,<2
+# often breaks CI on master, so pin and update deliberately, on our own terms
+mypy==1.11.2
```
Exirel, line 25: Is it acceptable to accept [compatible version](https://packaging.python.org/en/latest/specifications/version-specifiers/#compatible-release) for patch versions only (i.e. 1.11.3, .4, etc. but not 1.12)?

```suggestion
mypy~=1.11.2
```

SnoopJ, line 25: Nitpick: IMO it would be okay to allow patch versions in, but I don't feel super strongly about it.

```suggestion
mypy>=1.11, <1.12
```

SnoopJ (approved):

Exirel (approved):
I approve because I understand the issue. I offered an optional suggestion, and I wouldn't mind if you didn't accept it.

dev-requirements.txt
dgw: I don't expect the `mypy` project to release any more 1.11.x versions now that it's moved on to 1.12. Their [PyPI release history](https://pypi.org/project/mypy/#history) is quite linear.

And even "compatible versions" can change something, though I do have a _bit_ more confidence in `mypy`'s versioning than, say, `sphinxcontrib-autoprogram`… I'd feel more comfortable with an exact-version pin, because I'm _very_ tired of merging a green PR only for `master` to go red because something updated.

_With apologies to @SnoopJ: I resolved his thread without commenting, since I would say basically the same thing twice and we don't need that. :)_

PR #2633, opened 2024-10-18
plugins.handlers: fix some PyFilePlugin logic errors

### LINKED ISSUE #2619: `ModuleNotFoundError` on relative import in folder plugin
### Description

Python 3.12, at least, seems to have trouble with submodules of folder plugins, e.g.:

- `.sopel/plugins/`
  - `fun_test_plugin/`
    - `__init__.py`
    - `sample.py`

If `__init__.py` does `from .sample import *`, loading will fail with `ModuleNotFoundError: No module named 'fun_test_plugin.sample'`

### Reproduction steps

Create a plugin folder in `.sopel/plugins/fun_test_plugin/` using the file content here:

```py
# __init__.py
import sopel.plugin

from .sample import *

@sopel.plugin.rule(r'^samplemsg')
def samplemsg(bot, trigger):
    bot.say('sample!')
```

```py
# sample.py
import sopel.module

@sopel.module.rule(r'submsg')
def submsg(bot, trigger):
    bot.say('submsg!')
```

For simplicity, you can run `sopel-plugins enable -a fun_test_plugin` on a test install of Sopel, so it will _only_ load the test plugin.

### Expected behavior

Plugins should be able to do this. It seemed to work fine before 8.0, and I suspect it has to do with the switch from `imp` to `importlib` and how the manual import is handled now vs. before.

### Relevant logs

```shell
[2024-09-11 23:05:54,367] sopel.bot            INFO     - Loading plugins...
[2024-09-11 23:05:54,373] sopel.bot            ERROR    - Error loading fun_test_plugin: No module named 'fun_test_plugin.sample'
  File "/workspace/sopel/sopel/bot.py", line 279, in setup_plugins
    plugin_handler.load()
  File "/workspace/sopel/sopel/plugins/handlers.py", line 523, in load
    self._module = self._load()
                   ^^^^^^^^^^^^
  File "/workspace/sopel/sopel/plugins/handlers.py", line 493, in _load
    self.module_spec.loader.exec_module(module)
  File "<frozen importlib._bootstrap_external>", line 995, in exec_module
  File "<frozen importlib._bootstrap>", line 488, in _call_with_frames_removed
  File "/workspace/.sopel/plugins/fun_test_plugin/__init__.py", line 3, in <module>
    from .sample import *
ModuleNotFoundError: No module named 'fun_test_plugin.sample'
```

### Notes

Brought to us on IRC by @clone2727, at which point I verified it and did a little testing. Thanks for the report!

Seems worth investigating a little more, even if the end verdict is "we can't realistically fix this" (importlib + PYTHONPATH is a scary combo). Even if we don't try to fix The Whole Bug, I did notice that there's an assignment into `sys.modules` (L490 below) that probably should be moved to _after_ a `try..except`, so it doesn't insert the new modules before checking if it has a valid loader.

https://github.com/sopel-irc/sopel/blob/1c6aeb93a269a3997a9d55ec102a9ed77e393402/sopel/plugins/handlers.py#L489-L493

### Sopel version

8.0.0

### Installation method

`pip install`

### Python version

3.12.3

### Operating system

Ubuntu 22.04.3 LTS

### IRCd

_No response_

### Relevant plugins

See sample code for `fun_test_plugin`

dgw:
### Description

This fixes submodule search locations for `PyFilePlugin`s, checks whether the module spec is valid _before_ adding it to `sys.modules` (instead of checking after), and adds a regression test for the submodule search location fix.

Fixing `submodule_search_locations` resolves #2619.

### Checklist

### DIFF
--- a/test/plugins/test_plugins_handlers.py
+++ b/test/plugins/test_plugins_handlers.py
@@ -100,3 +100,45 @@ def test_get_label_entrypoint(plugin_tmpfile):
     assert meta['label'] == 'plugin label'
     assert meta['type'] == handlers.EntryPointPlugin.PLUGIN_TYPE
     assert meta['source'] == 'test_plugin = file_mod'
+
+
+MOCK_PARENT_MODULE = """
+from sopel import plugin
+
+from .sub import foo
+
+
+@plugin.command('mock')
+def mock(bot, trigger):
+    bot.say(foo)
+"""
+
+MOCK_SUB_MODULE = """
+foo = 'bar baz'
+"""
+
+
+@pytest.fixture
+def plugin_folder(tmp_path):
+    root = tmp_path / 'test_folder_plugin'
+    root.mkdir()
+
+    parent = root / '__init__.py'
+    with open(parent, 'w') as f:
+        f.write(MOCK_PARENT_MODULE)
+
+    submodule = root / 'sub.py'
+    with open(submodule, 'w') as f:
+        f.write(MOCK_SUB_MODULE)
+
+    return str(root)
+
+
+def test_folder_plugin_imports(plugin_folder):
+    """Ensure submodule imports work as expected in folder plugins.
+
+    Regression test for https://github.com/sopel-irc/sopel/issues/2619
+    """
+    handler = handlers.PyFilePlugin(plugin_folder)
+    handler.load()
+    assert handler.module.foo == 'bar baz'
--- a/sopel/plugins/handlers.py
+++ b/sopel/plugins/handlers.py
@@ -471,7 +471,7 @@ class PyFilePlugin(PyModulePlugin):
             spec = importlib.util.spec_from_file_location(
                 name,
                 os.path.join(filename, '__init__.py'),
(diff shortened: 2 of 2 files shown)


test/plugins/test_plugins_handlers.py
```diff
@@ -100,3 +100,44 @@ def test_get_label_entrypoint(plugin_tmpfile):
...
+    return str(root)
+
+
+def test_folder_plugin_imports(plugin_folder):
+    """Ensure submodule imports work as expected in folder plugins.
+
+    Regression test for https://github.com/sopel-irc/sopel/issues/2619
+    """
+    handler = handlers.PyFilePlugin(plugin_folder)
+    handler.load()
```
Exirel, line 143:
```suggestion
    handler.load()
    assert handler.module.foo == 'bar baz'
```

I'd feel better with an assert here!

Exirel (approved):
I'm going to approve because technically that's what we want. I'd feel better with the added assert but it's not absolutely necessary.

PR #2634, opened 2024-10-18
plugins: add more type annotations

dgw:
### Description

This does what it says on the tin, and is intended to land in 8.1.0.

Starting as a draft, because it has existing merge conflicts already (I think they hail from #2616, but haven't checked `git blame` for all of them) and also includes part of #2633 (which needs to land in 8.0.1 instead).

### Checklist

### Notes

I probably didn't open a PR from this branch last month because I wasn't sure it was done, but now that a good 4 weeks went by and seeing this in the list of branches made me think, "Wait, what's that for?" I think I'd better get in the habit of opening draft PRs sooner rather than later. #2633 was another forgotten branch that I thought had been turned into a PR but hadn't yet.

### DIFF
--- a/sopel/plugins/rules.py
+++ b/sopel/plugins/rules.py
@@ -38,8 +38,13 @@ from sopel.config.core_section import (
     COMMAND_DEFAULT_HELP_PREFIX, COMMAND_DEFAULT_PREFIX, URL_DEFAULT_SCHEMES)
 
 if TYPE_CHECKING:
-    from collections.abc import Generator, Iterable
+    from collections.abc import Generator, ItemsView, Iterable, Mapping
+    from re import Match, Pattern
+    from sopel.bot import Sopel, SopelWrapper
+    from sopel.config import Config
     from sopel.tools.identifiers import Identifier
+    from sopel.trigger import PreTrigger, Trigger
+    from typing import Callable
 
 
 __all__ = [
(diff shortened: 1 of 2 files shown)


sopel/plugins/rules.py
```diff
@@ -1017,30 +1048,30 @@ def from_callable_lazy(cls, settings, handler):
...
+                 channel_rate_message: Optional[str] = None,
+                 global_rate_message: Optional[str] = None,
+                 default_rate_message: Optional[str] = None,
+                 usages: Optional[tuple] = None,
+                 tests: Optional[tuple] = None,
+                 doc: Optional[str] = None):
         # core
         self._regexes = regexes
-        self._plugin_name = plugin
+        self._plugin_name = plugin or ''
```
Exirel, line 1074: Given ` plugin` is always a str now, there is no need for that (unless mypy complain? but that sounds weird):

```suggestion
        self._plugin_name = plugin
```

Note: I'm not even sure why I made `plugin`  optional in the first place.

Exirel (approved):
I really really like that. Only a nitpick.

dgw:
Hmm, I'm torn whether to resolve the conflicts on this. It's doing some unique stuff, but perhaps it's better to wait until after #2669 is done mucking about in `plugins/` and then go through `.handlers` & `.rules` for typing stuff again with fresh eyes.

Exirel:
@dgw so many conflicts now... :sob:

PR #2635, opened 2024-10-19
sopel: don't use bot.search_url_callbacks internally

Exirel:
### Description

Both `coretasks` and `url` used `bot.search_url_callbacks`, that was deprecated some time ago and wants to emit warning in Sopel 8.1. Because it was used, we didn't bump Sopel's version to 8.1.0.dev0 yet, and that was annoying.

Therefore, I replaced that by a direct access to the private attribute `_url_callbacks`, and then bumped to Sopel 8.1.0.dev0. And by the way, we should be able to remove `bot.call` in Sopel 9.

Thank you for your service, `register_url_callback`, but I can't wait to remove you and your sister methods!

### Checklist

### DIFF
--- a/sopel/bot.py
+++ b/sopel/bot.py
@@ -716,9 +716,14 @@ class Sopel(irc.AbstractBot):
         :param func: the function to call
         :type func: :term:`function`
         :param sopel: a SopelWrapper instance
-        :type sopel: :class:`SopelWrapper`
-        :param Trigger trigger: the Trigger object for the line from the server
-                                that triggered this call
+        :param trigger: the Trigger object for the line from the server that
+                        triggered this call
+
+        .. deprecated:: 8.1
+
+            This method is deprecated and will be removed in Sopel 9.0. The
+            new rules system uses :meth:`call_rule` instead.
+
         """
         nick = trigger.nick
         current_time = time.time()
@@ -1120,6 +1125,7 @@ class Sopel(irc.AbstractBot):
         # Avoid calling shutdown methods if we already have.
         self.shutdown_methods = []
 
+    # TODO: Remove in Sopel 9.0
     # URL callbacks management
 
     @deprecated(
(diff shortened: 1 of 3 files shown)


sopel/bot.py
```diff
@@ -714,11 +714,15 @@ def call(
         """Call a function, applying any rate limits or other restrictions.
 
         :param func: the function to call
-        :type func: :term:`function`
```
dgw, line 717: Removing this is an interesting choice when the type annotation still says `Any`

dgw (commented):
At a glance from my phone, this looks sane. TIL GitHub only shows the "Viewed" checkbox on files in landscape orientation. :/

> Thank you for your service, `register_url_callback`, but I can't wait to remove you and your sister methods!

I was pretty excited about that too… #2436 is biding its time since early last year (but obviously will need a rebase to move out of draft status; I knew the risks going in!)

sopel/bot.py
Exirel: Yeah bad idea.

dgw (approved):
LGTM. I wish I knew why it took me this long to re-review. 😅

dgw:
@Exirel you can squash out the fixup, and drop 4c624ac313be2bd9ffbec5eee183e3a8428992bb (`master` version was updated in 65fabe8ab7ec875a60d1bd2e8b168b182b4505d7 / #2645).

Exirel:
I squashed against itself, so let's see if the test suite pass properly before merging. I'm still not sure about that whole "don't rebase on master" thing, it feels wrong to me. :upside_down_face:

dgw:
> I'm still not sure about that whole \"don't rebase on master\" thing, it feels wrong to me

Ah, GitHub runs pull request jobs on the merge result of `BASE <- PR_HEAD`, so it's almost like rebasing. You still get all of the upstream changes! 😁

This one's checks passed and it's now ready to merge when _I_ get back to my PC in a few days (to do the backport/changelog stuff properly, very tricky to do on my tablet).

Issue #2425, opened 2023-03-18
Direct call to bot.register_* methods by plugin handler


Exirel:
### Context

When the new plugin system was put in place, it was important to keep the bot's interface as backward compatible as possible, and we did a pretty good job. Now it is time to think about the future, and remove the leftovers. For instance, the registration (and removal) of a plugin is a complex sequence:

* the bot in its setup phase, calls `setup_plugins` to register and loads its plugins
* this `setup_plugins` method calls the handler's register method (and provide the bot itself as argument)
* this `register` method calls `loader.clean_module`, which does a bunch of stuff we should not need, but for now we do
* then it calls `bot.add_plugin`, circling back to the bot...

This issue is a follow-up on #2341 that adds TODO to remove `add_plugin` and `remove_plugin`.

### Problems & Solutions

The problem is that `bot.add_plugin` must have the same input as the output of `loader.clean_module`, and adding any new type of plugin callables is more difficult because of that. Since the bot already has a bunch of specific ` register_*`  methods:

```python
self.register_callables(callables)
self.register_jobs(jobs)
self.register_shutdowns(shutdowns)
self.register_urls(urls)
```

we could reuse these directly. All that is left is to register the plugin handler itself (`self._plugins[plugin.name] = plugin`) which can be done directly by the `setup_plugins` method.

And then, we can deprecate ` add_plugin`, do the same thing for `remove_plugin`, and probably clean up some other things along the way (I'm looking at you, `sopel.loader`).

### Planning

- [x] Replace `add_plugin` and `remove_plugin` by direct call to specific registration methods (see #2636)
- [x] deprecate both in Sopel 8.1, to be removed in Sopel 9 (see #2636)

Removing the deprecated method will be done in #2640

PR #2636, opened 2024-10-19
bot, plugins: deprecate bot.{add|remove}_plugin

Exirel:
### Description

Related to #2425 (first two tasks).

Both methods to add/remove plugins rely on values direct from `loader.clean_module`, which is inconvenient at time. In order to rework smoothly this interface, these methods are now deprecated and they will be removed in Sopel 9.0.

We still need methods to record the plugin handler to the plugin name in `bot.plugins`, but they don't do more than that. Everything else is now directly handled and controlled by the plugin handler.

For good measure, I've fully type check `sopel.plugins.handlers`. This will create merge conflict with #2634 and I'm sorry about that, yet it was too good an opportunity to miss it.

### Checklist

### DIFF
--- a/sopel/bot.py
+++ b/sopel/bot.py
@@ -41,7 +41,9 @@ from sopel.trigger import Trigger
 if TYPE_CHECKING:
     from collections.abc import Iterable, Mapping
 
-    from sopel.plugins.handlers import AbstractPluginHandler
+    from sopel.plugins.handlers import (
+        AbstractPluginHandler, PluginMetaDescription,
+    )
     from sopel.trigger import PreTrigger
 
 
@@ -344,6 +346,37 @@ class Sopel(irc.AbstractBot):
 
     # plugins management
 
+    def set_plugin_handler(
+        self,
+        handler: AbstractPluginHandler,
+    ) -> None:
+        """Record a plugin ``handler``.
+
+        :param handler: the plugin handler to record
+
+        Recording a plugin handler associates only its name to :attr:`plugins`.
+        To register a plugin handler's callables, jobs, etc., you should use
+        its :meth:`~sopel.plugins.handlers.AbstractPluginHandler.register`
+        method.
+
+        .. versionadded:: 8.1
+        """
+        self._plugins[handler.name] = handler
+
+    def clear_plugin_handler(self, name: str) -> None:
+        """Remove the plugin handler for ``name``.
+
+        :param name: plugin name to forget
+
+        Removing a plugin handler removes only its name from :attr:``plugins``.
+        To unregister a plugin handler's callables, jobs, etc. you should use
+        its :meth:`~sopel.plugins.handlers.AbstractPluginHandler.unregister`
+        method.
+
+        .. versionadded:: 8.1
+        """
+        del self._plugins[name]
+
     def reload_plugin(self, name: str) -> None:
         """Reload a plugin.
 
@@ -359,10 +392,12 @@ class Sopel(irc.AbstractBot):
             raise plugins.exceptions.PluginNotRegistered(name)
 
         plugin_handler = self._plugins[name]
+
         # tear down
         plugin_handler.shutdown(self)
         plugin_handler.unregister(self)
         LOGGER.info("Unloaded plugin %s", name)
+
         # reload & setup
         plugin_handler.reload()
         plugin_handler.setup(self)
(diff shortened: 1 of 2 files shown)


sopel/bot.py
```diff
@@ -342,6 +344,37 @@ def post_setup(self) -> None:
 
     # plugins management
 
+    def store_plugin_handler(
```
dgw, line 347: Maybe this is a nitpick, but I would call this:

```suggestion
    def set_plugin_handler(
```

That is all it does. And if the handler name already exists in `self._plugins`, the existing handler will be overwritten, which seems _possibly_ bad (even though other parts of the plugin system are supposed to make sure only one plugin with a given name gets loaded).

sopel/bot.py
```diff
@@ -342,6 +344,37 @@ def post_setup(self) -> None:
...
+        Recording a plugin handler associates only its name to :attr:`plugins`.
+        To register a plugin handler's callables, jobs, etc., you should use
+        its :meth:`~sopel.plugins.handlers.AbstractPluginHandler.register`
+        method.
+
+        .. versionadded:: 8.1
+        """
+        self._plugins[handler.name] = handler
+
+    def remove_plugin_handler(self, name: str) -> None:
```
dgw, line 364: Depending on how you feel about `store_plugin_handler()` -> `set_plugin_handler()`, you might take this too:

```suggestion
    def clear_plugin_handler(self, name: str) -> None:
```

sopel/bot.py
```diff
@@ -342,6 +344,37 @@ def post_setup(self) -> None:
...
+        """Remove a plugin handler from ``name``.
```
dgw, line 365: Regardless of the method name, this wording implies that a plugin ``name`` can have multiple handlers, which it can't.

```suggestion
        """Remove the plugin handler for ``name``.
```

sopel/plugins/handlers.py
```diff
@@ -387,24 +384,36 @@ def register(self, bot: Sopel) -> None:
...
-        relevant_parts = loader.clean_module(self.module, bot.config)
-        for part in itertools.chain(*relevant_parts):
+        callables, jobs, _, urls = loader.clean_module(
+            self.module,
+            bot.config,
+        )
+
+        for part in itertools.chain(callables, jobs, urls):
             # annotate all callables in relevant_parts with `plugin_name`
             # attribute to make per-channel config work; see #1839
```
dgw, line 394:
```suggestion
            # annotate all plugin callables with a `plugin_name` attribute
            # to make per-channel config work; see #1839
```

sopel/plugins/handlers.py
```diff
@@ -387,24 +384,36 @@ def register(self, bot: Sopel) -> None:
             bot.cap_requests.register(self.name, cap_request)
 
         # plugin callables go through ``bot.add_plugin``
```
dgw, line 386:
```suggestion
```

Not any more!

dgw (commented):
Mostly documentation/comment tweaks, with a couple naming questions…

There were a couple of _really_ nitpicky things I decided not to bother with, since you've been waiting on review long enough.

sopel/plugins/handlers.py
```diff
@@ -50,16 +50,20 @@
...
 
 from sopel import __version__ as release, loader, plugin as plugin_decorators
 from . import exceptions
 
 
 if TYPE_CHECKING:
-    from sopel.bot import Sopel
     from types import ModuleType
 
+    from importlib_metadata import EntryPoint
```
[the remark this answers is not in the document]
Exirel: Oh, good point. Will see what I can do, yes.

Exirel:
Yay, an integration bug! So, on `master`, there is an unsafe section of code in `sopel/plugins/handlers.py`:

```python
        if (
            version is None
            and hasattr(self.entry_point, "dist")
            and hasattr(self.entry_point.dist, "name")
        ):
            dist_name = self.entry_point.dist.name
```

This is unsafe but undetected by mypy because `self.entry_point` isn't properly annotated. Because in my PR I added a type annotation on that attribute, mypy is now able to recognize the issue: `entry_point.dist` can be `None`.

Because that code is not present yet in this PR, I can either:

a. create a different PR to fix the type safety issue on master
b. rebase this PR then fix it

Option b is faster and more reliable.

sopel/plugins/handlers.py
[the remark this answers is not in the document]
dgw: Based on #2670 I suspect we cannot switch this over yet, so I'll resolve the thread.

dgw:
I kept getting interrupted, so the fix on `master` (#2670) took longer to review and ship than planned… But it should all work out next time @Exirel has a chance to work on this and squash the history.

Exirel:
@dgw I had to resolve a conflict to enable CI pipelines, so I'll probably rebase on master once you validate the PR in order to remove the merge commit.

dgw (approved):
Go ahead and rebase.

Thing is, since you force-pushed, GitHub already took away my "changes since your last review" button, so I ended up re-reviewing the whole patch anyway. 🤷‍♂️

PR #2637, opened 2024-10-21
Support `invite-notify` CAP

dgw:
### Description

Sopel now requests the `invite-notify` CAP. I couldn't think of anything useful for it to do with the notifications, except to log them like other channel events are, at the INFO level. A test case added for `coretasks` checks the expected log message.

The `admin` plugin's `invite_join` handler now ignores `INVITE`s not targeted at the bot. A new `builtins/test_builtins_admin.py` test suite validates its invite-handling behavior. Of the four test cases, `test_invite_accept_not_us` was the only one not already passing prior to patching the plugin.

Progresses #971.

### Checklist

### Notes

Discovering some other old branches that I'd never created PRs for got me thinking, and I looked in the Branches list for other things that I might've forgotten about. This was one of them. Then I realized that `admin` would be broken-ish without changes, so I added that to the old branch before PRing it.

Initially, I'm targeting this feature for Sopel 8.1—but it has _potential_ as a "breaking change" for third-party plugin code that also listens for the `'INVITE'` event, if (like `admin`) they aren't checking the target nick.

In the very most cautious of worlds, we'd have to hold this until 9.0. I'd rather not, though, if we can consider `'INVITE'` handlers niche enough that it's not worth delaying.

### DIFF
--- a/sopel/builtins/admin.py
+++ b/sopel/builtins/admin.py
@@ -291,7 +291,13 @@ def me(bot, trigger):
 @plugin.priority('low')
 def invite_join(bot, trigger):
     """Join a channel Sopel is invited to, if the inviter is an admin."""
+    nick = trigger.args[0]
     channel = trigger.args[1]
+
+    if nick != bot.nick:
+        # this handler only cares if the invitee was the bot
+        return
+
     if trigger.admin:
         LOGGER.info(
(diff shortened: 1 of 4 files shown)


Exirel (approved):
I like that. And yes, it makes total sense to me to fix the `admin` plugin here for Sopel 8.1.

Issue #2602, opened 2024-03-19
More missing `versionadded` annotations


dgw:
While looking up some things in the 8.0 docs for sopel-irc/sopel-help#10, I found that `AbstractBot.safe_text_length()` is missing a `versionadded` annotation. There are probably other pieces of the API that are missing these important version notes, too. We already did at least one reasonably big documentation pass to clean up stuff like this, but obviously there were omissions (or just new things that were added without annotations _after_ the cleanup had been done).

I expect this will take someone a bit of time, so I don't want to shove it into 8.0.0 and take our collective attention away from finalizing the changelog, upgrade guide, and one last tweak to `adminchannel` (#2601). Still, I don't want us to forget about it.

Documentation improvements like this are perfect candidates for patch releases, so 8.0.1 it is.

PR #2642, opened 2024-11-11
irc: `versionadded` notes + `Optional` annotation migration

dgw:
### Description

This takes care of #2602 as far as what directly prompted me to slot it into the 8.0.1 milestone, with new `versionadded` annotations in the docstrings of stuff that's new-ish (added in 7.x or later). I figure after this PR, that issue can become a general checklist of which submodules have been looked at again since 8.0.0 and live un-milestoned for a while.

Maybe we should also have a tracking issue for the second half of this PR, which starts the process of something we talked about in other patches: Replacing use of `typing.Optional` with `| None`. Right now I think "just migrate the files you touch" is better than a "touch ALL the things!" PR, but we still should keep track of what's been done.

### Checklist

  - Ran `make lint` only locally, as no runtime code is touched

### DIFF
--- a/sopel/irc/__init__.py
+++ b/sopel/irc/__init__.py
@@ -78,9 +78,12 @@ class AbstractBot(abc.ABC):
         self.backend: AbstractIRCBackend = UninitializedBackend(self)
         """IRC Connection Backend."""
         self._connection_registered = threading.Event()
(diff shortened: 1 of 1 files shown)


Exirel (approved):
That's a lot of work to track everything down to its version of origin. Well done!

PR #2647, opened 2025-01-02
build: pin urllib3<2.3 to accommodate our old fork of vcrpy

dgw:
### Description

I had the itch to work on a feature for 8.1, but noticed that unrelated tests were failing when I started validating my changes…

`urllib3` 2.3.0, released 22 December 2024, broke compatibility with `vcrpy` 5.x by using a `response.version_string` attribute that's missing from the mocked-up response objects. ([The diff](https://github.com/urllib3/urllib3/commit/f9d37add7983d441b151146db447318dff4186c9) looks pretty innocuous, but every change has the potential to [break _someone's_ workflow](https://xkcd.com/1172/), right?.)

Upstream has fixed this for vcrpy 7.x, but we're on an old fork of vcrpy 5.x due to incompatibilities with mixed `urllib` versions across older versions of Python (see #2456).

I'll add a note to #2456 with an update on the current state of this. We can remove the pin again when py3.9 support drops in October 2025. Original hope was to get a PR merged upstream in vcrpy that fixes the urllib version issues with Python < 3.10, but that hasn't gone anywhere.

### Checklist

- [x] I have read [CONTRIBUTING.md](https://github.com/sopel-irc/sopel/blob/master/CONTRIBUTING.md)
- [x] I can and do license this contribution under the EFLv2
- [x] No issues are reported by `make qa` (runs `make lint` and `make test`)
- [x] I have tested the functionality of the things this change touches

### Notes

Milestoned for 8.0.2 because it'll need backporting to the maintenance branch (otherwise CI there will be broken).

Issue #2456, opened 2023-05-21
Unpin transitive dependency `urllib3`


dgw:
Follow-up to #2455. We only want this pinned temporarily, as described in 6e64c965810c27a2b25d48d02d622124463cd9c0; when possible, the pin should be removed.

Relies on resolution of upstream issue, kevin1024/vcrpy#688

dgw:
#2519 made unpinning `urllib3` possible, and the upstream issue in `vcrpy` has been "closed". But we're still doing Kludgy Things related to both packages:

https://github.com/sopel-irc/sopel/blob/973a489355540d68b95db01a49e983ac7a740bcc/dev-requirements.txt#L20-L22

I will retitle this issue to reflect the current state, which is that `vcrpy` might or might not ever officially support using `urllib3` 2.x on Python <3.10 given problems with `vcrpy`'s test suite on those combinations. There's no guarantee that I'll ever manage to get my branch into a state acceptable to upstream (but the test failures happen only in the suite's "offline" mode that mocks out actual communication with a server, seeming to work just fine when talking to _real_ servers).

dgw:
Quick update summary:

- The upstream patch I made hasn't gone anywhere, because while it works for us it doesn't seem to be compatible with the full set of features and library combos vcrpy supports.
- We now have to pin urllib3 so it won't break (#2647) with vcrpy 5.x
- I now look forward to dropping Python 3.9 later this year… That's the last version affected by this issue.

I might formally propose skipping vcrpy tests on the older Python version anyway, if I get the bee in my bonnet to bring that component current sooner. The ~~house of cards~~ pile of pins can only get so tall before it really annoys me.

PR #2654, opened 2025-02-02
Release prep: 8.0.2

dgw:
Trying out a new workflow: I've merged the changelog and backport branches for 8.0.2 into this PR, which will both run CI on the final result and allow other maintainers to give the overall patch a once-over before it gets merged for final release.

If we like doing things this way, we can use it for 8.1.0—and create (an) appropriate label(s) in the tracker for this type of PR.

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -22,6 +22,7 @@ sphinx-rfcsection~=0.1.1
 vcrpy @ git+https://github.com/sopel-irc/vcrpy@uncap-urllib3
 # also have to use a version of urllib3 that doesn't use the `version_string`
 # attr of the response object, because vcrpy won't support it until 7.x
+# reverting to mainline vcrpy is tracked in #2456
 urllib3<2.3
 # type check
 mypy>=1.3,<2
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -20,6 +20,9 @@ sphinx-rfcsection~=0.1.1
 # use fork of vcrpy 5.x until kevin1024/vcrpy#777 is (hopefully) accepted
 # (or until py3.9 EOL... in 10/2025, I HOPE NOT)
 vcrpy @ git+https://github.com/sopel-irc/vcrpy@uncap-urllib3
+# also have to use a version of urllib3 that doesn't use the `version_string`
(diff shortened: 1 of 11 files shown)


dgw:
> which will both run CI on the final result

Well it's _supposed_ to do that. I will attempt to tweak this config and make the branch name `v` prefix optional:

https://github.com/sopel-irc/sopel/blob/44bc866bea3128c143acb5d2f5efe8301aa8d379/.github/workflows/ci.yml#L10-L13

Exirel (approved):

dev-requirements.txt
```diff
@@ -20,6 +20,9 @@ sphinx-rfcsection~=0.1.1
 # use fork of vcrpy 5.x until kevin1024/vcrpy#777 is (hopefully) accepted
 # (or until py3.9 EOL... in 10/2025, I HOPE NOT)
 vcrpy @ git+https://github.com/sopel-irc/vcrpy@uncap-urllib3
+# also have to use a version of urllib3 that doesn't use the `version_string`
+# attr of the response object, because vcrpy won't support it until 7.x
```
SnoopJ, line 24: We may as well refer to #2647 here too, I think?

SnoopJ (approved):

dev-requirements.txt
dgw: Hmm, it could but that was the PR that added the comment in question. I will add a reference to #2456.

dgw: f03fbae1aacd653f15e0bbc0b520d59c0d64b09e - cherry-picked to #2655 rather than creating a whole new PR for it.

dgw:
```
17:21:50 <+Sopel> [sopel] dgw commented on pull request #2655: I figure there's no need to go through a whole
new PR just to pull an added `dev-requirements.txt` comment line from #2654 upstream into `master`. I will
get @SnoopJ's sign-off on the new comment via IRC (it was his idea).
https://github.com/sopel-irc/sopel/pull/2655#issuecomment-2638257122
17:22:57 <+SnoopJ> :+1: 
17:24:08 <+dgw> goes for 2654 too I hope?
17:24:12 <+SnoopJ> ye
```

PR #2655, opened 2025-02-02
ci: make `v` prefix optional in maintenance branch names

dgw:
We aren't using the `v` prefix for maintenance branches. What we have are names like `7.1.x`, `8.0.x`, etc. and CI hasn't been running for 8.0 series maintenance patches like it should.

I already proposed this in the 8.0.2 release-prep PR (#2654), but it needs to be in the master branch too so future maintenance branches will get the new patterns automatically.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,12 +5,12 @@ on:
   push:
     branches:
       - master
-      - v[0-9]+.[0-9]+.x
+      - v?[0-9]+.[0-9]+.x
       - gha-*
   pull_request:
     branches:
       - master
-      - v[0-9]+.[0-9]+.x
+      - v?[0-9]+.[0-9]+.x
 
 jobs:
   tests:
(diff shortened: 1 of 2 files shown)


dgw:
Note: We could probably _remove_ the `v` prefix entirely. The distinction is supposed to be:

- `*.*.x`: maintenance _branch_
- `v*.*.*`: release _tag_

But I went with the smaller change for now. We're allowed to change the convention, and no release _tag_ should ever actually have a name ending in `.x`, so in theory we could use `v` prefix everywhere.

SnoopJ (approved):
I approve of adjusting the workflow to match what's been done in the most recent release, making it optional is a nice middle-ground.

Exirel (approved):

dgw:
I figure there's no need to go through a whole new PR just to pull an added `dev-requirements.txt` comment line from #2654 upstream into `master`. I will get @SnoopJ's sign-off on the new comment via IRC (it was his idea).

PR #2659, opened 2025-02-22
Introduce dedicated plugin abort exception

### LINKED ISSUE #2656: [proposal] introduce dedicated exception for bailing out of plugin functions with a user-customized error message
### Requested Feature

I propose the addition of a new exception `sopel.plugin.PluginError` or similar that can be used by plugin authors to abort plugin event handling while also sending a fully-custom string error message via IRC.

Prompted by discussion in `#sopel` on Libera.chat on 9 Feb 2025, with reference to PHP's `die()` mechanism and `sys.exit(), SystemExit` in the Python stdlib.

### Problems Solved

It is relatively common that plugin code wants to abort the handling of a command, event, etc. and it is also common that people¹ solve this problem by doing something like:

```python
from sopel import plugin

@plugin.command("thrust")
def thrust(bot, trigger):
    if cannot_continue_predicate(trigger):
        bot.say("Error: the turbo encabulator is not available")
        return False

    operate_novertrunnion()
```

This leaves a pitfall where people¹ may forget to terminate control flow:

```python
from sopel import plugin

@plugin.command("thrust")
def thrust(bot, trigger):
    if cannot_continue_predicate(trigger):
        bot.say("Error: the turbo encabulator is not available")
        # NOTE: NO RETURN! Control flow spills out of this block and will probably cause an exception.

    operate_novertrunnion()
```

I think it is reasonable to say that plugin authors _often_ want to simultaneously express the semantics of _"handling has stopped'_ and _"send an error message to IRC"_. I propose the introduction of a dedicated exception, so that this code can be written as in the following section.

---

¹ it's me, I'm people

### Alternatives

A function `plugin.abort()` or similar could be introduced to have the same effect. I think this may be a good idea _with_ introduction of an exception, in the same fashion that `sys.exit()` is a helper that raises `SystemExit` for you.

@dgw has some reservations about introducing such a helper in `plugin` so it is possible this has some other home. I argue that `plugin` is a reasonable place to put it because the proposed feature is very specific to plugin author code, but I can see a valid design concern in keeping the existing suite of decorators separated from other machinery. Perhaps `plugin` should be a package with a `plugin.error` and `plugin.decorators` gathered into the package root to preserve existing API?

## Implementation notes

In the bot core that handles plugin event dispatch, this feature couple be implemented by doing the moral equivalent of:

```python
try:
    call_handler(bot, trigger)
except PluginError as exc:
    # more type-safety/consistency checking is probably a good idea here, contingent on API
    bot.say(exc.args)
except Exception as exc:
    existing_behavior(exc)
```

## Example usage

```python
import random
from sopel import plugin

BARKS = ["arf!", "grrRRR", "yip yip!"]

@plugin.command("bark")
def bark(bot, trigger):
    bot.say(random_bark())


def random_bark() -> str:
    roll = random.randint(1, 4)
    if roll == 4:
        raise plugin.PluginError("Cannot locate bark")
    else:
        return random.choice(BARKS)
```

SnoopJ:
### Description

Closes #2656.

This changeset introduces a new exception: `sopel.plugins.exceptions.PluginAbort` (also exposed as `sopel.plugin.PluginAbort`) as well as the helper `sopel.plugin.abort()` for raising it.

This exception can be used by a plugin to immediately abort that plugin's handling of an event, optionally sending a message if an argument is given. I found myself for the N+1th time accidentally forgetting to `return` from a plugin after I had done `bot.say("This plugin definitely will not continue execution because I always remember the return")` and thinking that [there must be a better way](https://youtu.be/p33CVV29OG8?t=571).

Because this feature adds new major API, it will not be appropriate before Sopel 9.0 (if indeed it is wanted at all!)

### Feature demo

```python
# abort_demo.py
from sopel import plugin
from sopel.plugins.exceptions import PluginAbort


@plugin.command("abort1")
def abort1(bot, trigger):
    bot.say("this message should be visible")

    # ...
    # do some elaborate plugin things
    # ...
    
    # bail out if some of the elaborate things went wrong before this plugin
    # finishes and spits out the Answer to Life, the Universe, and Everything
    can_continue = False
    if not can_continue:
        raise PluginAbort("PluginAbort: this message should also be visible")

    bot.say("this message should NOT be visible")


@plugin.command("abort2")
def abort2(bot, trigger):
    bot.say("this message should be visible")
    plugin.abort("PluginAbort: this message should also be visible")
    bot.say("this message should NOT be visible")


@plugin.command("abort3")
def abort3(bot, trigger):
    bot.say("this message should be visible")
    plugin.abort([1, 2, 3])
    bot.say("this message should NOT be visible")
```

```python
<SnoopJ> ,abort1                                                                                                                                                        
<mediocrebot> this message should be visible                                                                                                                            
<mediocrebot> PluginAbort: this message should also be visible                                                                                                          
<SnoopJ> ,abort2                                                                                                                                                        
<mediocrebot> this message should be visible                                                                                                                            
<mediocrebot> PluginAbort: this message should also be visible                                                                                                          
<SnoopJ> ,abort3                                                                                                                                                        
<mediocrebot> this message should be visible                                                                                                                            
<mediocrebot> [1, 2, 3]
```

### Checklist

### DIFF
--- a/sopel/bot.py
+++ b/sopel/bot.py
@@ -35,6 +35,7 @@ from sopel.plugins import (
     jobs as plugin_jobs,
     rules as plugin_rules,
 )
+from sopel.plugins.exceptions import PluginAbort
 from sopel.tools import jobs as tools_jobs
 from sopel.trigger import Trigger
 
@@ -702,6 +703,15 @@ class Sopel(irc.AbstractBot):
 
         try:
             rule.execute(sopel, trigger)
+        except PluginAbort as abort:
+            LOGGER.warning("Plugin aborted while handling trigger %r", trigger, exc_info=abort)
+
+            if len(abort.args) == 1:
+                message = abort.args[0]
+                if not isinstance(message, str):
+                    message = repr(message)
+
+                self.say(message, trigger.sender)
         except KeyboardInterrupt:
             raise
         except Exception as error:
--- a/sopel/bot.py
+++ b/sopel/bot.py
@@ -707,7 +707,7 @@ class Sopel(irc.AbstractBot):
             LOGGER.warning("Plugin aborted while handling trigger %r", trigger, exc_info=abort)
 
             if abort.message:
-                self.say(message, trigger.sender)
+                self.say(abort.message, trigger.sender)
         except KeyboardInterrupt:
             raise
         except Exception as error:
--- a/sopel/bot.py
+++ b/sopel/bot.py
@@ -706,11 +706,7 @@ class Sopel(irc.AbstractBot):
         except PluginAbort as abort:
             LOGGER.warning("Plugin aborted while handling trigger %r", trigger, exc_info=abort)
 
-            if len(abort.args) == 1:
-                message = abort.args[0]
-                if not isinstance(message, str):
-                    message = repr(message)
-
+            if abort.message:
                 self.say(message, trigger.sender)
         except KeyboardInterrupt:
             raise
--- a/sopel/plugin.py
+++ b/sopel/plugin.py
@@ -43,8 +43,9 @@ if TYPE_CHECKING:
 __all__ = [
     # constants
     'NOLIMIT', 'VOICE', 'HALFOP', 'OP', 'ADMIN', 'OWNER', 'OPER',
-    # decorators
+    # functions
     'abort',
+    # decorators
     'action_command',
     'action_commands',
     'allow_bots',
--- a/sopel/plugin.py
+++ b/sopel/plugin.py
@@ -14,6 +14,7 @@ import functools
 import logging
 import re
 from typing import (
+    Any,
     Callable,
     Literal,
     Optional,
@@ -2029,6 +2030,6 @@ def output_prefix(prefix: str) -> Callable:
     return add_attribute
 
 
-def abort(*args, **kwargs) -> None:
+def abort(*args: Any, **kwargs: Any) -> None:
     """Abort execution of a plugin event"""
     raise PluginAbort(*args, **kwargs)
--- a/sopel/plugin.py
+++ b/sopel/plugin.py
@@ -16,6 +16,7 @@ import re
 from typing import (
     Callable,
     Literal,
+    NoReturn,
     Optional,
     Pattern,
     Protocol,
@@ -2030,6 +2031,6 @@ def output_prefix(prefix: str) -> Callable:
     return add_attribute
 
 
-def abort(msg: str | None = None) -> None:
+def abort(msg: str | None = None) -> NoReturn:
     """Abort execution of a plugin event"""
     raise PluginAbort(msg)
--- a/sopel/plugin.py
+++ b/sopel/plugin.py
@@ -23,6 +23,7 @@ from typing import (
     Union,
 )
 
+from sopel.plugins.exceptions import PluginAbort
 # import and expose privileges as shortcut
 from sopel.privileges import AccessLevel
 
@@ -42,6 +43,7 @@ __all__ = [
     # constants
     'NOLIMIT', 'VOICE', 'HALFOP', 'OP', 'ADMIN', 'OWNER', 'OPER',
     # decorators
+    'abort',
     'action_command',
     'action_commands',
     'allow_bots',
@@ -2025,3 +2027,8 @@ def output_prefix(prefix: str) -> Callable:
         function.output_prefix = prefix
         return function
     return add_attribute
+
+
+def abort(*args, **kwargs) -> None:
+    """Abort execution of a plugin event"""
+    raise PluginAbort(*args, **kwargs)
--- a/sopel/plugin.py
+++ b/sopel/plugin.py
@@ -14,7 +14,6 @@ import functools
 import logging
 import re
 from typing import (
-    Any,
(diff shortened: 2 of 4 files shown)


sopel/plugin.py
```diff
@@ -2029,6 +2030,6 @@ def add_attribute(function):
     return add_attribute
 
 
-def abort(*args, **kwargs) -> None:
+def abort(*args: Any, **kwargs: Any) -> None:
```
SnoopJ, line 2033: I can see room for an argument that the API of `PluginAbort` should be restricted to exactly two valid invocations: `PluginAbort()` and `PluginAbort(msg: str)` but I wired this up as truly arbitrary and I hate that I had to annotate `Any` here.

sopel/plugin.py
```diff
@@ -42,6 +44,7 @@
     # constants
     'NOLIMIT', 'VOICE', 'HALFOP', 'OP', 'ADMIN', 'OWNER', 'OPER',
     # decorators
+    'abort',
```
dgw, line 47: Nit: This isn't a decorator.

test/test_bot.py
```diff
@@ -1060,6 +1061,59 @@ def testrule(bot, trigger):
     assert items == [1], 'There must not be any new item'
 
 
+def test_call_rule_pluginabort_with_message(
+    mockbot: bot.Sopel,
+    match_hello_rule: typing.Callable,
```
dgw, line 1066: FYI you should not need to annotate test functions. I'm not sure it serves a purpose since these are fixtures.

sopel/plugin.py
dgw: IMO it's entirely reasonable to insist on providing One Common Pattern (`abort(msg: str | None)`) through this helper, and make plugin devs raise the exception themselves if they want something non-standard.

Especially because the intent of this feature, as I understand it, is as a substitute for `bot.reply("Whoops!"); return`. Supporting more than an optional abort message seems… "extra" of you 🤔

sopel/bot.py
```diff
@@ -702,6 +703,15 @@ def call_rule(
 
         try:
             rule.execute(sopel, trigger)
+        except PluginAbort as abort:
+            LOGGER.warning("Plugin aborted while handling trigger %r", trigger, exc_info=abort)
+
+            if len(abort.args) == 1:
```
dgw, line 709: I would go as far as formalizing the constructor signature & attributes of `PluginAbort()`:

```py
def __init__(self, message: str | None = None):
    self.message = message
    super().__init__(message)
```

As mentioned in other notes, the purpose of this feature is pretty specific. I don't understand why it needs to take arbitrary positional + keyword args.

dgw (commented):
Since this might wait until 9.0, we have a hot minute to talk about purpose and API design 😅

test/test_bot.py
SnoopJ: I think I used `test_call_rule()` as my template and took them from there. I don't mind doing it when it indicates the type provided by `pytest` for that fixture but I was just following style from what I copy-pasted :sweat_smile:

dgw: Yep, I see that some functions have and some don't. Doesn't matter because mypy isn't set to check this directory; just saying you can save some effort if not copy-pasting :)

sopel/bot.py
```diff
@@ -706,12 +706,8 @@ def call_rule(
...
             LOGGER.warning("Plugin aborted while handling trigger %r", trigger, exc_info=abort)
 
-            if len(abort.args) == 1:
-                message = abort.args[0]
-                if not isinstance(message, str):
-                    message = repr(message)
-
-                self.say(message, trigger.sender)
+            if abort.message:
+                self.say(abort.message, trigger.sender)
```
dgw, line 710: Now the only thing I have immediately is this idea that there should be a way to reply() instead. When I do the speak-and-return pattern this is intended to replace, I usually use reply.

Oh, and the subject of our IRC discussion just now, about how to control rate limiting behavior of the rule after this abort mechanism trips.

sopel/plugin.py
```diff
@@ -2025,3 +2028,8 @@ def add_attribute(function):
         function.output_prefix = prefix
         return function
     return add_attribute
+
+
+def abort(msg: str | None = None) -> None:
```
Exirel, line 2033: When a function never returns (because it'll always raise an exception), you should use `typing.NoReturn` (see https://docs.python.org/3/library/typing.html#typing.NoReturn).

SnoopJ: Thanks! :blue_heart:

sopel/bot.py
SnoopJ: Hmm, how would that work for events that weren't triggered by a channel message (i.e. something on an interval)?

I can see the case for making it reply-by-default where that is contextually meaningful, but it does seem like it would be a more elaborate implementation to get there.

test/test_bot.py
Exirel: I tend to type annotate in test because it helps my code editor with auto-completion.

SnoopJ:
Closing in anticipation of a feature redesign to target the decorator API, see [issue comment](https://github.com/sopel-irc/sopel/issues/2656)

PR #2661, opened 2025-02-25
cli.plugins: add `sopel-plugins new` subcommand

### LINKED ISSUE #2660: Idea: `sopel-plugins new` CLI
It's possible to [invoke `cookiecutter` programmatically](https://cookiecutter.readthedocs.io/en/stable/advanced/calling_from_python.html), meaning we could have a CLI command shipped with the bot that starts generating a new plugin. Couple preliminary details:

- `sopel-plugins new` is just my initial thought. Could be called something else, or live in a different namespace (e.g. separate this and any future developer-oriented commands into a `sopel-dev` CLI)
- While Sopel could just auto-install `cookiecutter` as a dependency, alternatively, this can be an optional feature:
  ```py
  try:
      from cookiecutter.main import cookiecutter
  except ImportError:
      print("To use this feature, install the `cookiecutter` package and try again")
  ```

_(This random idea brought to you by [streamlining the name](https://github.com/sopel-irc/plugin/commit/30fc14c6e5c7af5ae57292205e80fd8ac3a7e60e) of our cookiecutter template repo.)_

dgw:
### Description

This new `new` action for `sopel-plugins` creates a plugin using [our repo template](https://github.com/sopel-irc/plugin) and the [`cookiecutter`](https://pypi.org/project/cookiecutter/) tool.

Command options overview:

* Specifying a plugin name will pre-fill that `plugin_name` in the `cookiecutter` prompts\
  e.g. `sopel-plugins new foobar`
* The new repo is created in `.` by default, but the parent directory can be changed without `cd`ing using `-d`/`--directory`\
  e.g. `sopel-plugins new foobar -d ~/experiments`
* Supports using a customized template with the `-t`/`--template` flag
  e.g. `sopel-plugins new -t gh:ASopelHacker/fancy-template`

Using `sopel-plugins new` only requires the user to install the `cookiecutter` script entry point somewhere on `$PATH` where Sopel can find it using `subprocess.run()`. This differs from the approach I described in #2660 using `cookiecutter` programmatically as a Python module, because it's more flexible for the end user (they can use the recommended `pipx` method of installing `cookiecutter`, or maintain it however else they'd like).

Closes #2660.

### Checklist

### Notes

Yes, I implemented using `cookiecutter` programmatically first. I left the two approaches separate in the commit history in case there is a compelling reason to go back. Besides the "user convenience" thing about having `cookiecutter` installed wherever-you-want instead of forcing it to be installed in Sopel's (v)env, `cookiecutter` doesn't currently advertise itself as a typed library (neither does Sopel, to be fair), and I could not find stubs for it. Having that `# type: ignore` comment bugged me.

### DIFF
--- a/sopel/cli/plugins.py
+++ b/sopel/cli/plugins.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 import argparse
 import inspect
 import operator
+import subprocess
 
 from sopel import config, plugins
 from . import utils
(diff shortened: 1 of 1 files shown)


Exirel:
I'm on the fence for that one. I understand the convenience of exposing `sopel-plugins new`, however I tend to see anything that is just a proxy to something else as a burden for maintenance with little value.

From what I read, we could tell/direct people to the cookie cutter template instead in the documentation, and that would do the same thing, with less code.

Exirel:
After further considerations: I'd rather see more documentation around an official Sopel Plugin template for cookiecutter in the documentation than to add a new subcommand that is just a proxy to cookiecutter.

It's not a tool I'm really used to, but I think it's well worth it to document its usage within the context of Sopel plugin development for plugin authors.

PR #2669, opened 2025-03-20
bot, plugin, plugins: class based callables

Exirel:
### Description

Behold! Plugin's decorators are now creating plugin objects, instead of adding type unsafe attributes to functions.

This is theorically backward compatible with Sopel 8.0 and below's system of plugin callable (i.e. functions only).

The `sopel.loader` module is still present, and still works. It should be easier to deprecate it, to be removed in Sopel 9.

This is minor effects on the bot itself, this is more an enabler for what's to come, opening a wide arrayd of possibilities for plugin author in term of customization of plugin callables.

The `require_*` decorators are not properly handled yet, even tho they still work "as is", they might be too dependent on the order in which they are used to decorate a callable. This will be fixed in a future commit.

### Checklist

### DIFF
--- a/docs/source/package/plugins.rst
+++ b/docs/source/package/plugins.rst
@@ -18,3 +18,4 @@ Plugin Internal Machinery
     plugins/handlers
     plugins/jobs
     plugins/rules
+    plugins/callables
--- a/sopel/builtins/safety.py
+++ b/sopel/builtins/safety.py
@@ -360,7 +360,7 @@ def virustotal_lookup(
     }
     bot.memory[SAFETY_CACHE_KEY][url] = result
     if len(bot.memory[SAFETY_CACHE_KEY]) >= (2 * CACHE_LIMIT):
-        _clean_cache(bot)
+        _clean_cache(bot)  # type: ignore[arg-type]
     return result
 
 
--- a/sopel/builtins/tld.py
+++ b/sopel/builtins/tld.py
@@ -292,9 +292,9 @@ def _update_tld_data(bot, which, force=False):
 
 
 @plugin.interval(60 * 60)
-def update_caches(bot, force=False):
-    _update_tld_data(bot, 'list', force)
-    _update_tld_data(bot, 'data', force)
+def update_caches(bot):
+    _update_tld_data(bot, 'list', False)
+    _update_tld_data(bot, 'data', False)
 
 
 @plugin.command('tld')
@@ -383,7 +383,8 @@ def tld_cache_command(bot, trigger):
     subcommand = subcommand.lower()
     if subcommand == 'update':
         bot.reply("Requesting updated IANA list and Wikipedia data.")
-        update_caches(bot, force=True)
+        _update_tld_data(bot, 'list', True)
+        _update_tld_data(bot, 'data', True)
     elif subcommand == 'clear':
         bot.db.delete_plugin_value('tld', 'tld_data_cache')
         bot.memory['tld_data_cache'] = {}
--- a/sopel/loader.py
+++ b/sopel/loader.py
@@ -16,6 +16,11 @@ import logging
 import re
 
 from sopel.config.core_section import COMMAND_DEFAULT_HELP_PREFIX
+from sopel.plugins.callables import (
+    AbstractPluginObject,
+    PluginCallable,
+    PluginJob,
+)
 
 
 LOGGER = logging.getLogger(__name__)
@@ -232,24 +237,49 @@ def clean_module(module, config):
     callable, i.e. properties related to threading, docs, examples, rate
     limiting, commands, rules, and other features.
     """
-    callables = []
-    shutdowns = []
-    jobs = []
-    urls = []
+    callables: list[PluginCallable] = []
+    shutdowns: list = []
+    jobs: list[PluginJob] = []
+    urls: list[PluginCallable] = []
+    handler: PluginJob | PluginCallable
     for obj in vars(module).values():
         if callable(obj):
             is_sopel_callable = getattr(obj, '_sopel_callable', False) is True
             if getattr(obj, '__name__', None) == 'shutdown':
                 shutdowns.append(obj)
+                continue
             elif not is_sopel_callable:
                 continue
-            elif is_triggerable(obj):
-                clean_callable(obj, config)
-                callables.append(obj)
-            elif hasattr(obj, 'interval'):
-                clean_callable(obj, config)
-                jobs.append(obj)
-            elif is_url_callback(obj):
-                clean_callable(obj, config)
-                urls.append(obj)
+
+            if isinstance(obj, PluginJob):
+                obj.setup(config)
+
+                if obj.intervals:
+                    jobs.append(obj)
+
+            else:
+                if not isinstance(obj, AbstractPluginObject):
+                    # compatibility with old-style plugin callables
+                    clean_callable(obj, config)
+                    if getattr(obj, 'interval', []):
+                        handler = PluginJob.ensure_callable(obj)
+                    else:
+                        handler = PluginCallable.ensure_callable(obj)
+                elif isinstance(obj, PluginCallable):
+                    handler = obj
+                else:
+                    # it's a subclass of AbstractPluginObject that isn't a
+                    # plugin job or callable, and it cannot be handled
+                    continue
+
+                handler.setup(config)
+
+                if isinstance(handler, PluginJob):
+                    jobs.append(handler)
+                elif isinstance(handler, PluginCallable):
+                    if handler.is_triggerable:
+                        callables.append(handler)
+                    elif handler.is_url_callback:
+                        urls.append(handler)
+
     return callables, jobs, shutdowns, urls
--- a/sopel/plugin.py
+++ b/sopel/plugin.py
@@ -11,12 +11,14 @@ from __future__ import annotations
 
 import enum
 import functools
+import inspect
 import logging
 import re
 from typing import (
     Callable,
     Literal,
     Optional,
+    overload,
     Pattern,
     Protocol,
     TYPE_CHECKING,
@@ -24,6 +26,12 @@ from typing import (
 )
 
 # import and expose privileges as shortcut
+from sopel.plugins.callables import (
+    AbstractPluginObject,
+    PluginCallable,
+    PluginGeneric,
+    PluginJob,
+)
 from sopel.privileges import AccessLevel
 
 VOICE = AccessLevel.VOICE
@@ -37,6 +45,7 @@ OPER = AccessLevel.OPER
 if TYPE_CHECKING:
     from collections.abc import Iterable
     from sopel.bot import SopelWrapper
+
 
 __all__ = [
     # constants
@@ -368,17 +377,19 @@ def unblockable(
         Sopel's :meth:`~sopel.bot.Sopel.dispatch` method.
 
     """
-    def add_attribute(function):
-        function.unblockable = True
-        return function
+    def decorator(function: Callable) -> PluginCallable:
+        handler = PluginCallable.ensure_callable(function)
+        handler.unblockable = True
+        return handler
 
     # hack to allow both @unblockable and @unblockable() to work
     if callable(function):
-        return add_attribute(function)
-    return add_attribute
+        return decorator(function)
+
+    return decorator
 
 
-def interval(*intervals: Union[int, float]) -> Callable:
+def interval(*intervals: Union[int, float]) -> Callable[[Callable], PluginJob]:
     """Decorate a function to be called by the bot every *n* seconds.
 
     :param intervals: one or more duration(s), in seconds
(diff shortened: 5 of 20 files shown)


sopel/builtins/safety.py
```diff
@@ -360,7 +360,7 @@ def virustotal_lookup(
     }
     bot.memory[SAFETY_CACHE_KEY][url] = result
     if len(bot.memory[SAFETY_CACHE_KEY]) >= (2 * CACHE_LIMIT):
-        _clean_cache(bot)
+        _clean_cache(bot)  # type: ignore[arg-type]
```
dgw, line 363: Does this happen on `master` without the rest of this patch? We should pull it out into a separate PR if so.

sopel/builtins/tld.py
```diff
@@ -292,9 +292,9 @@ def _update_tld_data(bot, which, force=False):
 
 
 @plugin.interval(60 * 60)
-def update_caches(bot, force=False):
-    _update_tld_data(bot, 'list', force)
-    _update_tld_data(bot, 'data', force)
+def update_caches(bot):
+    _update_tld_data(bot, 'list', False)
+    _update_tld_data(bot, 'data', False)
```
dgw, line 297: Could be problematic if doing this is no longer supported, certainly if targeting this for a .1 release. In builtins and third-party plugins, having optional parameters to modify the default behavior of a `@plugin.interval` job isn't _terribly_ uncommon.

sopel/loader.py
```diff
@@ -232,24 +237,49 @@ def clean_module(module, config):
...
+                    if getattr(obj, 'interval', []):
+                        handler = PluginJob.ensure_callable(obj)
+                    else:
+                        handler = PluginCallable.ensure_callable(obj)
+                elif isinstance(obj, PluginCallable):
+                    handler = obj
+                else:
+                    # it's a subclass of AbstractPluginObject that isn't a
+                    # plugin job or callable, and it cannot be handled
+                    continue
```
dgw, line 273: Log something for this case?

sopel/loader.py
```diff
@@ -232,24 +237,49 @@ def clean_module(module, config):
...
-                jobs.append(obj)
-            elif is_url_callback(obj):
-                clean_callable(obj, config)
-                urls.append(obj)
+
+            if isinstance(obj, PluginJob):
+                obj.setup(config)
+
+                if obj.intervals:
+                    jobs.append(obj)
```
dgw, line 258: I'm a little confused why things are done in this order, such that there are two `isinstance(var, PluginJob)` checks with slightly different behavior.

Would it be reasonable to massage objects that aren't yet `AbstractPluginObject` subclasses into the appropriate types first, _then_ run through the subclass checks to set them up?

Maybe you're doing it this way to better isolate the old "badly named" attributes from the new code. It's just a weird speed-bump reading through and seeing the same `isinstance()` check happen again below.

Exirel:
The cyclic import alert raised by CodeQL are easy to fix: I'll copy the necessary code from `sopel.loader.clean_module` where I need it and the problem will disappear! I'm just glad the test suite passed without issues.

sopel/loader.py
```diff
@@ -232,24 +237,49 @@ def clean_module(module, config):
...
+
+                handler.setup(config)
+
+                if isinstance(handler, PluginJob):
+                    jobs.append(handler)
+                elif isinstance(handler, PluginCallable):
+                    if handler.is_triggerable:
+                        callables.append(handler)
+                    elif handler.is_url_callback:
+                        urls.append(handler)
```
dgw, line 283: I know it's already this way, but URL callbacks are totally "triggerable".

Since you're already shuffling things around, I hope you'll consider renaming `AbstractPluginObject.is_triggerable` in the interface definition. Unfortunately I don't yet have any great ideas for a replacement name that cover both generic Rule triggers (and find/search) _and_ the Command types. But that's how I understand the `triggerable` criteria:

Negatives, not any of:

* a startup()
* a job
* a URL callback
* a shutdown()

Positives, any one (or more) of:

* a command
* a nickname command
* an action command
* a generic rule (or a lazy one)
* a find rule (or a lazy one)
* a search rule (or a lazy one)
* a CTCP match
* any `@plugin.event` handler

It is quite literally just about anything that isn't a job, a URL callback, or a startup/shutdown handler. "Triggerable" isn't a great description of what those have in common when URL callbacks are also "triggerable".

Even some of my other ideas at present—`is_event_handler`, `is_input_handler`, `is_message_handler` (where you can also replace "handler" with "callback")—don't really sound any better. The old saying below seems truer every day. ☹️

> There are 2 hard problems in computer science: cache invalidation, **naming things**, and off-by-1 errors.

sopel/plugins/callables.py
```diff
@@ -0,0 +1,586 @@
...
+
+.. versionadded:: 8.1
+
+The core of a plugin consists of several objects, either plugin callables or
+plugin jobs. These objects are represented by instances of
+:class:`PluginCallable` and :class:`PluginJob` respectively.
+
+.. seealso::
+
+    The :mod:`sopel.plugin`'s decorator create these objects directly.
```
dgw, line 11: Englishing just a little :)

```suggestion
    The decorators in :mod:`sopel.plugin` create these objects directly.
```

sopel/plugins/callables.py
```diff
@@ -0,0 +1,586 @@
...
+
+.. important::
+
+    This is all relatively new. Its usage and documentation is for Sopel core
+    development and advanced developers. It is subject to rapid changes
```
dgw, line 16: We might consider dropping "and advanced developers" from this disclaimer in any new core stuff that's not intended as public API, since ideally plugin devs won't touch it no matter how "advanced" they think they are.

This is a public service announcement from the Sopel Crimes™ Stoppers Unit.

sopel/plugins/callables.py
```diff
@@ -0,0 +1,586 @@
...
+
+    @property
+    @deprecated(
+        reason='`thread` is replaced by the `threaded` attribute.',
+        version='8.1',
+        removed_in='9.0',
+    )
+    def thread(self):
+        # compatiblity property
+        # TODO: remove in Sopel 9
```
dgw, line 94: I could fix the typo, or we could just rely on the `@deprecated` decorator to imply all of this 😁

```suggestion
```

sopel/plugins/callables.py
```diff
@@ -0,0 +1,586 @@
...
+            self.nickname_commands,
+            self.action_commands,
+            self.url_regex,
+            self.url_lazy_loaders,
+        ])
+
+        return allowed
+
+    @property
+    def is_triggerable(self) -> bool:
```
dgw, line 326: I went into too much detail about this in an earlier file… Won't repeat myself.

But the gist is that I think this property needs a better name. URL callbacks seem "triggerable" too, in a way not meaningfully different from how `find_rules` work.

sopel/plugins/callables.py
```diff
@@ -0,0 +1,586 @@
...
+        """Optional setup.
+
+        :param settings: the bot's configuration
+
+        This method will be called by the bot before registering the job.
+
+        .. note::
+
+            This methods is a no-op. It exists for subclasses that may requires
+            to perform operation before registration.
```
dgw, line 584: Nitpick.

```suggestion
            This method is a no-op. It exists for subclasses that may need to
            perform operations before registration.
```

dgw (commented):
I skimmed through and only made notes on anything that grabbed my attention—without going too deep, as you suggested. :)

Hopefully the nitpicks are helpful, even if we wind up discarding any design-ish suggestions I made!

docs/source/package/plugins.rst
```diff
@@ -18,3 +18,4 @@ Plugin Internal Machinery
     plugins/handlers
     plugins/jobs
     plugins/rules
+    plugins/callables
```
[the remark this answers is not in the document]
Exirel: No strong opinion yet. I feel like this part of the documentation needs some care and love. I'll try to think more than the 0.001s I gave it so far. 🤣

sopel/builtins/safety.py
Exirel: No, because a. the function _clean_cache is not altered, and b. there is no type check until this.

That being said, as you can see it works at runtime, and it will be better once we remove the SopelWrapper class, because the PluginJob's call signature requires a Sopel instance instead.

sopel/builtins/tld.py
Exirel: Yeah I get you. I really don't like it, but I think I have a workaround, i.e. accept `*arg` and `**kwargs` in both PluginCallable and PluginJob's call methods.

sopel/builtins/safety.py
dgw: Oof, I've had a bug in my mental model for years: I thought SopelWrapper was a subclass of Sopel. Whoops! 😂

sopel/builtins/tld.py
Exirel: Fixed!

sopel/plugins/callables.py
Exirel: Done!

sopel/loader.py
dgw: The fox has nothing on you. You've clearly thought about it, which is enough for me at this stage :)

sopel/plugins/rules.py
```diff
@@ -553,23 +555,20 @@ class AbstractRule(abc.ABC):
...
-        Sopel's function-based rule handlers are simple callables, decorated
-        with :mod:`sopel.plugin`'s decorators to add attributes, such as rate
-        limit, threaded execution, output prefix, priority, and so on. In order
-        to load these functions as rule objects, this class method can be used;
-        it takes the bot's ``settings`` and a cleaned ``handler``.
+        .. seealso::
+
+            The :class:`~sopel.plugins.callables.PluginCallable` from the
+            :mod:`sopel.plugins.callables` modules define the interface and
+            properties that are expected from a plugin callable.
```
dgw, line 570:
```suggestion
        .. seealso::

            The :class:`~sopel.plugins.callables.PluginCallable` class from the
            :mod:`sopel.plugins.callables` module defines the interface and
            properties that are expected from a plugin callable.
```

sopel/plugins/rules.py
```diff
@@ -930,35 +928,35 @@ def kwargs_from_callable(cls, handler):
...
+            'doc': handler.doc,
         }
 
     @classmethod
-    def regex_from_callable(cls, settings, handler):
+    def regex_from_callable(
+        cls,
+        settings: Config,
+        handler: PluginCallable,
+    ) -> tuple[re.Pattern]:
```
dgw, line 959: ?


```suggestion
    ) -> tuple[re.Pattern, ...]:
```

sopel/plugin.py
```diff
@@ -1809,19 +1815,48 @@ def handle_example_bugs(bot, trigger):
...
+
+        argspec = inspect.getfullargspec(url_handler)
+
+        if len(argspec.args) >= match_count:
+            @functools.wraps(url_handler)
+            def wrapped_handler(bot, trigger):
+                return url_handler(bot, trigger, match=trigger)
+
+            if isinstance(function, AbstractPluginObject):
+                function._handler = wrapped_handler
```
dgw, line 1843: Pylance didn't like this when I opened VS Code to more easily look for where the `match` argument handling went.

In case it helps you further refine the WIP, here is the full table of problems I got (limited to open files, I believe, so this won't cover _everything_):

Image

dgw (commented):
I'm submitting a partial review because I don't have time to go through this _whole_ patch today, but I still want you to have some feedback to work on when you get the itch.

Hopefully all the pending comments come through, and me switching between "the new experience" and the old "Files changed" view hasn't made any disappear. (Would I notice which ones were missing, if any? Probably not.)

sopel/irc/utils.py
```diff
@@ -60,7 +60,8 @@ class CapReq:
 
     .. deprecated:: 8.0
 
-        This class is deprecated. See :class:`sopel.plugin.capability` instead.
+        This class is deprecated. See :function:`sopel.plugin.capability`
+        instead.
```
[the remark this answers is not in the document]
Exirel: Apparently it does appear somewhere:

image

sopel/plugins/rules.py
Exirel: Oh wow that's a very dumb mistake.

sopel/irc/utils.py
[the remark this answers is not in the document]
dgw: Wow, I couldn't find it with Sphinx's search. Carry on, then.

sopel/plugin.py
Exirel: Yeah I tried to fix that, I think I'm not done yet. However, I've added a method to replace the handler, so it's now type safe.

Except that I went into a rabbit hole of type hinting things after the other since mypy started complaining about some stuffs here and there.

I'm not too worried, this is an improvement! More to come soon.

Exirel:
And apparently, it will help to move forward #2603 as well!

```
python -m mypy --disallow-untyped-decorators sopel
```

This passed! :tada: 

I can do that here or in another PR. What do you think @dgw?

PR #2670, opened 2025-04-15
plugins: fix type safety issue on EntryPoint's dist attribute

Exirel:
### Description

There is a type safety issue in `sopel/plugins/handlers.py` where `entry_point.dist` could be None. In theory, this never happens because we checked if it has the `name` attribute, but that doesn't convince mypy that we are doing things right.

The solution is rather straightforward: properly annotate the entry point and check that its `dist` attribute isn't `None`, and voilà! The condition check is simpler and mypy is happy too, everyone win!

### Checklist

### DIFF
--- a/sopel/plugins/handlers.py
+++ b/sopel/plugins/handlers.py
@@ -58,6 +58,9 @@ from . import exceptions
 
 
 if TYPE_CHECKING:
+    # TODO: Replace by `importlib.metadata` from stdlib in Python 3.10+
+    from importlib_metadata import EntryPoint
+
     from sopel.bot import Sopel
     from types import ModuleType
 
@@ -607,8 +610,8 @@ class EntryPointPlugin(PyModulePlugin):
     should not be modified at runtime.
     """
 
-    def __init__(self, entry_point):
-        self.entry_point = entry_point
+    def __init__(self, entry_point: EntryPoint) -> None:
+        self.entry_point: EntryPoint = entry_point
         super().__init__(entry_point.name)
 
     def load(self):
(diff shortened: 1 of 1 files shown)


sopel/plugins/handlers.py
```diff
@@ -622,10 +624,14 @@ def get_version(self) -> Optional[str]:
         """
         version: Optional[str] = super().get_version()
 
+        # note: we need to check for attribute because of older Python version
```
dgw, line 627: Should say which versions, so future-us can easily know when it's safe to refactor into something simpler.

sopel/plugins/handlers.py
```diff
@@ -622,10 +624,14 @@ def get_version(self) -> Optional[str]:
...
+        # and we need to check if .dist is not None because hasattr does not
+        # type safeguard properly (or mypy doesn't care?)
         if (
             version is None
-            and hasattr(self.entry_point, "dist")
-            and hasattr(self.entry_point.dist, "name")
+            and hasattr(self.entry_point, 'dist')
+            and self.entry_point.dist is not None
+            and hasattr(self.entry_point.dist, 'name')
```
dgw, line 634: This could have been a +1 diff instead of -2/+3 😂

sopel/plugins/handlers.py
```diff
@@ -58,6 +58,8 @@
 
 
 if TYPE_CHECKING:
+    from importlib_metadata import EntryPoint
```
dgw, line 61: Here too, is it safe to comment something like:

```suggestion
    from importlib_metadata import EntryPoint
    # TODO: This can be switched to `importlib.metadata` from stdlib in Python 3.10+
```

?

dgw (commented):
What I have to say is pretty much only about the comments, because future-us will have forgotten what this is about when it's time to touch it again. 😉

sopel/plugins/handlers.py
Exirel: I think it's more like 3.11, but the doc is unclear about that.

sopel/plugins/handlers.py
Exirel: Will revert that.

dgw (approved):
Good enough for me! (We'll have some fun with this in future major versions, I can tell.)

PR #2676, opened 2025-05-18
coretasks: handle '*' placeholder for omitted channel in WHO/WHOX RPLs

### LINKED ISSUE #2675: WHO response processing must ignore channel placeholder
Sopel sends a WHO for each user joining a channel it already knows about:

https://github.com/sopel-irc/sopel/blob/97bb97c66c1d0816054263c4ad1e05400f229782/sopel/coretasks.py#L996-L998

This is the only place `_send_who()` is called with a nick and not a channel name.

When queried with a channel name, servers MUST return the same channel name in WHO replies. However, they may use an arbitrary channel the user has joined for nickname queries—or, crucially, [MUST\* use `*` as a placeholder](https://ircv3.net/specs/extensions/whox#rpl_whospcrpl-354-numeric-reply) if refusing to name a channel.

When recording WHO(X) replies, Sopel does not check for `*` as the channel name and could end up adding a channel named \* to its list:

https://github.com/sopel-irc/sopel/blob/97bb97c66c1d0816054263c4ad1e05400f229782/sopel/coretasks.py#L1552-L1558

----

\* — This "MUST" applies to WHOX, which was enabled in the reported log snippet. Regular WHO replies are less strictly specified but [the horse docs imply](https://modern.ircdocs.horse/#rplwhoreply-352) that they generally follow the same convention.

_I wish to thank IRC user `dyno10` for mentioning this oddity. It got me curious enough to recheck the specs._

dgw:
### Description

Tin. Fixes #2675.

Also explicitly tests the `0` placeholder for missing account name in WHOX (regular WHO doesn't send that). We do not request the other fields that could have placeholders (`ip` and `oplevel`) in WHOX.

### Checklist

### Notes

Yes, possibly the test assertions are a bit overkill. `assert`ing is cheap, and I would like to be extra sure.

### DIFF
--- a/sopel/coretasks.py
+++ b/sopel/coretasks.py
@@ -1537,6 +1537,11 @@ def _record_who(
         usr.away = away
     if is_bot is not None:
         usr.is_bot = is_bot
+
+    # `*` placeholder is returned for users with no visible channels; see #2675
+    if channel == '*':
+        return
+
     priv = 0
     if modes:
         mapping = {
@@ -1549,6 +1554,7 @@ def _record_who(
         }
         for c in modes:
             priv = priv | mapping[c]
+
     if channel not in bot.channels:
         bot.channels[channel] = target.Channel(
             channel,
--- a/test/test_coretasks.py
+++ b/test/test_coretasks.py
@@ -675,3 +675,49 @@ def test_handle_who_reply_botmode(mockbot):
         ':End of /WHO list.')
 
     assert mockbot.users['Internets'].is_bot is True
+
+
+def test_handle_who_reply_placeholder(mockbot):
+    """Ensure account & channel placeholders in WHO/WHOX reply are ignored."""
+    # no channel (WHO and WHOX both use `*` placeholder if channel is omitted)
+    # no account (regular WHO doesn't provide account name)
+    mockbot.on_message(
+        ':some.irc.network 352 Sopel * '
+        'nesbitt harlow.new.town * Mr_Nesbitt H '
+        ':0 Mr. Nesbitt of Harlow New Town')
+    # yes channel
+    # no account (WHOX uses `0` placeholder if account not logged in)
+    mockbot.on_message(
+        ':some.irc.network 354 Sopel 999 #channel '
+        'kandrews leighton.road.slough Ken_Andrews G '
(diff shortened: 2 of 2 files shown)


test/test_coretasks.py
```diff
@@ -675,3 +675,53 @@ def test_handle_who_reply_botmode(mockbot):
...
+    assert mockbot.users['Ken_Andrews'].realname == 'Ken Andrews'
+    assert mockbot.users['Ken_Andrews'].account is None
+    assert mockbot.users['Ken_Andrews'].away is True
+    assert mockbot.users['Ken_Andrews'].is_bot is None
+
+    assert len(mockbot.channels) == 1
+
+    assert '*' not in mockbot.channels
+    with pytest.raises(KeyError):
+        mockbot.channels['*'].add_user(mockbot.users['Mr_Nesbitt'])
```
Exirel, line 719: You can stop at this:

```suggestion
    with pytest.raises(KeyError):
        mockbot.channels['*']
```

(and yes, CodeQL will complain but whatever)

But I'm not sure it's that necessary because you already checked that `'*' not in mockbot.channels`.

sopel/coretasks.py
```diff
@@ -1549,7 +1549,11 @@ def _record_who(
         }
         for c in modes:
             priv = priv | mapping[c]
-    if channel not in bot.channels:
+
+    # `*` placeholder is returned for users with no visible channels; see #2675
+    if channel == '*':
+        return
+    elif channel not in bot.channels:
```
Exirel, line 1556: Since you are using `return` here you can simplify the if/elif (since you don't have a generic `else`) :

```suggestion
    if channel == '*':
        return

    if channel not in bot.channels:
```

Exirel: Also, you can put that `if *: return` before the `if modes` above!

test/test_coretasks.py
dgw: I guess we can trust the `in` operator (and by extension, `SopelIdentifierMemory.__contains__()`) since that's tested and covered separately under `test/tools/test_tools_memories.py`.

Exirel (approved):

PR #2681, opened 2025-06-15
Replace flake8-import-order with flake8-isort

half-duplex:
### Description

Replaces flake8-import-order with flake8-isort, configuring the latter to match our current style as closely as possible.

isort can do the sorting automatically and seems more comprehensive, more configurable without code, and can (now?) be configured to match our preferences. See also https://github.com/PyCQA/flake8-import-order/issues/163

@dgw would you prefer removing the no_lines_before? "I don't actually like squishing the PROJECT and LOCAL groups together" in https://github.com/sopel-irc/sopel/pull/1752#issuecomment-631117207

Alternative to #2680

### Checklist

- [x] I have read [CONTRIBUTING.md](https://github.com/sopel-irc/sopel/blob/master/CONTRIBUTING.md)
- [x] I can and do license this contribution under the EFLv2
- [x] No issues are reported by `make qa` (runs `make lint` and `make test`)
- [x] I have tested the functionality of the things this change touches

PR #2680, opened 2025-06-14
dev: cap `flake8-import-order` to avoid bugs & Python 3.8 deprecation

dgw:
### Description

flake8-import-order 0.19.0, released earlier this week, drops support for Python 3.7 & 3.8.

It also adds some new ordering & whitespace checks we need to fix in our code before upgrading, too.

### Checklist

_This PR is to fix a CI failure, and is therefore intentionally submitted without running the checks locally first._

- [x] I have read [CONTRIBUTING.md](https://github.com/sopel-irc/sopel/blob/master/CONTRIBUTING.md)
- [x] I can and do license this contribution under the EFLv2
- [ ] No issues are reported by `make qa` (runs `make lint` and `make test`)
- [x] I have tested the functionality of the things this change touches

PR #2682, opened 2025-06-15
Move flake8 and mypy config to pyproject.toml

half-duplex:
### Description

Moves the last stuff out of setup.cfg.
Annoyingly this requires flake8-pyproject as a dev-requirement.
And while I'm in here, adds .venv (used by `uv`) to flake8 ignores.

### Checklist


Exirel:
With #2681 or #2680 that solve the Py3.8 compatibility issue, I really like that change. Good call!

pyproject.toml
```diff
@@ -73,6 +73,36 @@ sopel-plugins = "sopel.cli.plugins:main"
 [project.entry-points.pytest11]
 pytest-sopel = "sopel.tests.pytest_plugin"
 
+[tool.flake8]
+max-line-length = 79
+no-accept-encodings = true
+type-checking-exempt-modules = "typing"
+exclude = [
```
dgw, line 80: Do these work differently in pyproject? Just wondering what happened to the trailing `/*`s.

pyproject.toml
```diff
@@ -73,6 +73,36 @@ sopel-plugins = "sopel.cli.plugins:main"
...
+    ".venv",
```
dgw, line 81: This one in particular might be user-specific, but idk how practical it is to layer a local config on top of the project one.

dgw (commented):
Not that much for me to say, other than (not your fault) flake8 shouldn't require a plugin to support reading its config from `pyproject.toml`

dgw (approved):
Previous line notes are nitpicks at most, no reason to keep holding up.

pyproject.toml
dgw: Might work even better to use [filename](https://flake8.pycqa.org/en/latest/user/options.html#cmdoption-flake8-filename) with a couple glob patterns matching the subdirectories we _do_ want (with `.py` suffix included, of course); then the exclusions become trivial.

Issue #2690, opened 2025-07-22
Python 3.14 support


SnoopJ:
### Requested Feature

Sopel should support Python 3.14, which [entered release candidate status today](https://www.python.org/downloads/release/python-3140rc1/). 3.14.0 is [expected](https://peps.python.org/pep-0745/#schedule) in the first week of Oct 2025.

SnoopJ:
Prognosis looks good from the standpoint of `make qa`.

# Lint

`flake8-future-import` is not compatible with Python 3.14 because of usage of `ast.Str`, which is removed in this release (among other node types).

This plugin appears to be abandoned (last commit 3 years ago) so if we plan to keep it, we should probably switch to consuming it as a [local plugin](https://flake8.pycqa.org/en/latest/user/configuration.html#using-local-plugins) and fix the problem in a fork. Project is licensed MIT, so no concerns there.

<details><summary>click for flake8 failure details</summary>

```
16:53 [jgerity@giskard ~/repos/sopel.git (master 2025-06-21)]
$ make lint
flake8 sopel/ test/
Traceback (most recent call last):
  File "/media/nas/repos/sopel.git/3.14.0rc1_venv/lib/python3.14/site-packages/flake8/plugins/finder.py", line 291, in _load_plugin
    obj = plugin.entry_point.load()
  File "/home/jgerity/.pyenv/versions/3.14.0rc1/lib/python3.14/importlib/metadata/__init__.py", line 179, in load
    module = import_module(match.group('module'))
  File "/home/jgerity/.pyenv/versions/3.14.0rc1/lib/python3.14/importlib/__init__.py", line 88, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
           ~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<frozen importlib._bootstrap>", line 1398, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1371, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1342, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 938, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 762, in exec_module
  File "<frozen importlib._bootstrap>", line 491, in _call_with_frames_removed
[...]

PR #2691, opened 2025-07-24
dev: replace flake8-future-import with fork, fix for Python 3.14

SnoopJ:
### Description

This changeset forks `flake8-future-import` to fix [an issue](https://github.com/xZise/flake8-future-import/issues/27) discovered while investigating #2690, which is likely the only resolution of that problem as the upstream project appears to have been abandoned.

It is possible that we will prefer to remove this plugin entirely in favor of some more simple check for `from __future__ import annotations`, in which case this PR should be closed.

### Checklist

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,5 +1,0 @@
-[flake8:local-plugins]
-extension =
-    FI = flake8_future_import:FutureImportChecker
-paths =
-    ./contrib
--- a/.flake8
+++ b/.flake8
@@ -1,0 +1,5 @@
+[flake8:local-plugins]
+extension =
+    FI = flake8_future_import:FutureImportChecker
+paths =
+    ./contrib
--- a/Makefile
+++ b/Makefile
@@ -5,7 +5,7 @@ qa: lint test coverages
 lint: lint-style lint-type
 
 lint-style:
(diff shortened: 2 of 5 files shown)


Exirel:
Given #2692 works as intended, I think we can close this! Well done diving into all the detail of the issues, that's impressive!

dgw:
> It is possible that we will prefer to remove this plugin entirely in favor of some more simple check for `from __future__ import annotations`, in which case this PR should be closed.

#2692 is such a "more simple check", in the form of a different flake8 plugin that _only_ concerns itself with `__future__.annotations`. And indeed, we probably should prefer that to adopting/vendoring a plugin.

The only thing that would make this solution preferable is enforcement of other (obsolete) `__future__`s' absence. `flake8-future-annotations` doesn't check presence or absence of _anything_ else.

We _have_ already converted the existing files to import _only_ `annotations`. I like to _think_ we're rigorous enough with our code reviews not to let other future imports sneak in by accident.

**Edit:** Ninja'd by @Exirel above, just as I was editing my last sentence. 😂

SnoopJ:
Agreed, the other PR is the best way to address the issue with no change in the lint profile.

PR #2700, opened 2025-08-24
Replace bot.config by bot settings

### LINKED ISSUE #2662: Inconsistent use of `bot.settings` in code and docs
After a question earlier today on IRC about how to access the bot config (`bot.settings.sectionname` vs `config.sectionname`), I took a brief look at some of our documentation about using settings.

First off, it's important to note that many `builtins` still reference `bot.config`—meaning anyone using the builtins as reference (which is part of what they're there for) will copy-paste the "old" style. Things should be updated to use `bot.settings`, and ultimately we could use a proper lifecycle plan for the `bot.config` attribute. It's ["TODO" deprecated](https://github.com/sopel-irc/sopel/blob/6528e72eba78b22830046ac4468edec4b6dbe9cb/sopel/irc/__init__.py#L133-L137), not actually deprecated; we should perhaps have `bot.settings` also be a property, and store the actual `Config` in a sunder attr. (`config` is already a property, so in theory it would be easy to slap a `@sopel.lifecycle.deprecated()` decorator on it.)

More importantly, any places in the docs or website that refer to `bot.config` should be updated.

And finally, [the `sopel.config.types` page](https://sopel.chat/docs/package/config/types) has a very unclear example right at the top, the source of our IRC visitor's confusion: it shows just using `config` on its own.

Image

It's not immediately clear to a reader that this is about working with the config types _outside_ the context of a `bot`. As we work on 8.1 we should keep an eye out for cases like this to clarify them and/or link to the relevant place in the Developer Overview section—here, [the "Configuration and plugin setup" tutorial](https://sopel.chat/docs/plugin/tutorials/configuration-and-setup).

Exirel:
### Description

Tin, fixes #2662 with hopefully slightly less confusing examples in the documentation.

Each commit modifies only one part of the bot e.g., `sopel.plugins`, `sopel.coretasks`, `sopel.builtins`. I hope this will help, at least a little, to review this PR.

### Checklist

### DIFF
--- a/sopel/builtins/admin.py
+++ b/sopel/builtins/admin.py
@@ -36,22 +36,26 @@ class AdminSection(types.StaticSection):
     """Auto-join channels when invited"""
 
 
-def configure(config):
+def configure(settings):
     """
     | name | example | purpose |
     | ---- | ------- | ------- |
     | hold\\_ground | False | Auto-rejoin the channel after being kicked. |
     | auto\\_accept\\_invite | True | Auto-join channels when invited. |
     """
-    config.define_section('admin', AdminSection)
-    config.admin.configure_setting('hold_ground',
-                                   "Automatically re-join after being kicked?")
-    config.admin.configure_setting('auto_accept_invite',
-                                   'Automatically join channels when invited?')
+    settings.define_section('admin', AdminSection)
+    settings.admin.configure_setting(
+        'hold_ground',
+        "Automatically re-join after being kicked?",
+    )
+    settings.admin.configure_setting(
+        'auto_accept_invite',
+        'Automatically join channels when invited?',
+    )
 
 
 def setup(bot):
-    bot.config.define_section('admin', AdminSection)
+    bot.settings.define_section('admin', AdminSection)
 
 
 class InvalidSection(Exception):
@@ -77,11 +81,11 @@ def _get_config_channels(channels):
 
 
 def _set_config_channels(bot, channels):
-    bot.config.core.channels = [
+    bot.settings.core.channels = [
         ' '.join([part for part in items if part])
         for items in channels.items()
     ]
-    bot.config.save()
+    bot.settings.save()
 
 
 def _join(bot, channel, key=None, save=True):
@@ -94,7 +98,7 @@ def _join(bot, channel, key=None, save=True):
         bot.join(channel, key)
 
     if save:
-        channels = dict(_get_config_channels(bot.config.core.channels))
+        channels = dict(_get_config_channels(bot.settings.core.channels))
         # save only if channel is new or key has been changed
         if channel not in channels or channels[channel] != key:
             channels[channel] = key
@@ -106,7 +110,7 @@ def _part(bot, channel, msg=None, save=True):
     bot.part(channel, msg or None)
 
     if save:
-        channels = dict(_get_config_channels(bot.config.core.channels))
+        channels = dict(_get_config_channels(bot.settings.core.channels))
         if channel in channels:
             del channels[channel]
             _set_config_channels(bot, channels)
@@ -297,7 +301,7 @@ def invite_join(bot, trigger):
         LOGGER.info(
             'Got invited to "%s" by an admin.', channel)
         bot.join(channel)
-    elif bot.config.admin.auto_accept_invite:
+    elif bot.settings.admin.auto_accept_invite:
         LOGGER.info(
             'Got invited to "%s"; admin.auto_accept_invite is on', channel)
         bot.join(channel)
@@ -321,7 +325,7 @@ def hold_ground(bot, trigger):
         return
 
     channel = trigger.sender
-    if bot.config.admin.hold_ground:
+    if bot.settings.admin.hold_ground:
         LOGGER.info('Got kicked from "%s"; admin.hold_ground is on.', channel)
         bot.join(channel)
     else:
@@ -403,9 +407,9 @@ def set_config(bot, trigger):
     If value is not provided, the current value will be displayed.
     """
     try:
-        section, section_name, static_sec, option, value = parse_section_option_value(bot.config, trigger)
+        section, section_name, static_sec, option, value = parse_section_option_value(bot.settings, trigger)
     except ValueError:
-        bot.say('Usage: {}set section.option [value]'.format(bot.config.core.help_prefix))
+        bot.say('Usage: {}set section.option [value]'.format(bot.settings.core.help_prefix))
         return
     except (InvalidSection, InvalidSectionOption) as exc:
         bot.say(exc.args[1])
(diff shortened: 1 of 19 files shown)


dgw (changes requested):
This is largely fine, and (as noted in file comments) I was going to just handle a merge conflict or two manually in the terminal so nobody has to mess with this patch any more.

However, there's a significant issue with the edited code example in `sopel/config/types.py`. This patch didn't _introduce_ the problem, but let's fix it as long as we're touching those lines anyway?

dgw:
Figured I'd go ahead and merge #2706 (removing `url`) as long as this already had a merge conflict with `wikipedia`. :)

sopel/builtins/admin.py
```diff
@@ -36,22 +36,26 @@ class AdminSection(types.StaticSection):
     """Auto-join channels when invited"""
 
 
-def configure(config):
+def configure(settings):
```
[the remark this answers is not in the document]
Exirel: Yeah... I've been quite inconsistent with that. I don't even know why. Maybe a naming conflict with an import while doing things but then it became irrelevant further down the edit? I don't really remember!

sopel/config/types.py
```diff
@@ -11,13 +11,15 @@
...
-    >>> config.spam.eggs = ['goose', 'turkey', 'duck', 'chicken', 'quail']
-    >>> print(config.spam.eggs)
+    >>> settings.spam.eggs = ['goose', 'turkey', 'duck', 'chicken', 'quail']
+    >>> print(settings.spam.eggs)
     ['goose', 'turkey', 'duck', 'chicken', 'quail']
-    >>> config.spam.eggs = 'herring'
+    >>> settings.spam.eggs = 'herring'
     Traceback (most recent call last):
         ...
     ValueError: ListAttribute value must be a list.
```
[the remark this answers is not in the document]
dgw: I think it should mention that no one is expected to define new sections after plugins' `setup()`s run, or at least make it very clear that this is for experimentation only.

Giving examples that can cause problems at runtime if used incorrectly makes me nervous 😅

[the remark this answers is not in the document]
Exirel: I get what you are saying. What I want is to explain to the reader how they can use this feature "in isolation", because maybe someone will want to write some command line or some website to work with Sopel's configuration system. However, I completely see the problem you are pointing out: it lacks context, and the most common context is within a plugin (either `setup` or `configure` hook).

Well. Maybe I can add a `.. note:` here? Making sure to express "hey, the usual way in a plugin is through `setup`, go read *that doc* about it" should be good, shouldn't it?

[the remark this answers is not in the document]
dgw: Yeah, that works for making it clear this is just for demonstration.

sopel/config/types.py
```diff
@@ -5,24 +5,45 @@
...
+
+Without that, Sopel's configuration won't recognize the type of ``spam.eggs``
+properly, which can lead to invalid values.
+
+.. note::
+
+    The :meth:`~sopel.config.Config.define_section` is usually used in the
+    plugin's ``setup`` and ``configure`` functions. You can read more about
+    the ways to :ref:`configure your plugin <plugin-anatomy-config>` in the
+    plugin section of the documentation.
```
dgw: Adding "method" is grammatically necessary, but modifying the hyperlink text was a style choice. :)

```suggestion
    The :meth:`~sopel.config.Config.define_section` method is usually used
    in the plugin's ``setup`` and ``configure`` functions. You can read more
    about the :ref:`ways to configure your plugin <plugin-anatomy-config>`
    in the plugin section of the documentation.
```

Note: Width of re-wrap is "close enough" but not verified in a real text editor.

sopel/tests/__init__.py
```diff
@@ -18,3 +24,33 @@ def rawlist(*args):
         [b'PRIVMSG :Hello!\\r\\n']
     """
     return ['{0}\r\n'.format(arg).encode('utf-8') for arg in args]
+
+
+def on_message(bot: Sopel, message: str) -> None:
```
dgw: This equally applies to what's happening in #2716, but this PR is ground zero for noticing the buggy behavior that led to the other patch…

I would much prefer to wrap this functionality in the mock `bot` one gets from the `botfactory`, rather than a top-level function in `tests`. It's far more ergonomic to use `mockbot.on_message()` in a test case.

Is there ever a reason _not_ to wait for running threads to finish before continuing with a given test that triggers plugin callables with an incoming message(s)? Thing is, the `MockIRCServer` makes that behavior globally toggle-able (with the default being On). Nice as it is to have the "real" `Sopel` used in tests, I see a strong case here for a thin `MockSopel` wrapper (or if you can stand it, monkeypatching wrappers for just the few bot methods that spawn threads).

sopel/tests/__init__.py
```diff
@@ -18,3 +24,33 @@ def rawlist(*args):
...
+    """Send the message and wait for all running threads to end.
+
+    When triggering plugin callable with
+    :meth:`bot.on_message <sopel.irc.AbstractBot.on_message>` Sopel may execute
+    them in their own threads. This is the normal and expected behavior, but it
+    isn't practical for testing purpose.
```
dgw:
```suggestion
    isn't practical for testing purposes.
```

sopel/tests/mocks.py
```diff
@@ -114,23 +122,81 @@ class MockIRCServer:
...
+        :param user: the user sending the ``INVITE`` message
+        :param nick: the nick of the invited user
+        :param channel: where the nick is invited to
+        :param blocking: whether to block until all triggered threads
+                         have finished (optional)
+
+        This will send one event: an ``INVITE`` event from ``user`` to ``nick``
+        to join the given ``channel``.
+
+        User this to emulate when a user invite another one to a channel::
```
dgw:
```suggestion
        Use this to emulate when a user invites someone else to a channel::
```

dgw (commented):
Nitpicking again, I guess it's what I do. 😁

sopel/config/types.py
```diff
@@ -5,24 +5,47 @@
...
-As an example, if one wanted to define the ``[spam]`` section as having an
-``eggs`` option, which contains a list of values, they could do this:
+As an example, if one wanted to create a section with an ``eggs`` option, which
+contains a list of values, they can create a subclass of
+:class:`StaticSection`::
 
-    >>> class SpamSection(StaticSection):
+    >>> from sopel.config import types
+    >>> class SpamSection(types.StaticSection):
     ...     eggs = ListAttribute('eggs')
```
```suggestion
    ...     eggs = types.ListAttribute('eggs')
```

dgw (commented):
Man, I don't know if I missed this before or if it crept in during the rebase, but there's _one little issue_ in the `config.types` example. So close!

Here, too, I don't want to mess with the history and break @Exirel's commit signatures.

PR #2701, opened 2025-08-24
tools.identifiers: correct handling of empty and missing CHANTYPES

### LINKED ISSUE #2591: Sane-ify default CHANTYPES
A clarification (ircdocs/modern-irc#230) to the "Modern IRC" docs we (maintainers) know and love will invalidate some of Sopel's default assumptions if accepted.

If the server sends no CHANTYPES token in ISUPPORT, we should assume only `#&`. Some places in Sopel's code assume e.g. `#&+!` instead, which is not supported by the RFCs.

For now, let's plan on reassessing this for 8.1.

dgw:
### Description

The fix here is twofold.

If CHANTYPES is not advertised, assume only the RFC1459 baseline, '#&'.

If CHANTYPES is advertised but without a value, ensure that `Identifier`
can handle the missing value, which the `_optional(tuple)` ISUPPORT
token parser parses as `None`.

Tests updated to correspond, of course.

Closes #2591.

### Checklist

### DIFF
--- a/sopel/tools/identifiers.py
+++ b/sopel/tools/identifiers.py
@@ -50,7 +50,7 @@ RFC1459_STRICT_TABLE = str.maketrans(
     string.ascii_uppercase + '[]\\',
     string.ascii_lowercase + '{}|',
 )
-DEFAULT_CHANTYPES = ('#', '&', '+', '!')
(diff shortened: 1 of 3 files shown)


Exirel (approved):
Clean. Very nice. LGTM. :ship:

PR #2707, opened 2025-09-21
plugins.handlers: PyFilePlugin version falls back to file's mtime

dgw:
### Description

Alternative to the suggested implementation from #2498, putting this fallback in the plugin machinery instead of a special display-only case in `version`.

`PyFilePlugin` keeps track of the modification time when it (re)loads its `.py` file and uses that to derive a datever if there isn't a "real" version in the module.

Starting as draft… If this idea has legs I'll finish up with any needed test cases.

### Checklist

  - Didn't run the checks yet. See note above; this is starting as a proof of concept.
  - Yes, but just manually.

### DIFF
--- a/sopel/plugins/handlers.py
+++ b/sopel/plugins/handlers.py
@@ -44,6 +44,7 @@ away from the rest of the application.
 from __future__ import annotations
 
 import abc
+from datetime import datetime
 import importlib
 import importlib.util
 import inspect
@@ -503,6 +504,7 @@ class PyFilePlugin(PyModulePlugin):
         self.filename = filename
         self.path = filename
         self.module_spec = spec
+        self.file_mtime: float | None = None
 
         super().__init__(name)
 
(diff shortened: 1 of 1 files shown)


sopel/plugins/handlers.py
```diff
@@ -540,8 +542,26 @@ def get_meta_description(self) -> PluginMetaDescription:
...
+        :return: the plugin's version string
+
+        If the plugin module doesn't have a ``__version__`` attribute, this
+        method uses the file's last modification date as the version string.
+        """
+        version = super().get_version()
+        if version is None and self.file_mtime is not None:
+            # try to get version from file metadata
+            dt = datetime.fromtimestamp(self.file_mtime)
+            version = dt.strftime("%Y.%m.%d")
```
SnoopJ, line 557: Nitpick: it feels a little weird to me to call that "getting" rather than constructing the default version. I guess maybe "try" is throwing me since I can't see any reason these operations might fail with the other changes.

sopel/plugins/handlers.py
```diff
@@ -540,8 +542,26 @@ def get_meta_description(self) -> PluginMetaDescription:
...
+    def get_version(self) -> str | None:
+        """Retrieve the plugin's version.
+
+        :return: the plugin's version string
+
+        If the plugin module doesn't have a ``__version__`` attribute, this
+        method uses the file's last modification date as the version string.
+        """
+        version = super().get_version()
+        if version is None and self.file_mtime is not None:
```
dgw: If the above suggestion to always initialize `file_mtime` goes in, this check no longer makes sense.

```suggestion
        if version is None:
```

sopel/plugins/handlers.py
```diff
@@ -503,6 +504,7 @@ def __init__(self, filename: str):
         self.filename = filename
         self.path = filename
         self.module_spec = spec
+        self.file_mtime: float | None = None
```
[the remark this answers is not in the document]
dgw: Hm, I don't see version numbers in `list`, but `show` does include it. Good enough reason to have the mtime always set, thanks.

Exirel (approved):

PR #2709, opened 2025-09-27
docs: update Python for unstable build to 3.13

dgw:
We can't use Python 3.8 any more for a few reasons.

One, it's EOL, so we probably shouldn't use it for "real" tasks other than CI tests.

Two, we're using the `--group` argument to `pip` now for dev requirements, and the last version of `pip` installable on Python 3.8 doesn't support that.

In a way this ties in with sopel-irc/sopel.chat#57, since we have to start building the whole site on newer-than-py3.8 soonish due to deprecation of Netlify's Ubuntu Focal-based build image. If the unstable docs work OK with py3.13, then the main site should as well. (Outside of the `/docs` path, we're still using Jekyll; testing _that_ on a newer build image is a different task that will involve Ruby fun-times.)

### Checklist

  - _Doesn't touch actual code, so new `make qa` issues are impossible._
  - I need the build to run on Netlify in order to assess this; it can't be tested locally.


dgw:
I'll ready this for review now; the deploy preview works after tweaking `SPHINXBUILD` to invoke `python3 -m sphinx` instead of `sphinx-build` directly. Must be some issue with how PATH is derived on Netlify's newer build image. 🫤

dgw:
Thanks, both of you. Going to try merging with `[skip actions]`—which _I think_ will skip the (broken) GHA CI tests while still deploying the unstable docs. Time to validate that theory.

dgw:
Theory validated. No Actions workflow ran, but Netlify did deploy:

image

Thanks again for the reviews!

PR #2714, opened 2025-10-03
isupport: parse CLIENTTAGDENY to a custom type that understands `*`

### LINKED ISSUE #2501: Should include CLIENTTAGDENY parsing in ISUPPORT
Coretasks asks for `message-tags` now, and with that comes the world of _all_ tag types—client-only tags included. (Even if coretasks didn't, plugins can pretty easily request the capability themselves thanks to #2341.)

Part of the message tag specification [is `CLIENTTAGDENY`](https://ircv3.net/specs/extensions/message-tags.html#rpl_isupport-tokens), a comma-separated list of client-only tag names (without the `+` prefix) that the server will ignore (remove) when forwarding messages to other clients. The list can also start with `*` to block all client-only tags and exempt just a few tag names with the `-` prefix.

_Note: I typed this on a "smart"phone, which tried to auto-incorrect many things as I went. Hopefully none of its attempts escaped my notice and thus became successful._

### Current state
Sopel's `ISupport` object handles the `CLIENTTAGDENY` token value as a plain string right now:

```pycon
>>> i.CLIENTTAGDENY
'typing,draft/foobar'
```

Plugin code can certainly use this to check whether a tag is denied, but it's not as convenient as it could be.

### Basic improvement
The most basic parsing needed to make things simpler for plugin code is to split the value on `,` into an actual `list`, i.e. the result above would become:

```pycon
>>> i.CLIENTTAGDENY
['typing', 'draft/foobar']
```

All tags denied:

```pycon
>>> i.CLIENTTAGDENY
['*']
```

Most tags denied, with exceptions:

```pycon
>>> i.CLIENTTAGDENY
['*', '-typing', '-draft/foobar']
```

That would simplify checking whether:

- the tag a plugin wants to use is denied
  (`'*' not in bot.isupport.CLIENTTAGDENY and 'tagname' in bot.isupport.CLIENTTAGDENY`)
- all tags are denied but the tag a plugin wants is exempted
  (`'*' in bot.isupport.CLIENTTAGDENY and '-tagname' in bot.isupport.CLIENTTAGDENY`)

#### Implementation considerations
An empty value is allowed, but servers are recommended to omit the `CLIENTTAGDENY` token entirely when empty.

With Sopel's current `ISupport` behavior, an empty token and an omitted token are _not_ the same, and plugins would need to check `if 'CLIENTTAGDENY' in bot.isupport` in conjunction with the sample conditions I wrote above (or use `bot.isupport.get('CLIENTTAGDENY', [])` to ensure they will not accidentally try to check if a tag name is `in None`).

### Further ideas
A method that takes just a tag name and distills all of the checks described above down to a boolean (e.g. `client_tag_is_denied(tag_name: str) -> bool`) could be a thing, too. I just don't have a solid idea yet where such a method would live in the bot's API structure.

_Inspired by [a recent patch](https://github.com/kiwiirc/irc-framework/pull/366) submitted to irc-framework._

dgw:
This replaces directly passing through the CLIENTTAGDENY value as a plain string.

Consumers of the CLIENTTAGDENY token can either check whether a tag name is directly contained in the set, or go through the `is_denied()` method to apply wildcard logic if appropriate.

The `ClientTagDeny` object's membership tests are case insensitive.

Resolves #2501 in, possibly, "overkill" fashion. I like offering plugin authors some convenience, but I'll understand if it would be preferred to just use a simple list or set.

### Checklist

  - There _were_, but this time I actually remembered to lint and test locally _before_ opening a PR 😅

### DIFF
--- a/sopel/irc/isupport.py
+++ b/sopel/irc/isupport.py
@@ -100,6 +100,81 @@ def _parse_prefix(value):
     return tuple(zip(modes, prefixes))
 
 
+class ClientTagDeny:
+    """Storage for CLIENTTAGDENY ISUPPORT parameter values.
+
+    This class behaves more or less like a set, but is case-insensitive when
+    checking membership and stores all elements in lowercase.
+
+    If the special wildcard ``*`` is present, the :meth:`is_denied` method will
+    return ``True`` for any tag name, except if the explicit negation of that
+    tag name (``-tagname``) is also present.
+    """
+    def __init__(self, iterable=None):
+        self._data = set(x.lower() for x in iterable) if iterable else set()
+
+    def __contains__(self, element: str) -> bool:
+        return element.lower() in self._data
+
+    def __eq__(self, other):
+        if not isinstance(other, ClientTagDeny):
+            return NotImplemented
+        return self._data == other._data
+
+    def __or__(self, other):
+        result = self.__class__(self)
+        result.update(other)
+        return result
+
+    def __ior__(self, other):
+        self.update(other)
+        return self
+
+    def add(self, element: str) -> None:
+        self._data.add(element.lower())
+
+    def remove(self, element: str) -> None:
+        self._data.remove(element.lower())
+
+    def discard(self, element: str) -> None:
+        self._data.discard(element.lower())
+
+    def clear(self):
+        self._data.clear()
+
+    def update(self, *others):
+        for other in others:
+            self._data.update(x.lower() for x in other)
+
+    def __iter__(self):
+        return iter(self._data)
+
+    def __len__(self):
+        return len(self._data)
+
+    def __repr__(self):
+        return f"{self.__class__.__name__}({list(self._data)!r})"
+
+    def is_denied(self, tagname: str) -> bool:
+        """Check if the given ``tagname`` is denied.
+
+        :param tagname: the tag name to check
+        :return: ``True`` if the tag name is denied, ``False`` otherwise
+
+        Supports wildcard and explicit logic; checks are case-insensitive.
+        """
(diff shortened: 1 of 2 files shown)


dgw:
I am trying to keep the interface simple, using the `in` operator where possible. In truth, I don't think my original implementation actually needed an `is_denied()` method… just a smarter `__contains__()`. Oops.

Also oops: Tag names MUST be treated as "[case-sensitive opaque identifiers](https://ircv3.net/specs/extensions/message-tags#format)", so the case-insensitive comparison I added for convenience is spec-breaking.

Making the object read-only makes sense to me. It works as long as a plugin doesn't grab and hold a long-lived reference to it (since the server can re-advertise the CLIENTTAGDENY token if its configuration changes).

I don't have the energy to do it _right now_, but this needs a rewrite with a closer eye to the spec. (Another example: `*` MUST be the first item in the list, so there's no reason to use a condition like `'*' in self._data`.) For now, back to draft it goes.

Exirel:
> I am trying to keep the interface simple, using the `in` operator where possible. In truth, I don't think my original implementation actually needed an `is_denied()` method… just a smarter `__contains__()`. Oops.

The more I think about the issue, the more I think the bot itself should be the one dealing with allowed/denied tags, and offer a convenient interface, and less ISUPPORT itself.

But for that, I need to pursue my exploration of implementing client-only tag in Sopel... and I've barely touched the surface so far.

> It works as long as a plugin doesn't grab and hold a long-lived reference to it

Exactly like anything that is in ISUPPORT: a plugin shall not make a copy and use that as a reference, since it can change at anytime.

PR #2716, opened 2025-11-14
tests: mock IRC server's new shortcut methods

Exirel:
### Description

When Sopel handles a trigger, it may execute commands in their own threads. This is the expected behavior. However, when testing Sopel's various commands, this may result in race conditions.

One use case is the INVITE event handler in the `admin` builtin plugin: because the handler is executed in a thread, the tests would sometimes result in a failure that was quite hard to understand at first.

To solve that, two new methods on the mock IRC server class has been added:

* `invite`: this method will send the appropriate message to the bot, then wait to join all the running thread (unless told not to)
* `message`: this method is a generic "send raw message from the server to the bot" method which will deal properly with the running thread logic

I took that opportunity to write some more about test tools, improving docstrings along the way.

#### Bonus

I also added a new test helper function: `sopel.tests.on_message`. This function will call bot's `on_message` method then join all running trigger threads.

And because I could, I also type annotated some stuffs. ~~By the way, we may need to change the config factory to use `tmp_path`  instead of ` tmpdir`  since Pytest recommend to do so, and it will be easier to type annotate this way.~~ And I replaced the `tmpdir` by a standard `pathlib.Path` object, with more documentation, because really there is no reason not to remove an unwanted dependency.

### Checklist

### DIFF
--- a/sopel/tests/factories.py
+++ b/sopel/tests/factories.py
@@ -154,7 +154,7 @@ class ConfigFactory:
         :return: an instance of test configuration
         """
         tmpfile = self.tmpdir / name
-        tmpfile.write_text(data, encoding='utf-8')
+        tmpfile.write_text(data)
         return config.Config(str(tmpfile))
 
 


dgw (commented):
Not super sure about this approach, if I'm honest. I did a little brainstorming, and I'm hoping it'll inspire you in turn. :)

Exirel:
@dgw this PR is taking a turn! From "let's just fix some race conditions reached by Py3.14" to "better testing tools and better documentation and also let's fix that".

Note that I changed your `versionadded` of a parameter to `versionchanged`: the methods were added in 7.0 but the new parameter is a change added in 7.1. Hope that clarifies that!

sopel/tests/factories.py
```diff
@@ -27,15 +43,13 @@ class BotFactory:
     def preloaded(
         self,
         settings: config.Config,
-        preloads: Optional[Iterable[str]] = None,
```
SnoopJ: Interesting that we had not imported `Optional` but weren't seeing an error? I guess we weren't type-checking here maybe? Anyway, I like the change.

sopel/tests/factories.py
```diff
@@ -67,6 +81,16 @@ def preloaded(
...
     def __call__(self, settings: config.Config) -> bot.Sopel:
+        """Create a test Sopel instance.
+
+        :param settings: test settings used by the test bot
+        :return: an instance of Sopel ready for testing purpose
+
+        This method will create an instance of :class:`~sopel.bot.Sopel` using
+        the ``settings`` provided and a
+        :class:`~sopel.tests.mocks.MockIRCBackend` that can be used to fake
+        messages received from an IRC server.
```
SnoopJ: The mention of the backend swap-out here feels a little oblique to me. I would perhaps add a mention of the fake backend to the one-line summary.

sopel/tests/factories.py
```diff
@@ -75,22 +99,44 @@ def __call__(self, settings: config.Config) -> bot.Sopel:
...
+
     .. seealso::
 
         The :func:`~sopel.tests.pytest_plugin.configfactory` fixture can be
         used to instantiate this factory.
     """
-    def __init__(self, tmpdir):
-        self.tmpdir = tmpdir
+    def __init__(self, tmpdir: pathlib.Path) -> None:
+        self.tmpdir: pathlib.Path = tmpdir
```
SnoopJ: Doubling up on the annotation looks a little weird to me, was `mypy` complaining?

sopel/tests/factories.py
```diff
@@ -75,22 +99,44 @@ def __call__(self, settings: config.Config) -> bot.Sopel:
 class ConfigFactory:
     """Factory to create settings.
 
+    :param tmpdir: a test folder to store documentation files
+
+    An instance of this factory can be used as a callable to create an instance
+    of :class:`~sopel.config.Config`.
+
+    .. versionchanged:: 8.1
```
SnoopJ: Is it also worth mentioning that we now explicitly use UTF-8 when writing the config data out? I don't entirely understand the original implementation but it seems we were at most getting this encoding implicitly.

sopel/tests/factories.py
Exirel: Sphinx was! For some reason `from pathlib import Path` confuses Sphinx that cannot find the reference. 😓

sopel/tests/__init__.py
```diff
@@ -18,3 +24,39 @@ def rawlist(*args):
...
+
+    .. seealso::
+
+        A test IRC server can be used instead with its
+        :meth:`~sopel.tests.mocks.MockIRCServer.message` method.
+
+    """
+    bot.on_message(message)
+
+    while threads := bot.running_triggers:
```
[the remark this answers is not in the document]
Exirel: There could be! With the `echo` mechanism (which I tend to disprove of).

sopel/tests/factories.py
Exirel: And you just discovered a behavior that shouldn't be implicit! Welp... I don't know what to do right now.

On one end, you are right, before we used the default (i.e., let the interpreter decides on the fly), but that can lead to unexpected behavior. On the other end, if here we decide to use utf-8 explicitly, we should also do that when reading the file...

😭

sopel/tests/factories.py
```diff
@@ -27,15 +43,13 @@ class BotFactory:
...
         settings: config.Config,
-        preloads: Optional[Iterable[str]] = None,
+        preloads: Iterable[str] | None = None,
     ) -> bot.Sopel:
         """Create a bot and preload its plugins.
 
         :param settings: Sopel's configuration for testing purposes
-        :type settings: :class:`sopel.config.Config`
-        :param list preloads: list of plugins to preload, setup, and register
+        :param preloads: list of plugins to preload, setup, and register
```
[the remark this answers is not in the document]
dgw: I don't mind trying for language other than "list of" when the object doesn't need to be a `list`, but it reads poorly if the parameter description is just "plugins to preload, setup, and register". My personal style guide won't allow it!

There's "iterable of", but that sounds even weirder than leaving it off completely. To me it's fine to _suggest_ that the parameter be a list while letting the caller use a different Iterable type if they have a reason to.

sopel/tests/factories.py
dgw: If it helps, this kind of nicely fits into the [robustness principle](https://en.wikipedia.org/wiki/Robustness_principle). When Sopel writes out a config file on the user's behalf, yes, it should be correctly encoded. Reading _should_ accept any valid encoding, as long as the interpreter can determine it.

Though I guess that's the issue: Sometimes the correct encoding to use when interpreting a given input isn't possible to determine unambiguously.

At any rate it's out of scope for this PR—but this might be a good suggestion for 9.0 or 10.0, to be stricter about the config and explicitly specifying that it must be UTF-8 and will refuse to load if the file can't be parsed as such.

sopel/tests/factories.py
```diff
@@ -14,44 +24,71 @@
...
         :return: a test instance of the bot
-        :rtype: :class:`sopel.bot.Sopel`
 
-        This will instantiate a :class:`~sopel.bot.Sopel` object, replace its
-        backend with a :class:`~.mocks.MockIRCBackend`, and then preload
-        plugins. This will automatically load the ``coretasks`` plugin, and
-        every other plugin from ``preloads``::
+        This method will create an instance of :class:`~sopel.bot.Sopel` using
+        the ``settings`` provided and a fake backend which collect all messages
```
dgw: First nemesi**S** of 2026, sort of. :P

```suggestion
        the ``settings`` provided and a fake backend which collects all messages
```

sopel/tests/factories.py
```diff
@@ -14,44 +24,71 @@
...
+
+            The created instance will use the
+            :class:`~sopel.tests.mocks.MockIRCBackend` which does not generate
+            network activity and stores messages sent by the bot in-memory.
+
         .. note::
 
             This will automatically setup plugins: be careful with plugins that
-            require access to external services on setup.
+            require access to external services or network connectivity on
```
dgw: Not your word choice, but as long as we're touching it:


```suggestion
            require access to external services or network connectivity during
```

sopel/tests/factories.py
```diff
@@ -67,6 +104,21 @@ def preloaded(
         return mockbot
 
     def __call__(self, settings: config.Config) -> bot.Sopel:
+        """Create a test bot with a fake backend without any plugins.
+
+        :param settings: test settings used by the test bot
+        :return: an instance of Sopel ready for testing purpose
```
dgw:
```suggestion
        :return: an instance of Sopel ready for use in tests
```

sopel/tests/mocks.py
```diff
@@ -4,9 +4,17 @@
 """
 from __future__ import annotations
 
+from typing import Iterable, NoReturn, TYPE_CHECKING
+
 from sopel.irc.abstract_backends import AbstractIRCBackend
 
 
+if TYPE_CHECKING:
+    from sopel.bot import Sopel
```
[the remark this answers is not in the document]
dgw: Silly code-style bot doesn't understand `if TYPE_CHECKING:`, sad.

sopel/tests/mocks.py
```diff
@@ -87,50 +104,170 @@ class MockIRCServer:
...
+    or its more specific methods).
+
+    .. versionchanged:: 7.1
+
+        The ``join_threads`` parameter has been added.
+
+    .. seealso::
+
+        The :class:`~sopel.tests.factories.IRCFactory` factory can be used to
+        create such mock object, either directly or by using ``pytest`` and the
```
dgw:
```suggestion
        create such a mock object, either directly or by using ``pytest`` and the
```

sopel/tests/mocks.py
```diff
@@ -87,50 +104,170 @@ class MockIRCServer:
...
-    in response to messages sent via this ``MockIRCServer`` are finished
-    running before execution can continue. If set to ``False``, the mock
-    server will not wait for the bot to finish processing threaded
-    :term:`callables <Plugin callable>` before returning.
+        While this is not ideal for testing purpose, this behavior can be
+        controlled in two ways:
+
+        * set :attr:`join_threads` to ``True`` (the default) to automatically
+          join threads after sending messages
+        * use the ``blocking`` optional argument of each methods to override
```
dgw:
```suggestion
        * use the ``blocking`` optional argument of each method to override
```

sopel/tests/mocks.py
```diff
@@ -87,50 +104,170 @@ class MockIRCServer:
...
+    """
+    bot: Sopel
+    """The bot instance used by the server to send messages.
 
     .. note::
 
-        You can override ``join_threads`` on a per-method-call basis with the
-        ``blocking`` arguments to the instance methods below.
+        The bot instance should use a :class:`MockIRCBackend` for testing
+        purpose.
```
dgw:
```suggestion
        The bot instance should use a :class:`MockIRCBackend` for testing.
```

sopel/tests/mocks.py
```diff
@@ -87,50 +104,170 @@ class MockIRCServer:
...
+
+        .. note::
+
+            To add **the bot** to a channel after using this method, you should
+            use the :meth:`channel_joined` method.
+
+        .. note::
+
+            To add a user (that is **not** the bot itself) to a channel after
+            using this method, you should use the :meth:`join` method.
```
dgw: I don't think this needs to be two separate notes.


```suggestion
        .. note::

            To add **the bot** to a channel after using this method, you should
            use the :meth:`channel_joined` method.

            To add **a different user** to a channel after using this method,
            you should use the :meth:`join` method.
```

dgw (changes requested):
Man I _finally_ got around to this!!! It probably looks like a lot of line notes, as per usual—and as per usual, it's largely nitpicks in docstrings.

I'm perfectly happy to have nitpicks, mind you, since the last CI run on this PR was a while ago and pushing some fixups will smoke-test everything with whatever dependencies have released updates in the past couple months.

sopel/tests/mocks.py
```diff
@@ -87,50 +104,170 @@ class MockIRCServer:
...
+        these threads after sending messages to the bot.
 
-    The default ``join_threads`` behavior is suitable for testing most common
-    plugin callables, and ensures that all callables dispatched by the ``bot``
-    in response to messages sent via this ``MockIRCServer`` are finished
-    running before execution can continue. If set to ``False``, the mock
-    server will not wait for the bot to finish processing threaded
-    :term:`callables <Plugin callable>` before returning.
+        While this is not ideal for testing purpose, this behavior can be
+        controlled in two ways:
```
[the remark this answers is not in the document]
Exirel: Ah! That's because, in my head, I was talking about "sending a message may result in running threads": the echo feature does that!

docs/source/tests/pytest.rst
```diff
@@ -3,4 +3,4 @@ Fixtures with pytest
 ====================
 
 .. automodule:: sopel.tests.pytest_plugin
-   :members:
+   :members: botfactory,configfactory,triggerfactory,ircfactory,userfactory
```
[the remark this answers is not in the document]
Exirel: Not adding to the batch commit so I can revisit that later.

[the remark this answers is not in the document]
Exirel: That should do it.

image

dgw (approved):
I think we should :shipit: — will allow a brief opportunity to rewrite history if you want, but honestly the follow-up commits look like they'd take a bit of effort to split into fixups and it's probably not worth it. :)

PR #2723, opened 2026-02-04
irc: make `log_raw()` method testable, and test it

dgw:
### Description

With the call to `self.bot.log_raw()` living in the `AsyncioBackend`'s read handler, there was no way to trigger it from tests. (Well… no way without doing _a lot_ of other work.)

Moving the call inside `AbstractBot.on_message()` (which is called for every incoming line anyway) makes the raw log testable via a `caplog` fixture, so we can prove that the bot is logging IRC traffic.

The new test case proves a point in #2721, but does not close any actual issues.

### Checklist

  - Honestly, I was wary of tripping up another of the few test cases that also uses `caplog`, but no: All passed locally.
  - It's literally a new test case :)

### Notes

There was one wee oddity with literal `\r\n` appearing in the outgoing line but not in the incoming one. I added a fix commit addressing it, in lieu of scope creep (making the `MockIRCServer` ensure a trailing `\r\n`).

### DIFF
--- a/test/test_irc.py
+++ b/test/test_irc.py
@@ -518,7 +518,7 @@ def test_log_raw(configfactory, botfactory, ircfactory, caplog):
 
     with caplog.at_level(logging.DEBUG, logger='sopel.raw'):
         bot.write(('WHOIS', 'datboi'))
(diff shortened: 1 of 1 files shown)


Exirel (approved):
OK on second reading it totally makes sense.

PR #2725, opened 2026-02-21
tools, plugins: remove tools.jobs and move Job to plugins.jobs

### LINKED ISSUE #2476: Type conflict between base `tools.jobs.Scheduler` class and `plugins.jobs.Scheduler` subclass
Raised in #2471. Band-aided with a `type: ignore[assignment]` comment, but we should really redesign this part of the API.

@Exirel pointed out that Sopel doesn't really need a non-plugin job scheduler any more, since it can just use the asyncio event loop now. Since `tools.jobs` is already documented as internal tools that plugin developers shouldn't use, any reason we shouldn't just drop the `tools.jobs` module and move the still-useful parts into `plugins.jobs`?

(`tools.jobs` says "As of Sopel 5.3, this is an internal tool […]", so that's several _years_ of warning that plugin authors shouldn't use it. The `Scheduler` class also used to be a `JobScheduler` before some refactoring years ago, so code that tries to use the even older version that wasn't explicitly documented as being internal will fail.)

Exirel:
### Description

Fix #2476.

As per the discussion in the issue, I removed `sopel.tools.job` and moved the Job class to `sopel.plugins.jobs`. While I was there, I added a couple of tests (which required small adjustments here and there), and made sure to add `versionadded`  or ` versionchanged` in the docstrings.

I may want to add more tests in the future, but testing the time related multithreading thing is annoying to automate.

### Checklist


sopel/bot.py
```diff
@@ -619,15 +618,11 @@ def register_callables(self, callables: Iterable[PluginCallable]) -> None:
...
-    def register_jobs(self, jobs: Iterable) -> None:
+    def register_jobs(self, jobs: Iterable[PluginJob]) -> None:
         for func in jobs:
-            job = tools_jobs.Job.from_callable(self.settings, func)
+            job = plugin_jobs.Job.from_callable(self.settings, func)
             self._scheduler.register(job)
 
-    def unregister_jobs(self, jobs: Iterable) -> None:
-        for job in jobs:
-            self._scheduler.remove_callable_job(job)
```
dgw, line 629: Aaaa, removing something with no deprecation warning in a minor version 🙀

But it's not advertised in the API docs (no docstring), so… it's probably fine? 😅

sopel/coretasks.py
```diff
@@ -177,15 +177,14 @@ def setup(bot: Sopel) -> None:
...
-            doc=None,
+
+        handler = plugin.interval(wait_interval)(
+            plugin.label('throttle_join')(
+                _join_event_processing
+            )
         )
-        bot.scheduler.register(job)
+        handler.plugin_name = 'coretasks'
+        bot.register_jobs([handler])
```
dgw, line 187: If I'm honest, this feels a bit too "voodoo" for long-term use, even though conceptually it's just "using decorators without `@` syntactic sugar". Can we make a better interface?

sopel/plugins/jobs.py
```diff
@@ -2,6 +2,12 @@
 
 .. versionadded:: 7.1
 
+.. versionchanged:: 8.1
+
+    This module used to depends on ``sopel.tools.job`` but this module and
+    its content have been removed or moved. The class ``Job`` is now defined
+    here as :class:`sopel.plugins.jobs.Job`.
```
dgw:
```suggestion
    This module used to depend on ``sopel.tools.job`` but this module and
    its content have been removed or moved. The class ``Job`` is now defined
    here as :class:`sopel.plugins.jobs.Job`.
```

sopel/plugins/jobs.py
```diff
@@ -18,54 +24,120 @@
...
         :class:`plugin rules manager <sopel.plugins.rules.Manager>`.
 
     .. important::
 
-        This is an internal tool used by Sopel to manage its jobs. To register
-        a job, plugin authors should use :func:`sopel.plugin.interval`.
+        This is an internal tool used by Sopel to manage its jobs and
+        should not be used by plugin authors. Its usage and documentation is
+        for Sopel core development and advanced developers. It is subject to
+        rapid changes between versions without much (or any) warning.
```
dgw:
```suggestion
        to support Sopel core development. It is subject to rapid changes
        between versions without much (or any) warning.
```

sopel/plugins/jobs.py
```diff
@@ -85,28 +157,346 @@ def unregister_plugin(self, plugin_name):
...
+
+    :param intervals: set of intervals; each is a number of seconds between
+                      calls to ``handler``
+    :type intervals: :term:`iterable`
+    :param str plugin: optional plugin name to which the job belongs
+    :param str label: optional label (name) for the job
+    :param handler: function to be called when the job is ready to execute
+    :type handler: :class:`sopel.plugins.callables.PluginJob`
+    :param bool threaded: run the job in a separate thread
+    :param str doc: optional documentation for the job
```
dgw: Type hints make these redundant.

```suggestion
    :param intervals: set of intervals; each is a number of seconds between
                      calls to ``handler``
    :param plugin: optional plugin name to which the job belongs
    :param label: optional label (name) for the job
    :param handler: function to be called when the job is ready to execute
    :param threaded: run the job in a separate thread
    :param doc: optional documentation for the job
```

sopel/plugins/jobs.py
```diff
@@ -85,28 +157,346 @@ def unregister_plugin(self, plugin_name):
...
+
+    Job is a simple structure that holds information about when a function
+    should be called next. They are best used with a :class:`Scheduler`
+    that will manage job execution when they are ready.
```
dgw:
```suggestion
    A ``Job`` is a simple structure that holds information about when a
    function should be called next. They are best used in conjunction with
    a :class:`Scheduler` that will manage job execution when they are ready.
```

sopel/plugins/jobs.py
```diff
@@ -85,28 +157,346 @@ def unregister_plugin(self, plugin_name):
...
+    When a job is ready, you can execute it by calling its :meth:`execute`
+    method (providing the appropriate ``manager`` argument)::
+
+        if job.is_ready_to_run(time.time()):
+            job.execute(manager)  # marked as running
+            # "next times" have been updated; the job is not running
+
+    In that case, ``execute`` takes care of the running state of the job.
+
+    Alternatively, you can use a ``with`` statement to perform action before
```
dgw: Nitpick.

```suggestion
    Alternatively, you can use a ``with`` statement to perform action(s) before
```

sopel/plugins/jobs.py
```diff
@@ -85,28 +157,346 @@ def unregister_plugin(self, plugin_name):
...
+    .. versionadded:: 8.1
+    """
+    @classmethod
+    def kwargs_from_callable(cls, handler: PluginJob) -> dict:
+        """Generate the keyword arguments to create a new instance.
+
+        :param handler: callable used to generate keyword arguments
+        :type handler: :term:`function`
+        :return: a map of keyword arguments
+        :rtype: dict
```
dgw: Type hints = redundant markup.

```suggestion
        :return: a map of keyword arguments
```

sopel/plugins/jobs.py
```diff
@@ -85,28 +157,346 @@ def unregister_plugin(self, plugin_name):
...
+        }
+
+    @classmethod
+    def from_callable(cls, settings: Config, handler: PluginJob) -> Job:
+        """Instantiate a Job from the bot's ``settings`` and a ``handler``.
+
+        :param settings: bot's settings
+        :type settings: :class:`sopel.config.Config`
+        :param handler: callable used to instantiate a new job
+        :type handler: :term:`function`
```
dgw:
```suggestion
        :param settings: bot's settings
        :param handler: callable used to instantiate a new job
```

sopel/plugins/jobs.py
```diff
@@ -85,28 +157,346 @@ def unregister_plugin(self, plugin_name):
...
+        """
+        return self._threaded
+
+    def is_ready_to_run(self, at_time: int | float) -> bool:
+        """Check if this job is (or will be) ready to run at the given time.
+
+        :param int at_time: Timestamp to check, in seconds
+        :return: ``True`` if the job is (or will be) ready to run, ``False``
+                 otherwise
+        :rtype: bool
```
dgw:
```suggestion
        :param at_time: timestamp to check, in seconds
        :return: ``True`` if the job is (or will be) ready to run, ``False``
                 otherwise
```

sopel/plugins/jobs.py
```diff
@@ -85,28 +157,346 @@ def unregister_plugin(self, plugin_name):
...
+        """
+        return not self.is_running.is_set() and any(
+            (next_time - at_time) <= 0
+            for next_time in self.next_times.values()
+        )
+
+    def next(self: Self, current_time: int | float) -> Self:
+        """Update :attr:`next_times`, assuming it executed at ``current_time``.
+
+        :param int current_time: timestamp of the current time
```
dgw:
```suggestion
        :param current_time: timestamp of the current time
```

sopel/plugins/jobs.py
```diff
@@ -85,28 +157,346 @@ def unregister_plugin(self, plugin_name):
...
+            # if last time + interval is in the future, it's used
+            # else, try to run it asap
+            self.next_times[interval] = max(last_time + interval, current_time)
+
+        return self
+
+    def execute(self, manager: Sopel) -> Any:
+        """Execute the job's handler and return its result.
+
+        :param object manager: used as argument to the job's handler
```
dgw:
```suggestion
        :param manager: used as argument to the job's handler
```

sopel/plugins/jobs.py
```diff
@@ -85,28 +157,346 @@ def unregister_plugin(self, plugin_name):
...
+
+    The :term:`function` to execute is the ``handler``, which must be a
+    callable with this signature::
+
+        def handler(manager):
+            # perform action periodically
+            # return is optional
+
+    The ``manager`` parameter can be any kind of object; usually it's an
+    instance of :class:`sopel.bot.Sopel`.
```
dgw, line 283: Is this still true given type hints for `execute()` enforcing _only_ `Sopel`? (Line 487)

test/plugins/test_plugins_jobs.py
```diff
@@ -22,12 +29,69 @@ class WithJobMockException(Exception):
...
+
+
+def test_jobscheduler_unregister(mockbot: Sopel) -> None:
+    job = jobs.Job.from_callable(mockbot.settings, mock_job)
+    plugin_name = job.get_plugin_name()
+
+    scheduler = jobs.Scheduler(mockbot)
+    assert len(scheduler.jobs) == 0, 'Sanity check failed'
+    assert scheduler.unregister_plugin(plugin_name) == 0, (
+        'No job registered to unregister for that plugin'
```
dgw:
```suggestion
        'There should be no job to unregister for that plugin_name'
```

test/plugins/test_plugins_jobs.py
```diff
@@ -22,12 +29,69 @@ class WithJobMockException(Exception):
...
+    )
+
+    scheduler.register(job)
+    assert len(scheduler.jobs) == 1, 'Sanity check failed'
+    assert scheduler.unregister_plugin(plugin_name) == 1, (
+        'There was one registered job to unregister'
```
dgw:
```suggestion
        'There must be one job to unregister for that plugin_name'
```

dgw (changes requested):
I have checked over perhaps 90% of this, mostly skimming over some of the trivial test changes.

Since we _have_ the tests, most of the comments are expectedly related to docs.

sopel/coretasks.py
Exirel: Hm... on one hand, yes it's just how you can use decorators. And the other hand, the PluginJob new class is not public yet, and I'm not sure it's ready to be, even tho it's probably a very simple interface to use.

So... it's kinda hard to say what's best here.

I also want to point out that, until the PluginJob/PluginCallable rework, you had to use `clean_callable` to make sure it was possible to register the job: there is now one less step to do so!

sopel/plugins/jobs.py
Exirel: Ooooooh good catch. I need to see what I can do with that. Since it's now in `sopel.plugins.jobs`... I'd argue it should always require Sopel as the manager.

sopel/coretasks.py
Exirel: I made it more clear what it does by first generating local decorators, then applying them. And to make sure it was all clear: I added some inline comments to explain that process.

sopel/bot.py
Exirel: Yeah, it's fiiiiiiine...

Exirel:
@dgw made things clearer! Should be good for a last round.

docs/source/conf.py
```diff
@@ -55,8 +55,7 @@
     ('py:class', 'callable'),
     ('py:class', 'depends on subclass'),
     ('py:class', 'mixed'),
-    ('py:class', 'sopel.tools.jobs.Job'),
-    ('py:class', 'sopel.tools.jobs.JobScheduler'),
+    ('py:class', 'mappingproxy'),
```
dgw, line 58: Nitpicking the nitpick config: alphabetical order :)

```suggestion
    ('py:class', 'mappingproxy'),
    ('py:class', 'mixed'),
```

I'm not sure whether GH will actually allow using/applying this as a suggestion since it didn't give me the button to create one. I made it myself.

sopel/bot.py
```diff
@@ -34,14 +34,13 @@
...
 )
-from sopel.tools import jobs as tools_jobs
 from sopel.trigger import Trigger
 
 
 if TYPE_CHECKING:
     from collections.abc import Iterable, Mapping
 
-    from sopel.plugins.callables import PluginCallable
+    from sopel.plugins.callables import PluginCallable, PluginJob
```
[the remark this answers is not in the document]
dgw: @Exirel this cyclic-import stuff doesn't matter inside `if TYPE_CHECKING:` does it?

Sure is convenient if we can just ignore the code-scanner here.

sopel/plugins/jobs.py
```diff
@@ -18,54 +24,120 @@
...
-class Scheduler(jobs.Scheduler):
+class Scheduler(threading.Thread):
     """Plugin job scheduler.
 
     :param manager: bot instance passed to jobs as argument
     :type manager: :class:`sopel.bot.Sopel`
 
-    Scheduler that stores plugin jobs and behaves like its
-    :class:`parent class <sopel.tools.jobs.Scheduler>`.
+    Scheduler is a :class:`thread <threading.Thread>` that keeps track of
```
dgw, line 54:
```suggestion
    Scheduler is a :class:`~threading.Thread` that keeps track of
```

Really nitpick-level. I just figure if it's going to render as a classname, the displayed string might as well _be_ the class name.

sopel/plugins/jobs.py
```diff
@@ -18,54 +24,120 @@
...
+        removed in Sopel 8.1.
+
     .. note::
 
-        This class is a specific implementation of the scheduler, made to store
-        jobs by their plugins and be used by the bot (its ``manager``).
-        It follows a similar interface as the
+        This class is made to store jobs by their plugins and be used by the
+        bot (its ``manager``). It follows a similar interface as the
         :class:`plugin rules manager <sopel.plugins.rules.Manager>`.
```
dgw, line 94: This is another nitpick-level tweak, rephrasing to allow the class name itself as display text for the link.

```suggestion
        :class:`~sopel.plugins.rules.Manager` for plugin rules.
```

sopel/plugins/jobs.py
```diff
@@ -85,28 +157,339 @@ def unregister_plugin(self, plugin_name):
...
+        it retrieves the jobs that are ready for execution, and executes them.
+        See the :meth:`Job.execute` method for more information.
+
+        Internally, it loops forever until its :attr:`stopping` event is set.
 
-    def _get_ready_jobs(self, now):
+        .. note::
+
+            This should not be called directly, as it will be done by the
+            :meth:`threading.Thread.start` method.
```
dgw, line 198: Display nitpick, again.

```suggestion
            :meth:`~threading.Thread.start` method.
```

sopel/plugins/jobs.py
```diff
@@ -85,28 +157,339 @@ def unregister_plugin(self, plugin_name):
...
+
+        @plugin.interval(60)
+        def job_callable(bot: Sopel) -> None:
+            bot.say('#example', 'Ping channel every 60s')
+
+    When a job is ready, you can execute it by calling its :meth:`execute`
+    method (providing the appropriate ``manager`` argument)::
+
+        if job.is_ready_to_run(time.time()):
+            job.execute(manager)  # marked as running
```
dgw, line 289: For future consideration: It might be useful to have the manager's (un)register-jobs method handle setting/removing an attribute on the `Job` instead of always having to pass an argument like this.

Definitely outside the scope of this PR. Just thought about it during my final review pass and seems worth thinking about more at a later date.

Or: Is the current state required for the flexibility of using a `with` statement as demonstrated below? The `manager` argument isn't passed there.

sopel/plugins/jobs.py
```diff
@@ -85,28 +157,339 @@ def unregister_plugin(self, plugin_name):
...
+        :return: a map of keyword arguments
+
+        This classmethod takes the ``handler``'s attributes to generate a map
+        of keyword arguments for the class. This can be used by the
+        :meth:`from_callable` classmethod to instantiate a new rule object.
+
+        The expected attributes are the ones set by decorators from the
+        :mod:`sopel.plugin` module.
+        """
+
```
dgw, line 322: This blank seems like a stray. I thought our linter would catch these, but maybe it's not set up.


```suggestion
```

dgw (approved):
Approving with nitpicks attached; you can take care of whichever ones you like/agree with while squashing\* to get rid of that fixup commit 😛

And yes, I realize it's been a _really_ long time since I last looked at this. It still makes perfect sense to me after reading through the diffs again, so that's a good sign.

\* — Reminder: If you need to rebase on `master`, please force-push it separately from a history rewrite. Especially if you are applying nitpicks/suggestions. Otherwise the little changes get lost in a huge diff from the changed base commit. 🙂

PR #2726, opened 2026-02-22
sopel: replace `typing.Optional` and `typing.Union` with `|` for type annotations

### LINKED ISSUE #2643: Replace `Optional` type annotations with `| None`
What it says on the tin. We've talked about this a few times in PR reviews, and on IRC; I kicked it off in #2642 and decided we could use a tracking issue for the rest.

Edit: it was all done in one go instead of having several PR for that.

Exirel:
### Description

Fixes #2643 and add deal with `Union` too for good measures.

The only exception is type alias with Optional or Union, which can't be replaced until Python 3.10 and above. I haven't tested this with Py3.8 but I'll trust our CI for that.

### Checklist

### DIFF
--- a/sopel/db.py
+++ b/sopel/db.py
@@ -526,8 +526,8 @@ class SopelDB:
         self,
         nick: str,
         key: str,
-        default: typing.Optional[typing.Any] = None
-    ) -> typing.Optional[typing.Any]:
+        default: typing.Any = None
+    ) -> typing.Any:
         """Get a value from the key-value store for ``nick``.
 
         :param nick: the nickname whose values to access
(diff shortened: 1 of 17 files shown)


sopel/db.py
```diff
@@ -526,8 +526,8 @@ def get_nick_value(
         self,
         nick: str,
         key: str,
-        default: typing.Optional[typing.Any] = None
-    ) -> typing.Optional[typing.Any]:
+        default: typing.Any | None = None
+    ) -> typing.Any | None:
```
dgw: This is a mismatch with the other value getters below, which use only `typing.Any` for both `default` and the return value.

Personally I think it's better to treat `None` as its own thing (and therefore change the _other_ getters to use `typing.any | None`), but what I noticed _first_ was the inconsistency.

dgw (changes requested):
I hesitate to call these "nitpicks" per se, because I think the semantic issue is more significant than that word generally means. 😅

But it's not a major change request, so you'll probably make quick work of it.

dgw (approved):
I will now give you an opportunity to squash, since one of the new commits explicitly says "fixup" in its message. :)

Issue #2587, opened 2023-12-18
Undocumented and contradictory hostmask support in `owner` and `admins` settings


dgw:
There is a contradiction between Sopel's docs for the `core.owner` and `core.admins` settings, and how those values are actually used.

## What's wrong?
According to the docs, under both "The [core] configuration section" and "Core Config Section" pages, examples show only nicknames under both of these settings. But internally, a `Trigger` _can_ treat them as `user@host` regex patterns, too:

https://github.com/sopel-irc/sopel/blob/6fc9eee6585082f8794d91dabde2d8ffb1e1d447/sopel/trigger.py#L548-L563

This ability doesn't appear anywhere in the documentation, perhaps because it breaks intended usage of the setting values in other places. `core.owner`, for example, is used in `coretasks` and third-party plugins as the target for important messages about the bot's configuration or errors. That value _must_ be a nickname, not a `user@host` regex, for those uses to work.

## How should we fix it?
We already have `owner_account` and `admin_accounts`, which are used if the server supports `account-tag` or otherwise attaches a services account name to messages in a way that Sopel can parse.

As a middle ground, there should exist `owner_host` and `admin_hosts` settings. The names are subject to revision, but their function is clear: Stricter controls on owner/admin identification beyond simple nickname matching, where account matching isn't available.

I propose a tiered enforcement system like this, described for `owner` but applied to both (just to save me some typing):

- Only `owner` set: Match on nickname
  - Here is the reason for putting this in 8.1.0: We need to plan ahead, add the new settings, and raise deprecation warnings for detectable cases of behavior that will be removed in 9.0, meaning `@` or other nick-invalid characters in one of the `owner` or `admins` values.
- `owner` and `owner_host` set: Match on `user@host`
- `owner` and `owner_account` set: Match on account name
- `owner`, `owner_host`, and `owner_account` set: Match account name

In summary, if either `host` or `account` settings exist, nickname matching is ignored, period. The `account` always takes priority, same as it does now; if the IRCd has services accounts, we defer to its authorization mechanism(s).

### But what about admins?
Yes, I said that the above would apply to admins, but there's another consideration: Sopel can have only one owner, but multiple admins. We might need to validate the configuration by comparing the length of `admins` and `admin_accounts`, though there are plenty of cases I could see where `admin_accounts` or `admin_hosts` would _not_ be the same length as an `admins` list based on nicks only. (Grouped nicks and common host cloaks are just two of many possibilities.)

[...]

PR #2733, opened 2026-04-08
config, doc, test: clarify `nick@host` format of `owner, admins` settings

SnoopJ:
### Description

This changeset clarifies in the documentation that the core config field `owner` may be given as `nick@host` as well as a nick, and also covers this previously-untested usage in the `Trigger` tests. (Edit: this applies to the `admins` field as well)

### Checklist

  - `1451 passed, 8 xfailed, 1 warning in 47.36s`

### Remarks for reviewer

In moving the dummy configs to a fixture, I originally wanted to move the template out of the module namespace and into the fixture, relying on `textwrap.dedent()` to take care of the indentation:

```python
@pytest.fixture
def tmpconfig(configfactory) -> Config:
    return textwrap.dedent(f"""
        ...
    """
```

But this approach can cause trouble with trying to add a field later if also using an f-string, so I settled on a fixture that returns source. In particular, the thing I stumbled on was wanting to write:

```python
def test_ircv3_account_tag_trigger(nick, tmpconfig_src, configfactory):
    line = '@account=bar :Nick_Is_Not_Foo!foo@example.com PRIVMSG #Sopel :Hello, world'
    pretrigger = PreTrigger(nick, line)

    # NOTE: only first line of tmpconfig_src has the expected indent level
    # so the result ends up exactly the same as without the dedent()
    tmp_config_account = dedent(f"""
        {tmpconfig_src}
        owner_account = bar
    """)
```

I don't feel strongly about eliminating the module-level `TMP_CONFIG` and if desired I can update this PR to preserve it.

### DIFF
--- a/sopel/config/core_section.py
+++ b/sopel/config/core_section.py
@@ -79,7 +79,7 @@ class CoreSection(StaticSection):
     """
 
     admins = ListAttribute('admins')
-    """The list of people (other than the owner) who can administer the bot.
+    """List of nicks or ``nick@host`` who can administer the bot (other than the owner).
 
     Example:
 
@@ -1114,7 +1114,7 @@ class CoreSection(StaticSection):
     """
 
     owner = ValidatedAttribute('owner', default=NO_DEFAULT)
-    """The IRC name of the owner of the bot.
+    """The IRC nick or ``nick@host`` of the owner of the bot.
 
     **Required** even if :attr:`owner_account` is set.
     """
--- a/test/test_trigger.py
+++ b/test/test_trigger.py
@@ -18,6 +18,14 @@ admins =
 """
 
 
+TMP_CONFIG_HOSTMASK = """
+[core]
+owner = Foo@example.com
+admins =
+    Bar
+"""
+
+
 TMP_CONFIG_ACCOUNT = """
 [core]
 owner = Foo
@@ -289,11 +297,12 @@ def test_ctcp_action_pretrigger(nick):
     assert pretrigger.status_prefix is None
 
 
-def test_ctcp_action_trigger(nick, configfactory):
+@pytest.mark.parametrize("configsrc", [TMP_CONFIG, TMP_CONFIG_HOSTMASK])
+def test_ctcp_action_trigger(nick, configsrc, configfactory):
     line = ':Foo!bar@example.com PRIVMSG #Sopel :\x01ACTION Hello, world\x01'
     pretrigger = PreTrigger(nick, line)
 
-    config = configfactory('default.cfg', TMP_CONFIG)
+    config = configfactory('default.cfg', configsrc)
     fakematch = re.match('.*', line)
 
     trigger = Trigger(config, pretrigger, fakematch)
@@ -335,11 +344,12 @@ def test_ircv3_extended_join_pretrigger(nick):
     assert pretrigger.status_prefix is None
 
 
-def test_ircv3_extended_join_trigger(nick, configfactory):
+@pytest.mark.parametrize("configsrc", [TMP_CONFIG, TMP_CONFIG_HOSTMASK])
+def test_ircv3_extended_join_trigger(nick, configsrc, configfactory):
     line = ':Foo!foo@example.com JOIN #Sopel bar :Real Name'
     pretrigger = PreTrigger(nick, line)
 
-    config = configfactory('default.cfg', TMP_CONFIG)
+    config = configfactory('default.cfg', configsrc)
 
     fakematch = re.match('.*', line)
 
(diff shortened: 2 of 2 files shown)


sopel/config/core_section.py
```diff
@@ -1114,7 +1114,7 @@ def homedir(self):
     """
 
     owner = ValidatedAttribute('owner', default=NO_DEFAULT)
-    """The IRC name of the owner of the bot.
+    """The IRC nick or ``nick@hostmask`` of the owner of the bot.
```
dgw: I think this should just say "host". A "hostmask" is the three-part `nick!user@host` string.

```suggestion
    """The IRC nick or ``nick@host`` of the owner of the bot.
```

test/test_trigger.py
```diff
@@ -289,11 +297,12 @@ def test_ctcp_action_pretrigger(nick):
...
 
 
-def test_ctcp_action_trigger(nick, configfactory):
+@pytest.mark.parametrize("config", [TMP_CONFIG, TMP_CONFIG_HOSTMASK])
+def test_ctcp_action_trigger(nick, config, configfactory):
     line = ':Foo!bar@example.com PRIVMSG #Sopel :\x01ACTION Hello, world\x01'
     pretrigger = PreTrigger(nick, line)
 
-    config = configfactory('default.cfg', TMP_CONFIG)
+    config = configfactory('default.cfg', config)
```
dgw: This pattern is the same in each modified test, but I will only leave the comment once:

I don't like shadowing the parameter name with a local.

Lowest-impact change would be to call the new parameter something else like `configfile`, leaving the local as `config` (which is accurate; it's a config object created by the `configfactory` function).

dgw (changes requested):
In addition to the style tweaks mentioned as line notes, let's either include `core.admins` in this as well ([it also uses `match_host_or_nick()`](https://github.com/sopel-irc/sopel/blob/f8d536dff1dcdf23b6c85ed04ee1d799a3a984f0/sopel/trigger.py#L559-L563)) or immediately follow up with another PR to do so.

test/test_trigger.py
SnoopJ: You know, I'd entirely missed the collision of names here, but I agree that there shouldn't be one. Resolved in 
ebb98a6

dgw (approved):
Discretizing tests for the different cases will probably happen in whatever patch addresses #2734 (thanks for opening that). For now this is a good step forward.

10 commits is a lot, with the fixups and stuff, so this is your 👍 from me to squash things. 😸

SnoopJ:
I have a smaller patch for #2734 that "just" adds a new test, will open that in a moment, but I don't think it's any big problem if we change the doc in one changeset and address the problem of test coverage in another, since that gap has been there a while anyway.

dgw:
GH doesn't support relationships between PRs and issues (why?) but consider this blocked by #2736 for now. (@Exirel you're still welcome to review; certainly the tests are useful unless we get rid of the behavior entirely.)

dgw:
After what one might call _much_ discussion on IRC, we seem to have an alternative plan involving a new config value and deprecating certain *value patterns* of the existing `owner` setting. I will describe it in #2736, perhaps mirrored in #2587 (or just linked from there) as it's all related.

sopel/config/core_section.py
```diff
@@ -79,7 +79,7 @@ class CoreSection(StaticSection):
     """
 
     admins = ListAttribute('admins')
-    """The list of people (other than the owner) who can administer the bot.
+    """List of nicks or ``nick@host`` who can administer the bot (other than the owner).
```
Exirel, line 82: Let's upgrade the code-block bellow with a hostmask example!

sopel/config/core_section.py
```diff
@@ -1114,7 +1114,7 @@ def homedir(self):
     """
 
     owner = ValidatedAttribute('owner', default=NO_DEFAULT)
-    """The IRC name of the owner of the bot.
+    """The IRC nick or ``nick@host`` of the owner of the bot.
```
Exirel, line 1117: As mentioned on IRC (not by me but I agree), I think it's best not to allow nick@host in `core.owner`.

Exirel (changes requested):
As said before me: I agree that it's not a good idea to allow a hostmask for `core.owner` as this is used and documented as used to ping the bot's owner.

I prefer the suggested solution (IIRC) of adding a new setting for owner's hostmask. Therefore, I'm not sure we should add these tests now.

dgw:
With #2735 merged and #2740 we're on the way to addressing the results of discussion on this patch regarding `core.owner`.

I think #2587 can cover any further discussion about how to deal with `core.admins`. But I'm afraid this patch is still sort of in limbo, because we can't realistically update the docs for using (or not using) hostnames in `core.admins` until we decide whether or not mixing the two formats in a single list is acceptable. 😅

PR #2735, opened 2026-04-12
test: cover admin functionality of `Trigger`

### LINKED ISSUE #2734: `core.admins` functionality is not tested
On #2733, @dgw pointed out that the `core.admins` config field also should be documented as accepting the `nick@host` format, which made me realize that the functionality of this field (with respect to `Trigger`) is not being tested.

The proof of the pudding is in the `pytest`ing, so here's a diff that deletes part of the check:

```diff
diff --git a/sopel/trigger.py b/sopel/trigger.py
index 0fae5bc3..e9ac461d 100644
--- a/sopel/trigger.py
+++ b/sopel/trigger.py
@@ -557,7 +557,5 @@ class Trigger(str):
         else:
             self._owner = match_host_or_nick(settings.core.owner)
         self._admin = (
-            self._owner or
-            self.account in settings.core.admin_accounts or
-            any(match_host_or_nick(item) for item in settings.core.admins)
+            self._owner
         )
```

With this diff applied, the test suite experiences no failures:

```
$ PYTEST_ADDOPTS="-x" make test                                                                                                                                                              
coverage run -m pytest -v
==================================================================================== test session starts ====================================================================================
platform linux -- Python 3.9.16, pytest-7.4.4, pluggy-1.6.0 -- /var/lib/gitea/data/gitea-repositories/snoopj/sopel.git/env/3.8/bin/python3                                                   
cachedir: .pytest_cache
rootdir: /var/lib/gitea/data/gitea-repositories/snoopj/sopel.git                              
configfile: pyproject.toml
testpaths: test, sopel   
plugins: sopel-8.1.0.dev0, requests-mock-1.9.3, vcr-1.0.2                                     
collected 1453 items

... snip ...

=================== 1445 passed, 8 xfailed, 1 warning in 60.56s (0:01:00) ===================
```

We should add tests to `test_trigger.py` that cover this feature (for both a list of nicks and a list of ``nick@host``)

SnoopJ:
### Description

Closes #2734 by adding a new test that checks that `Trigger.admin` is correct for a basic `PRIVMSG`. Could be expanded for other types of messages but I figured this was a good start.

### Checklist

### DIFF
--- a/test/test_trigger.py
+++ b/test/test_trigger.py
@@ -433,3 +433,27 @@ def test_statusmsg_trigger(nick, configfactory):
     assert trigger.sender == '#channel'
     assert trigger.sender == Identifier('#channel')
     assert trigger.status_prefix == '@'
+
+
+@pytest.mark.parametrize("senderhostmask,isadmin", [
+    ("Foo!bar@example.com", True),
+    ("Bar!bar@example.com", True),
+    ("Stranger!bar@example.com", False),
+    ("Bar!bar@stranger.domain", False)
+])
+def test_privmsg_admin(nick, senderhostmask, isadmin, configfactory):
+    line = f':{senderhostmask} PRIVMSG #Sopel :Hello world!'
+    pretrigger = PreTrigger(nick, line)
+
+    config = configfactory('default.cfg', """
+        [core]
+        owner = SomeoneElse
+        admins =
+            Foo@example.com
+            Bar@example.com
+    """)
+    fakematch = re.match('.*', line)
+
+    trigger = Trigger(config, pretrigger, fakematch)
+    assert trigger.sender == '#Sopel'
+    assert trigger.admin is isadmin, "Admin privilege does not match expectation"
--- a/test/test_trigger.py
+++ b/test/test_trigger.py
@@ -441,7 +441,7 @@ def test_statusmsg_trigger(nick, configfactory):
     ("Stranger!bar@example.com", False),
     ("Bar!bar@stranger.domain", False)
 ])
-def test_privmsg_admin(nick, senderhostmask, isadmin, configfactory):
+def test_privmsg_admin_match_host(nick, senderhostmask, isadmin, configfactory):
     line = f':{senderhostmask} PRIVMSG #Sopel :Hello world!'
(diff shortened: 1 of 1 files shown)


dgw:
This is getting at what I meant [here](https://github.com/sopel-irc/sopel/pull/2733#pullrequestreview-4095670186) about "discretizing" different cases.

Parametrized tests make each combination a unique test that can pass or fail independently, which I (at least) find helpful when something regresses. That's great :)

I would just ask you to consider adding nick-only matching to this matrix for completeness. Possibly also wildcard handling, since that's a feature of the underlying pattern generator:

https://github.com/sopel-irc/sopel/blob/f8d536dff1dcdf23b6c85ed04ee1d799a3a984f0/sopel/tools/__init__.py#L82-L93

test/test_trigger.py
```diff
@@ -433,3 +433,27 @@ def test_statusmsg_trigger(nick, configfactory):
...
     assert trigger.status_prefix == '@'
+
+
+@pytest.mark.parametrize("senderhostmask,isadmin", [
+    ("Foo!bar@example.com", True),
+    ("Bar!bar@example.com", True),
+    ("Stranger!bar@example.com", False),
+    ("Bar!bar@stranger.domain", False)
+])
+def test_privmsg_admin(nick, senderhostmask, isadmin, configfactory):
```
Exirel: A bit nitpicky: we are testing only config with hostmask declaration, so `test_privmsg_admin`  is a bit too generic a name to describe the test case.

Exirel (approved):
I approve. I added a nitpick you can either try to fix in some way or ignore.

I concur with @dgw : having a very clear and direct test with the parameters on top makes it easy to debug in case the behavior change and/or a bug is introduced while working on it later.

test/test_trigger.py
dgw: I went with `test_privmsg_admin_match_host` in a follow-up commit, admittedly without pre-discussion. 😅

This doesn't worry me very much in the long term because we'll need to rework the test cases some as we tease host matching out of `core.owner` and possibly `core.admins` and into dedicated settings.

Exirel (approved):

Issue #2737, opened 2026-05-07
Python 3.15 support


SnoopJ:
### Requested Feature

Sopel should support Python 3.15, which [entered beta status today](https://www.python.org/downloads/release/python-3150b1/). 3.15.0 is [expected](https://peps.python.org/pep-0790/#schedule) in the first week of Oct 2026.

### Notes

Same idea as #2690, #2608. We might need to change stuff, we might not need to do anything. Likely to end up waiting on upstreams for a while.

SnoopJ:
Looks like we're in the clear for 3.15.0b1, `make qa` passes! 🥳

```
===== 1449 passed, 8 xfailed, 5 warnings in 57.38s =====
```

4 of those warnings are from passing `count` to `re.sub()` by position, which is deprecated since Python 3.13, and the other warning is a SQLAlchemy deprecation we've been seeing for a while.

```
test/builtins/test_builtins_find.py::test_valid_replacements[ABabAB-s/b/c/i-A\x02c\x02abAB]                                                                                                  
test/builtins/test_builtins_find.py::test_valid_replacements[ABabAB-s/b/c/i-A\x02c\x02abAB]                                                                                                  
test/builtins/test_builtins_find.py::test_valid_replacements[ABabAB-s/b/c/ig-A\x02c\x02a\x02c\x02A\x02c\x02]
test/builtins/test_builtins_find.py::test_valid_replacements[ABabAB-s/b/c/ig-A\x02c\x02a\x02c\x02A\x02c\x02]
  /var/lib/gitea/data/gitea-repositories/snoopj/sopel.git/sopel/builtins/find.py:187: DeprecationWarning: 'count' is passed as positional argument
    return re.sub(regex, subst, line, count == 1)
```

I do also see a bunch of jank coming from some of the Sphinx extensions, but these errors are nonfatal to `make qa` and to `make build_docs` as well (though I did not check the generated documentation). Will check if those have been reported upstream, but no action required from us AFAICT.

<details><summary>click to see reported errors</summary>

```
[...]

PR #2738, opened 2026-05-07
find: use non-deprecated re.sub() API

SnoopJ:
### Description

This changeset resolves deprecated use of `re.sub()` in the `find` built-in. Passing the `count` parameter by position has been deprecated since Python 3.13.

Noticed this in the course of filing #2737.

### Checklist

  - ```===== 1445 passed, 8 xfailed, 1 warning in 44.19s =====``` (down from 5 warnings before this changeset)

### DIFF
--- a/sopel/builtins/find.py
+++ b/sopel/builtins/find.py
@@ -182,9 +182,11 @@ def findandreplace(bot, trigger):
     # i flag turns off case sensitivity. re.U turns on unicode replacement.
     if 'i' in flags:
         regex = re.compile(re.escape(old), re.U | re.I)
+        # re.sub() uses count=0 to mean "replace all" and str.replace() uses -1, so we must translate here
+        re_count = int(count == 1)
 
         def repl(line, subst):
-            return re.sub(regex, subst, line, count == 1)
+            return re.sub(regex, subst, line, count=re_count)
     else:
         def repl(line, subst):
             return line.replace(old, subst, count)


sopel/builtins/find.py
```diff
@@ -182,9 +182,11 @@ def findandreplace(bot, trigger):
     # i flag turns off case sensitivity. re.U turns on unicode replacement.
     if 'i' in flags:
         regex = re.compile(re.escape(old), re.U | re.I)
+        # re.sub() uses count=0 to mean "replace all" and str.replace() uses -1, so we must translate here
+        re_count = int(count == 1)
```
SnoopJ, line 186: I did think about other ways to re-arrange here but settled on something relatively close to the original and a comment explaining why it looks so weird.

dgw: Honestly we (the project, not necessarily those personnel currently discussing) should have untangled this ages ago. At least there's a comment explaining use of `==` now, but holy— why optimize out the use of `re.sub()`? Regex operations aren't _that_ slow.

In a hot-path context I would have a different opinion, but in this case one of us should feel free to simplify things here and unify on `re.sub()` to avoid the hard-to-read voodoo. Maintainability trumps performance in this case, IMO.

dgw (approved):
Approving with the caveat of needing future simplification, as addressed in the line notes at L+185. This does fix the deprecation, though.

_(Note to self: See `simplify-findandreplace` branch for a patch built on top of this branch.)_

dgw:
Logic simplification now opened: #2739

PR #2739, opened 2026-05-10
find: simplify `findandreplace()` logic

dgw:
### Description

Streamline the logic in `find` plugin's `findandreplace()` function:

* Remove the dance of choosing `re.sub()` or `str.replace()` depending on input. Just use `re.sub()` always.
* Bonus: Use `@plugin.require_chanmsg` decorator instead of an inline `if` block.

In follow-up to https://github.com/sopel-irc/sopel/pull/2738/files/eaf6ae38e6477aa97cbb560e25018a67665e59ae#r3204111000

Performance isn't a big concern here, so I think we're much better off with an easier-to-read function that takes slightly longer to run when a user invokes it. Chances are no one will notice the added delay.

### Checklist

  - I'm pleased to report that this feature is already covered by `test/builtins/test_builtins_find.py` and none of the tests broke with the patch here 🥳

### DIFF
--- a/sopel/builtins/find.py
+++ b/sopel/builtins/find.py
@@ -152,9 +152,9 @@ def findandreplace(bot, trigger):
     # Correcting other person vs self.
     rnick = bot.make_identifier(trigger.group('nick') or trigger.nick)
 
-    # only do something if there is conversation history to work with
     history = bot.memory['find_lines'].get(trigger.sender, {}).get(rnick, None)
     if not history:
+        # No conversation history to potentially correct; bail out.
         return
 
     sep = trigger.group('sep')
@@ -171,8 +171,11 @@ def findandreplace(bot, trigger):
 
     # If g flag is given, replace all. Otherwise, replace once.
     count = 0 if 'g' in flags else 1
+
     # If i flag is given, ignore case when replacing.
-    regex_flags = re.U | (re.I if 'i' in flags else 0)
+    regex_flags = re.U
+    if 'i' in flags:
+        regex_flags |= re.I
 
     # Precompile the regex with its flags
     regex = re.compile(re.escape(old), regex_flags)
--- a/sopel/builtins/find.py
+++ b/sopel/builtins/find.py
@@ -150,7 +150,11 @@ def kick_cleanup(bot, trigger):
 @plugin.require_chanmsg
 def findandreplace(bot, trigger):
     # Correcting other person vs self.
-    rnick = bot.make_identifier(trigger.group('nick') or trigger.nick)
+    correcting_self = True
+    rnick = trigger.nick
+    if trigger.group('nick'):
+        correcting_self = False
+        rnick = bot.make_identifier(trigger.group('nick'))
 
     history = bot.memory['find_lines'].get(trigger.sender, {}).get(rnick, None)
     if not history:
@@ -162,7 +166,6 @@ def findandreplace(bot, trigger):
 
     old = escape_sequence_pattern.sub(decode_escape, trigger.group('old'))
     new = trigger.group('new')
-    me = False  # /me command
     flags = trigger.group('flags') or ''
 
     # only clean/format the new string if it's non-empty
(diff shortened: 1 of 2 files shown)


sopel/builtins/find.py
```diff
@@ -173,23 +170,16 @@ def findandreplace(bot, trigger):
         new = escape_sequence_pattern.sub(decode_escape, new)
 
     # If g flag is given, replace all. Otherwise, replace once.
-    if 'g' in flags:
-        count = -1
-    else:
-        count = 1
+    count = 0 if 'g' in flags else 1
+    # If i flag is given, ignore case when replacing.
+    regex_flags = re.U | (re.I if 'i' in flags else 0)
```
SnoopJ: I would personally prefer to elaborate that ternary, it's a bit dense.

```suggestion
    regex_flags = re.U
    if 'i' in flags:
        regex_flags |= re.I
```

sopel/builtins/find.py
```diff
@@ -200,14 +190,16 @@ def repl(line, subst):
...
         else:
             me = False
-        replaced = repl(line, new)
-        if replaced != line:  # we are done
+        replaced = do_replacement(line, new)
+        if replaced != line:
+            # we are done
             new_line = replaced
-            new_display = repl(line, bold(new))
+            new_display = do_replacement(line, bold(new))
```
SnoopJ, line 204: I didn't really pay attention to this when last we looked at it, but doing the replacement twice to apply the bold is… huh.

It smells like it could be simplified further, but even if so, it's not something that's particularly worth doing in this changeset. Just sharing my thoughts, no action required.

sopel/builtins/find.py
```diff
@@ -147,15 +147,12 @@ def kick_cleanup(bot, trigger):
...
 def findandreplace(bot, trigger):
-    # Don't bother in PM
-    if trigger.is_privmsg:
-        return
-
     # Correcting other person vs self.
     rnick = bot.make_identifier(trigger.group('nick') or trigger.nick)
 
-    # only do something if there is conversation to work with
+    # only do something if there is conversation history to work with
```
SnoopJ: I'd be tempted to stick this in the `if not history` block and rewrite it to say something like `# no lines from that sender, bail out`

SnoopJ (approved):
One nit and two remarks, but willing to approve now because it's definitely simpler.

sopel/builtins/find.py
dgw: …it was like that initially and I chose to be clever just before committing lmao

sopel/builtins/find.py
dgw: This second call exists because of #2589, and I couldn't immediately think of a better alternative. If there are two different versions of the replaced string, it seems (almost?) inevitable that we have to replace twice.

As one of my occasional Tiny Experiments With """AI""", I asked VS Code if it could reduce the two replacements to one. Best it ("Auto" was selected so I didn't catch which model) could do was one of the obvious solutions: move the second replacement inside `do_replacement()` and have it return a 2-tuple of `replaced, display`. We still end up replacing twice, so `do_replacement()` gets longer without shortening any of the logic outside it.

sopel/builtins/find.py
```diff
@@ -173,23 +170,19 @@ def findandreplace(bot, trigger):
...
-            return re.sub(regex, subst, line, count=re_count)
-    else:
-        def repl(line, subst):
-            return line.replace(old, subst, count)
+    # Precompile the regex with its flags
+    regex = re.compile(re.escape(old), regex_flags)
+
+    # Dynamically defined replacement function makes later calls a bit clearer
+    def do_replacement(line, subst):
+        return re.sub(regex, subst, line, count=count)
```
Exirel: As per the documentation of [Pattern.sub](https://docs.python.org/3/library/re.html#re.Pattern.sub):

```suggestion
    # Dynamically defined replacement function makes later calls a bit clearer
    def do_replacement(line, subst):
        return regex.sub(subst, line, count=count)
```

sopel/builtins/find.py
Exirel: Oh yeah, you can do something like that:

```python
def replace_highlight(regex: re.Pattern, line: str, subst: str) -> tuple[str, str|None]:
    new_line = regex.sub(subst, line, count=count)
    if new_line == line:
        return new_line, None
    return new_line, regex.sub(bold(subst), line, count=count)
```

This way you can 1. have this function declared outside the find command, 2. make it returns the replaced line and the display version and 3. have unit tests for it!

You can then adapt this code with:

```python
for line in history:
    # ... <snip> ...
    new_line, new_display = replace_highlight(regex, line, new)
    if new_display is not None:
        break

if new_display is None:
    # Didn't find anything
    return
```

From the type checking perspective, it should be even more correct, because new_line can never be None, and new_display is properly checked for None (it wasn't before).

Exirel (approved):
OK, hear me out: it's fine, but it could have more over-engineering prowess on top. What do you say?

sopel/builtins/find.py
Exirel: Another idea...

You can do more:

```python
for line in history:
    # ... <snip> ...
    new_line, new_display = replace_highlight(regex, line, new)
    if new_display is not None:
        break
else:
    return
```

Because of the [for-in-else statement](https://docs.python.org/3/reference/compound_stmts.html#for), you don't even need that second if!

dgw:
@Exirel it's your fault that I got sucked into even more tweaks here. Spending more time in the file made me notice other things that annoyed me 😝

SnoopJ (approved):
Changes look good to me, no further remarks

Exirel:
> Spending more time in the file made me notice other things that annoyed me 😝

But what do you think of my idea of `def replace_highlight(regex: re.Pattern, line: str, subst: str) -> tuple[str, str|None]:`?

dgw:
> > Spending more time in the file made me notice other things that annoyed me 😝
> 
> But what do you think of my idea of `def replace_highlight(regex: re.Pattern, line: str, subst: str) -> tuple[str, str|None]:`?

I can't speak for the me who pushed the last set of amendments to this patch, but the me of today thinks that would complicate implementation of #2741 (opened about a week later) 😅

Which is to say, either we'd end up throwing away the default display copy in some cases, or we'd have to give that nice tidy function access to the bot's settings.

PR #2747, opened 2026-07-15
db: return get_*_value() default as-is instead of deserializing it

### LINKED ISSUE #2672: bot.db.get_channel_value(chan, key, False) is returning a string instead of a bool
### Description

If you have data in the default database and you try to pull it out using:

`bot.db.get_channel_value(chan, key, False)`

The docs say, https://sopel.chat/docs/package/db.html#sopel.db.SopelDB.get_channel_value

"default – value to return if key does not have a value set (optional)"

If you put in `False` as the third parameter, I'd expect a `bool` to be returned, but rather it is a string of "False"

Relevant code snippets:

* https://github.com/sopel-irc/sopel/blob/86c73dbb9040351e48c453266d9c0e241302e252/sopel/db.py#L826
* https://github.com/sopel-irc/sopel/blob/86c73dbb9040351e48c453266d9c0e241302e252/sopel/db.py#L48

Thank you for taking a look. I love what Sopel as become!

### Reproduction steps

1. `bot.db.set_channel_value("##test", "blah")`
2. `bot.db.get_channel_value("##test", "blah", False)`

### Expected behavior

I expect to get a type of bool of `False` rather than a string of "False"

### Relevant logs

```shell

```

### Notes

_No response_

### Sopel version

8.0.2

### Installation method

`pip install`

### Python version

3.11.2

### Operating system

Debian 12

### IRCd

Libera.Chat

### Relevant plugins

custom plugin

TowyTowy:
### Description

`get_nick_value()`, `get_channel_value()`, and `get_plugin_value()` passed the caller-supplied `default` through `_deserialize()` along with values read from the database. Since `_deserialize()` runs `json.loads(str(value))`, a non-string default whose `str()` isn't valid JSON was corrupted — most visibly a `False` default came back as the string `"False"` (because JSON's boolean is lowercase `false`).

Deserialization only makes sense for values coming out of the store; a caller-supplied default is already a Python object and should be returned verbatim. This changes the three getters to only deserialize DB results and return the default unchanged.

Fixes #2672

### Checklist

- [x] I have read CONTRIBUTING.md


dgw:
To me, this is a big enough behavior change that we shouldn't consider it until the next major version. I don't think it needs any fancy logic to warn about it in 8.1.x (how would we even do that?), yet springing it on people in a minor release doesn't feel right.

Plus, it gives us time to really think through the potential consequences.

